PR #222, opened 2016-11-02
Set default URLs to point at weave cloud

squaremo:
Also, allow the token to be supplied via the environment variable `FLUX_SERVICE_TOKEN`, for convenience.


squaremo:
(CI is actually green, but the rename tricked github)

PR #223, opened 2016-11-02
Rename to flux

squaremo:
`mv flux{y,}`


squaremo:
No-one to review but it has to happen, so.

PR #227, opened 2016-11-04
Better secrets

squaremo:
Replaces the "******" output with things that make it easier for users to troubleshoot, e.g., a public key fingerprint.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,5 +1,7 @@
 package flux
 
+const secretReplacement = "******"
+
 // Instance configuration, mutated via `fluxctl config`. It can be
 // supplied as YAML (hence YAML annotations) and is transported as
 // JSON (hence JSON annotations).
@@ -29,4 +31,15 @@ type InstanceConfig struct {
 	Git      GitConfig      `json:"git" yaml:"git"`
 	Slack    SlackConfig    `json:"slack" yaml:"slack"`
 	Registry RegistryConfig `json:"registry" yaml:"registry"`
+}
+
+func (c InstanceConfig) HideSecrets() InstanceConfig {
+	if c.Git.Key != "" {
+		c.Git.Key = secretReplacement
+	}
+	for host, auth := range c.Registry.Auths {
+		auth.Auth = secretReplacement
+		c.Registry.Auths[host] = auth
+	}
+	return c
 }
--- a/config.go
+++ b/config.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rsa"
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/json"
 	"golang.org/x/crypto/ssh"
 	"strings"
 )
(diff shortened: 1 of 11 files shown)


config.go
```diff
@@ -20,13 +30,62 @@ type RegistryConfig struct {
...
+		c.Registry.Auths[host] = auth.HidePassword()
+	}
+	return c
+}
+
+func (a Auth) HidePassword() Auth {
+	if a.Auth == "" {
+		return a
+	}
+	bytes, err := base64.StdEncoding.DecodeString(a.Auth)
```
paulbellamy: but... Why do we base64 encode passwords?

config.go
```diff
@@ -20,13 +30,62 @@ type RegistryConfig struct {
...
+	}
+
+	pubKey, err := ssh.NewPublicKey(&privKey.PublicKey)
+	if err != nil {
+		g.Key = secretReplacement
+		return g
+	}
+
+	hash := sha256.Sum256(pubKey.Marshal())
+	g.Key = "SHA256: " + strings.TrimRight(base64.StdEncoding.EncodeToString(hash[:]), "=") + " (RSA)"
```
paulbellamy: You can't generate the private key from a public one, so why not just print the public key? too long? Is this a standard fingerprint format?

server/server.go
```diff
@@ -307,7 +305,7 @@ func (s *Server) GetConfig(instID flux.InstanceID, includeSecrets bool) (flux.In
 
 	config := fullConfig.Settings
 	if !includeSecrets {
-		removeSecrets(&config)
+		config = config.HideSecrets()
```
paulbellamy: It would be nice it no secrets was the default, and you had to "ExposeSecrets()"

config.go
squaremo: Because that's what the docker config does, for better or worse, and I chose to accept creds in that form to make it easier to supply them via clipboard (or just pointing at the file).

server/server.go
squaremo: I hear ya -- safety by default. Maybe some tricky thing whereby the usual config value marshals with hidden secrets, but you can get another type that marshals *with* secrets?

(cross fingers there ends up being no good reason to ever transmit secrets, in which case this goes away)

config.go
paulbellamy: > directly comparable to github & ssh-keygen
👍

server/server.go
paulbellamy: > Maybe some tricky thing whereby the usual config value marshals with hidden secrets, but you can get another type that marshals with secrets?

yep.
```
type ConfigWithSecrets Config

func (c ConfigWithSecrets) MarshalJSON() {
...
}
```
type thing.

squaremo: > cross fingers there ends up being no good reason to ever transmit secrets

Actually I don't think there is any good reason to ever send secrets back; and, having implemented the marshalling trick discussed above, I can report that it's just confusing.

So I think I will remove the secrets option, *always* hide secrets, and simply put the public key in the Key field (you can always generate the fingerprint if you want to check, or we can add an option for it).

squaremo:
@paulbellamy PTAL. I've removed the ability to ask for secrets back again, and put the fingerprinting in the client as an option, so you can get the whole public key back.

PR #240, opened 2016-11-16
Platform presence

squaremo:
This adds a method `IsDaemonConnected` to the `DaemonAPI` interface, for checking whether a particular daemon is known to be connected, and exposes it in the HTTP transport.

It's not in the ClientAPI interface because I don't expect it to be called from the command-line client (although that's somewhat arbitrary at this point, anyway it can be moved later).

`IsDaemonConnected` asks the message bus, which checks for the presence of a platform subscribed as the instance given. "Presence" means that we can ping it via an RPC call -- necessitating a type representing a platform that can be pinged, `RemotePlatform`.

Why ping? In some ways, just answering "yes" if there's a connection registered would be neater and faster. However, we don't in general know if connections are live, since they may not shut down cleanly -- and we only find that out after attempting to write to (and read from) the socket. A secondary benefit of using a ping is that it demonstrates that the process on the other side is still minimally
active.

(It also fixes a bug in which the test for cleaning up a connection compared two interface pointers that would never be equal.)

To make testing easier, I've changed the signature of Subscribe so that it accepts a channel on which to synchronise, and returns immediately instead of blocking. This involves a goroutine to
clean up after, and forward any error.

### DIFF
--- a/http/transport.go
+++ b/http/transport.go
@@ -22,6 +22,7 @@ import (
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/api"
 	"github.com/weaveworks/flux/http/websocket"
+	"github.com/weaveworks/flux/platform"
 	"github.com/weaveworks/flux/platform/rpc"
 )
 
@@ -39,6 +40,7 @@ func NewRouter() *mux.Router {
 	r.NewRoute().Name("GetConfig").Methods("GET").Path("/v4/config")
 	r.NewRoute().Name("SetConfig").Methods("POST").Path("/v4/config")
 	r.NewRoute().Name("RegisterDaemon").Methods("GET").Path("/v4/daemon")
+	r.NewRoute().Name("IsConnected").Methods("HEAD", "GET").Path("/v4/ping")
 	return r
 }
 
@@ -56,6 +58,7 @@ func NewHandler(s api.FluxService, r *mux.Router, logger log.Logger, h metrics.H
 		"GetConfig":      handleGetConfig,
 		"SetConfig":      handleSetConfig,
 		"RegisterDaemon": handleRegister,
+		"IsConnected":    handleIsConnected,
 	} {
 		var handler http.Handler
 		handler = handlerFunc(s)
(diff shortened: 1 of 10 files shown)


platform/standalone.go
```diff
@@ -34,39 +34,58 @@ func (s *StandaloneMessageBus) Connect(inst flux.InstanceID) (Platform, error) {
...
+		done:   done,
 	}
 	s.Unlock()
 
-	// Wait to be kicked, or an error to happen
-	err := <-done
+	// The only way we detect remote platforms closing are if an RPC
+	// is attempted and it fails. When that happens, clean up behind
+	// us.
+	go func() {
```
squaremo: Pretty sure this can be done better

squaremo:
@paulbellamy Do you mind taking a look -- especially at the bit I've commented on above (the last commit, in effect).

platform/standalone.go
paulbellamy: AFAICT, everywhere we call this we do:
```Go
foo.Subscribe(blah, complete)
err := <-complete
```
So, why not just have this block and return the error? Then you don't need this to be in a goroutine.

platform/platform.go
```diff
@@ -24,17 +24,32 @@ type Platform interface {
 	Regrade([]RegradeSpec) error
 }
 
+type RemotePlatform interface {
+	Platform
+	Ping() error
```
paulbellamy: Platform could have a `Ping() error` too, and have standalone platform return nil (or contact k8s). Then we could maintain a single interface.

squaremo: Hmm, yeah I guess so; RPC would then call that Ping, rather than implementing it inside the mechanism.

platform/standalone.go
squaremo: Because something needs to see the error and do the cleanup before continuing.

squaremo: Or do you mean, why not just go back to how it was, where it blocked inside the method? Because then it's very difficult to write a test that uses Subscribe -- that was the motivation for this change.

platform/platform.go
squaremo: (I am in two minds about this, but I did end up folding Ping into the Platform interface.)

http/transport.go
```diff
@@ -636,6 +639,25 @@ func handleRegister(s api.FluxService) http.Handler {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		inst := getInstanceID(r)
+
+		switch s.IsDaemonConnected(inst) {
+		case platform.ErrPlatformNotAvailable:
+			w.WriteHeader(http.StatusNotFound)
+		case nil:
+			w.WriteHeader(http.StatusNoContent)
+		default:
+			w.WriteHeader(http.StatusInternalServerError)
```
paulbellamy: maybe, log the error here?

squaremo: There's no log to log logs in

paulbellamy: > (do we really want to send internal errors to the user?)
Noooooooooope

squaremo: Ops! That happens all through this file.

squaremo: OK removing that again.

Issue #239, opened 2016-11-16
Could not find image name


paulbellamy:
```
$ fluxctl release --all --update-image=quay.io/weaveworks/billing-usage:0.3.0
Submitting release job...
Release job submitted, ID 3b85d97f-3545-5b09-1156-3d0569d826c9
Status: Update 1 images(s) in the resource definition file for billing/billing-usage: billing-usage (quay.io/weaveworks/billing-usage:0.2.0 -> quay.io/weavewoStatus: Failed: updating pod controller for quay.io/weaveworks/billing-usage:0.3.0: Could not find image name

Here's as far as we got:
 1) Submitted job.
 2) Calculating release actions.
 3) Release quay.io/weaveworks/billing-usage:0.3.0 to all services
 4) service default/kubernetes does not have images associated: empty selector
 5) service kube-system/scope-probe does not have images associated: no matching replication controllers or deployments
 6) Clone the config repo.
 7) Clone OK.
 8) Update 1 images(s) in the resource definition file for billing/billing-usage: billing-usage (quay.io/weaveworks/billing-usage:0.2.0 -> quay.io/weaveworks/billing-usage:0.3.0).
 9) updating pod controller for quay.io/weaveworks/billing-usage:0.3.0: Could not find image name
 10) Failed: updating pod controller for quay.io/weaveworks/billing-usage:0.3.0: Could not find image name
Took 29.028194s
```

It might be because the image tag is all numbers and dots.

squaremo:
See PR #241

squaremo:
Fixed by #241.

PR #241, opened 2016-11-18
Allow dots and underscores in (yamel) image tags

squaremo:


squaremo:
Fixes #239.

platform/kubernetes/update.go
```diff
@@ -84,7 +84,7 @@ func tryUpdate(def, newImageStr string, trace io.Writer, out io.Writer) error {
 	imageRE := multilineRE(
 		`      containers:.*`,
 		`(?:      .*\n)*(?:  ){3,4}- name:\s*"?([\w-]+)"?(?:\s.*)?`,
-		`(?:  ){4,5}image:\s*"?(`+newImage.Repository()+`:[\w-]+)"?(\s.*)?`,
+		`(?:  ){4,5}image:\s*"?(`+newImage.Repository()+`:[._\w-]+)"?(\s.*)?`,
```
paulbellamy: Why not: https://github.com/docker/distribution/blob/master/reference/regexp.go#L36 ?

Issue #216, opened 2016-10-26
Messaging system to intermediate between daemons and service


squaremo:
In #199 daemons connect to a single fluxsvc instance with a websocket. Operations needing to ask the platform for something use a Connecter to get a handle to the RPC mechanism; that RPC mechanism is simply to serialise the request to the websocket (as a JSON-RPC), and decode the response.

Once there is more than one instance of fluxsvc this will not be adequate. Requests from one instance must be conveyed to wherever the desired daemon is connected (which is still via websockets).

One fairly obvious way to do it is with a messaging system to route the requests and responses. This does not have to be reliable, since we expect RPCs to fail from time to time anyway. RPCs get put into the system with the daemon (instance) name along with a "reply to", and relayed by the fluxsvc to which that daemon is connected, along with the "reply to". The responses include the "reply to", and relayed by the fluxsvc to the messaging system with the "reply to" as the destination.

(Another way is to simply have a directory to look up the address the fluxsvc to which the daemon is connected, and have that daemon relay requests along with a return address.)

paulbellamy:
Next up, is to deploy nats and switch over, right? Anything else outstanding?

paulbellamy:
Before deploying, please check https://github.com/weaveworks/flux/pull/314 hasn't broken this.

paulbellamy:
Blocked from going to prod on: https://github.com/weaveworks/flux/issues/420

rade:
I know we've discussed what's left to do here, but I keep forgetting.... so could somebody please list that here?

paulbellamy:
We've changed the rate limits in fluxd, but need to have people deploy that, otherwise switching this  on will make it timeout for them.

paulbellamy:
Discussed f2f: Releasing this means we need to tell everyone (all 5 of them) to upgrade their fluxd. That's easy, but let's wait for a bigger release to minimize disruption. Moved into :octocat: sync-loop milestone to be deployed when we do that release.

squaremo:
Imma just close this -- we're using NATS for the (>=v6 API) service.

PR #246, opened 2016-11-21
Use NATS as a message bus

squaremo:
In short: this provides an implementation of `MessageBus` which sends requests over NATS.

What's the point? If we are serving daemon connections from more than one pod (i.e., if we scale the fluxsvc deployment to > 1) then the pod that answers a client request may be different to the pod that is serving the daemon's connection. Using a message bus (i.e., NATS) connects them up.

It may be possible to make the `net/rpc` abstractions work with NATS (though I didn't find a library for it). However, we have a requirement which is difficult to reconcile with `net/rpc`: each request must be addressed and routed to a specific daemon. For that reason I implemented a simple scheme using NATS operations (Subscribe, Request) directly.

### DIFF
--- a/platform/rpc/nats/bus.go
+++ b/platform/rpc/nats/bus.go
@@ -2,7 +2,7 @@ package nats
 
 import (
 	"errors"
-	"fmt"
+	"strings"
 	"time"
 
 	"github.com/nats-io/nats"
@@ -15,11 +15,22 @@ import (
 const (
 	timeout      = 5 * time.Second
 	presenceTick = 10 * time.Millisecond
+	encoderType  = nats.JSON_ENCODER
+
+	methodPresence     = ".Platform.Presence"
+	methodAllServices  = ".Platform.AllServices"
+	methodSomeServices = ".Platform.SomeServices"
+	methodRegrade      = ".Platform.Regrade"
 )
 
 type NATS struct {
-	url  string
-	conn *nats.EncodedConn
+	url string
+	// It's convenient to send (or request) on an encoding connection,
+	// since that'll do encoding work for us. When receiving though,
+	// we want to decode based on the method as given in the subject,
+	// so we use a regular connection and do the decoding ourselves.
+	snd *nats.EncodedConn
+	rcv *nats.Conn
 }
 
 var _ platform.MessageBus = &NATS{}
@@ -29,13 +40,14 @@ func NewMessageBus(url string) (*NATS, error) {
 	if err != nil {
 		return nil, err
 	}
-	encConn, err := nats.NewEncodedConn(conn, nats.JSON_ENCODER)
+	encConn, err := nats.NewEncodedConn(conn, encoderType)
 	if err != nil {
 		return nil, err
 	}
 	return &NATS{
-		url:  url,
-		conn: encConn,
+		url: url,
+		rcv: conn,
+		snd: encConn,
 	}, nil
 }
 
@@ -50,18 +62,13 @@ func (n *NATS) AwaitPresence(instID flux.InstanceID, timeout time.Duration) erro
 	for {
 		select {
 		case <-attempts.C:
-			if err := n.conn.Request(string(instID)+".Platform.Presence", Presence{}, &pres, presenceTick); err == nil {
+			if err := n.snd.Request(string(instID)+methodPresence, Presence{}, &pres, presenceTick); err == nil {
 				return nil
 			}
 		case <-timer:
-			return errors.New("timeout")
+			return errors.New("presence timeout")
 		}
 	}
-}
-
-type requester struct {
-	conn    *nats.EncodedConn
-	subject string
 }
 
 type Presence struct{}
@@ -95,9 +102,16 @@ func maybeString(err error) string {
 	return ""
 }
 
+// requester just collect the things you need to make a request
+// together
+type requester struct {
+	conn     *nats.EncodedConn
+	instance string
+}
+
 func (r *requester) AllServices(ns string, ig flux.ServiceIDSet) ([]platform.Service, error) {
 	var response AllServicesResponse
-	if err := r.conn.Request(r.subject+".Platform.AllServices", rpc.AllServicesRequest{ns, ig}, &response, timeout); err != nil {
+	if err := r.conn.Request(r.instance+methodAllServices, rpc.AllServicesRequest{ns, ig}, &response, timeout); err != nil {
 		return nil, err
 	}
 	return response.Services, maybeError(response.Error)
@@ -105,7 +119,7 @@ func (r *requester) AllServices(ns string, ig flux.ServiceIDSet) ([]platform.Ser
 
 func (r *requester) SomeServices(incl []flux.ServiceID) ([]platform.Service, error) {
 	var response SomeServicesResponse
-	if err := r.conn.Request(r.subject+".Platform.SomeServices", incl, &response, timeout); err != nil {
+	if err := r.conn.Request(r.instance+methodSomeServices, incl, &response, timeout); err != nil {
 		return nil, err
 	}
 	return response.Services, maybeError(response.Error)
@@ -113,7 +127,7 @@ func (r *requester) SomeServices(incl []flux.ServiceID) ([]platform.Service, err
 
 func (r *requester) Regrade(specs []platform.RegradeSpec) error {
 	var response RegradeResponse
-	if err := r.conn.Request(r.subject+".Platform.Regrade", specs, &response, timeout); err != nil {
+	if err := r.conn.Request(r.instance+methodRegrade, specs, &response, timeout); err != nil {
 		return err
 	}
 	if len(response.Result) > 0 {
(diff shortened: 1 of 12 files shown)


platform/rpc/nats/bus.go
```diff
@@ -0,0 +1,234 @@
...
+				err = errors.New("unknown message: " + request.Subject)
+			}
+			if err != nil {
+				sub.Unsubscribe()
+				close(requests)
+				done <- err
+				return
+			}
+		}
+	}()
```
paulbellamy: This giant switch *might* be easier to clean up if you separate out the "decode the thing", from the "do the thing" phases. Not sure.

platform/rpc/nats/bus.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	return nil
+}
+
+func maybeString(err error) string {
+	if err != nil {
+		return err.Error()
+	}
+	return ""
+}
```
peterbourgon: I find these "helper" methods unhelpful and confusing, and remember refactoring to remove a similar construct in the past. Can we please inline this logic instead?

platform/rpc/nats/bus.go
```diff
@@ -0,0 +1,246 @@
...
+// to talk to a particular instance.
+func (n *NATS) Connect(instID flux.InstanceID) (platform.Platform, error) {
+	return &requester{
+		conn:     n.snd,
+		instance: string(instID),
+	}, nil
+}
+
+// Subscribe registers a remote platform.Platform implementation as
+// representing a particular instance ID, blocking indefinitely.
```
peterbourgon: Is the "blocking indefinitely" part of this comment still accurate?

platform/rpc/nats/bus.go
```diff
@@ -0,0 +1,246 @@
...
+	Services []platform.Service
+	Error    string
+}
+
+type RegradeResponse struct {
+	Result rpc.RegradeResult
+	Error  string
+}
+
+type ping struct{}
```
peterbourgon: It looks like this is meant to be a placeholder. But that's a bit confusing; seeing a type definition, I expect some methods or something. What do you think about inlining `struct{}{}` in the places you use `ping{}`?

platform/rpc/nats/bus.go
squaremo: I don't understand what is confusing about using these functions. Is it the names? The particular use of functions to replace common bits of code? What is the difficulty here.

platform/rpc/nats/bus.go
squaremo: Nope it's not accurate, it was carried across from an older comment on `platform/standalone.go`. Ops! I'll mend it.

platform/rpc/nats/bus.go
peterbourgon: At a glance, it is not apparent what `return ..., maybeError(struct.Field)` will return, or is doing. In general, it's nonidiomatic to hide or abstract error control flow with helper functions. Another example of that would be

```go
func main() {
    f, err := getSomething()
    checkErr(err)
    v, err := doSomething()
    checkErr(err)
    fmt.Println(v)
}

// Almost never the right way to do things.
func checkErr(err error) { 
    if err != nil {
        panic(err)
    }
}
```

squaremo: They don't abstract error control flow, so perhaps the confusion arises from the names. They are simply for transporting errors, which don't travel well.

If they were named `errorAsString` and `stringAsError`, would that be less misleading do you think?

peterbourgon: Well, if they were conversion functions, then perhaps; but they can also produce nil errors, which is a change in semantic meaning, so that's also misleading.

I'm telling you that

```go
func foo() {
    // ...
    return stringAsError(response.Error)
}
```

is somewhat confusing and nonidiomatic, and asking you to please write it out as

```go
func foo() {
    // ...
    if response.Error != "" {
        return errors.New(response.Error)
    }
    return nil
}
```

at each site.

platform/rpc/nats/bus.go
peterbourgon: Finding `struct{}{}` in a method call conveys an obvious meaning to my weary Go eyes: that this parameter is almost certainly an empty interface, and that it's being decoded-into or encoded-from, and that in this case it doesn't need to be anything in particular, because `struct{}` has no methods by definition. Finding `ping{}`, on the other hand, is an unknown: it opens the possibility that the parameter should satisfy some interface, which I'm now obliged to track down.

Do as you like.

platform/rpc/nats/bus.go
paulbellamy: fwiw, I agree with @peterbourgon, that you should inline it, because it's non-standard.

platform/rpc/nats/bus.go
paulbellamy: I prefer ping as a way of applying specific meaning to the message. The interpretation of `struct{}{}` is unspecified, while `ping{}` can be inferred to have some specific meaning. Much like `interface{}{}` vs a concrete type.

platform/rpc/nats/bus.go
squaremo: > Well, if they were conversion functions

`stringAsError` is a function from `string` to `error` (`error` includes `error(nil)`).

But anyway, taking the charge of readability seriously, I rewrote a procedure to do the same conversation explicitly inline:

```
func (r *requester) AllServices(ns string, ig flux.ServiceIDSet) ([]platform.Service, error) {
	var response AllServicesResponse
	var err error
	if err = r.conn.Request(r.instance+methodAllServices, rpc.AllServicesRequest{ns, ig}, &response, timeout); err != nil {
		return nil, err
	}
	if response.Error != "" {
		err = errors.New(response.Error)
	}
	return response.Services, err
}
```

Honestly, I do not see how this makes it more readable -- it just adds some noise, in the form of unimportant detail, which must be analysed each time to make sure it's not doing something important (or subtly different to the last such block). I'm not a zealot who wants to factor everything into noun soup. I really think in this instance that putting the conversion into a function and using it in value position makes things easier to follow.

squaremo:
As it stands, this will have problems with long-running regrades, since they take an unbounded amount of time, and the timeouts here will look like failures.

squaremo:
- [ ] Increase or remove timeout on Regrade, or otherwise figure out what to do with them

squaremo:
*Deployment notes*

This can be deployed without changing config (it will default to using the standalone bus). To use NATS:

 - deploy NATS (a minimal configuration is fine to start with)
 - add a `--nats-url` argument to fluxsvc's config
 - bump up the replicas!

platform/rpc/nats/bus.go
```diff
@@ -0,0 +1,276 @@
...
+	if _, ok := err.(platform.FatalError); ok {
+		resp.Fatal = true
+	}
+	resp.Error = err.Error()
+	return resp
+}
+
+// requester just collect the things you need to make a request
+// together
+type requester struct {
```
paulbellamy:
```
type platform struct {
  instanceID string
  conn       *nats.EncodedConn
```
maybe? Would match the interface it is implementing?

platform/rpc/nats/bus.go
```diff
@@ -0,0 +1,276 @@
...
+		instance: string(instID),
+	}, nil
+}
+
+// Subscribe registers a remote platform.Platform implementation as
+// the daemon for an instance (identified by instID). Any
+// platform.FatalError returned when processing requests will result
+// in the platform being deregistered, with the error put on the
+// channel `done`.
+func (n *NATS) Subscribe(instID flux.InstanceID, remote platform.Platform, done chan<- error) {
```
paulbellamy: It seems like this doesn't kick any existing connected remote platforms? Could this mean you will have multiple fluxsvcs listening and responding to rpcs?

platform/rpc/nats/bus.go
squaremo: Good idea.

platform/rpc/nats/bus.go
paulbellamy: is there a way of a fluxd to uniquely identify its cluster? a clusterID of some sort, it could send?

paulbellamy:
Fixes #216

platform/rpc/nats/bus.go
squaremo: Not sure, but either way, just identifying each subscriber uniquely, and publishing to a topic named for the instance, is enough to implement the kick function.

(We may want the cluster ID to be able to distinguish between daemons in the same cluster conflicting, and daemons in different clusters.)

platform/rpc/nats/bus_test.go
```diff
@@ -0,0 +1,190 @@
+package nats
```
paulbellamy: It would be nice if we didn't *need* nats running to run the tests. Maybe a build flag: `// +build integration` ? I don't feel strongly about it, though, wdyt?

squaremo: Yeah don't feel strongly either. If `gnatsd` was more trouble to run than `go get ... ; gnatsd &` then I'd be more inclined to make the tests opt-in.

squaremo:
Going to wait until a weekday to merge ..

PR #250, opened 2016-11-23
Version release binaries

squaremo:
Add the command `fluxctl version`, and give the release binaries a version based on the state of the git repo (as for image tags).

The purpose is to be able to identify the binary used by a user, e.g., in bug reports.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,12 @@ MIGRATIONS:=$(shell find db/migrations -type f)
 all: $(GOPATH)/bin/fluxctl $(GOPATH)/bin/fluxd $(GOPATH)/bin/fluxsvc build/.fluxd.done build/.fluxsvc.done
 
 .PHONY: release-bins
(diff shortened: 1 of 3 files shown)


peterbourgon:
LGTM otherwise.

Makefile
```diff
@@ -19,7 +19,12 @@ MIGRATIONS:=$(shell find db/migrations -type f)
 all: $(GOPATH)/bin/fluxctl $(GOPATH)/bin/fluxd $(GOPATH)/bin/fluxsvc build/.fluxd.done build/.fluxsvc.done
 
 .PHONY: release-bins
-release-bins: build/fluxctl-linux-amd64 build/fluxctl-darwin-amd64
+release-bins:
```
squaremo: Nope; if it did, then `make release-bins` would not always build binaries with the current version baked in (sometimes it would look at the dependencies and decide it didn't need to).

PR #252, opened 2016-11-24
Correct lookup of "library" images

### LINKED ISSUE #245: Error in list-images
```
$ fluxctl —token <token>   list-images
Error: executing HTTP request: reading HTTP response: 500 Internal Server Error (getting images for services: fetching image metadata for nats: Get https://index.docker.io/v2/nats/manifests/latest: http: non-successful response (status=401 body="{\"errors\":[{\"code\":\"UNAUTHORIZED\",\"message\":\"authentication required\",``\"detail\":[{\"Type\":\"repository\",\"Name\":\"nats\",\"Action\":\"pull\"}]}]}\n")
```

squaremo:
Some image names do not include an org; e.g., "nats:0.8.0". These are,
by convention, kept on docker hub in the org "library". However, in
general we want to deal with image names as they are supplied --
`nats:0.8.0` rather than `library/nats:0.8.0`.

The convention only really matters when looking up images in the
registry; so, this commit simply changes the code there to distinguish
between the name used to look up the image and the name to be used in
the result.

Fixes #245.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -112,17 +112,21 @@ func (c *Client) GetRepository(repository string) ([]flux.ImageDescription, erro
 		return nil, err
 	}
 
-	return c.tagsToRepository(cancel, client, repository, tags)
+	// the hostlessImageName is canonicalised, in the sense that it
+	// includes "library" as the org, if unqualified -- e.g.,
+	// `library/nats`. We need that to fetch the tags etc. However, we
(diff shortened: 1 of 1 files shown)


peterbourgon:
LGTM pending Circle.

squaremo:
Thanks for the review @peterbourgon

Issue #256, opened 2016-11-29
Release Jobs are not recoverable


squaremo:
In some circumstances a release job will never be marked as complete. We do report back if we *see* a failure, so a person can retry; however, we do not see all failures, and possibly worse, we do not see all successes.

Here are some of the things that can go wrong during a release, and the outcome.

*1. Error at any point up to pushing a commit to the config repo*

The failure is reported back to the user (or to the logs and history, if an automated release). No lasting effects.

*2. The service is restarted during a regrade*

The goroutine waiting on the regrade result is running in the service; when the service pod is terminated, the result of the regrade is lost. The user, or the client, may retry `CheckRelease`, but it will never be updated. No history or notification is recorded.

*3. The daemon terminates or loses the connection during the regrade*

In this case we don't know the result of the regrade. An error is reported back to the user, who can e.g., restart the daemon, etc., then review the state of the system to reconstruct what happened. The release is recorded as failed, though it may actually have succeeded.

------

The first is basically business as usual.

To mitigate 2) and to some extent 3), we can make the regrade operation non-blocking -- calling Regrade kicks a regrade off (i.e., calls only `kubectl apply`) and returns only any immediate problems (e.g., malformed yamel). The history and notifications are recorded at this point.

Subsequent invocations of `CheckRelease` look at the "rollout state" of the service by asking the daemon, rather than simply reporting on what's in the release job record, since there is no process actively updating it. (The staleness criteria for release jobs would need to be changed.)

We'd also need to try and make the release workers shut down gracefully, i.e., not leave incomplete records, although this will never be perfect of course.

The main problem with 3) is that successes can look like failures. With the mitigation above, the window for this to happen is smaller, but it can still happen. Possibly we can report this to the user in such a way as to make it clear that the outcome is unknown, and please check later.

squaremo:
Note transplanted from #261: the worker can become a worker pool, managing its own goroutines. This would help tidy up main.go.

paulbellamy:
This (combined with: https://github.com/weaveworks/flux/pull/311) would fix long-running deploys (e.g. cortex)

squaremo:
Another problematic scenario: if a job panics, it will not be completed, and our rate limiting will mean no further jobs for that instance will be attempted.

philwinder:
Need to take another look at this after the move to git-sync. We will probably have similar issues, but they might be different.

Also, this issue is too vague to take any action. Needs some refining.

squaremo:
[...]

PR #261, opened 2016-11-30
Make fluxsvc regrade async, gracefully stop

squaremo:
When upgrading fluxsvc itself, we effectively ask Kubernetes to restart the pod(s) that's watching the regrade process, and thereby lose the result.

This commit makes `fluxsvc` a special case which is run asynchronously. That is, the regrade is started but the service does not wait for it to complete before reporting back to the user.

There is still a race between reporting back to the user, and the pod being restarted. For that reason, the service attempts to shut down gracefully, waiting until all release processes are completed, with a generous timeout. This means the release procedure has a chance to record what happened.

This commit also changes the worker loop so that it polls for the next job with a backoff, rather than polling on a tick.

### DIFF
--- a/cmd/fluxsvc/main.go
+++ b/cmd/fluxsvc/main.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 	"time"
 
@@ -28,6 +29,8 @@ import (
 	"github.com/weaveworks/flux/release"
 	"github.com/weaveworks/flux/server"
 )
+
+const shutdownTimeout = 30 * time.Second
 
 func main() {
 	// Flag domain.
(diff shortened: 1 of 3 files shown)


cmd/fluxsvc/main.go
```diff
@@ -181,11 +184,25 @@ func main() {
...
+		workers := sync.WaitGroup{}
+
+		// for each worker goroutine
+		{
+			workers.Add(1)
+			worker := release.NewWorker(rjs, instancer, releaseMetrics, logger)
+			go func() {
+				defer workers.Done()
+				worker.Work(shutdownWorkers)
+			}()
```
paulbellamy: What about encapsulating the goroutines inside the worker? so it just has a single poller looking at the db, then passes jobs off to goroutines?

A Worker.Stop() method could then do the shutdown/wait dance.

WDYT?

release/releaser.go
```diff
@@ -582,11 +589,19 @@ func (r *releaser) releaseActionRegradeServices(services []flux.ServiceID, msg s
...
+						ServiceID:     service,
+						NewDefinition: def,
+					})
+				} else {
+					rc.Instance.LogEvent(namespace, serviceName, "Starting regrade "+cause)
+					specs = append(specs, platform.RegradeSpec{
+						ServiceID:     service,
+						NewDefinition: def,
+					})
+				}
```
paulbellamy: Might be cleaner if `Async: true`, were part of the spec? But I guess we want to do all the sync spec, *then* any async ones after, so... fine.

squaremo: The decision about whether to treat the release as async could be made higher up and passed to us, that's true. I liked this way around simply because it kept the hack to one file :)

cmd/fluxsvc/main.go
squaremo: Mmm yes, that would keep main.go tidier, which would be pleasing. I was trying not to intervene too much, leaving larger changes for the larger overhaul (probably that which addresses #256).

squaremo:
Thanks @paulbellamy!

PR #266, opened 2016-12-01
Detect when service unavailable and retry

squaremo:
In some cases, particularly when fluxsvc is itself restarting, `fluxctl check-release` will encounter 50x responses from a proxy. These are usually transient, so the thing to do is to keep checking back.

### DIFF
--- a/cmd/fluxctl/check_release_cmd.go
+++ b/cmd/fluxctl/check_release_cmd.go
@@ -44,7 +44,7 @@ func (opts *serviceCheckReleaseOpts) Command() *cobra.Command {
 	return cmd
 }
 
-func (opts *serviceCheckReleaseOpts) RunE(_ *cobra.Command, args []string) error {
+func (opts *serviceCheckReleaseOpts) RunE(cmd *cobra.Command, args []string) error {
 	if len(args) != 0 {
 		return errorWantedNoArgs
 	}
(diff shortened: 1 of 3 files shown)


squaremo:
> It should probably eventually fail altogether, no?

You can always interrupt it .. but I suppose one might switch away to some other task, and forget about it.

paulbellamy:
It is difficult for a user to know if it is still working, or if it should be interrupted.

squaremo:
When trying this out by killing authfe, I noticed that a reverse proxy can take an awfully long time to report a 502 Gateway Error, meaning that the client ends up stuck waiting for the result (and saying nothing) much longer than it spends counting down to a retry.

Now I'm thinking it would be better just to exit the first time, and say "please check later".

paulbellamy:
I'm assuming that if the service comes back, there is no 502 served to the client?

paulbellamy:
We could consider X retries, or X seconds, whichever comes first...

squaremo:
> I'm assuming that if the service comes back, there is no 502 served to the client?

When I restart authfe, it still seems to often pause then return a 502; though it does eventually recover and report the result.

squaremo:
> We could consider X retries, or X seconds, whichever comes first...

This is a good idea -- in fact, just having an overall time limit is much better than retries with backoff, since the attempts themselves take a variable amount of time.

Issue #262, opened 2016-11-30
Automation of missing services interrupts wss connection constantly


squaremo:
If you automate a service that doesn't exist, it will constantly force the daemon's websocket to reconnect.

This is because the semantics for the RPC-over-websocket are that any error returned closes the connection and forces the daemon to reconnect. This is not strictly necessary -- it's done because any error may indicate a fatal problem with the websocket, so it's safer to drop it. In combination with an automated operation that is guaranteed to cause an error, though, it's problematic.

squaremo:
Fixed by #269.

PR #269, opened 2016-12-02
Don't drop websocket on application error

squaremo:
At present, any error returned by a daemon will cause the connection
to be dropped: the error is fed back to the subscribe method, which
closes its completion channel, and the HTTP handler exits, closing the
connection.

Most of these errors will be *application* level errors, e.g,. "such
and such a service does not exist", rather than transport level
errors. So it's reasonable to keep the connection open in these cases.

Happily, the RPC mechanism transmits application errors as strings,
and reconstitutes them by simply casting them to `rpc.ServerError`, so
we can detect them. Everything else becomes a `FatalError`, which will
close down the connection as before.

### DIFF
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -25,6 +25,16 @@ type Platform interface {
 	Ping() error
 }
 
+// Wrap errors in this to indicate that the platform should be
+// considered dead, and disconnected.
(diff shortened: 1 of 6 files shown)


squaremo:
(Fixes #262)

PR #270, opened 2016-12-02
Detect when the client v service API is mismatched

squaremo:
Getting a 404 result from the client usually means that there's a
mismatch between the client and the service. This commit makes that
situation detectable by returning a particular type for transport
errors (as opposed to application errors), and treating 404 as a
special case in the client.

### DIFF
--- a/cmd/fluxctl/main.go
+++ b/cmd/fluxctl/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"io"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


squaremo:
> In general, would having the client call a /version endpoint, and a single API version. Then having the client fail early, be a better plan?

Honestly, I don't think so. Having a single API version means every change to the API obsoletes all `fluxctl`s everywhere.

Not that we have been doing anything much better. Probably we need to sit down and figure out the versioning conventions, e.g., adding an optional parameter doesn't need a new version; changing the return type does, that sort of thing.

squaremo:
In the meantime, is "definitely better than what we have" good enough?

PR #281, opened 2016-12-08
Rework the automator to use scheduled jobs, so it can be distributed.

paulbellamy:


jobs/db_store.go
```diff
@@ -127,6 +127,18 @@ func (s *DatabaseStore) PutJob(inst flux.InstanceID, job Job) (JobID, error) {
 	}
 
 	err = s.Transaction(func(s *DatabaseStore) error {
+		if job.Key != "" {
+			var count int
+			err := s.conn.QueryRow(`
+			SELECT count(*) FROM jobs WHERE key = $1 AND claimed_at IS NULL
```
paulbellamy: I initially had `finished_at IS NULL`, but that precludes re-scheduling the current job. (You have to mark it finished prematurely, which is a bit weird). Hence I went with `claimed_at` instead.

paulbellamy:
You're right, it is a bit fragile. Will rework to make it less so.

automator/automator.go
```diff
@@ -41,7 +41,7 @@ func (a *Automator) Start(errorLogger log.Logger) {
 		for _, inst := range insts {
 			for service, conf := range inst.Config.Services {
 				if conf.Policy() == flux.PolicyAutomated {
-					a.cfg.Jobs.PutJob(inst.ID, jobs.Job{
+					a.cfg.Jobs.PutJobIfNotExists(inst.ID, jobs.Job{
```
squaremo: I expected these to be the other way around -- that is, `PutJob` checks if the job already exists as it does now, and a variant that doesn't do the duplicate check for use if you're already in a job and you want to reschedule it -- simply because that would keep most invocations as they were.

PR #291, opened 2016-12-09
[WIP] Add support for DaemonSets as pod controllers

squaremo:
NB: this itself is not sufficient, because DaemonSets do not have support for rollout as deployments do. We may be able to fake it by deleting the pods in question and letting them be created; otherwise, we'll have to wait until there's a rollout thingy.

Most relevant seeming issue in k8s: https://github.com/kubernetes/kubernetes/issues/22543


squaremo:
Old and unworkable. I mean the PR ...

PR #293, opened 2016-12-09
Build a fluxctl image, same as fluxsvc and fluxd

paulbellamy:
A bit weird with how the release-bins are built, but...


docker/Dockerfile.fluxctl
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.3
```
squaremo: I don't think this needs any of the things in alpine, since it's a static binary. It's a static binary, right?

paulbellamy: yes, actually.

paulbellamy:
This was so you could mount the binary into another container but that is awkward on k8s, so closed.

Issue #258, opened 2016-11-29
Remove semaphore from Release


squaremo:
I don't think there's any point to it, given that the model is surely to have a pool of workers that run through jobs one each at a time, and no need for mutual exclusion except on release jobs themselves.

squaremo:
They* are running on a ticker as well, for some reason. They should run like this:

```
Is there a job to do?
  Yes: claim it
  No: wait for a bit, then go to the start
```

*There's actually only one, but there's no reason why there couldn't be more.

squaremo:
Fixed by #296

PR #296, opened 2016-12-12
Remove semaphore from releaser

squaremo:
The releaser used a channel as a mutex, allowing only one release to
happen at a time. When there was one fluxsvc and one platform in
operation, this made some sense.

However, we now have multiple platforms (one per instance), and want
to run more than one fluxsvc. Having a mutex guarding multiple
unrelated platforms is counter-productive; and anyway irrelevant if
there are multiple running services.

Regrades are still serialised in the platform, so there is still a
guard against running interacting regrades.

### DIFF
--- a/release/releaser.go
+++ b/release/releaser.go
@@ -24,10 +24,7 @@ const FluxServiceName = "fluxsvc"
 type Releaser struct {
 	instancer instance.Instancer
 	metrics   Metrics
-	semaphore chan struct{}
 }
-
-const maxSimultaneousReleases = 1
 
 type Metrics struct {
 	ReleaseDuration metrics.Histogram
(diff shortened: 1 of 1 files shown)


paulbellamy:
Fixes #258

release/releaser.go
```diff
@@ -24,7 +24,6 @@ const FluxServiceName = "fluxsvc"
 type Releaser struct {
 	instancer instance.Instancer
 	metrics   Metrics
-	semaphore chan struct{}
 }
 
 const maxSimultaneousReleases = 1
```
paulbellamy: Can nuke this as well.

squaremo: Good spot, ta

PR #303, opened 2016-12-12
Quell traces from kubeservice, and correctly interpret its output

### LINKED ISSUE #254: Lots of false kubeservice failures in logs
`kubeservice` is designed to find the service that corresponds to a given deployment; we use it to find the deployment that corresponds to a given service. Although this is the wrong way around (see #139), it does give the right answers.

However, each time it's used, it outputs tens of lines of error reports -- even in the course of a successful use. Very occasionally these are useful for debugging, when a deployment is expected to match but doesn't. Most of the time it just obscures whatever else is going on.

### LINKED ISSUE #301: If kubeservice finds more than one service matching, `FilesFor` fails
`kubeservice` names the service(s) matching a given deployment or RC file. In fluxsvc, it's used in the procedure `FilesFor` to find the deployment that "supplies" pods to a given service. This is done by checking each file to see if it gives the right answer, where the right answer is exactly the service given.

Sometimes there is more than one service that selects the pods from a deployment, and in these cases, `kubeservice` returns more than one service name, separated by newlines. But `FilesFor` treats this as a failure, since it doesn't match the expected answer. Oh no!

squaremo:
Fixed #301. Fixes #254.
See commits.

### DIFF
--- a/platform/kubernetes/files_for.go
+++ b/platform/kubernetes/files_for.go
@@ -53,9 +53,11 @@ func FilesFor(path, namespace, service string) (filenames []string, err error) {
 		if err := cmd.Run(); err != nil {
 			continue
 		}
-		out := strings.TrimSpace(stdout.String())
-		if out == tgt { // kubeservice output is "namespace/service", same as ServiceID
-			winners = append(winners, file)
+		for _, out := range strings.Split(strings.TrimSpace(stdout.String()), "\n") {
+			if out == tgt { // kubeservice output is "namespace/service", same as ServiceID
+				winners = append(winners, file)
+				break
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


platform/kubernetes/files_for.go
```diff
@@ -46,23 +46,18 @@ func FilesFor(path, namespace, service string) (filenames []string, err error) {
...
-			fmt.Fprintf(os.Stderr, stderr.String()+"\n")
-			fmt.Fprintf(os.Stderr, "kubeservice FAILED: %v\n", err)
 			continue
 		}
-		out := strings.TrimSpace(stdout.String())
-		fmt.Fprintf(os.Stderr, "kubeservice SUCCESS: %s -> %s\n", file, out)
-		if out == tgt { // kubeservice output is "namespace/service", same as ServiceID
-			winners = append(winners, file)
+		for _, out := range strings.Split(strings.TrimSpace(stdout.String()), "\n") {
+			if out == tgt { // kubeservice output is "namespace/service", same as ServiceID
```
paulbellamy: Won't this mean it still only returns a single service?

squaremo: We're trying to find the files that match a given service. For each file examined, all we care about is if the file (a deployment, say) matches _that_ service. `kubeservice` outputs all the services that a deployment matches. So, we check if any of those match the one given.

Issue #306, opened 2016-12-13
Releases of fluxd look like failures because they kill fluxd


squaremo:
In Slack: `fluxy/fluxd: Regrade due to "Release latest images to fluxy/fluxd": failed: unexpected EOF`

squaremo:
Fixed by #307.

PR #307, opened 2016-12-13
Treat fluxd as something to upgrade async

squaremo:
This helps avoid the "Unexpected EOF" failure messages we otherwise see.


squaremo:
Fixes #306.

Issue #297, opened 2016-12-12
Add alerting on our local fluxd being running and connected


paulbellamy:
## TODO:

- [ ] Expose the metrics in prom (https://github.com/weaveworks/flux/pull/308 and https://github.com/weaveworks/service-conf/pull/437)
- [ ] Set up an alert

PR #308, opened 2016-12-13
Add prom metrics on platforms (both fluxd and fluxsvc)

paulbellamy:
First part of #297

### DIFF
--- a/cmd/fluxsvc/main.go
+++ b/cmd/fluxsvc/main.go
@@ -107,6 +107,7 @@ func main() {
 			Name:      "history_duration_seconds",
 			Help:      "History method duration in seconds.",
 		}, []string{"service_spec", "success"})
+		serverMetrics.PlatformMetrics = platform.NewMetrics()
 		releaseMetrics.ReleaseDuration = prometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
 			Namespace: "flux",
 			Subsystem: "fluxsvc",
--- a/cmd/fluxsvc/main.go
+++ b/cmd/fluxsvc/main.go
@@ -88,30 +88,35 @@ func main() {
 			Subsystem: "fluxsvc",
 			Name:      "http_request_duration_seconds",
 			Help:      "HTTP request duration in seconds.",
+			Buckets:   stdprometheus.DefBuckets,
 		}, []string{"method", "status_code"})
 		serverMetrics.ListServicesDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
 			Namespace: "flux",
 			Subsystem: "fluxsvc",
 			Name:      "list_services_duration_seconds",
 			Help:      "ListServices method duration in seconds.",
+			Buckets:   stdprometheus.DefBuckets,
 		}, []string{"namespace", "success"})
 		serverMetrics.ListImagesDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
 			Namespace: "flux",
 			Subsystem: "fluxsvc",
 			Name:      "list_images_duration_seconds",
 			Help:      "ListImages method duration in seconds.",
+			Buckets:   stdprometheus.DefBuckets,
 		}, []string{"service_spec", "success"})
 		serverMetrics.HistoryDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
 			Namespace: "flux",
 			Subsystem: "fluxsvc",
 			Name:      "history_duration_seconds",
 			Help:      "History method duration in seconds.",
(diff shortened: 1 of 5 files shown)


platform/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+type Metrics struct {
+	AllServicesDuration  metrics.Histogram
+	SomeServicesDuration metrics.Histogram
+	RegradeDuration      metrics.Histogram
+	PingDuration         metrics.Histogram
+}
+
+func NewMetrics() Metrics {
+	return Metrics{
+		AllServicesDuration: prometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
```
squaremo: Do we definitely want summaries here? As I understand it, they cannot be aggregated when querying.

platform/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+			"success", fmt.Sprint(err == nil),
+		).Observe(time.Since(begin).Seconds())
+	}(time.Now())
+	return i.p.SomeServices(ids)
+}
+
+func (i *instrumentedPlatform) Regrade(spec []RegradeSpec) (err error) {
+	defer func(begin time.Time) {
+		i.m.SomeServicesDuration.With(
+			"specs", fmt.Sprint(len(spec)),
```
squaremo: I don't think we really care about this as a dimension for the time series.

platform/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+			"success", fmt.Sprint(err == nil),
+		).Observe(time.Since(begin).Seconds())
+	}(time.Now())
+	return i.p.AllServices(maybeNamespace, ignored)
+}
+
+func (i *instrumentedPlatform) SomeServices(ids []flux.ServiceID) (svcs []Service, err error) {
+	defer func(begin time.Time) {
+		i.m.SomeServicesDuration.With(
+			"ids", fmt.Sprint(len(ids)),
```
squaremo: I don't think we really care about this as a dimension for the time series (although granted, it may affect the duration).

platform/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+
+func Instrument(p Platform, m Metrics) Platform {
+	return &instrumentedPlatform{p, m}
+}
+
+func (i *instrumentedPlatform) AllServices(maybeNamespace string, ignored flux.ServiceIDSet) (svcs []Service, err error) {
+	defer func(begin time.Time) {
+		i.m.AllServicesDuration.With(
+			"namespace", maybeNamespace,
+			"ignored", fmt.Sprint(len(ignored)),
```
squaremo: I don't think we need to care about this as a dimension.

squaremo:
Oh I meant that to be Comment rather than unqualified approval .. oh well, see the comments ...

http/daemon.go
```diff
@@ -79,6 +89,24 @@ func (a *Daemon) connect() error {
...
+	// Instrument connection lifespan
+	connectedAt := time.Now()
+	disconnected := make(chan struct{})
+	defer close(disconnected)
+	go func() {
+		t := time.NewTicker(1 * time.Second)
+		for {
+			select {
+			case now := <-t.C:
+				a.setConnectionDuration(now.Sub(connectedAt).Seconds())
```
paulbellamy: good queestion. I assumed the ticker values would be less skewed... will check.

platform/metrics.go
paulbellamy: Oh yeah. should be histogram. will fix!

platform/metrics.go
paulbellamy: I was hoping it would help us find any sort of "accidentally quadratic" errors.

http/daemon.go
paulbellamy: [docs say](https://golang.org/pkg/time/#Ticker):
> It adjusts the intervals or drops ticks to make up for slow receivers.

platform/metrics.go
squaremo: This one doesn't affect the complexity in the same way as the others, so I think it can be removed.

cmd/fluxsvc/main.go
```diff
@@ -83,49 +83,62 @@ func main() {
 		helperDuration metrics.Histogram
 	)
 	{
-		httpDuration = prometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
+		httpDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
```
paulbellamy: Eugh, yeah you probably need to set up some buckets.

paulbellamy: done.

paulbellamy:
Discussed last changes in person with @squaremo

PR #311, opened 2016-12-16
Fair Scheduling

### LINKED ISSUE #298: Throttle running jobs per instance
Our scarce resources are, more or less, computer time and outgoing requests. These are consumed by ReleaseJobs. A crude way of sharing the resources out fairly among instances is to allow only a small number of jobs per instance to be running at a time (perhaps even just one).

paulbellamy:
Fixes #298 

Only run one job at a time per instance.

Caveat is that if there is a crash which leaves a job claimed, but not finished, no more jobs will happen. We should probably timeout/expire old jobs based on heartbeat! Maybe that plays into recoverable jobs? or we could just GC old jobs which were claimed but not heartbeated in a long time. @squaremo, I added a commit of GCing old jobs, wdyt?

### DIFF
--- a/jobs/db_store.go
+++ b/jobs/db_store.go
@@ -166,8 +166,8 @@ func (s *DatabaseStore) PutJob(inst flux.InstanceID, job Job) (JobID, error) {
 		if job.Key != "" {
 			var count int
 			err = s.conn.QueryRow(`
-				SELECT count(*) FROM jobs WHERE key = $1 AND finished_at IS NULL
-			`, job.Key).Scan(&count)
+				SELECT count(1) FROM jobs WHERE instance_id = $1 AND key = $2 AND finished_at IS NULL
+			`, string(inst), job.Key).Scan(&count)
 			if err != nil {
 				return errors.Wrap(err, "looking for existing job")
 			}
--- a/jobs/db_store.go
+++ b/jobs/db_store.go
@@ -405,8 +405,10 @@ func (s *DatabaseStore) GC() error {
 
 		if _, err := s.conn.Exec(`
 			DELETE FROM jobs
-						WHERE finished_at IS NOT NULL
-							AND submitted_at < $1
+						WHERE (finished_at IS NOT NULL AND submitted_at < $1)
+						   OR (claimed_at IS NOT NULL
+							 AND claimed_at < $1
+							 AND (heartbeat_at IS NULL OR heartbeat_at < $1))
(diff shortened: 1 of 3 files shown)


squaremo:
> We should probably timeout/expire old jobs based on heartbeat! Maybe that plays into recoverable jobs?

Exactly -- job recovery would look for anything that's missed heartbeats, and unclaimify it.

> I added a commit of GCing old jobs, wdyt?

That would mean jobs that are claimed but no longer running will get cleared up eventually, which is good (as long as "eventually" is fairly long). If someone leaves `fluxctl release ...` open on such a job, they'll eventually get a 404 or whatever (it's probably a 500, isn't it), which is not ideal, but I can live with it for now.

jobs/db_store_test.go
```diff
@@ -339,3 +350,155 @@ func TestDatabaseStoreScheduledJobs(t *testing.T) {
...
+	if !job1.Done || !job1.Success {
+		t.Errorf("expected job to have been marked as done")
+	}
+
+	// Take another
+	// - It should be instance1's next job
+	job2, err := db.NextJob(nil)
+	bailIfErr(t, err)
+	// - It should be the highest priority
+	if job2.ID != job2ID {
```
squaremo: The comment above doesn't seem to match what's happening here.

jobs/db_store.go
```diff
@@ -209,15 +209,34 @@ func (s *DatabaseStore) NextJob(queues []string) (Job, error) {
...
+			-- Don't make jobs available until after they are scheduled
+			AND scheduled_at <= $1
+
+			-- Only one job at a time per instance
+			AND instance_id NOT IN (
+				SELECT instance_id
+				FROM jobs
+				WHERE claimed_at IS NOT NULL
+				AND finished_at IS NULL
+				GROUP BY instance_id
```
paulbellamy: Trying to get it working on pg and ql (which has broken joins), was such a pain. I wanted to get a `running_jobs_per_instance` count, so we could later on do set a `running_job_per_instnace < instance.max_concurrency`, and manage it that way, but, left that for later.

paulbellamy: Trying to get it working on pg and ql (which has broken joins), was such a pain. I wanted to get a `running_jobs_per_instance` count, so we could later on do set a `running_job_per_instnace < instance.max_concurrency`, and manage it that way, but, left that for later.

PR #314, opened 2016-12-19
Regrade -> Release

paulbellamy:
@squaremo wdyt?

I picked release because it matches `fluxctl release`. Deploy might be
more meaningful, but this makes it consistent everywhere, and Release is
still more meaningful than regrade.

---

There are a couple (tangentially related) slack message changes in the first commit, also, which change:
```
default/users: Regrade due to "Release latest images to default/users": done
default/users: Regrade due to "Apply latest config to default/users": done
```
to
```
default/users: Release latest images to default/users. done
default/users: Apply latest config to default/users. done
```
Could take the de-dupe even further, with the service name, but it depends how we want to consolidate output for multiple services, and image versions, etc.

### DIFF
--- a/platform/kubernetes/kubernetes.go
+++ b/platform/kubernetes/kubernetes.go
@@ -36,10 +36,10 @@ type apiObject struct {
 	} `yaml:"metadata"`
 }
 
-type releaseExecFunc func(*Cluster, log.Logger) error
+type applyExecFunc func(*Cluster, log.Logger) error
 
-type release struct {
-	exec    releaseExecFunc
+type apply struct {
+	exec    applyExecFunc
 	summary string
 }
 
@@ -171,7 +171,7 @@ func (c *Cluster) AllServices(namespace string, ignore flux.ServiceIDSet) (res [
 
 func (c *Cluster) makeService(ns string, service *api.Service, controllers []podController) platform.Service {
 	id := flux.MakeServiceID(ns, service.Name)
-	status, _ := c.status.getReleaseProgress(id)
+	status, _ := c.status.getApplyProgress(id)
 	return platform.Service{
 		ID:         id,
 		IP:         service.Spec.ClusterIP,
@@ -353,14 +353,14 @@ func (c *Cluster) Apply(defs []platform.ServiceDefinition) error {
 					continue
 				}
 
-				plan, err := controller.newRelease(newDef)
+				plan, err := controller.newApply(newDef)
 				if err != nil {
 					applyErr[def.ServiceID] = errors.Wrap(err, "creating release")
 					continue
 				}
 
-				c.status.startRelease(def.ServiceID, plan)
-				defer c.status.endRelease(def.ServiceID)
+				c.status.startApply(def.ServiceID, plan)
+				defer c.status.endApply(def.ServiceID)
 
 				logger := log.NewContext(c.logger).With("method", "Apply", "namespace", namespace, "service", serviceName)
 				if err = plan.exec(c, logger); err != nil {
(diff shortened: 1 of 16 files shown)


paulbellamy:
Big downside of this is it changes the API, which breaks fluxctl clients. which is starting to become more of an issue.

squaremo:
I agree with taking "Regrade" out of the user interface (including notifications and event log messages). It's not really part of user vocabulary -- why should they need to know the difference between a release and a regrade; it's just one action so far as they are concerned.

I am OK with keeping the term "regrade" in the code and API -- it is no less accurate than "release" for the purpose to which it's put, and there is a distinction we can draw between "regrade" (tell the platform to apply new definitions) and "release" (calculate what needs to change, update config, do a regrade, and send notifications). If we're deciding that "regrade" is in itself confusing, I'd rather change it to something like "ApplyDefinitions", i.e., to be more precise.

With that in mind I don't think it's worth changing the API just to switch terms. I have some sympathy if you would still like to do so -- but I think we need to address compatibility issues, if so.

paulbellamy:
> I am OK with keeping the term "regrade" in the code and API

To minimize backwards compatibility issues, sure. It's just ongoing tech debt/overhead.

> it is no less accurate than "release" for the purpose to which it's put

Disagree.

> If we're deciding that "regrade" is in itself confusing

Yes. It doesn't mean anything. it could be replaced with "verb"

> I'd rather change it to something like "ApplyDefinitions"

I'd agree. Based on your definition of what it does, then I'd say `Apply`, which leads to a natural division of a `release` into several sections:
```
Release:
  1) Gather Materials
    a) Clone the repo
    b) Check registry for new tags
  2) Update Manifests
    a) Find manifest files
    b) Update them
    c) Commit & Push
  3) Apply New Manifests
    a) Rolling-update to Kubernetes
```

squaremo:
> Yes. It doesn't mean anything. it could be replaced with "verb"

That's not really true at all -- "regrade" means something: to grade (specify the version of) again. But I'd accept the argument that it's obscure -- you don't want to send people to a dictionary when they're reading code.

> Based on your definition of what it does, then I'd say Apply, which leads to a natural division of a release into several sections

`Apply` works for me (though it _does_ suffer a little from being non-specific :). Using `Release` to refer to a subsection of release is just making a mistake of the same magnitude, in the other direction.

paulbellamy:
> Using Release to refer to a subsection of release is just making a mistake of the same magnitude

Agreed, that's why in the sub-section list above I tried to avoid using it again. Will update code to match. :+1:

paulbellamy:
@squaremo ok, I think I've got it all sorted. Could you give it a read to make sure it makes sense, and give it a try to make sure I've not broken anything?

platform/kubernetes/release.go
```diff
@@ -16,13 +16,13 @@ import (
 	"github.com/weaveworks/flux/platform"
 )
 
-func (c podController) newRegrade(newDefinition *apiObject) (*regrade, error) {
+func (c podController) newRelease(newDefinition *apiObject) (*release, error) {
```
squaremo: Should these be `apply` too?

release/releaser.go
```diff
@@ -215,16 +215,16 @@ func (r *Releaser) releaseImages(method, msg string, inst *instance.Instance, ki
 	}
 
 	stage.ObserveDuration()
-	stage = metrics.NewTimer(base.With("stage", "calculate_regrades"))
+	stage = metrics.NewTimer(base.With("stage", "calculate_releases"))
```
squaremo: `apply`?

platform/kubernetes/kubernetes.go
```diff
@@ -36,10 +36,10 @@ type apiObject struct {
 	} `yaml:"metadata"`
 }
 
-type regradeExecFunc func(*Cluster, log.Logger) error
+type releaseExecFunc func(*Cluster, log.Logger) error
 
-type regrade struct {
-	exec    regradeExecFunc
+type release struct {
```
squaremo: Still some `release`s in here, should they be `apply`s?

release/releaser.go
```diff
@@ -612,31 +612,31 @@ func (r *Releaser) releaseActionRegradeServices(services []flux.ServiceID, msg s
...
+			// Report individual service release results.
 			for _, service := range services {
 				namespace, serviceName := service.Components()
 				switch serviceName {
 				case FluxServiceName, FluxDaemonName:
 					continue
 				default:
 					if err := results[service]; err == nil { // no entry = nil error
-						rc.Instance.LogEvent(namespace, serviceName, "Regrade due to "+cause+": done")
+						rc.Instance.LogEvent(namespace, serviceName, msg+". done")
```
squaremo: Will these definitely start with `"Release"`, as the slack notifications expect?

squaremo:
Looks right, thanks for the extra go-around Paul.

I tried it locally and Slack messages appear to work when releasing new images, but not when applying existing config (`--no-update`). I think for now we could just change the message so that it's caught by the regexp.

paulbellamy:
Rebased for nats changes.

PR #322, opened 2016-12-20
Instrument image registry operations

squaremo:
As a prelude to introducing a cache, or otherwise improving on the
fetching of image metadata, this commit adds some metrics to the
registry client.

We're interested in

 * how long fetching image metadata takes overall
 * how many requests are made, and of what kind (fetching the list of
   tags, and fetching each tags' metadata)

firstly to see if they are really problematic, and secondarily since
the main improvements will likely come from lowering these.

### DIFF
--- a/platform/metrics.go
+++ b/platform/metrics.go
@@ -9,11 +9,7 @@ import (
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
 
 	"github.com/weaveworks/flux"
-)
-
-const (
-	LabelMethod  = "method"
-	LabelSuccess = "success"
+	fluxmetrics "github.com/weaveworks/flux/metrics"
 )
 
 type Metrics struct {
@@ -28,7 +24,7 @@ func NewMetrics() Metrics {
 			Name:      "request_duration_seconds",
 			Help:      "Request duration in seconds.",
 			Buckets:   stdprometheus.DefBuckets,
-		}, []string{LabelMethod, LabelSuccess}),
+		}, []string{fluxmetrics.LabelMethod, fluxmetrics.LabelSuccess}),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


registry/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+type Metrics struct {
+	// Latency of image fetch, that is getting *all* information about
+	// an image
+	FetchDuration metrics.Histogram
+	// Counts of particular kinds of request
+	RequestCount metrics.Counter
+}
+
+const (
+	LabelInstanceID  = "instanceID"
```
paulbellamy: I believe in other places I've used `snake_case` for `instance_id`

registry/registry.go
```diff
@@ -107,6 +116,11 @@ func (c *Client) GetRepository(repository string) ([]flux.ImageDescription, erro
 	}
 
 	tags, err := client.Tags(hostlessImageName)
+	c.Metrics.RequestCount.With(
+		LabelRepository, repository,
+		LabelRequestKind, RequestKindTags,
+		LabelSuccess, strconv.FormatBool(err == nil),
+	).Add(1)
```
paulbellamy: Maybe request duration histograms? then we can get count as well, I believe? Would let us break down the FetchDuration above into "tags" and "metadata" requests.

platform/metrics.go
```diff
@@ -26,28 +27,28 @@ func NewMetrics() Metrics {
 			Name:      "all_services_duration_seconds",
 			Help:      "AllServices method duration in seconds.",
 			Buckets:   stdprometheus.DefBuckets,
-		}, []string{"namespace", "success"}),
+		}, []string{"namespace", fluxmetrics.LabelSuccess}),
```
paulbellamy: I'd rather either, all the labels live in fluxmetrics, or none. The half-half thing is weird.

PR #323, opened 2016-12-20
Try a 'fluxctl status' command

### LINKED ISSUE #244: Status monitoring
It would be useful to have a way to confirm flux is running properly. I would propose:

* A Status command, contacting the backend and showing everything is OK.
* Prometheus metrics (and grafana graphs).

paulbellamy:
Fixes #244

@squaremo, wdyt?

I'm not really convinced on the distinction between this and `IsConnected`, but this format gives us more room for the status of various bits.. It's also a bit unclear on the distinction between this and `GetConfig`. Suggestions?

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -7,6 +7,7 @@ import (
 )
 
 type ClientService interface {
+	Status(inst flux.InstanceID) (flux.Status, error)
 	ListServices(inst flux.InstanceID, namespace string) ([]flux.ServiceStatus, error)
 	ListImages(flux.InstanceID, flux.ServiceSpec) ([]flux.ImageStatus, error)
 	PostRelease(flux.InstanceID, jobs.ReleaseJobParams) (jobs.JobID, error)
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -68,6 +68,7 @@ func (opts *rootOpts) Command() *cobra.Command {
 
 	cmd.AddCommand(
 		newVersionCommand(),
+		newStatus(opts).Command(),
 		newServiceShow(svcopts).Command(),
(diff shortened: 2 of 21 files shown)


squaremo:
Could it be implemented just by calling `GetConfig` and `IsConnected` from the client? I guess that would mean it's more difficult to add information in.

paulbellamy:
Hmm, I added it as a separate method, so we could try other things to verify it was set up correctly, like cloning the repo. I've added that into this PR.

So if you have misconfigured your git public key you'll get:
```
$ fluxctl status
fluxd:
  connected: true
git:
  configured: true
  error: 'git clone: exit status 128'
```

paulbellamy:
Rebased

paulbellamy:
As raised in: https://github.com/weaveworks/flux/issues/337 it would be good to output the *reason* for the git failure.

Multiline strings get rendered well by the go yaml library we're using, e.g.
```Go
yaml.Marshal(map[string]string{
	"field1": "1234",
	"field2": "abcd\nefgh\nijkl",
})
```
renders...
```yaml
field1: "1234"
field2: |-
  abcd
  efgh
  ijkl
```
So we could put out the git stderr into the status response, and it should render well.

paulbellamy:
Have added git clone stderr into the `status` reponse:
```
$ fluxctl status
fluxd:
  connected: true
  version: HEAD-00fa07e-WIP
git:
  configured: true
  error: |
    Cloning into '/tmp/flux-gitclone819220243/repo'...
    Warning: Permanently added 'github.com,192.30.253.113' (RSA) to the list of known hosts.
    Permission denied (publickey).
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights
    and the repository exists.
```

PR #324, opened 2016-12-20
Add fluxsvc db query metrics

paulbellamy:


squaremo:
If we put all the operations in their own metric, does this make it impossible to look at say, the mean over all operations?

paulbellamy:
D'oh. Right. Not impossible, but a pain. Will fix that. Made the same
mistake on platform.Metrics

PR #326, opened 2016-12-21
WIP -- Cut down the number of image registry requests for automated services

paulbellamy:
@squaremo, approach is as discussed on slack.

I added a new AutomatedInstanceJob, to check all images for an instance.

Also, reworked the job.Handler, to return a list of "follow-up jobs" which makes testing easier, though I'm not 100% sure on that change.

## TODO

- [ ] Do release W image to x, y, z as a single release job.
- [ ] Consider making the main automator task of scheduling AutomatedInstanceJob a job, so it only scans the config db from a single fluxsvc (probably a separate thing to do).


squaremo:
Side note: it occurs to me that a major saving comes from looking at the services that are defined first, rather than the services running, since we anyway discard any of the latter that aren't in the former, so their image metadata is useless.

paulbellamy:
It also lets us automate (and release) services which are not *already* running.

paulbellamy:
I've merged the regrade -> release PR here, as it touched a lot of the same code.

This has gotten quite large, so I'm debating how to break it up...

paulbellamy:
Broken up, and superceded by various PRs.

PR #350, opened 2017-01-10
READY - Push binaries to GH releases upon a release

### LINKED ISSUE #248: Automated, versioned releases
It would be helpful for users if there was a stable, published URL where they could get the latest binary; and it would be helpful for us if those included a version we could ask people for, when troubleshooting for example.

There's a target in the makefile which builds "release binaries" (just fluxctl binaries for `{linux, darwin} x {amd64}`). I propose, then:

 1. Adding a version command which reports the baked-in version of the binary
 2. Baking in a version, at least when building release binaries
 3. Automating releases so that a stable URL is maintained (but not necessarily driving releases from CI)

philwinder:
Requires a: `GITHUB_TOKEN` env var, which is a GH personal access token, to be added to the circle environment vars.

Creating a new release will prompt circle to upload binaries to release.
Added release docs.
Added changelog.
Added script to upload files to GH releases.
Alterted circle.yml to build on tagged releases

Fixes #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,10 @@ $ gvt fetch <dependency>
 recursively; so, it should be sufficient to just add packages you
 import.
 
+### Releasing
+
+See the [release docs](./docs/releasing.md) for instructions about how to release a version of flux.
+
 ### Contribution
 
 Flux follows a typical PR workflow.
--- a/bin/upload-binaries
+++ b/bin/upload-binaries
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+## Uploads the built binaries to the GH release
+## Requires github-release
+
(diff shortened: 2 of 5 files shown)


philwinder:
@paulbellamy Who's GH personal access token do we want to use. Yours?

paulbellamy:
Do we have to use a personal one? is there a "repo" or "org"-level one we can use?

philwinder:
Doesn't appear to be possible. Has to be personal.

http://stackoverflow.com/questions/31159275/how-do-i-generate-the-github-oauth-token-for-organization-accounts

paulbellamy:
Added.

PR #355, opened 2017-01-10
Move api endpoints to /api/flux/...

### LINKED ISSUE #217: Move api under /api/flux prefix
Will make the service play better with authfe.

And by "play better" I mean not expose prometheus metrics to the world.

philwinder:
Fixes #217
@paulbellamy Is this what you mean?

However, this makes the api endpoints rather verbose.

### DIFF
--- a/http/transport.go
+++ b/http/transport.go
@@ -29,6 +29,7 @@ import (
 
 func NewRouter() *mux.Router {
 	r := mux.NewRouter()
+	r = r.PathPrefix("/api/flux").Subrouter()
 	r.NewRoute().Name("ListServices").Methods("GET").Path("/v3/services").Queries("namespace", "{namespace}") // optional namespace!
 	r.NewRoute().Name("ListImages").Methods("GET").Path("/v3/images").Queries("service", "{service}")
 	r.NewRoute().Name("PostRelease").Methods("POST").Path("/v4/release").Queries("service", "{service}", "image", "{image}", "kind", "{kind}")


paulbellamy:
it is exactly, there'll need to be a matching change in authfe, unfortunately, which makes deployment a bit tricky.

paulbellamy:
Oh will need updating for status stuff merged. Also, is it possible to do the url changes in one place at all? instead of in each route?

philwinder:
Prefix URL now set in one place.

http/transport.go
```diff
@@ -27,8 +27,11 @@ import (
 	"github.com/weaveworks/flux/platform/rpc"
 )
 
+const URL_PREFIX = "/api/flux"
```
paulbellamy: eh, only used in one place, just inline it.

http/transport.go
```diff
@@ -27,8 +27,11 @@ import (
...
+
 func NewRouter() *mux.Router {
 	r := mux.NewRouter()
+	r = r.PathPrefix(URL_PREFIX).Subrouter()
```
paulbellamy: With a subrouter, check that the router lookup route-by-name thing we do in the client still works.

philwinder: I tested manually and it works. But I can't find that in the code. Where is it?

paulbellamy: https://github.com/weaveworks/flux/blob/master/http/transport.go#L759

philwinder: Yeah it works fine.

PR #372, opened 2017-01-13
Rate limit registry requests

paulbellamy:


registry/backoff.go
```diff
@@ -0,0 +1,108 @@
...
+func HostBackoffRoundTripper(r http.RoundTripper, maxBackoff time.Duration) http.RoundTripper {
+	return &hostBackoffRoundTripper{
+		roundTripper: r,
+		max:          maxBackoff,
+		backoffs:     map[string]*backoff{},
+	}
+}
+
+func (c *hostBackoffRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	key := backoffKey{host: r.URL.Host}
```
philwinder: `backoffKey` doesn't appear to be included anywhere.

registry/backoff.go
```diff
@@ -0,0 +1,108 @@
...
+	if user, pass, ok := r.BasicAuth(); ok {
+		key.user = user
+		key.pass = pass
+	}
+	c.Lock()
```
philwinder: Can you do that? Where is that lock coming from? The RoundTripper?

What are you locking here? The backoff map?

Might be worth using an external read/write mutex, so if it is used elsewhere (either in your code now, or in the future) it makes it a bit more obvious that the map read/write is not thread safe.

registry/backoff.go
```diff
@@ -0,0 +1,108 @@
...
+	b, ok := c.backoffs[key]
+	if !ok {
+		b = &backoff{max: c.max}
+		c.backoffs[key] = b
```
philwinder: I'm not sure why you need to store the backoffs.

The simpler and automatically thread safe option, would be to create a new backoff object for each call. Then you don't need any of this map or locking stuff.

Why do you want to store it? And what benefit do you get from reusing it? Does it need to be that complex?

registry/backoff.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"errors"
+	"net/http"
+	"sync"
+	"time"
+)
+
+var (
+	// For testing
+	now = time.Now
```
philwinder: This is actually used in the code below. I assume you're going to do something about that.

registry/backoff.go
```diff
@@ -0,0 +1,108 @@
...
+	// Ratio of success/failure on scale from [0, 1] where 0 means all success.
+	ratio float64
+	// last time a request was started
+	lastStarted time.Time
+	sync.Mutex
+}
+
+// Fail should be called each time a request succeeds.
+func (b *backoff) Success() {
+	b.update(0.0)
```
philwinder: What/where is the update method?

registry/backoff.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Failure should be called each time a request fails.
+func (b *backoff) Failure() {
+	b.update(1.0)
+}
+
+// finish is a helper for success and fail.
```
philwinder: Doesn't look like finish is ever used. And I'm unsure of what it's meant to do.

registry/backoff.go
```diff
@@ -0,0 +1,93 @@
...
+	initial time.Duration
+	max     time.Duration
+
+	current time.Duration
+	sync.RWMutex
+}
+
+// Failure should be called each time a request fails.
+func (b *backoff) Failure() {
+	b.Lock()
```
philwinder: You can remove the locking from b too. It's only accessed in the function above and that's sequential.

registry/backoff.go
```diff
@@ -0,0 +1,93 @@
...
+	if b.current == 0 {
+		b.current = b.initial
+	} else if b.current > b.max {
+		b.current = b.max
+	}
+}
+
+// Wait how long to sleep before *actually* starting the request.
+func (b *backoff) Wait() time.Duration {
+	b.RLock()
```
philwinder: Same

registry/backoff.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// backoff calculates an exponential backoff. This is used to
+// calculate wait times for future requests.
+type backoff struct {
+	initial time.Duration
+	max     time.Duration
+
+	current time.Duration
+	sync.RWMutex
```
philwinder: Need for locking?

registry/backoff.go
```diff
@@ -0,0 +1,93 @@
...
+		resp, err := c.roundTripper.RoundTrip(r)
+		switch {
+		case err != nil && strings.Contains(err.Error(), "Too Many Requests (HAP429)."):
+			// Catch the terrible dockerregistry error here. Eugh. :(
+			fallthrough
+		case resp != nil && resp.StatusCode == http.StatusTooManyRequests:
+			// Request rate-limited, backoff and retry.
+			b.Failure()
+			// Wait until the next time we are allowed to make a request
+			c.clock.Sleep(b.Wait())
```
philwinder: Since i don't think there's any need to lock the backoff, I don't think we need a fancy sleep?

registry/backoff.go
paulbellamy: belt and braces in case it were used from multiple threads. will remove.

registry/backoff.go
paulbellamy: Probably not, but makes testing a bit easier.

PR #377, opened 2017-01-17
Refactor registry

philwinder:


registry/remote_factory.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Credentials to a (Docker) registry.
+type Credentials struct {
+	m map[string]creds
+}
+
+type RemoteClientFactory interface {
+	Create(id Image) (Remote, error)
+}
```
paulbellamy: I like what this is doing, but I think the naming and organization needs a bit more work.

paulbellamy: Maybe there are 2 different "things" which the main registry takes:

```
type HostResolver interface {
  Resolve(host string) Remote
}
type CredentialSource interface { // Credentialer? /shrug
  CredentialsFor(host string) // This is almost what Credentials already implements
}
// Then the main registry does something like this?
host := image.Host
remote := hostResolver.Resolve(host)
creds := credentials.CredentialsFor(host)
remote.GetRepository(creds, image.Repository)
```
Just trying some ideas...

registry/remote_test.go
```diff
@@ -0,0 +1,100 @@
...
+	r := remote{
+		cancel: func() { didCancel = true },
+	}
+	r.Cancel()
+	if !didCancel {
+		t.Fatal("Expected it to call the cancel func")
+	}
+}
+
+func TestRemoteClient_ErrorsForCoverage(t *testing.T) {
```
paulbellamy: ErrorsForCoverage isn't a great name. What failure scenarios are you testing?

registry/remote_test.go
```diff
@@ -0,0 +1,100 @@
...
+	if err == nil {
+		t.Fatal("Expected error")
+	}
+	_, err = c.Manifest(img)
+	if err == nil {
+		t.Fatal("Expected error")
+	}
+}
+
+func TestRemoteClient_TestNew(t *testing.T) {
```
paulbellamy: maybe some comments on what the properties you are trying to assert are...

registry/remote_factory.go
```diff
@@ -0,0 +1,131 @@
...
+			}),
+			Jar:     jar,
+			Timeout: requestTimeout,
+		},
+		Logf: dockerregistry.Quiet,
+	}
+	return
+}
+
+// --- Credentials
```
paulbellamy: Maybe a separate file for these?

registry/image.go
```diff
@@ -0,0 +1,128 @@
...
+	HostOrgRepo() string
+	OrgRepo() string
+	CreatedAt() *time.Time
+	Clone() Image
+}
+
+type image struct {
+	host, org, repo, tag string
+	createdAt            *time.Time `json:",omitempty"`
+}
```
philwinder: Interfaces are better for testing and easier to refactor.

An image belongs with a registry in the domain. It doesn't belong in the root. And also, you'd get circular dependencies (flux.Service->registry.x->flux.Image)

philwinder: Although I could be persuaded, as it's almost a primitive that can be mocked by simply instantiating. But I would argue that you might want to return something which it doesn't return in a normal situation, like an invalid repo.

registry/image.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (i *image) Components() (host, org, repo, tag string) {
+	return host, org, repo, tag
+}
+func (i *image) Host() string {
+	return i.host
+}
+func (i *image) Org() string {
+	return i.org
+}
```
philwinder: 😬  Like most things it's called something different all over. No real consensus, but the first hit was: https://access.redhat.com/documentation/en/red-hat-enterprise-linux-atomic-host/7/paged/recommended-practices-for-container-development/chapter-4-image-naming-conventions
Which uses <user>/<repo>.

https://docs.docker.com/engine/reference/commandline/tag/
This page uses the word "repository" to refer to the first part. And `Image` to refer to both parts.

https://docs.docker.com/engine/getstarted/step_six/
This page calls the first part a `Namespace`.

So, wtf. I agree with you in that namespace seems like a good name.

registry/image.go
```diff
@@ -0,0 +1,128 @@
...
+func (i *image) Repo() string {
+	return i.repo
+}
```
philwinder: Because of the confusion (see above), I don't think that repo should be /../../. I'll have a method more explicitly named to do that.

registry/image.go
```diff
@@ -0,0 +1,128 @@
...
+func (i *image) Tag() string {
+	return i.tag
+}
```
philwinder: Docker calls it a tag? https://docs.docker.com/engine/reference/commandline/tag/

registry/image.go
```diff
@@ -0,0 +1,128 @@
...
+func (i *image) CreatedAt() *time.Time {
+	return i.createdAt
+}
+
+// Fully qualified name
+func (i *image) FQN() string {
+	return fmt.Sprintf("%s/%s/%s:%s", i.host, i.org, i.repo, i.tag)
+}
```
philwinder: Yes

instance/multi.go
```diff
@@ -43,11 +43,16 @@ func (m *MultitenantInstancer) Get(instanceID flux.InstanceID) (*Instance, error
...
 	}
-	regClient := registry.NewClient(
-		creds,
-		log.NewContext(instanceLogger).With("component", "registry"),
-		m.RegistryMetrics.WithInstanceID(instanceID),
-	)
+	var reg registry.Registry
+	{
+		reg = registry.NewRegistry(
+			registry.NewRemoteClientFactory(creds),
```
philwinder: We need it because we're creating a new remote on every request.

registry/registry.go
```diff
@@ -2,68 +2,70 @@
...
-	dockerHubLibrary = "library"
-
 	requestTimeout = 10 * time.Second
 )
 
-type creds struct {
-	username, password string
+// Old registry interface. Used as an adaptor to the new class
+type Client interface {
+	GetRepository(repository string) ([]flux.ImageDescription, error)
```
philwinder: For the moment, yes. This is an adapter between the old code and the new one. We can remove it shortly.

registry/registry.go
```diff
@@ -2,68 +2,70 @@
...
-func NewClient(c Credentials, l log.Logger, m Metrics) Client {
-	return &client{
-		Credentials: c,
-		Logger:      l,
-		Metrics:     m,
-	}
+// New registry interface
+type Registry interface {
+	GetRepository(repository Image) ([]Image, error)
+	GetImage(repository Image) (Image, error)
```
philwinder: 😬  Rather not string. No constraints on a string, so you'd get people using `host/namespace/repo` or `namespace/repo` or `namespace/repo:tag`. Better to be explicit with a type.

And although rather redundant (all the info is in an image), I like the naming of Repository. It reads nicer.

registry/remote.go
```diff
@@ -0,0 +1,64 @@
...
+	Manifest(img Image) (Image, error)
+	Cancel()
+}
+
+type remote struct {
+	client dockerRegistryInterface
+	cancel context.CancelFunc
+}
+
+func newRemote(client *dockerregistry.Registry, cancel context.CancelFunc) Remote {
```
philwinder: Sorry, I don't understand this. cancel is called by code using the Remote interface. It cancels any current operations. So the Tags method doesn't need to know any cancel func because it never needs to call cancel itself.

registry/remote.go
```diff
@@ -0,0 +1,64 @@
...
+		client: client,
+		cancel: cancel,
+	}
+}
+
+func (rc *remote) Tags(id Image) (_ []string, err error) {
+	return rc.client.Tags(id.OrgRepo())
+}
+
+func (rc *remote) Manifest(img Image) (Image, error) {
```
philwinder: Its the remote's responsibility to convert the raw Manifest into a domain object (an image).

If you want more distinction, then we could do the same as above? Change it to:

`+func (rc *remote) Manifest(repo Repository, ref string) (Image, error)`

registry/image.go
paulbellamy: > Interfaces are better for testing and easier to refactor.

Only when the thing might have different implementations, or depend on some external resource. Image does neither of those, it's just a set of pure functions.

paulbellamy: > (flux.Service->registry.x->flux.Image)

Does anything in root package depend on registry? That is surprising.

registry/image.go
paulbellamy: Eugh.. docker (as usual) can't make up their mind on what to call it. as long as *we* are internally consistent, whatever is fine.

instance/multi.go
paulbellamy: Right. I was more objecting to the naming. (And if there is any way we can avoid the factory altogether)

registry/registry.go
paulbellamy: Right, can't we just do that here?

registry/image.go
philwinder: Consider myself persuaded.

registry/image.go
```diff
@@ -0,0 +1,128 @@
...
+	img := i.Clone().(*image)
+	img.tag = t
+	return img
+}
+
+func (i *image) WithCreatedAt(c *time.Time) Image {
+	img := i.Clone().(*image)
+	img.createdAt = c
+	return img
+}
```
philwinder: Ok, but this will break encapsulation rules. E.g. in the old version it was impossible to have an image without a name. Now it is possible because users have access to `img.Name`.

philwinder: And doesn't explicitly remind you that you need to clone. E.g. I just got a race in one of my commits because I forgot to copy.

paulbellamy: > race

eh? we don't really need to pass the image id around as `*Image` do we?

philwinder: img was passed in.
In a goroutine I accidentally reused the img object by:
`img.Tag = "someNewTag"`
Hence, the goroutines were all editing the same object at the same time. I.e. you need to copy for each go routine. It's more obvious when you have a `Copy()` or `Clone()` like method to call, that's all.

registry/registry.go
```diff
@@ -74,79 +76,19 @@ func (f roundtripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
-		host = parts[0]
-		org = parts[1]
-		image = parts[2]
-	default:
-		return nil, fmt.Errorf(`expected image name as either "<host>/<org>/<image>", "<org>/<image>", or "<image>"`)
+func (c *registry) GetRepository(img Image) (_ []Image, err error) {
+	var remote Remote
+	{
+		remote, err = c.factory.Create(img)
+		remote = NewRemoteMonitoringMiddleware(c.Metrics)(remote)
```
philwinder: A simpler way? This is the bestest way evaa! Middlewares FTW. It means that the remote doesn't need to know anything about monitoring. Decoupled and cohesive.

paulbellamy: I meant the wiring up and factory stuff. the monitoring middleware is :fire:

registry/remote_factory.go
philwinder: Ok, so above you've extracted out the creds. That's cool. But you've effectively just renamed the factory method to resolve. Since you have an aversion to the word factory, I will blast all mentions off the face of the codebase. :-)

paulbellamy: sure

instance/instance.go
```diff
@@ -106,22 +106,47 @@ func (h *Instance) CollectAvailableImages(services []platform.Service) (ImageMap
...
+			ID:        flux.ParseImageID(im.String()),
+			CreatedAt: im.CreatedAt,
+		}
+	}
+	return
 }
 
-// Create an image map containing exact images. At present this
-// assumes they exist; but it may in the future be made to verify so.
+// CreateFor an image map containing exact images.
```
paulbellamy: Seems like this comment is a bit garbled?

instance/instance.go
```diff
@@ -140,20 +165,17 @@ func (h *Instance) ExactImages(images []flux.ImageID) (ImageMap, error) {
 
 // Checks whether the given image exists in the repository.
 // Return true if exist, false otherwise
-func (h *Instance) imageExists(image flux.ImageID) (exists bool, err error) {
-	// Get a list of images
-	images, err := h.registry.GetRepository(image.Repository())
+func (h *Instance) imageExists(image flux.ImageID) (bool, error) {
+	img, err := registry.ParseImage(string(image), nil)
```
paulbellamy: Why do we need to parse this here? We got passed in an `ImageID` right?

registry/cache.go
```diff
@@ -14,28 +14,32 @@ import (
...
-func NewCache(b backend, creds Credentials, cache MemcacheClient, expiry time.Duration, logger log.Logger) *Cache {
-	return &Cache{
-		backend: b,
-		creds:   creds,
-		expiry:  expiry,
-		Client:  cache,
-		logger:  logger,
+type CachedDockerRegistry func(dockerRegistryInterface) dockerRegistryInterface
+
+func NewCache(creds Credentials, cache MemcacheClient, expiry time.Duration, logger log.Logger) CachedDockerRegistry {
```
paulbellamy: We always call `NewCache(...)(client)`, so let's take `client` as a regular argument, and make it:
```
func NewCache(next dockerRegistryInterface, creds Credentials, expiry time.Duration, Client MemcacheClient, logger log.Logger) dockerRegistryInterface
```

registry/cache_test.go
```diff
@@ -56,18 +56,19 @@ func TestCache(t *testing.T) {
...
 
 	manifestCalled := 0
-	c := NewCache(
-		&MockBackend{
-			manifest: func(repo, ref string) ([]schema1.History, error) {
-				manifestCalled++
-				return []schema1.History{{`{"test":"json"}`}}, nil
-			},
+
+	mock := &MockBackend{
```
paulbellamy: you mean `MockDockerRegistry`?

registry/credentials.go
```diff
@@ -0,0 +1,58 @@
...
+		hosts = append(hosts, host)
+	}
+	return hosts
+}
+
+type roundtripperFunc func(*http.Request) (*http.Response, error)
+
+func (f roundtripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return f(r)
+}
```
paulbellamy: You probably didn't mean to bring this along with the credentials?

registry/metrics.go
```diff
@@ -35,19 +34,19 @@ func NewMetrics() Metrics {
...
 			Subsystem: "registry",
 			Name:      "request_duration_seconds",
-			Help:      "Duration of HTTP requests made in the course of fetching image metadata",
+			Help:      "Duration of HTTP requests made in the course of fetching Image metadata",
 		}, []string{fluxmetrics.LabelInstanceID, LabelRepository, LabelRequestKind, fluxmetrics.LabelSuccess}),
 	}
 }
 
-func (m Metrics) WithInstanceID(instanceID flux.InstanceID) Metrics {
+func (m Metrics) WithInstanceID(instanceID string) Metrics {
```
paulbellamy: why loosen this up?

registry/cache_test.go
paulbellamy: or `MockDockerClient` (which is a thing now)?

registry/registry.go
```diff
@@ -2,80 +2,37 @@
...
-
-// Credentials to a (Docker) registry.
-type Credentials struct {
-	m map[string]creds
-}
-
-// Client is a handle to a bunch of registries.
-type Client interface {
-	GetRepository(repository string) ([]flux.ImageDescription, error)
+// New registry interface
```
paulbellamy: is there an old one somewhere?

registry/registry.go
```diff
@@ -2,80 +2,37 @@
...
 }
 
-// client is a handle to a registry.
-type client struct {
-	Credentials    Credentials
-	Logger         log.Logger
-	Metrics        Metrics
-	MemcacheClient MemcacheClient
-	CacheExpiry    time.Duration
+// registry is a handle to a registry.
```
paulbellamy: eh.. what? Doesn't seem to be... I'm not really sure what this comment is trying to tell me.

registry/registry.go
```diff
@@ -84,86 +41,15 @@ func (f roundtripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
-
-	hostlessImageName := fmt.Sprintf("%s/%s", org, image)
-	httphost := "https://" + host
-
-	// quay.io wants us to use cookies for authorisation, so we have
-	// to construct one (the default client has none). This means a
-	// bit more constructing things to be able to make a registry
-	// client literal, rather than calling .New()
-	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+func (c *registry) GetRepository(img Repository) (_ []Image, err error) {
```
paulbellamy: `c *registry`? (and a few other places)

registry/registry.go
```diff
@@ -172,65 +58,49 @@ func (c *client) GetRepository(repository string) (_ []flux.ImageDescription, er
...
+		image Image
 		err   error
 	}
 
 	fetched := make(chan result, len(tags))
 
 	for _, tag := range tags {
 		go func(t string) {
-			img, err := c.lookupImage(client, lookupName, imageName, t)
+			i, err := remote.Manifest(repository, t)
```
paulbellamy: is `i` an image here? `i` is an int iterator just below, so that's a bit confusing.

registry/registry.go
```diff
@@ -172,65 +58,49 @@ func (c *client) GetRepository(repository string) (_ []flux.ImageDescription, er
...
 			if err != nil {
 				c.Logger.Log("registry-metadata-err", err)
 			}
-			fetched <- result{img, err}
+			fetched <- result{i, err}
 		}(tag)
 	}
 
-	images := make([]flux.ImageDescription, cap(fetched))
+	images := make([]Image, cap(fetched))
```
paulbellamy: Wait, this isn't an ImageDescription? I don't understand the reason for this change.

registry/cache_test.go
philwinder: No, this is your mock that you created. I left it as the same for now.

registry/registry.go
philwinder: Yeah, comments often get garbled when refactoring method/class names.

registry/registry.go
philwinder: Because ImageDescription is bogus type. All the registry code doesn't use any of the external image types any more.

Makefile
```diff
@@ -20,6 +20,7 @@ all: $(GOPATH)/bin/fluxctl $(GOPATH)/bin/fluxd $(GOPATH)/bin/fluxsvc build/.flux
 
 .PHONY: release-bins
 release-bins:
+	rm -r ./vendor/github.com/heroku/docker-registry-client/vendor || true # Remove the heroku vendor folder. Won't build without this.
```
philwinder: Agree would like a fix, but I don't know what that fix would be.

instance/instance.go
philwinder: Because you wanted the interface to be `GetImage(Repository, Tag)`, so the ParseImage parses the image so we can get a tag.

philwinder: Suppose I could use the ImageId.Components method. But I don't like that as it never errors, doesn't do any checking and doesn't have any tests.

philwinder: In fact, no, I can't use it. Because one of the bug reports was that it didn't error when the tag/image was malformed. It was because it never errors in the ImageID stuff.

So I'm keeping this as is, because it is safer.

registry/cache.go
philwinder: Don't see any reason to do this. This is a common middleware pattern.

registry/metrics.go
philwinder: To remove the dependency on the flux God-package. 
I just checked and the new cache re-introduces the dependency again.

instance/instance.go
paulbellamy: This is where I was saying the validating `ImageID` stuff probably should go in the `flux` package, so we don't have 2 (subtly) different versions of the same thing.

registry/cache.go
paulbellamy: I don't see any reason to do this. This is an uncommon code pattern.

registry/cache_test.go
paulbellamy: I know. I mean can we both use the same one?

registry/metrics.go
paulbellamy: Having a package which has all the domain model structs is hardly what I'd call a "God-package".

instance/instance.go
philwinder: Yes, I tried that. After about a day of attempting, I gave up. That ImageId string is in nearly every file in the repo. I.e. nearly every file is directly coupled to flux package through this.

And there are major issues with changing it. Since some files expect it not to have a tag, other expect it to have a tag. I think I've even seens places that expect it to be with/without the host.

So, I could go through and change everything all at once, but it would be a massive PR and would probably break everything.

registry/cache.go
philwinder: All the metrics stuff uses it. And all the http stuff uses it. Google golang middleware and you can read for days. And GoKit uses it for nearly everything. From logging to rate limiting.

registry/cache_test.go
philwinder: Possibly. Let me take a look.

registry/metrics.go
philwinder: Ok, adding it back in. But I think that having every file directly coupled to this file is a bad idea.

For example, imagine that we wanted to change the definition of what ImageID is, to change it to be a proper object with a CreatedAt field? It's hard, because of all the coupling (it would take too long and probably break things), so you end up creating another thing instead (called ImageDescription).

And then there's circular dependencies too. But anyway. Added.

registry/registry.go
philwinder: Just to elaborate. ImageDescription is bad because there isn't an entity called ImageDescription. Image HAS A Description, yes, but ImageDescription isn't a thing. And in this case, it's got nothing to do with a description. Its a "created at" time.

It should be something like:
```Go
type Image struct {
    id string
    createdAt time.Time
}
```
I would have made this change, but as mentioned in another comment, ImageID is too coupled to the rest of the codebase and is too ill-defined to state it's use (it's just a string).

paulbellamy:
I still have some remaining issues with this (factory naming, Image being in registry package), but overall it is good enough that we can merge it in and fix those later. Let's fix up the conflicts (can help with that), rebase away the circle/heroku things, and merge it!

PR #383, opened 2017-01-19
Limit max concurrent tag fetching. There can be a lot of tags

paulbellamy:
So we don't fire off 2k+ simultaneous requests to dockerhub (or memcached).

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -9,6 +9,7 @@ import (
 
 const (
 	requestTimeout = 10 * time.Second
+	maxConcurrency = 10 // Chosen arbitrarily
 )
 
 // The Registry interface is a domain specific API to access container registries.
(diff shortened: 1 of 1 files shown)


registry/registry.go
```diff
@@ -218,17 +219,24 @@ func (c *client) tagsToRepository(cancel func(), client backend, lookupName, ima
...
+				fetched <- result{img, err}
 			}
-			fetched <- result{img, err}
-		}(tag)
+		}()
+	}
+	for _, tag := range tags {
+		toFetch <- tag
 	}
+	close(toFetch)
```
philwinder: I need to check this, but I think you need to wait here.

I think there's a chance that you will close the channel before the channel has finished emptying.

Either use a `waitgroup.Wait()` or some othe channel to signal that it is finished.

philwinder: Yeah I think that's right. Check out https://play.golang.org/p/wZTj9i2ZXe.

philwinder: Working example with waitgroup: https://play.golang.org/p/6ldraf4zq1

paulbellamy: Closing a chan doesn't stop you receiving already sent values, it just stops you sending *more* values. Your examples work because you are waiting for the goroutines to finish consuming before the program exits.

But at the point we close the chan it looks like `["tag1", "tag2", "tag3", <close>]`. The worker goroutines process through all the jobs on the chan, then receive the `<close>` sigil which tells the range loop to finish.

We could have also divided up the `tags` array into `n` slices, and created a goroutine for each of them, but I thought I was more likely to write an off-by-one bug and panic that way.

philwinder: Correct. My mistake.

paulbellamy:
rebased

PR #392, opened 2017-01-20
Ensure registry cache does not interfere with returned errors

paulbellamy:


paulbellamy:
If there is a cache miss, and an upstream error we were hiding the upstream error.

paulbellamy:
Bug introduced in: https://github.com/weaveworks/flux/pull/376

PR #395, opened 2017-01-23
Finish the bus implementation of platform.Version

squaremo:
The NATS bus implementation has two main parts: a client that encodes each method as a message and sends it as a request over NATS, and a server which listens for all requests to a certain instance, and relays it over the RPC connection.

This commit completes the implementation of Version in the bus by adding the server part, and a test for it.


paulbellamy:
Part of: https://github.com/weaveworks/flux/issues/216

PR #396, opened 2017-01-23
Create a binary to query a SQL db and expose for prometheus.

philwinder:
This adds a simple binary that connects to an Postgres-like DB (e.g. RDS), performs a set of user-specified queries and exposes them on the `/metrics` endpoint.

The binary is added to the fluxsvc dockerfile. To use, alter the entrypoint to call the `fluxmon` application.

Queries are specified in a yaml file.

See `--help` for all options.

### DIFF
--- a/cmd/fluxmon/main.go
+++ b/cmd/fluxmon/main.go
@@ -1,0 +1,164 @@
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics/prometheus"
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/weaveworks/flux/db"
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+)
+
+const (
+	// Parameters
+	databaseSourceParam = "database-source"
+	listenParam         = "listen"
+	queriesParam        = "queries"
+	tickParam           = "tick"
+	namespaceParam      = "prometheus-namespace"
+	subsystemParam      = "prometheus-subsystem"
+	nameParam           = "prometheus-name"
+
+	// Defaults
+	defaultServerPort     string = ":80"
+	defaultDatabaseSource string = "file://fluxy.db"
+	defaultQueries        string = "queries.yaml"
+	defaultTick           string = "5s"
+	defaultNamespace      string = "flux"
+	defaultSubsystem      string = "jobs"
+	defaultName           string = "db_status_count"
+
+	LabelName = "name"
+)
+
+// Represents the names and queries that the user wants to perform on the database
+type queries struct {
+	Queries []struct {
+		Name  string
+		Query string
+	}
+}
+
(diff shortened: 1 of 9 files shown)


squaremo:
> The binary is added to the fluxsvc dockerfile.

Sorry if this is recapitulating a discussion elsewhere, but what's the rationale for including it in this image?

cmd/fluxmon/main.go
```diff
@@ -0,0 +1,164 @@
...
+					err = conn.QueryRow(q.Query).Scan(&count)
+					if err != nil {
+						logger.Log("stage", "query", "name", q.Name, "query", q.Query, "err", err)
+						errc <- err
+					}
+					logger.Log("stage", "query", "name", q.Name, "result", fmt.Sprintf("%v", count))
+					jobStatus.With(
+						LabelName, q.Name,
+					).Set(float64(count))
+				}
```
paulbellamy: I thought this was going to happen on-demand when there was a request?

philwinder:
@squaremo No reason. It was the easiest first thing to do.

cmd/fluxmon/queries.yaml
```diff
@@ -0,0 +1,11 @@
+queries:
```
paulbellamy: could really genericize it and even do:
```
gauges:
  flux_job_states:
    labels: ["name"]
    queries:
    - name: "scheduled"
      query: "blah"
    - ...
```
maybe that's too much...

philwinder:
@squaremo I'm starting to refactor out now. Any recommendations? I.e. new dockerfile? Should it be pushed? New repository?

deploy/standalone/flux-deployment.yaml
```diff
@@ -11,13 +11,13 @@ spec:
     spec:
       containers:
       - name: fluxd
-        image: quay.io/weaveworks/fluxd:master-6cc08e4
+        image: weaveworks/fluxd:test
```
squaremo: Is this change (and the analogous one just below) deliberate?

philwinder: :-( Oops. Development environment.

cmd/fluxmon/queries.yaml
philwinder: Cool!

cmd/fluxmon/main.go
philwinder: Bit confused here.

I thought we wanted a separate process that would monitor the DB and expose some metrics to prom. Let me review the issue again.

philwinder: Ah yes, I remember now:
> whenever a prom scrape is requested.

Will implement that.

cmd/fluxmon/queries.yaml
squaremo: One thing I noticed is you can only select a single number for your gauge. We might, for example, want to have per-instance gauges for job queue length, or per queue (~job type). But this is a great start, and we can do that later since this is already useful.

squaremo:
> I'm debating if fluxmon should live in this repo, or in monitoring as jobmon

It certainly seems like it's not specific to flux -- I'd support a separate (public) repo, or moving to monitoring.

philwinder:
Was fixed by addition of external microservice. See
https://github.com/weaveworks/service-conf/pull/556
https://github.com/weaveworks/service-conf/pull/558
https://github.com/weaveworks/service-conf/pull/560

PR #397, opened 2017-01-23
Api/flux prefix migration

paulbellamy:
Make flux respond both at `/` and at `/api/flux` so we can swap authfe over, then remove support for `/`

Part of https://github.com/weaveworks/flux/issues/217


squaremo:
Will this also need changes to standalone instructions?

paulbellamy:
at this stage no, but quite right.

paulbellamy:
In trying to figure out how to migrate this, I've realized that changing the router also changes the client. So, if we change the router to have a prefixed subrouter, the client will begin adding `/api/flux` to each request route. This makes the migration (as far as what we tell users to do, and whether they should upgrade fluxctl) really hard to predict.

PR #401, opened 2017-01-25
updated to the latest versions of fluxd and fluxsvc

nustiueudinastea:
The deploy instructions for the standalone Flux were failing because they were running an old version of Flux. This PR updates both containers to their latest version.


paulbellamy:
Nice one! Thanks so much!

PR #402, opened 2017-01-26
memcache instruction

nustiueudinastea:
Added instruction to run memcache, which is required but not mentioned in the doc.


paulbellamy:
Thanks!

PR #405, opened 2017-01-26
Depend on k8s.io/client-go

### LINKED ISSUE #277: Use k8s.io/kubernetes/client-go instead of the full kubernetes API
This may drag in less stuff, and thereby speed up `gvt restore` and consequently CircleCI builds.

awh:
Depend on the new k8s client library rather than k8s entire, reducing `gvt restore` times by ~75%.

I have tested `fluxctl list-services` and `fluxctl release` with cloud.weave.works which I believe exercises most of the k8s API integration in fluxd.

Fixes #277.


squaremo:
:1st_place_medal: :1st_place_medal: :1st_place_medal:

PR #414, opened 2017-02-01
Refactor out image id stuff to fix broken dockerhub updates

paulbellamy:
Fixes https://github.com/weaveworks/flux/issues/413


image.go
```diff
@@ -0,0 +1,23 @@
+package flux
+
+import (
+	"time"
+)
+
+// Image can't really be a primitive string only, because we need to also
+// record information about it's creation time. (maybe more in the future)
```
squaremo: its

image_id.go
```diff
@@ -0,0 +1,103 @@
+package flux
```
squaremo: Perhaps it's misplaced parsimony, but I'd be happier if this was all just in `image.go`.

image_id.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	var ta string
+	if i.Tag != "" {
+		ta = fmt.Sprintf(":%s", i.Tag)
+	}
+	return fmt.Sprintf("%s%s", i.Repository(), ta)
+}
+
+// Repository returns the short version of an image's repository (trimming if dockerhub)
+func (i ImageID) Repository() string {
```
squaremo: Usually we want one of three things:

  1. are these images equivalent? (compare canonical IDs)
  2. are these images from the same repo? (compare canonical repository name)
  3. show the repository name as given

This suggests to me that the most important methods are, say, `Equivalent(imageID)` and `SameRepository(imageID)`. I'm not sure we ever display canonical names/IDs, so those might not even need to be calculated outside of those methods.

Since we mostly want to display image names as given, we probably don't want to add defaults in when parsing (only when comparing).

paulbellamy:
> Rationalised
Not fully. `registry.Repository` is still a (slightly odd) thing.

image_id.go
paulbellamy: I was debating that as well, will do.

image_id.go
paulbellamy: The point of the bug is that we need to use the short "repository" in the k8s update regex. But yeah, these methods could probably be further simplified.

PR #415, opened 2017-02-01
Better slack notifications

paulbellamy:
Fixes: https://github.com/weaveworks/flux/issues/390

# TODO:

- [x] Clean up `TODO`s in the code
- [x] Some sort of migration from jobs? or fill the release struct to send to notifiers
- [x] Make *sure* the serviceSpec stuff migration works, or move it to separate pr.
- [x] Figure out how to integrate it with michael's changes for job results...


paulbellamy:
serviceSpec stuff moved to: https://github.com/weaveworks/flux/pull/423

paulbellamy:
Note, without more info from the release, this doesn't currently include any results in the output.

release.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Release describes a release
+type Release struct {
+	ID        ReleaseID            `json:"id"`
+	CreatedAt time.Time            `json:"createdAt"`
+	StartedAt time.Time            `json:"startedAt"`
+	EndedAt   time.Time            `json:"endedAt"`
+	Done      bool                 `json:"done"`
+	Priority  int                  `json:"priority"`
+	Status    ServiceReleaseStatus `json:"status"`
```
squaremo: Is this supposed to be an enum (one of those consts above) or a free-form string?

jobs/job.go
```diff
@@ -150,16 +150,7 @@ func (j *Job) UnmarshalJSON(data []byte) error {
...
-	ServiceSpecs []flux.ServiceSpec
-	ImageSpec    flux.ImageSpec
-	Kind         flux.ReleaseKind
-	Excludes     []flux.ServiceID
-
-	// Backwards Compatibility, remove once no more jobs
-	// TODO: Remove this once there are no more jobs with ServiceSpec, only ServiceSpecs
-	ServiceSpec flux.ServiceSpec
-}
+type ReleaseJobParams flux.ReleaseSpec
```
squaremo: Since it's all serialised in JSON, does there need to be an alias?

paulbellamy: Probably not. Thought it might be cleaner to have a alias here so the param types are all consistent.

release.go
paulbellamy: enum from above.

paulbellamy:
I re-did the default template, to use the spec instead of the results to output stuff.

It means the output is based on what was requested, not what was done, but... that is what we have for now (until the result format changes from @squaremo's work).

for example:
```
Release no updates to flux-example/nginx. done
Release all latest to flux-example/nginx. done
Release quay.io/weaveworks/helloworld:master-a000001 to all. done
Release all latest to flux-example/nginx. some error here. failed
Release all latest to all. done // This one is a bit unsatisfying...
```

paulbellamy:
Also, I think all the `TODO`s left (now) are places where we need to integrate this with your changes, @squaremo, or reminders to remove deprecated stuff.

release/releaser.go
```diff
@@ -230,10 +231,43 @@ func (r *Releaser) releaseWithoutUpdate(method, msg string, inst *instance.Insta
...
+		// table, and the job should really just have a pointer to that.
+		err2 = notifications.Release(cfg, flux.Release{
+			ID:        flux.ReleaseID(job.ID),
+			CreatedAt: job.Submitted,
+			StartedAt: job.Claimed,
+			// TODO: fetch the job and look this up so it matches
+			EndedAt:  time.Now().UTC(),
+			Done:     true,
+			Priority: job.Priority,
+			Status:   flux.ServiceReleaseStatus(job.Status),
```
squaremo: Should this be one of the enum values? Since we only use this release for the notification, and the notification doesn't use this field, it doesn't make a difference to the outcome. But it may avoid a slip-up when this code is adapted later.

squaremo:
> I re-did the default template, to use the spec instead of the results to output stuff.

Does this more or less replicate the current behaviour?

PR #418, opened 2017-02-02
Less cardinality on registry metrics, as prom can't handle it

paulbellamy:
Fixes https://github.com/weaveworks/flux/issues/417

### DIFF
--- a/platform/metrics.go
+++ b/platform/metrics.go
@@ -99,10 +99,10 @@ func NewBusMetrics() BusMetrics {
 			Subsystem: "bus",
 			Name:      "kick_total",
 			Help:      "Count of bus subscriptions kicked off by a newer subscription.",
-		}, []string{fluxmetrics.LabelInstanceID}),
+		}, []string{}),
 	}
 }
 
 func (m BusMetrics) IncrKicks(inst flux.InstanceID) {
-	m.KickCount.With(fluxmetrics.LabelInstanceID, string(inst)).Add(1)
+	m.KickCount.Add(1)
 }
--- a/registry/metrics.go
+++ b/registry/metrics.go
@@ -21,8 +21,7 @@ type Metrics struct {
 }
 
 const (
-	LabelRepositoryHost = "repository_host"
-	LabelRequestKind    = "kind"
+	LabelRequestKind = "kind"
 
 	RequestKindTags     = "tags"
 	RequestKindMetadata = "metadata"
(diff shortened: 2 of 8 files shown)


platform/metrics.go
```diff
@@ -99,10 +99,10 @@ func NewBusMetrics() BusMetrics {
 			Subsystem: "bus",
 			Name:      "kick_total",
 			Help:      "Count of bus subscriptions kicked off by a newer subscription.",
-		}, []string{fluxmetrics.LabelInstanceID}),
+		}, []string{}),
```
jml: Prometheus has a different API for … oh, never mind.

tomwilkie:
Its pretty well documented that you shouldn't do high cardinality:
- https://prometheus.io/docs/practices/naming/#labels
- https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels

registry/metrics.go
```diff
@@ -37,20 +36,13 @@ func NewMetrics() Metrics {
 			Name:      "fetch_duration_seconds",
 			Help:      "Duration of Image metadata fetches, in seconds.",
 			Buckets:   stdprometheus.DefBuckets,
-		}, []string{fluxmetrics.LabelInstanceID, LabelRepository, fluxmetrics.LabelSuccess}),
+		}, []string{LabelRepositoryHost, fluxmetrics.LabelSuccess}),
```
tomwilkie: Whats the possible cardinality of the repository host?

registry/metrics.go
```diff
@@ -37,20 +36,13 @@ func NewMetrics() Metrics {
...
 			Help:      "Duration of HTTP requests made in the course of fetching Image metadata",
-		}, []string{fluxmetrics.LabelInstanceID, LabelRepository, LabelRequestKind, fluxmetrics.LabelSuccess}),
-	}
-}
-
-func (m Metrics) WithInstanceID(instanceID flux.InstanceID) Metrics {
-	return Metrics{
-		FetchDuration:   m.FetchDuration.With(fluxmetrics.LabelInstanceID, string(instanceID)),
-		RequestDuration: m.RequestDuration.With(fluxmetrics.LabelInstanceID, string(instanceID)),
+		}, []string{LabelRepositoryHost, LabelRequestKind, fluxmetrics.LabelSuccess}),
```
tomwilkie: Whats the cardinality of LabelRequestKind?

paulbellamy: 2

awh:
LGTM.

registry/metrics.go
paulbellamy: I was debating that. at the moment it's things like `index.docker.io`, `quay.io`, and `gcr.whatever`.
🤔 
I suppose if users are using self-hosted registries, it could be quite a few. Will disable for now until we know more.

paulbellamy:
> `service_spec_all` in particular doesn't communicate much to me.

It doesn't communicate much to anyone, but such is the granularity of data we're allowed to have. I suspect that the durations for that are a bimodal distribution, so `service_spec_all` should help us differentiate those.

PR #419, opened 2017-02-02
Standardize http metrics, to service_request_duration

paulbellamy:
Should be the same as we use for other services.

### DIFF
--- a/cmd/fluxsvc/main.go
+++ b/cmd/fluxsvc/main.go
@@ -99,7 +99,7 @@ func main() {
 		busMetrics       platform.BusMetrics
 		helperDuration   metrics.Histogram
 		historyMetrics   history.Metrics
-		httpDuration     metrics.Histogram
+		httpDuration     *stdprometheus.HistogramVec
 		instanceMetrics  instance.Metrics
 		jobWorkerMetrics jobs.WorkerMetrics
 		registryMetrics  registry.Metrics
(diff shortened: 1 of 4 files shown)


paulbellamy:
To match queries in: https://github.com/weaveworks/monitoring/pull/109

cmd/fluxsvc/main.go
```diff
@@ -99,21 +99,20 @@ func main() {
...
+		httpDuration     *stdprometheus.HistogramVec
 		instanceMetrics  instance.Metrics
 		jobWorkerMetrics jobs.WorkerMetrics
 		registryMetrics  registry.Metrics
 		releaseMetrics   release.Metrics
 		serverMetrics    server.Metrics
 	)
 	{
-		httpDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
+		httpDuration = stdprometheus.NewHistogramVec(stdprometheus.HistogramOpts{
```
squaremo: One thing this loses is the request _counts_ (and therefore rates). Does the standard instrumentation not care about those?

paulbellamy: ? This doesn't lose that?

paulbellamy: for a histogram prom creates the three: `flux_request_duration_seconds_(count|bucket|sum)`

squaremo: I beg your puddin', I thought that was something gokit's metrics did. As you were ...

PR #421, opened 2017-02-03
Better release output

squaremo:
The releaser code works by accumulating a list of "release action"
structures, then running through them.

But the actions mostly do not correspond with what we want to tell the
user about -- most of them are just preparation (e.g., "clone the
repo"), which does not have a place in the plan we report to the
user. This means the useful information is interleaved with irrelevant
details, and decisions made during the process (e.g., about which
services to ignore) are obscured.

A better structure is to simply do all the preparation, which is for
the most part the same each time, in one place, while keeping an
explicit record of the decisions about what to do in the execution
phase.

These commits change the planning code so that git and the
instance config are examined before the running system. This means we
don't waste time (and the user's attention) on services that can never
be affected.

 - [x] restore instrumentation (metrics and logging)
 - [x] rejig the automator to work with this
 - [x] change the release job and command to report on the `ReleaseResult`

*Review notes*

For the files in `release/`, it is best to look at the end state, since it is largely a rewrite. The control flow is almost the same as before; the main difference is the removal of the "accumulate actions" structure, as described above. The changes to `automator/automator.go` follow from these, but can be reviewed as a diff, since they are not extensive. The changes in `platform/kubernetes` are to support the rewritten releaser, and should be read alongside. There are also follow-on changes in the job database and `cmd/fluxctl`, to store and output the result that is the object of this PR.

The changes in `release/` don't have that much effect on the code outside that package (!). The remainder of the changes are mostly about making various things more amenable to testing. One other change of note of note is the metrics we collect about releases, which I have simplified as noted in the relevant commit.


release/releaser.go
```diff
@@ -44,233 +42,376 @@ func NewReleaser(
...
+			}
+			id, err := flux.ParseServiceID(string(s))
+			if err != nil {
+				return nil, err
+			}
+			ids = append(ids, id)
+		}
+		return rc.SelectExactServices(ids, lockedSet, excludedSet, logStatus)
+	case flux.ServiceSpecAll:
+		return rc.SelectServices(nil, lockedSet, excludedSet, logStatus)
```
paulbellamy: A slightly simpler way of handling backwards compat might be:
```
if releaseJob.ServiceSpec != "" {
  releaseJob.ServiceSpecs = append(releaseJob.ServiceSpecs, releaseJob.ServiceSpec)
}

// Then only handle the array case.
// But it assumes that only one or the other will be set.
```

release/releaser.go
```diff
@@ -44,233 +42,376 @@ func NewReleaser(
...
 
-func (r *Releaser) releaseImages(method, msg string, inst *instance.Instance, getServices ServiceSelector, getImages ImageSelector) ([]ReleaseAction, error) {
-	var res []ReleaseAction
-	res = append(res, r.releaseActionPrintf(msg))
-
-	var (
-		base  = r.metrics.StageDuration.With("method", method)
-		stage *metrics.Timer
-	)
+// --- Additions to ReleaseContext for above
```
paulbellamy: should probably live in `release/context.go`

release/releaser.go
```diff
@@ -44,233 +42,376 @@ func NewReleaser(
...
+	logStatus("Finding defined services.")
+	var updates []*ServiceUpdate
+	if updates, err = r.selectServices(rc, &params, logStatus); err != nil {
+		return nil, err
+	}
 
-	case params.ServiceSpec == flux.ServiceSpecAll && params.ImageSpec == flux.ImageSpecNone:
-		releaseType = "release_all_without_update"
-		actions, err = r.releaseWithoutUpdate(releaseType, msg, inst, services)
+	logStatus("Looking up images.")
```
paulbellamy: This could be inside the if below

release/releaser.go
```diff
@@ -44,233 +42,376 @@ func NewReleaser(
...
-	res = append(res, r.releaseActionClone())
+// execute is called if the release is *not* a dry run; i.e., it
+// should actually effect the changes calculated.
+func (r *Releaser) execute(rc *ReleaseContext, updates []*ServiceUpdate, releaseJob *jobs.ReleaseJobParams, logStatus statusFn) error {
+	// Write changes, commit and push
+	logStatus("Writing changes to files.")
+	err := rc.WriteUpdates(updates)
+	if err != nil {
+		return err
+	}
```
paulbellamy: Need to differentiate between:

a) there are no updates. (bail, and noop)
b) we're applying latest config (no commit & push to do, but continue)

release/releaser.go
squaremo: Yep, just collecting them here until I've got all the ones I need, so less chance of leaving dead code around.

squaremo:
> initial thoughts. I didn't see the "action -> result" thing you were talking about

The log is the unstructured "here are some actions and some results", the ReleaseResult is the (new) structured "here are the results".

release/releaser.go
paulbellamy: or (post: https://github.com/weaveworks/flux/pull/423), might not be needed...

paulbellamy:
ah right, I had expected the structured log to be more around steps of the deploy, but doing by service makes sense too.

release.go
```diff
@@ -7,12 +7,32 @@ import (
 	"github.com/weaveworks/flux/guid"
 )
 
+// ReleaseKind says whether a release is to be planned only, or planned then executed
+type ReleaseKind string
```
squaremo: Moved here, since these types pertain to releases.

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
-		res = append(res, r.releaseActionPrintf(format, args...))
-	})
-
-	if len(updateMap) <= 0 {
-		res = append(res, r.releaseActionPrintf("All selected services are running the requested images. Nothing to do."))
-		return res, nil
+	// At this point we have have filtered the updates we can do down
+	// to nothing. Check and exit early if so
+	if len(updates) == 0 {
+		logStatus("No updates to do, finishing.")
```
paulbellamy: This isn't really true if it is a no_update change, right?

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
+		return nil, nil
 	}
 
-	stage.ObserveDuration()
-	stage = metrics.NewTimer(base.With("stage", "finalize"))
-
-	// We have identified at least 1 release that needs to occur. Releasing
-	// means cloning the repo, changing the resource file(s), committing and
-	// pushing, and then making the release(s) to the platform.
+	if spec.Kind == flux.ReleaseKindExecute {
```
paulbellamy: Would it be cleaner to invert this, return early, and avoid the nesting?

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
+				return nil, err
+			}
+		}
 
-	res = append(res, r.releaseActionClone())
-	for service, applies := range updateMap {
-		res = append(res, r.releaseActionUpdatePodController(service, applies))
+		logStatus("Applying changes.")
+		timer = r.metrics.NewStageTimer("apply_changes")
+		err = applyChanges(rc.Instance, updates, results)
```
paulbellamy: this err is ignored.

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
+	if err != nil {
+		return nil, err
+	}
+	lockedSet := LockedServices(conf)
+
+	excludedSet := flux.ServiceIDSet{}
+	excludedSet.Add(spec.Excludes)
+
+	// For backwards-compatibility, there's two fields: ServiceSpec
+	// and ServiceSpecs. An entry in ServiceSpec takes precedence.
```
paulbellamy: Isn't that already dealt with by this point? handled in http/transport?

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
-				updateJob(result)
-			}
-			actions[i].Result = result
+// Find all the image updates that should be performed, and do
+// replacements. (For a dry-run, we don't strictly need to do the
+// replacements, since we won't be committing any changes back;
+// however we do want to see if we *can* do the replacements, because
+// if not, it indicates there's likely some problem with the running
+// system vs the definitions given in the repo.)
+func calculateImageUpdates(inst *instance.Instance, updateable []*ServiceUpdate, spec *flux.ReleaseSpec, results flux.ReleaseResult, logStatus statusFn) ([]*ServiceUpdate, error) {
```
paulbellamy: might be slightly cleaner if this took a list of image IDs, and a list of services, then did the calculation. Right now it conflates "go look at the registries" with "figure out what to do".

paulbellamy: also easier to test, probably.

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
+	}
 
-func CalculateUpdates(services []platform.Service, images instance.ImageMap, printf func(string, ...interface{})) map[flux.ServiceID][]ContainerUpdate {
-	updateMap := map[flux.ServiceID][]ContainerUpdate{}
-	for _, service := range services {
-		containers, err := service.ContainersOrError()
+	// Look through all the services' containers to see which have an
+	// image that could be updated.
+	var updates []*ServiceUpdate
+	for _, update := range updateable {
```
paulbellamy: s/updateable/candidates/ ? maybe?

release/releaser.go
```diff
@@ -45,539 +36,318 @@ func NewReleaser(
...
-			// finished recording what happened, as part of a graceful
-			// shutdown. So the only thing that goes missing is the
-			// result from this release call.
-			if len(asyncDefs) > 0 {
-				rc.Instance.PlatformApply(asyncDefs)
-			}
-
-			return "", transactionErr
-		},
+func commitMessageFromReleaseSpec(spec *flux.ReleaseSpec) string {
```
paulbellamy: I felt like we had a thing like this for notifications. Maybe we could refactor and share?

paulbellamy: Ah, though maybe the notification should be based on what it *actually* did, not just the intent. Since the actual results are obvious from the git commit, this can just be the intent.

paulbellamy:
Am getting a slightly weird newline in the output for one image id:
```
$ fluxctl release --all --update-image=quay.io/weaveworks/helloworld:master-a000002 --dry-run
Submitting dry-run release job...
Release job submitted, ID e7137339-de70-9b24-3d0e-aa24f89bc1cb
Status: Complete.

Here's the plan:
default/helloworld  pending
                              helloworld: quay.io/weaveworks/helloworld:master-a000001 -> master-a000002
```

paulbellamy:
Also, please squash the commits.

paulbellamy:
Do we need to make changes to the UI to support this?

platform/kubernetes/kubernetes.go
```diff
@@ -126,9 +126,8 @@ func (c *Cluster) SomeServices(ids []flux.ServiceID) (res []platform.Service, er
 		for _, name := range names {
 			service, err := services.Get(name)
 			if err != nil {
-				return nil, errors.Wrapf(err, "finding service %s among services for namespace %s", name, ns)
+				continue
```
squaremo: We report it in the results as "not in running system"

circle.yml
```diff
@@ -19,6 +19,9 @@ checkout:
     - mv ${HOME}/flux ${GOPATH}/src/github.com/weaveworks/
     - ln -s ${GOPATH}/src/github.com/weaveworks/flux ${HOME}/flux # Circle needs this to be here, apparently
     - cd ${GOPATH}/src/github.com/weaveworks/flux
+    - sudo pip install pyyaml
+    - git config --global user.email "example@example.com"
+    - git config --global user.name "Weave Flux test user"
```
squaremo: Git will refuse to push if you don't have them (yes you can probably supply them in the invocation, but this is easier)

jobs/db_store.go
```diff
@@ -338,11 +348,32 @@ func (s *DatabaseStore) scanParams(method string, params []byte) (interface{}, e
 	}
 }
 
+func (s *DatabaseStore) scanResult(method string, result []byte) (interface{}, error) {
+	if result == nil {
+		return nil, nil
```
squaremo: Good question -- otherwise `if scannedresult == nil` might not work (differently typed nils are not equal)?

jobs/job.go
```diff
@@ -109,14 +111,15 @@ func (j *Job) UnmarshalJSON(data []byte) error {
...
-		Log       []string  `json:"log,omitempty"`
-		Status    string    `json:"status"`
-		Done      bool      `json:"done"`
-		Success   bool      `json:"success"` // only makes sense after done is true
+		Submitted time.Time       `json:"submitted"`
+		Claimed   time.Time       `json:"claimed,omitempty"`
+		Heartbeat time.Time       `json:"heartbeat,omitempty"`
+		Finished  time.Time       `json:"finished,omitempty"`
+		Log       []string        `json:"log,omitempty"`
+		Result    json.RawMessage `json:result`
```
squaremo: eep. Seems to work, but probably only by coincidence.

platform/kubernetes/test/data.go
```diff
@@ -0,0 +1,91 @@
+package test
```
squaremo: yep!

release/print.go
```diff
@@ -0,0 +1,27 @@
...
+				continue
+			}
+		}
+		fmt.Fprintf(w, "%s\t%s\t%s\n", serviceID, result.Status, result.Error)
+		for _, update := range result.PerContainer {
+			fmt.Fprintf(w, " \t \t %s: %s -> %s\n", update.Container, update.Current.FullID(), update.Target.Tag)
+		}
+	}
+	w.Flush()
+}
```
squaremo: :+1: for sure

release/releaser.go
squaremo: If it's a "no update change", we will not have executed the previous block which looks for new images and filters `updates` down to just those things with image updates.

Is the layout of this code a bit misleading, perhaps -- I can try and rejig it.

release/releaser.go
squaremo: Yes, probably -- will rejig as above.

release/releaser.go
squaremo: ops.

release/releaser.go
squaremo: It is now, yeah -- not sure if we have to deal with "old" jobs (probably not). I can take this out, or we can have a PR which just removes that field altogether, if we don't think there are any uses left.

release/releaser.go
squaremo: Hmm, yes it could be split this way:

 1. collect all the image IDs you care about
 2. go look the image tags up
 3. figure out which things have updates.

If you provide image IDs, the first step is done, perforce (though possibly, work can be avoided by checking if any of the services actually use them). Otherwise you need to collect all the images from the services given.

Then the rest proceeds as before. One thing that might trip this up (or make it a better idea!) is other code which uses the image map; i.e., the automator.

release/releaser.go
squaremo: :+1:
Probably ought to stick to consistent names throughout, since many of these procedures have the same pattern of taking candidate updates in, and filtering them.

release/releaser.go
squaremo: That was my thinking, although it's a tricky balance.

More info can go in the commit body (later).

squaremo:
> Do we need to make changes to the UI to support this?

In the same way old clients should be able to use the new service (and just get slightly more helpful output), the UI should cope. But I will check!

jobs/db_store.go
paulbellamy: I was thinking that then a `result.(ReleaseResult)` would always work, even if metadata were nil.

release/releaser.go
paulbellamy: Only uses are from old fluxctl clients, so was leaving the field in for a bit until they're gone.

paulbellamy: Could quarantine it to the http/transport, with a
```
type LegacyReleaseSpec struct {
  flux.ReleaseSpec
  ServiceSpec flux.ServiceSpec `json:"serviceSpec"`
}
```
then convert it or w/e. maybe that is worth doing... :thinking:

release/releaser.go
paulbellamy: :+1: for dealing with this later.

jobs/db_store.go
```diff
@@ -351,10 +382,10 @@ func (s *DatabaseStore) UpdateJob(job Job) error {
 	return s.Transaction(func(s *DatabaseStore) error {
 		if res, err := s.conn.Exec(`
 			UPDATE jobs
-				 SET params = $1, log = $2, status = $3
-			 WHERE id = $4
-				 AND instance_id = $5
-		`, string(paramsBytes), string(logBytes), job.Status, string(job.ID), string(job.Instance)); err != nil {
+				 SET params = $1, result = $2, log = $3, status = $4
```
squaremo: It was like that when I got here

jobs/db_store.go
squaremo: Is the same true of params?

jobs/db_store.go
paulbellamy: so?

jobs/db_store.go
paulbellamy: Probably?

squaremo: Unmarshalling `[]byte(nil)` results in an empty value, so this is actually _less_ code than before :1st_place_medal:

jobs/db_store.go
squaremo: waaaaah

squaremo: Actually I'm not sure what the complaint is -- all the queries seem to be formatting in this style

paulbellamy: oh, the complaint was spaces vs tabs. to be fair, it's probably my vim messing it up.

jobs/db_store.go
squaremo: Ah. It results in an empty value, and in an error. So, no medal.

release/releaser.go
squaremo: Oh, no it's not ignored. It is passed into `sendNotification(...)`, which sends a notification of whether the released completed or failed based on it, then spits it back.

release/releaser.go
squaremo: Can I do this refactor in another PR please? I feel like I've rewritten enough (the status quo at least has the benefit of mostly leaving other packages alone)

squaremo:
> Do we need to make changes to the UI to support this?
> 
> In the same way old clients should be able to use the new service (and just get slightly more helpful output), the UI should cope. But I will check!

Yes, the web UI outputs the logs as it gets them, similar to the CLI. Once this is deployed, it could show the ReleaseResult as it updates, as well.

platform/kubernetes/kubernetes.go
squaremo: :heavy_check_mark:

release/print.go
squaremo: :heavy_check_mark:

release/releaser.go
squaremo: ... in another PR ;)

squaremo:
Commits squished to a more manageable few, and parts redone per review (as indicated in my replies above). @paulbellamy, mind having another look to see if there's blockers left?

paulbellamy:
Fixes https://github.com/weaveworks/flux/issues/358

release/instance.go
```diff
@@ -0,0 +1,105 @@
...
+		switch serviceName {
+		case FluxServiceName, FluxDaemonName:
+			inst.LogEvent(namespace, serviceName, "Starting. (no result expected)")
+			asyncDefs = append(asyncDefs, platform.ServiceDefinition{
+				ServiceID:     update.ServiceID,
+				NewDefinition: update.ManifestBytes,
+				Async:         true,
+			})
+		default:
+			inst.LogEvent(namespace, serviceName, "Starting")
```
paulbellamy: Starting what? This makes the history a bit weird. The only thing the history ever says now is: `helloworld: Starting`

PR #422, opened 2017-02-03
Need to serialize/deserialize image ids as strings for backwrds compat with fluxctl

paulbellamy:
Was broken by: https://github.com/weaveworks/flux/pull/414


awh:
LGTM.

PR #427, opened 2017-02-06
Process NATS messages concurrently

### LINKED ISSUE #420: Bus subscriber blocks on Apply
The NATS bus which conveys RPCs to where they need to go, includes a loop that deals with relaying requests and responses. The idea of having a loop was simply to be able to use a single (wildcard) subscription, and thus avoid complicated {set,clean}up.

However, because the loop deals with one request at a time, it can block subsequent requests indefinitely. For almost everything, this is not a problem, since most RPC calls should be quick. But `Apply` (new definitions) can take an arbitrary duration to complete.

This problem likely shows up as RPCs (e.g., ping) timing out while a release is in progress.

awh:
Fixes #420.

### DIFF
--- a/platform/rpc/nats/bus.go
+++ b/platform/rpc/nats/bus.go
@@ -240,75 +240,110 @@ func (n *NATS) Subscribe(instID flux.InstanceID, remote platform.Platform, done 
 	myID := guid.New()
 	n.raw.Publish(string(instID)+methodKick, []byte(myID))
 
-	go func() {
+	errc := make(chan error)
+
+	processRequest := func(request *nats.Msg) {
 		var err error
-		for request := range requests {
-			switch {
-			case strings.HasSuffix(request.Subject, methodKick):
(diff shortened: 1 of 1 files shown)


platform/rpc/nats/bus.go
```diff
@@ -241,74 +241,98 @@ func (n *NATS) Subscribe(instID flux.InstanceID, remote platform.Platform, done
...
-				switch applyErr := err.(type) {
-				case platform.ApplyError:
-					result := fluxrpc.ApplyResult{}
-					for s, e := range applyErr {
-						result[s] = e.Error()
+			case request := <-requests:
+				// Some of these operations (Apply in particular) can block for a long time;
+				// dispatch in a goroutine and deliver any errors back to us so that we can
+				// clean up on any hard failures.
+				go func() {
```
paulbellamy: maybe this could be a method? reduce the nesting a bit?

PR #431, opened 2017-02-08
api to provide automated deploy key insertion

philwinder:
This code provides an api for inserting a deploy key.

It is intended that this code is forwarded by authfe. The GH authorisation token is injected by authfe.

But the code can be called manually with:
```
curl -XPOST -H "GithubToken: ${TOKEN}" http://192.168.99.100:31470/v5/integrations/github\?owner\=philwinder\&repository\=flux-example -v
```
Where `${TOKEN}` is your token and the repository has been altered.

This will insert a new deploy key called `flux-generated`. If the key already exists, it will be overwritten (by deleting and recreating).

This has been tested alongside the change to authfe (https://github.com/weaveworks/service/pull/1102)

### DIFF
--- a/http/transport.go
+++ b/http/transport.go
@@ -23,6 +23,7 @@ import (
 	"github.com/weaveworks/flux/api"
 	"github.com/weaveworks/flux/http/error"
 	"github.com/weaveworks/flux/http/websocket"
+	"github.com/weaveworks/flux/integrations/github"
 	"github.com/weaveworks/flux/jobs"
 	"github.com/weaveworks/flux/platform"
 	"github.com/weaveworks/flux/platform/rpc"
@@ -43,6 +44,7 @@ func NewRouter() *mux.Router {
 	r.NewRoute().Name("GetConfig").Methods("GET").Path("/v4/config")
 	r.NewRoute().Name("SetConfig").Methods("POST").Path("/v4/config")
 	r.NewRoute().Name("GenerateDeployKeys").Methods("POST").Path("/v5/config/deploy-keys")
+	r.NewRoute().Name("PostIntegrationsGithub").Methods("POST").Path("/v5/integrations/github").Queries("owner", "{owner}", "repository", "{repository}")
 	r.NewRoute().Name("RegisterDaemon").Methods("GET").Path("/v4/daemon")
 	r.NewRoute().Name("IsConnected").Methods("HEAD", "GET").Path("/v4/ping")
 	return r
@@ -50,21 +52,22 @@ func NewRouter() *mux.Router {
 
 func NewHandler(s api.FluxService, r *mux.Router, logger log.Logger, h *stdprometheus.HistogramVec) http.Handler {
 	for method, handlerFunc := range map[string]func(api.FluxService) http.Handler{
-		"ListServices":       handleListServices,
-		"ListImages":         handleListImages,
-		"PostRelease":        handlePostRelease,
-		"GetRelease":         handleGetRelease,
-		"Automate":           handleAutomate,
-		"Deautomate":         handleDeautomate,
-		"Lock":               handleLock,
-		"Unlock":             handleUnlock,
-		"History":            handleHistory,
-		"Status":             handleStatus,
-		"GetConfig":          handleGetConfig,
-		"SetConfig":          handleSetConfig,
-		"GenerateDeployKeys": handleGenerateKeys,
-		"RegisterDaemon":     handleRegister,
-		"IsConnected":        handleIsConnected,
+		"ListServices":           handleListServices,
+		"ListImages":             handleListImages,
+		"PostRelease":            handlePostRelease,
+		"GetRelease":             handleGetRelease,
+		"Automate":               handleAutomate,
+		"Deautomate":             handleDeautomate,
+		"Lock":                   handleLock,
+		"Unlock":                 handleUnlock,
+		"History":                handleHistory,
+		"Status":                 handleStatus,
+		"GetConfig":              handleGetConfig,
+		"SetConfig":              handleSetConfig,
+		"GenerateDeployKeys":     handleGenerateKeys,
+		"PostIntegrationsGithub": handlePostIntegrationsGithub,
+		"RegisterDaemon":         handleRegister,
+		"IsConnected":            handleIsConnected,
 	} {
 		var handler http.Handler
 		handler = handlerFunc(s)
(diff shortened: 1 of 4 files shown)


foot:
Yep this looks great!

squaremo:
It's a bit strange to me that the header is named `Authorization`, since 1. it's not used as authorisation for the resource requested, and 2. we use the header for other things, like authorisation using a service token. Is there a rationale behind that choice?

The route used is a bit misleading, or at least overly generic. What about something like (off the top of my head) `/integrations/github`?

philwinder:
Rational for Auth header was that GH uses Authorization as it's header. Happy to chage to GithubToken or whatever, because we're using it internally only.

And happy to change the path. I put it on config because we are affecting the config.

squaremo:
> Rational for Auth header was that GH uses Authorization as it's header. Happy to chage to GithubToken

OK cool -- I am more worried about it colliding than it being semantically sound, on reflection. Thinking about the change required in an authenticating proxy (which will surely be configurable), or in fluxctl (which we control ourselves), I think it's OK to use a specific name.

philwinder:
Ok changing api to: `/v5/integrations/github` and header to `GithubToken`. Thanks for the input!

philwinder:
Relies on merging https://github.com/weaveworks/flux/pull/437 first.

git/github.go
```diff
@@ -0,0 +1,85 @@
...
+	return g.doInsertDeployKeyRequest(req)
+}
+
+func (g *Github) encodeInsertDeployKeyRequest(inst flux.InstanceID, deployKey string) (io.Reader, error) {
+	var keyReq bytes.Buffer
+	keysBody := struct {
+		Title string `json:"title"`
+		Key   string `json:"key"`
+	}{
+		fmt.Sprintf("Flux-%s", inst),
```
squaremo: This is going to end up being something like "Flux-32" (instanceIDs are numbers). It's useful to have an identifier in there for troubleshooting, but I worry about leaking internal identifiers. I don't think we need more than one deploy key in any given repo, do we?

git/github.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+
+	// Set token in header
+	req.Header.Set("Authorization", fmt.Sprintf("token %s", tok))
+
+	return g.doInsertDeployKeyRequest(req)
+}
+
+func (g *Github) encodeInsertDeployKeyRequest(inst flux.InstanceID, deployKey string) (io.Reader, error) {
```
squaremo: This procedure and `decodeInsertDeployKeyError` don't appear to need `*Github` as a parameter.

http/transport.go
```diff
@@ -668,6 +671,64 @@ func invokeGenerateKeys(client *http.Client, t flux.Token, router *mux.Router, e
...
+		}
+
+		// Obtain the generated key
+		cfg, err := s.GetConfig(inst)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, err.Error())
+			return
+		}
+		publicKey := cfg.Git.HideKey().Key
```
squaremo: Perhaps we should rename `.HideKey().Key` to `PublicKey()`

squaremo: or the other permutation!

http/transport.go
```diff
@@ -668,6 +671,64 @@ func invokeGenerateKeys(client *http.Client, t flux.Token, router *mux.Router, e
...
+		gh := git.Github{
+			Client: http.Client{
+				Timeout: 10 * time.Second,
+			},
+		}
+		err = gh.InsertDeployKey(inst, owner, repo, tok, publicKey)
+		if err != nil {
+			httpErr, isHttpErr := err.(*httperror.APIError)
+			if isHttpErr {
+				w.WriteHeader(httpErr.StatusCode)
```
squaremo: Do we definitely want to pass on exactly the code GitHub returned? (Does this _happen_ to be correct, or is it a shortcut ..)

git/github.go
squaremo: Using the external instance ID ("smokey-water-72") would be great, but I don't think we have access to that.

philwinder: Note: Change key name to `"flux deploy"`.
Add "please use one repo per instance for now" to https://github.com/weaveworks/flux/pull/411.

paulbellamy:
What happens if a user is running in standalone mode?

foot:
I was getting 404s from this response so poked around w/ the auth stuff.

I can get a 200 if I change the new scopes in service from `write:public_key` to `repo` (full repo access), can you confirm this works for you too or if I'm doing something funny?

philwinder:
@foot I think you must have been doing something wrong. These are the scopes requested by service: https://github.com/weaveworks/service/pull/1102/files#diff-4c9a2f04e3d00c8925a3a5500eef166cR24
And they work for me?

@paulbellamy This isn't meant for standalone. This is for the weave cloud use case only. If you wanted to make it work with standalone then we'd have to add parameters for GH oauth tokens, etc.

http/transport.go
philwinder: I would have thought it was better to pass on the reason for the error?

git/github.go
philwinder: After discussion decided to keep it fixed as something like: `flux auto-generated key`

http/transport.go
philwinder: It's used elsewhere to hide the key. So I'd rather not.

philwinder:
@squaremo Review items fixed.
I've updated the description at the top.

git/deploy_key.go
```diff
@@ -0,0 +1,101 @@
...
+	k := deployKey{
+		Github: g,
+		owner:  o,
+		repo:   r,
+		key:    key,
+	}
+	// Set the key values from the remote
+	err := k.Get()
+	if err != nil {
+		return &k, err
```
squaremo: Should this be `nil`? (if err is checked, it doesn't matter I guess, but perhaps more idiomatic)

git/deploy_key.go
```diff
@@ -0,0 +1,101 @@
+package git
```
squaremo: I think this goes in `github/` rather than here.

git/github.go
```diff
@@ -0,0 +1,174 @@
...
+type owner struct {
+	name  string
+	token string
+}
+
+type repo struct {
+	name string
+}
+
+type deployKey struct {
```
squaremo: The type and its operations ought to be in the same file, I reckon.

squaremo: Actually I'm not sure why a deploy key isn't just a value, with Github having the operations to get, create, etc.

git/deploy_key.go
squaremo: Or even `integrations/github/`

git/deploy_key.go
philwinder: Yeah, Get isn't quite the right name. It has no return values. Will change name.

philwinder:
@squaremo Moved packaged and renamed `Get`. I don't know what you mean by "most of the operations could be inlined". Yes, I'm sure they could. But why would you want to? Clean code, et al.?

And I also didn't understand what you wanted me to do about the comment "deploy key just a value". It split out the deploy key actions because that was a good split in the domain. The different levels between Github and the deploy key. It helped to reduce the amount of code in Github.

squaremo:
> And I also didn't understand what you wanted me to do about the comment "deploy key just a value".
>  I don't know what you mean by "most of the operations could be inlined".

The thing this code needs to accomplish is to find a deploy key in github and replace it with one that's been generated. To do this,

 1. you create a `Github{}` and call `InsertDeployKey` on it; then
 2. the `Github{}` allocates a few values, then calls `newDeployKey`, to which it supplies itself as an argument; as part of that call,
 3. a `deployKey` struct is allocated with the `Github` in it; it calls `Synchronise`, before returning, which
 4. uses the supplied `Github` to request the keys from github, calling `newGetKeysRequest` on the `deployKey` to make one of the arguments; the call to Github is just a typed wrapper around using the HTTP client, which happens to be in the Github. When this done, it mutates some fields of the newly allocated `deployKey` and returns, at which point
 5. the Github (still in that first invocation) checks a field in the deployKey and deletes the key is it exists (let's ignore that for now); then it calls `Create` on the key to insert the deploy key, which
 6. .. is just a wrapper for calling Github back again.

To compile that description I had to shuttle backwards and forwards between those two files at pretty much every step. Wouldn't it be much simpler (and much less code) if there was just one type that did the work?

squaremo:
> Wouldn't it be much simpler (and much less code) if there was just one type that did the work?

Or put it this way -- it doesn't seem like `deployKey` is any use independent of `Github`, so why not just put it in the same place and inline the methods that are really just there to make it its own thing.

I think this would result in less and simpler code, but if you think that's not likely, or something of more value would be lost, please merge as-is.

philwinder:
Ok, will have a go at inlining and see what it looks like.

PR #432, opened 2017-02-08
Async releases and shutdown

paulbellamy:
Unfortunately, [this](https://github.com/weaveworks/flux/pull/415) added a race with the slack notification when releasing flux itself. This fixes that.

Need to move the yaml changes out to everywhere (launch-generator, and service-conf).

### DIFF
--- a/deploy/standalone/flux-deployment.yaml
+++ b/deploy/standalone/flux-deployment.yaml
@@ -4,11 +4,25 @@ metadata:
   name: flux
 spec:
   replicas: 1
+
+  # Having maxSurge 0 and maxUnavailable 1 means the deployment will update one
+  # replica at a time as it will have to stop one (making one unavailable)
+  # before it can start one (surge of zero)
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+
   template:
     metadata:
       labels:
         name: flux
     spec:
+      # Give fluxsvc up to 10 minutes grace to finish jobs in progress and exit
+      # cleanly. Service is available during this time, as long as we don't
+      # stop too many replicas at once.
+      terminationGracePeriodSeconds: 600
+
       containers:
       - name: fluxd
         image: quay.io/weaveworks/fluxd:master-952c028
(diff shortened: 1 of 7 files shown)


squaremo:
How will this behave with `fluxd`'s that ignore the extra argument? (Do we care? Maybe not much)

paulbellamy:
As long as they are in a deployment, `kubectl apply` should reboot them. They will fail to respond to the RPC and disconnect, though. hmmm...

squaremo:
> How will this behave with fluxd's that ignore the extra argument?

Ah, I missed that it only sends it for fluxsvc and fluxd anyway (and we will be running a new fluxd ourselves).

paulbellamy:
Yes, and we've never told anyone (afaik) to install fluxd with a replicationcontroller, (only deployments, which are well behaved).

deploy/standalone/flux-deployment.yaml
```diff
@@ -4,11 +4,25 @@ metadata:
   name: flux
 spec:
   replicas: 1
+
+  # Having maxSurge 0 and maxUnavailable 1 means the deployment will update one
```
squaremo: But why is this important or necessary? (some explanation in the comment would be good)

platform/kubernetes/release.go
```diff
@@ -86,27 +86,38 @@ func (c *Cluster) doApplyCommand(logger log.Logger, newDefinition *apiObject, ar
...
 	return func(c *Cluster, logger log.Logger) error {
-		return c.doApplyCommand(
-			logger,
-			newDef,
-			"rolling-update",
-			"--update-period", "3s",
-			def.Name,
-			"-f", "-", // take definition from stdin
-		)
+		err := make(chan error)
```
squaremo: `errc` for consistency ...

deploy/standalone/flux-deployment.yaml
paulbellamy: If we have more than one replica, this will ensure we deploy 1-at-a-time so they stay available while deploying.

PR #435, opened 2017-02-09
Panic! at the slack templating

paulbellamy:
Easier to go through each commit separately.
The first one is the important one, so if the second is contentious, I'll break it out.

### DIFF
--- a/cmd/fluxctl/release_cmd.go
+++ b/cmd/fluxctl/release_cmd.go
@@ -60,16 +60,19 @@ func (opts *serviceReleaseOpts) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if err := checkExactlyOne("--service=<service>, or --all", len(opts.services) > 0, opts.allServices); err != nil {
-		return err
+	if len(opts.services) <= 0 && !opts.allServices {
+		return newUsageError("please supply either --all, or at least one --service=<service>")
 	}
 
 	var services []flux.ServiceSpec
 	if opts.allServices {
 		services = []flux.ServiceSpec{flux.ServiceSpecAll}
 	} else {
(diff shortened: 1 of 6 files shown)


cmd/fluxctl/release_cmd.go
```diff
@@ -60,16 +60,23 @@ func (opts *serviceReleaseOpts) RunE(cmd *cobra.Command, args []string) error {
...
 
-	service, err := parseServiceOption(opts.service) // will be "" iff opts.allServices
-	if err != nil {
-		return err
+	var services []flux.ServiceSpec
+	if opts.allServices {
+		services = []flux.ServiceSpec{flux.ServiceSpecAll}
+	} else {
+		for _, spec := range opts.services {
+			services = append(services, flux.ServiceSpec(spec))
```
squaremo: Would it prevent a few mishaps if this parsed the spec as a `ServiceID` before turning it into a `ServiceSpec`? Not a big deal, since this will happen at the server anyway.

cmd/fluxctl/release_cmd.go
```diff
@@ -60,16 +60,23 @@ func (opts *serviceReleaseOpts) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if err := checkExactlyOne("--service=<service>, or --all", opts.service != "", opts.allServices); err != nil {
+	if err := checkExactlyOne("--service=<service>, or --all", len(opts.services) > 0, opts.allServices); err != nil {
```
squaremo: What we want to check is either `--all` or _at least one_ `--service` -- this will check that, but the error message (if you supply nothing, for example) will be "please supply exactly one of .." rather than "please supply either ...".

PR #444, opened 2017-02-14
add getting help channels

abuehrle:
@squaremo added new slack channels and other info on getting help.


squaremo:
Good call adding these @abuehrle, thank you!

PR #449, opened 2017-02-17
refactor metrics instantiation in fluxsvc for easier testing

philwinder:
Created a new method to instantiate all metrics for ease of use.
Moved hardcoded metrics to dedicated metrics files in the same package as the rest of the code.


paulbellamy:
Could take this even further, ditch all the `newFoo` methods and just declare each as a global `var` in their respective packages. They're always the same, and `newFoo` is a bit of a lie since it modifies global state in the prom client anyway.

PR #450, opened 2017-02-20
Add component tests to fluxsvc

philwinder:
The goal is to fully test the fluxsvc API. (Thereby decoupling fluxctl and fluxd from fluxsvc).
The fluxsvc server will be started and all dependencies are in memory.
I've reduced the ssh key size again because it was taking too long to generate a key of length 4096.
Most of the tests are using the various `Invoke*` methods in `transport.go`, rather than calling the API directly.
Where possible, the effects are tested, e.g. insertion into DB.


squaremo:
> Most of the tests are using the various Invoke* methods in transport.go, rather than calling the API directly.

How come this, rather than creating a client and using that? (And leaving all the invoke* procedures as package level)

philwinder:
@squaremo Yes, should have done that. Don't know why I didn't. Will change.

squaremo:
Rebassed and ready to go.

cmd/fluxsvc/fluxscv_test.go
```diff
@@ -0,0 +1,511 @@
+package main
```
paulbellamy: typo in filename.

git/keys.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // keySize is the size of generated private keys.
-var keySize = 4096
+var keySize = 2048
```
paulbellamy: why downgrade this?

philwinder: Because:
a) the latency for the user was too long (seconds)
b) some of the tests failed because of the delay
c) the tests took ages

cmd/fluxsvc/fluxscv_test.go
philwinder: Damn you're good.

philwinder:
Thanks for review Paul.
Yes, you're right about being coupled. But it does use dummy data to enact those apis. It is using the apis as users would use them. So I think it will catch issues like messing with the api methods, and changing the JSON definitions and things like that.

philwinder:
Just noticed that @squaremo has done some tinkering. Thank you!

git/keys.go
squaremo: a) that is OK if 4096 it's what github recommends
b) fine; set it for the tests only
c) set it for the tests only

PR #454, opened 2017-02-22
Scaffold for humane errors

### LINKED ISSUE #56: Humane errors
We should use some kind of enhanced error package ([e.g.](https://github.com/pkg/errors)) to produce humane i.e. human-readable errors at the fluxctl command line. I'm thinking something multi-line and literate: this has happened, here are the likely causes, here are some avenues to pursue for debugging/resolution.

- [ ] if the platform is unavailable, we should explain it might not be connected, or if it is (how can you check?), it could be worth trying the operation again (this includes "nats timeout")
- [ ] if the git repo can't be read, that often means the deploy key is missing, or the git URL is wrong
- [ ] fetching image metadata often fails. In this case, trying again may well work; or check the specific repository exists etc.
- [x] `GetRelease` returns a 500 when a job id doesn’t exist: [source](https://github.com/weaveworks/flux/blob/master/http/transport.go#L296)

squaremo:
We would like to take an incremental approach to improving the errors returned via the API. For that reason, we have to deal with any old error coming through, while doing better where we can.

The approach taken here is to test for some variety of flux.BaseError, and encode into the HTTP response accordingly. Content negotiation is used to determine whether to include a JSON error. Older clients won't include an Accept header, and so will get the output they expect.

The humane errors come with help text which can be printed out verbatim (most use for the CLI), and the original error for logging etc.

Fixes #56. Checklist moved from there:

- [x] if the platform is unavailable, we should explain it might not be connected, or if it is (how can you check?), it could be worth trying the operation again (this includes "nats timeout")
- [x] if the git repo can't be read, that often means the deploy key is missing, or the git URL is wrong
- [ ] fetching image metadata often fails. In this case, trying again may well work; or check the specific repository exists etc. (less of a problem nowadays? can be done in another PR, anyway)
- [x] `GetRelease` returns a 500 when a job id doesn’t exist: [source](https://github.com/weaveworks/flux/blob/master/http/transport.go#L296)


errors.go
```diff
@@ -0,0 +1,95 @@
...
+	return &BaseError{
+		Err: err,
+		Help: `An error occured for which we don't have a specific message.
+
+If you see this, it means we need to come up with a better message! It
+would help us if you log an issue at
+https://github.com/weaveworks/flux/issues saying what you were doing
+when you saw this, and quoting the following:
+
+    ` + err.Error(),
```
paulbellamy: I'd favour this more if the actual error message were more prominent, and how to report it was more secondary... maybe? Need an example of where this would come up...

git/errors.go
```diff
@@ -0,0 +1,73 @@
...
+	return flux.UserConfigProblem{&flux.BaseError{
+		Err: actual,
+		Help: `Problem cloning your git repository
+
+There was a problem cloning your git repository,
+
+    ` + url + `
+
+This may be because you have not supplied a valid deploy key, or
+because the repository has been moved or deleted (or never existed!).
```
paulbellamy: `[...] has been moved, deleted, or never existed.`

git/errors.go
```diff
@@ -0,0 +1,73 @@
...
+		Err: actual,
+		Help: `Problem committing and pushing to git repository.
+
+There was a problem with committing changes and pushing to the git
+repository. Since the repository had to be cloned to get to this
+point, this probably means that the repository exists but the deploy
+key provided doesn't have write permission.
+
+In GitHub, please check via the repository settings that the deploy
+key is "Read/write". You can cross-check the fingerprint with that
```
paulbellamy: `Read/Write`

http/transport.go
```diff
@@ -502,11 +518,14 @@ func handleIsConnected(s api.FluxService) http.Handler {
...
-			w.WriteHeader(http.StatusNotFound)
-		case nil:
+		err := s.IsDaemonConnected(inst)
+		if err == nil {
 			w.WriteHeader(http.StatusNoContent)
+			return
+		}
+		switch err.(type) {
+		case flux.UserConfigProblem:
+			w.WriteHeader(http.StatusNotFound)
```
paulbellamy: duplicate of `func errorResponse`

http/transport.go
```diff
@@ -557,6 +576,63 @@ func getInstanceID(req *http.Request) flux.InstanceID {
...
+	case flux.ServerException:
+		code = http.StatusInternalServerError
+		outErr = err.BaseError
+	default:
+		code = http.StatusInternalServerError
+		outErr = flux.CoverAllError(apiError)
+	}
+
+	writeError(w, r, code, outErr)
+}
```
paulbellamy: When doing this elsewhere, I've added a `http/render` package with `render.Error(w, r, err)`.

Aside: and generally a `render.Json(w, r, statusCode, obj)` as well

git/errors.go
squaremo: In the Github UI which I copied from, it's "Read/write".

errors.go
squaremo: Can turn it around to e.g.,

```
Error: platform unavailable: nats timeout

We don't have any help specific to this error. If you see this, it would help us if you log it an issue at

    https://github.com/weaveworks/flux/issues

saying what you were doing when you saw this, and quoting the message at the above.
```

git/errors.go
squaremo: A bit less dramatic, true ...

http/transport.go
squaremo: Not quite; `errorResponse` would write a 422 Unprocessable Entity (as a stand in for "your request is fine, I just can't do it right now"), whereas this is expected to return No Content, Not Found, or Aaaaargh.

git/errors.go
paulbellamy: :(

PR #439, opened 2017-02-13
More-detailed events

paulbellamy:
Fixes: https://github.com/weaveworks/flux/issues/374

Proposed format is:

```
{
  "id": 1234,

  "instanceID": "fuzzy-lunch-14",

  "serviceIDs": ["default/helloworld", "fluxy/fluxsvc"],

  // release|automate|deautomate|lock|unlock
  "type": "release",

  "startedAt": "2017-01-02T01:02:03.456789Z",

  // endedAt will be the same as StartedAt (except for new releases)
  "endedAt": "2017-01-02T01:12:03.456789Z",

  // debug|info|warn|error (for color-coding)
  "logLevel": "info",

  // pre-formated string for errors and other stuff, included for
  // backwards compatibility, but won't be included for new events.
  // Basically, they'll either have "metadata" (below), or a "message".
  "message": "old event message here",

  // event-type specific metadata, might be undefined
  "metadata": {
    "job": { ... }, // Or something else?
    "imageIDs": ["nginx:latest"], // For now there will only be one thing here, but...
    "error": "" // Error string, will be undefined if there wasn't one.
  }
}
```

## TODO

- [x] Figure out how to include the release spec and result in the format
  - [x] Integrate changes from https://github.com/weaveworks/flux/pull/421 for new releaseresult stuff
- [x] Ask simon if the format works for him
- [x] Check this doesn't break the UI (or fix the UI at the same time)
- [x] Double-check backwards-compatibility
- [ ] ~~Do fluxctl release when merging this~~
- [ ] Drop the old history table (https://github.com/weaveworks/flux/pull/470)

PR #470, opened 2017-02-24
Drop history table

paulbellamy:
Follow on from: https://github.com/weaveworks/flux/pull/439


paulbellamy:
Don't merge until #439 has been released through to prod and checking history still works!

PR #472, opened 2017-02-27
Refactor http transport

paulbellamy:
* Refactored out a findBinary method for kubeservice, to make it's usage clearer
* Refactored metrics, so that it is easier to query them in aggregate.
* Refactored http/transport to inline `invokeFoo` functions into the client (saves ~200 repeated lines)


http/client.go
```diff
@@ -27,53 +27,156 @@ func NewClient(c *http.Client, router *mux.Router, endpoint string, t flux.Token
...
+	if err != nil {
+		return "", errors.Wrap(err, "executing HTTP request")
+	}
+	defer resp.Body.Close()
+
+	var res postReleaseResponse
+	if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
+		return "", errors.Wrap(err, "decoding response from server")
+	}
+	return res.ReleaseID, nil
```
paulbellamy: Annoyingly, `PostRelease` is the only post with a response body. Would be nice if they could all use the same `c.post` method. Suggestions on how to do that, welcome...

server/metrics.go
```diff
@@ -7,41 +7,13 @@ import (
 )
 
 var (
-	statusDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
+	requestDuration = prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
```
squaremo: and analogues below :100: :+1: :1st_place_medal:

squaremo: Do we even look at these -- they are quite close to duplicating the API request metrics.

paulbellamy: True

PR #479, opened 2017-03-06
Refactor transport again

squaremo:
* Collect the `api.FluxService -> http.HandlerFunc` into `http.HandlerFunc`-typed methods on a struct, as a simplification.
 * Admitting a small amount of commonality, the HTTP server and client can be in different
packages.

(Could more be done?)

### DIFF
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -14,6 +14,7 @@ import (
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/api"
 	transport "github.com/weaveworks/flux/http"
+	"github.com/weaveworks/flux/http/client"
 )
 
 type rootOpts struct {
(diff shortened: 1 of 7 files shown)


cmd/fluxctl/root_cmd.go
```diff
@@ -91,7 +92,7 @@ func (opts *rootOpts) PersistentPreRunE(cmd *cobra.Command, _ []string) error {
 		return errors.Wrapf(err, "parsing URL")
 	}
 	opts.Token = getFromEnvIfNotSet(cmd.Flags(), "token", envVariableToken, opts.Token)
-	opts.API = transport.NewClient(http.DefaultClient, transport.NewRouter(), opts.URL, flux.Token(opts.Token))
+	opts.API = client.NewClient(http.DefaultClient, transport.NewRouter(), opts.URL, flux.Token(opts.Token))
```
paulbellamy: `client.New(...)`

cmd/fluxsvc/main.go
```diff
@@ -234,7 +235,7 @@ func main() {
 		logger.Log("addr", *listenAddr)
 		mux := http.NewServeMux()
 		mux.Handle("/metrics", promhttp.Handler())
-		handler := transport.NewHandler(server, transport.NewRouter(), logger)
```
paulbellamy: It would be nice to use the common http server at some point for common logging and instrumentation, but fine for now.

cmd/fluxctl/root_cmd.go
squaremo: Good call

PR #482, opened 2017-03-06
Implement `fluxctl save`

### LINKED ISSUE #460: bootstrap/save functionality
Will need fluxd changes to AllServices, so that it returns enough information to generate a manifest.

Undecided is whether this should commit straight into git, or should save them locally, then let the user commit and push them.

If we are doing it locally, we should detect and warn the user if it isn't in a git repo..

awh:
Implement `fluxctl save` command that exports k8s objects to YAML files in the current directory:

* Handles Namespaces, Deployments, Services and ReplicationControllers
* Filters out add-ons from `kube-system`
* Creates directory structure as per service-conf/k8s
* Attributes that should not be persisted in version control (e.g status, uid, automatically maintained annotations, creationTimestamp etc) are filtered prior to writing

Sample output:
```
$ fluxctl save
Saving Namespace 'default' to default-ns.yaml
Saving Deployment 'flux' to default/flux-dep.yaml
Saving Deployment 'memcached' to default/memcached-dep.yaml
Saving Service 'fluxsvc' to default/fluxsvc-svc.yaml
Saving Service 'kubernetes' to default/kubernetes-svc.yaml
Saving Service 'memcached' to default/memcached-svc.yaml
Saving Namespace 'kube-system' to kube-system-ns.yaml
```

Outstanding questions:

* Do we want to add any CLI options, e.g. ability to specify a target directory, include add-ons, don't filter attributes
* Do we want to tackle implementing Export as a new protocol (as discussed below) as part of this PR?
* Should Export include type & version information so that consumers can change their behaviour accordingly?
* Are there any other things that we should filter out of objects?
* Should we abort if target files already exist? Current implementation overwrites without warning

Fixes #460.

### DIFF
--- a/cmd/fluxctl/save_cmd.go
+++ b/cmd/fluxctl/save_cmd.go
@@ -1,0 +1,215 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	yaml "gopkg.in/yaml.v2"
+)
+
+type saveOpts struct {
+	*rootOpts
+	output string
+}
+
+func newSave(parent *rootOpts) *saveOpts {
+	return &saveOpts{rootOpts: parent}
+}
+
+func (opts *saveOpts) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "save",
+		Short: "save service definitions to local files in platform-native format",
+		Example: makeExample(
+			"fluxctl save",
+		),
+		RunE: opts.RunE,
+	}
+	return cmd
+}
+
+// Deliberately omit fields (e.g. status, metadata.uid) that we don't want to save
+type saveObject struct {
+	APIVersion string `yaml:"apiVersion,omitempty"`
+	Kind       string `yaml:"kind,omitempty"`
+
+	Metadata struct {
+		Annotations map[string]string `yaml:"annotations,omitempty"`
+		Labels      map[string]string `yaml:"labels,omitempty"`
+		Name        string            `yaml:"name,omitempty"`
+		Namespace   string            `yaml:"namespace,omitempty"`
+	} `yaml:"metadata,omitempty"`
+
+	Spec map[interface{}]interface{} `yaml:"spec,omitempty"`
+}
+
+func (opts *saveOpts) RunE(_ *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return errorWantedNoArgs
+	}
+
+	config, err := opts.API.Export(noInstanceID)
+	if err != nil {
+		return errors.Wrap(err, "exporting config")
+	}
+
+	yamls := bufio.NewScanner(bytes.NewReader(config))
+	yamls.Split(splitYAMLDocument)
+
+	for yamls.Scan() {
+		var object saveObject
+		// Most unwanted fields are ignored at this point
+		if err := yaml.Unmarshal(yamls.Bytes(), &object); err != nil {
+			return errors.Wrap(err, "unmarshalling exported yaml")
+		}
+
+		// Filter out remaining unwanted keys from unstructured fields
+		// e.g. .Spec and .Metadata.Annotations
+		filterObject(object)
+
+		if err := saveYAML(object); err != nil {
+			return errors.Wrap(err, "saving yaml object")
+		}
+	}
+
+	if yamls.Err() != nil {
+		return errors.Wrap(yamls.Err(), "splitting exported yaml")
+	}
+
+	return nil
+}
+
+// Remove any data that should not be version controlled
+func filterObject(object saveObject) {
+	delete(object.Metadata.Annotations, "deployment.kubernetes.io/revision")
+	delete(object.Metadata.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+	deleteNested(object.Spec, "template", "metadata", "creationTimestamp")
+	deleteEmptyMapValues(object.Spec)
+}
+
+// Recurse through nested maps to remove a key
+func deleteNested(m map[interface{}]interface{}, keys ...string) {
+	switch len(keys) {
+	case 0:
+		return
+	case 1:
+		delete(m, keys[0])
+	default:
+		if v, ok := m[keys[0]].(map[interface{}]interface{}); ok {
+			deleteNested(v, keys[1:]...)
+		}
+	}
+}
+
+// Recursively delete map keys with empty values
+func deleteEmptyMapValues(i interface{}) bool {
+	switch i := i.(type) {
+	case map[interface{}]interface{}:
+		if len(i) == 0 {
+			return true
+		} else {
+			for k, v := range i {
+				if deleteEmptyMapValues(v) {
+					delete(i, k)
+				}
+			}
+		}
+	case []interface{}:
+		if len(i) == 0 {
+			return true
+		} else {
+			for _, e := range i {
+				deleteEmptyMapValues(e)
+			}
+		}
+	case nil:
+		return true
+	}
+	return false
+}
+
+// Save YAML to directory structure
+func saveYAML(object saveObject) error {
+	var path string
+	if object.Kind == "Namespace" {
+		path = fmt.Sprintf("%s-ns.yaml", object.Metadata.Name)
+	} else {
+		dir := object.Metadata.Namespace
+		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+			return errors.Wrap(err, "making directory for namespace")
+		}
+
+		shortKind := abbreviateKind(object.Kind)
+		path = fmt.Sprintf("%s/%s-%s.yaml", dir, object.Metadata.Name, shortKind)
+	}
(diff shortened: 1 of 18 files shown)


squaremo:
> any advances on Export for the new platform method?

I think that name is fine; what it does is what I'd expect.

platform/rpc/client.go
```diff
@@ -91,6 +91,20 @@ func (p *RPCClient) Version() (string, error) {
 	return version, err
 }
 
+// Export is used to get service configuration in platform-specific format
+func (p *RPCClient) Export(ids []flux.ServiceID) (string, error) {
+	var config string
+	err := p.client.Call("RPCServer.Export", ids, &config)
+	if _, ok := err.(rpc.ServerError); !ok && err != nil {
+		return "", platform.FatalError{err}
+	} else if err != nil && err.Error() == "rpc: can't find method RPCServer.Export" {
```
squaremo: I would much rather `Export` formed part of a new protocol, so there weren't special cases like this. (If we change more things, which seems inevitable .. more special cases ...)

platform/rpc/client.go
```diff
@@ -91,6 +91,20 @@ func (p *RPCClient) Version() (string, error) {
...
+		// "Export" is not supported by this version of fluxd (it is old). Fail
+		// gracefully.
+		return "unknown", nil
```
squaremo: This does not seem very graceful to me. Won't it mean clients don't see any error, but have to check explicitly for a result of "unknown"?

server/server.go
```diff
@@ -475,3 +475,12 @@ func (p *loggingPlatform) Version() (v string, err error) {
 	}()
 	return p.platform.Version()
 }
+
+func (p *loggingPlatform) Export(ids []flux.ServiceID) (config string, err error) {
+	defer func() {
+		if err != nil {
+			p.logger.Log("method", "Export", "error", err, "config", config)
```
squaremo: `config` could be pretty large here, I'm not sure if we really want to log it even on error.

platform/rpc/client.go
awh: I don't understand what you're proposing here - what does 'form part of a new protocol' look like in concrete terms?

platform/rpc/client.go
awh: That seems fair, but perhaps `Version` should return an error too then (or form part of a new protocol?)

platform/rpc/client.go
squaremo: Yes that was a bit vague, sorry. I meant,

 - that there's a new interface under platform that includes this extra method (and possibly other changes);
 - daemons that support the new interface connect to a new websocket endpoint, and are represented on the server side by a new implementation;
 - daemons that only support the old interface connect to the old endpoint, and are represented on the server side by an implementation that adapts where possible and returns an error otherwise.

This is a lot of trouble to go to, to avoid this one special case! But I expect more changes of this kind, and it would be good to get scaffolding in at some point. If not here, then very soon.

Also, failing the RPC is a fatal error and will kill the websocket connection. Probably not a big problem in practice, since it will reconnect, but unnecessarily messy.

platform/rpc/client.go
squaremo: Yes, it probably should. Version is often made a special case (usually because it's thought of after the fact, as it was here; and sometimes because it's used for control flow, which it isn't here). But it'd be cleaner to do it as a protocol change, I think.

awh:
Sample output:
```
$ fluxctl save
Saving Namespace 'default' to default-ns.yaml
Saving Deployment 'flux' to default/flux-dep.yaml
Saving Deployment 'memcached' to default/memcached-dep.yaml
Saving Service 'fluxsvc' to default/fluxsvc-svc.yaml
Saving Service 'kubernetes' to default/kubernetes-svc.yaml
Saving Service 'memcached' to default/memcached-svc.yaml
Saving Namespace 'kube-system' to kube-system-ns.yaml
```

cmd/fluxctl/save_cmd.go
```diff
@@ -0,0 +1,215 @@
...
+	if object.Kind == "Namespace" {
+		path = fmt.Sprintf("%s-ns.yaml", object.Metadata.Name)
+	} else {
+		dir := object.Metadata.Namespace
+		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+			return errors.Wrap(err, "making directory for namespace")
+		}
+
+		shortKind := abbreviateKind(object.Kind)
+		path = fmt.Sprintf("%s/%s-%s.yaml", dir, object.Metadata.Name, shortKind)
```
squaremo: Is `filepath.Join` is more ... accurate, or something. Probably does not make a practical difference mind you.

cmd/fluxctl/save_cmd.go
```diff
@@ -0,0 +1,215 @@
...
+
+func abbreviateKind(kind string) string {
+	switch kind {
+	case "Service":
+		return "svc"
+	case "ReplicationController":
+		return "rc"
+	case "Deployment":
+		return "dep"
+	default:
```
squaremo: I think we have
```
  case "DaemonSet": return "ds"
  case "ConfigMap": return "config"
  case "Secret": return "secret"
```
too. (Could lowercase in the default case to cover that last, and be consistent otherwise.)

cmd/fluxctl/save_cmd.go
```diff
@@ -0,0 +1,215 @@
...
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	yaml "gopkg.in/yaml.v2"
+)
+
+type saveOpts struct {
+	*rootOpts
+	output string
```
squaremo: This doesn't appear to be used

cmd/fluxctl/save_cmd.go
```diff
@@ -0,0 +1,215 @@
...
+	if len(args) > 0 {
+		return errorWantedNoArgs
+	}
+
+	config, err := opts.API.Export(noInstanceID)
+	if err != nil {
+		return errors.Wrap(err, "exporting config")
+	}
+
+	yamls := bufio.NewScanner(bytes.NewReader(config))
```
squaremo: This is spelled `yamels` :)

http/server/server.go
```diff
@@ -388,6 +389,17 @@ func (s HTTPService) IsConnected(w http.ResponseWriter, r *http.Request) {
...
 
+func (s HTTPService) Export(w http.ResponseWriter, r *http.Request) {
+	inst := getInstanceID(r)
+	status, err := s.service.Export(inst)
+	if err != nil {
+		errorResponse(w, r, err)
+		return
+	}
+
+	jsonResponse(w, r, status)
```
squaremo: Is this going to be a whole bunch of bytes written out as an array of integers?

platform/kubernetes/kubernetes.go
```diff
@@ -435,6 +437,79 @@ func (c *Cluster) Version() (string, error) {
...
+		for _, service := range services.Items {
+			if isAddon(&service) {
+				continue
+			}
+			err := appendYAML(&config, "v1", "Service", service)
+			if err != nil {
+				return nil, errors.Wrap(err, "marshalling service to YAML")
+			}
+		}
+	}
```
squaremo: Ah OK, just sticking with deployments, namespaces and services for now, that explains the cases when picking a short name, above. We can add to these later, although perhaps we should talk about protocol changes first.

http/server/server.go
```diff
@@ -388,6 +389,17 @@ func (s HTTPService) IsConnected(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s HTTPService) Export(w http.ResponseWriter, r *http.Request) {
+	inst := getInstanceID(r)
+	status, err := s.service.Export(inst)
```
squaremo: `status`?

squaremo:
> Do we want to add any CLI options, e.g. ability to specify a target directory, include add-ons, don't filter attributes

I reckon add the first, target directory. The others we can live without, until we can't.

> Do we want to tackle implementing Export as a new protocol (as discussed below) as part of this PR?

Yes, I think so; we can collect it into a provisional new daemon protocol version.

> Should Export include type & version information so that consumers can change their behaviour accordingly?

Type and version of what? (`Kubernetes, v1.6`? or something else)

> Are there any other things that we should filter out of objects?
> Should we abort if target files already exist? Current implementation overwrites without warning

Yes, definitely. I'd include a `--force,-f`.

cmd/fluxctl/save_cmd.go
```diff
@@ -0,0 +1,243 @@
...
+// Save YAML to directory structure
+func saveYAML(object saveObject, out string) error {
+	buf, err := yaml.Marshal(object)
+	if err != nil {
+		return errors.Wrap(err, "marshalling yaml")
+	}
+
+	// to stdout
+	if out == "-" {
+		fmt.Fprintln(os.Stdout, "---")
```
paulbellamy: please add prefixing `---` when to file also.

paulbellamy: please add prefixing `---` when to file also.

PR #485, opened 2017-03-09
if we are pid 1, reap child processes, as git leaves zombies

### LINKED ISSUE #484: Shelling out to git leaves orphan zombie processes
Git expects that it can leave orphan processes (e.g. git-remote-http, and sh) around and that init will reap them.

Because we run fluxsvc as pid1 in our fluxsvc container (not init), we do not reap child processes correctly. This leads to a build-up of defunct git and sh processes like so:
```
fluxsvc-+-git---git-remote-http
        |-40400*[git-remote-http]
        `-2123*[sh]
```

More info: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

This is the cause of: https://github.com/weaveworks/service/issues/1144

paulbellamy:
Fixes #484 
Fixes https://github.com/weaveworks/service/issues/1144


paulbellamy:
Hmm.. doing it on sigchild (or at least this library) has introduced a race condition. odd.

PR #486, opened 2017-03-10
Package and command for diffing Kubernetes objects

squaremo:
This adds some machinery for diffing generic objects, then specialises that to the Kubernetes objects that we typically deal with, then use that to implement a `fluxctl diff` command.

The package `diff` provides generic representations of diffs, and some algorithms for diffing values of various kinds. There is a specialisation hook: implementing the interface `Differ` for a type means that will get used instead of the generic diff algorithm.

The package `platform/kubernetes/diff` contains some of the types used in Kubeternetes config, and some procedures for loading them from a file or directory, and some specialisations via `Differ` (e.g., for secrets). 

The idea here is to build the types out of the fields that matter when applying -- then it can be used to see if, for example, a commit contains a "meaningful" change, and therefore whether to apply it to the cluster. I have probably not captured all such fields; it may take a bit of trial and error to figure out what we care about.

Lastly, this adds a `fluxctl diff` command, which just checks one file (or directory) against another. This is mainly just to prove the use of the packages, but may eventually be useful in itself.

### DIFF
--- a/cmd/fluxctl/diff_cmd.go
+++ b/cmd/fluxctl/diff_cmd.go
@@ -1,11 +1,11 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
+	"io"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
-	yaml "gopkg.in/yaml.v2"
 
 	"github.com/weaveworks/flux/diff"
 	"github.com/weaveworks/flux/platform/kubernetes/resource"
@@ -13,7 +13,7 @@ import (
 
 type diffOpts struct {
 	*rootOpts
-	format string
+	quiet bool
 }
 
 func newDiff(root *rootOpts) *diffOpts {
@@ -26,48 +26,13 @@ func (opts *diffOpts) Command() *cobra.Command {
 		Short: "Show differences between one platform config and another",
 		RunE:  opts.RunE,
 	}
-	cmd.Flags().StringVarP(&opts.format, "output", "o", "text", "(yaml|json|text) whether to output differences in YAML or JSON, or just summarise in text")
+	cmd.Flags().BoolVarP(&opts.quiet, "quiet", "q", false, "just report which files differ")
 	return cmd
 }
 
 func (opts *diffOpts) RunE(cmd *cobra.Command, args []string) error {
 	if len(args) != 2 {
 		return newUsageError("please supply two filenames")
-	}
-
-	output := func(diffs diff.ObjectSetDiff) error {
-		diffs.Summarise(cmd.OutOrStdout())
-		return nil
-	}
-
-	switch opts.format {
-	case "text":
-		// already output
-	case "raw":
-		output = func(diffs diff.ObjectSetDiff) error {
-			fmt.Fprintf(cmd.OutOrStdout(), "%#v\n", diffs)
-			return nil
-		}
-	case "json":
-		output = func(diffs diff.ObjectSetDiff) error {
-			bytes, err := json.Marshal(diffs)
-			if err != nil {
-				return err
-			}
-			cmd.OutOrStdout().Write(bytes)
-			return nil
-		}
-	case "yaml":
-		output = func(diffs diff.ObjectSetDiff) error {
-			bytes, err := yaml.Marshal(diffs)
-			if err != nil {
-				return err
-			}
-			cmd.OutOrStdout().Write(bytes)
-			return nil
-		}
-	default:
-		return newUsageError("output format --output,-o must be 'raw', 'text', 'json' or 'yaml'")
 	}
 
 	a, err := resource.Load(args[0])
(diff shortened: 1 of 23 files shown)


cmd/fluxctl/diff_cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	if len(args) != 2 {
+		return newUsageError("please supply two filenames")
+	}
+
+	output := func(diffs diff.ObjectSetDiff) error {
+		diffs.Summarise(cmd.OutOrStdout())
+		return nil
+	}
+
+	switch opts.format {
```
squaremo: This is a bit pointless; I mainly wanted it for seeing the raw (`printf(%s)`) output for debugging.

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+// Difference represents a "chunk" of difference between two
+// `Object`s.
+type Difference interface {
+	Summarise(out io.Writer)
+	AtPath() string
+}
+
+type Chunk struct {
+	Deleted []interface{}
+	Added   []interface{}
```
philwinder: To have better type safety, could the added/deleted be strings? Or io.Readers?
Do they need to be interfaces?

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+	"reflect"
+	"sort"
+	"strings"
+)
+
+// Difference represents a "chunk" of difference between two
+// `Object`s.
+type Difference interface {
+	Summarise(out io.Writer)
+	AtPath() string
```
philwinder: Does a difference always have a path? 
E.g. if there is no file in the repository then there won't be a path, right?

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+
+type Object interface {
+	ID() ObjectID
+	Source() string
+}
+
+// ObjectSet is a set of several objects which can be diffed
+// collectively.
+type ObjectSet struct {
+	Source  string
```
philwinder: What is Source?

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+func MakeObjectSet(source string) *ObjectSet {
+	return &ObjectSet{
+		Source:  source,
+		Objects: map[ObjectID]Object{},
+	}
+}
+
+type ObjectSetDiff struct {
+	A, B      *ObjectSet
+	OnlyA     []Object
```
philwinder: Why do we need OnlyA/B? Isn't that information contained within the Differences.Added?

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+type OpaqueChunk struct {
+	Path string
+}
+
+func (c OpaqueChunk) AtPath() string {
+	return c.Path
+}
+
+// Objects are considered unique by {Namespace, Kind, Name}
+type ObjectID struct {
```
philwinder: General question. What happens if the user alters the namespace or name in the repository (unlikely to happen in the cluster)?
Then all elements in the repo will be marked as repo only, and the same for the cluster, but cluster only.
Probably fine.

diff/diff.go
philwinder: Yeah AFAICT they are always `[]Difference` types.

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	return diffValue(reflect.ValueOf(a), reflect.ValueOf(b), typA, "")
+}
+
+var differType = reflect.TypeOf((*Differ)(nil)).Elem()
+
+// Compare two reflected values and compile a list of differences
+// between them.
+func diffValue(a, b reflect.Value, typ reflect.Type, path string) ([]Difference, error) {
+	if typ.Implements(differType) {
```
philwinder: Could all the diff*'s implement the differType? Then you wouldn't need the switch below?

diff/diff.go
```diff
@@ -0,0 +1,293 @@
...
+			diffs = append(diffs, fieldDiffs...)
+		}
+	}
+	return diffs, nil
+}
+
+// diff each element, and include over- or underbite. TODO report an
+// array of ground values as a single chunk, rather than recursing.
+func diffArrayOrSlice(a, b reflect.Value, sliceTyp reflect.Type, path string) ([]Difference, error) {
+	var changed []Difference
```
philwinder: Grim. Well done, but grim. I wonder if there's a library to do this for you?

diff/diff_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	expected := []Difference{
+		Added("shamu", "map[four]"),
+		Removed("baz", "map[three]"),
+		Changed("bar", "bart", "map[two]"),
+	}
+	if !reflect.DeepEqual(expected, diffs) {
+		t.Errorf("expected diff:\n%#v\ngot:\n%#v\n", expected, diffs)
+	}
+}
```
philwinder: 💯

platform/kubernetes/resource/load.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	if err := chunks.Err(); err != nil {
+		return objs, fmt.Errorf(`scanning multidoc from "%s": %s`, source, err.Error())
+	}
+	return objs, nil
+}
+
+// ---
+// Taken directly from https://github.com/kubernetes/apimachinery/blob/master/pkg/util/yaml/decoder.go.
```
philwinder: Can't you reuse that package then?

platform/kubernetes/resource/spec.go
```diff
@@ -0,0 +1,106 @@
+package resource
+
+import (
+	"fmt"
+
+	"github.com/weaveworks/flux/diff"
+)
+
+// Types that daemonsets, deployments, and other things have in
```
philwinder: I'm worried that we're coupling ourselves to the k8s API here. I.e. we can't support older clusters (e.g. they don't have annotations) and we'd have to update if they add anything.

philwinder: And also we have the tedious task of going through each K8s type and adding it to this list.

squaremo:
Taking this off the radar (but leaving the branch for now)

PR #489, opened 2017-03-14
Protocol for syncing changes to the cluster

squaremo:
Adds a new method to platform, `Sync`, which is designed to be used for synchronising the cluster with git. Reimplements Apply in terms of Sync (vice-versa is not in general possible).

This has the side-effect of making all releases asynchronous, i.e., they succeed as soon as `kubectl apply` returns. If this is not actually what we want (I'm undecided), synchronisation can be restored by going back to the previous definition of Apply, or by using (async) Apply then asking the platform for the services in question until they match the versions required. Or something.

This PR dodges the question of what to do when applying replication controllers and daemonsets (they do not automatically start new pods, unlike deployments). They get `kubectl apply`d, which means the definition will be updated but the pods left alone. While sync is only used to implement releases, this won't matter, since they only affect deployments (assuming replication controllers are deprecated). But we should be careful to emit warnings, errors, or implement something sensible if we want to use sync in a more general way.

### DIFF
--- a/platform/kubernetes/kubernetes.go
+++ b/platform/kubernetes/kubernetes.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"os"
 	"os/exec"
-	"sync"
 
 	k8syaml "github.com/ghodss/yaml"
 	"github.com/go-kit/kit/log"
@@ -41,13 +40,6 @@ type apiObject struct {
 		Name      string `yaml:"name"`
 		Namespace string `yaml:"namespace"`
 	} `yaml:"metadata"`
-}
-
-type applyExecFunc func(*Cluster, log.Logger) error
-
-type apply struct {
-	exec    applyExecFunc
-	summary string
 }
 
 // --- add-ons
@@ -87,7 +79,6 @@ type Cluster struct {
 	config  *rest.Config
 	client  extendedClient
 	kubectl string
-	status  *statusMap
 	actionc chan func()
 	version string // string response for the version command.
 	logger  log.Logger
@@ -117,7 +108,6 @@ func NewCluster(config *rest.Config, kubectl, version string, logger log.Logger)
 		config:  config,
 		client:  extendedClient{client.Discovery(), client.Core(), client.Extensions()},
 		kubectl: kubectl,
-		status:  newStatusMap(),
 		actionc: make(chan func()),
 		version: version,
 		logger:  logger,
@@ -212,7 +202,8 @@ func (c *Cluster) AllServices(namespace string, ignore flux.ServiceIDSet) (res [
 
 func (c *Cluster) makeService(ns string, service *v1.Service, controllers []podController) platform.Service {
 	id := flux.MakeServiceID(ns, service.Name)
-	status, _ := c.status.getApplyProgress(id)
+	// FIXME go look in the k8s API
+	var status = ""
 	return platform.Service{
 		ID:         id,
 		IP:         service.Spec.ClusterIP,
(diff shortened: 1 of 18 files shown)


platform/kubernetes/kubernetes.go
```diff
@@ -79,43 +76,35 @@ func isAddon(obj namespacedLabeled) bool {
 
 // --- /add ons
 
+type Applier interface {
+	Delete(logger log.Logger, def []byte) error
+	Create(logger log.Logger, def []byte) error
+	Apply(logger log.Logger, def []byte) error
```
paulbellamy: what's the difference between create and apply?

platform/kubernetes/kubernetes.go
```diff
@@ -210,14 +199,24 @@ func (c *Cluster) AllServices(namespace string, ignore flux.ServiceIDSet) (res [
...
 	id := flux.MakeServiceID(ns, service.Name)
-	status, _ := c.status.getApplyProgress(id)
-	return platform.Service{
-		ID:         id,
-		IP:         service.Spec.ClusterIP,
-		Metadata:   metadataForService(service),
-		Containers: containersOrExcuse(service, controllers),
-		Status:     status,
+
+	// FIXME go look in the k8s API
```
paulbellamy: fixme

platform/kubernetes/release.go
```diff
@@ -86,49 +73,43 @@ func (c *Cluster) doApplyCommand(logger log.Logger, newDefinition *apiObject, ar
...
-				"rollout", "status",
-				"deployment", newDef.Metadata.Name,
-				"--namespace", newDef.Metadata.Namespace,
-			}
-			cmd := c.kubectlCommand(args...)
-			logger.Log("cmd", strings.Join(args, " "))
-			err = cmd.Run()
-		}
+func (c *Kubectl) Create(logger log.Logger, def []byte) error {
+	return c.doCommand(logger, def, "create", "-f", "-")
```
paulbellamy: discussed f2f whether to ditch create, or keep it. @squaremo to have a think.

platform/kubernetes/kubernetes.go
squaremo: The difference is that if you try to Create something that already exists, kubernetes will balk. So, if you calculate that you need to create something, but it breaks, your calculation was wrong and you should probably know about it.

A counter-argument is that actually, why should you care -- the goal is to make the cluster look like the config, and if applying something works whether or not it exists, just do that.

But the _counter_-counter argument is that applying something that doesn't exist and applying something that does do not _always_ have the same result. For example, with daemonsets, in the first case you get a daemonset configured how you expected; in the second case you get what you had already, which is different to the config (whereas using create would have told you something was wrong).

squaremo: -counter is that you can absorb that logical distinction by defining sync.Apply an "upsert" for everything, always, even if that means extra code (although we cannot really do it properly for daemonsets at the minute).

In other words, although `kubectl apply` does not do the same thing if the resource exists or doesn't exist, `SyncAction.Apply` can. Try to. Where possible.

squaremo: So the reason for distinguishing between Create and Apply comes down to: is it useful and possible to _just have upsert_?

squaremo: I cannot think of a concrete situation in which one is better than the other :S In lieu of that though, Paul suggests that declarations {Apply, Delete} may be easier to adapt to other platforms than operations {Create, Update, Delete}.

PR #491, opened 2017-03-15
[WIP] Implement versioned fluxd protocol

awh:
WIP scaffolding for versioning the fluxd protocol. TBD:

* 'NeedsUpgrade' error to be returned to platform clients (e.g. fluxsvc on behalf of fluxctl) when an old fluxd can't support the requested operation
* Sensible fluxd behaviour when it tries to talk to a deprecated RegisterDaemon endpoint


squaremo:
> 'NeedsUpgrade' error to be returned to platform clients (e.g. fluxsvc on behalf of fluxctl) when an old fluxd can't support the requested operation

Done neatly.

> Sensible fluxd behaviour when it tries to talk to a deprecated RegisterDaemon endpoint

Not a problem we'll have until we actually deprecate an endpoint, but we should build it in at least a version before we do that :)

Issue #139, opened 2016-09-14
Accidentally quadratic


squaremo:
- [x] https://github.com/weaveworks/fluxy/blob/master/helper.go#L86 `helper.AllReleasableImages`

Asks for each service in turn, and fetches the containers for each with `platform.ContainersFor(...)`, which itself fetches all the replication controllers and deployments and examines each one to see if it matches the service.
- [x] https://github.com/weaveworks/fluxy/blob/master/release/release.go#L154 `release.releaseAllToLatest`

Fetches the image metadata for each image mentioned by a container used by a service, and these images could (in principle) all be the same, and are certainly repeated in practice.
- [x] https://github.com/weaveworks/fluxy/blob/master/server.go#L68 `server.ListImages` and `server.ListServices`

These both ask for all the services, then iterate through all the rcs/deployments similar to the first item above.

philwinder:
Regarding [previous comment](https://github.com/weaveworks/flux/issues/139#issuecomment-249526428)

I investigated `kubeservice` and the python code is operating on the files that have been downloaded from the GH repo. I.e the files are local. Yes we iterate over all files to find a service, but there's no network latency involved with that.

So I don't see any value in attempting to reverse the way we do it because it would require a fairly large rewrite (and a lot of prior knowledge) and it seems to work well enough for the moment.

Closing because of this conclusion.

PR #493, opened 2017-03-15
WIP Replace kubeservice

### LINKED ISSUE #490: Cannot distinguish information kubeservice output from errors
We use a script `kubeservice` to find the services that a deployment (or replication controller) supplies pods to. What `kubeservice` _does_ is find the services given a particular deployment (or rc) file; what we _want_ is to find all the services, and for each, the deployment (or rc) files. This entails scanning the files n^2 times; but never mind, that's not why we're here.

The problem at hand is that kubeservice operates this way:
 1. try to load the file specified; if it's not a deployment or replication controller, exit with an error.
 2. for every other file in the directory, try to load it; if you can't output an error and continue. If the file is a service, add it to the list of candidates.
 3. for each candidate service, see if it matches the given deployment or rc and output it if so.

This is fine for the intended use, since the result will be either success: zero or more service names output; or failure, the given file was bad.

But in our use, kubeservice is invoked for every file, and we cannot distinguish between failing because a file isn't yaml (in which case, move on to the next) and failing because there's a "real" problem.

This came up because a user had an invalid yaml file in their repo, and flux would balk at it on every run, and report 0 services found.

squaremo:
Fixes #490 and an item of #139 by replacing the use of `kubeservice` with a bit of Go code. This relies on code from #486 at the minute, so will need to be rebased later.


squaremo:
Taking this off the radar.

PR #496, opened 2017-03-22
Deprecate updating ReplicationControllers

squaremo:
There are a few places where we can detect replication controllers and
return an error:

 * when listing the services or images
 * when updating a manifest
 * when applying an updated manifest

In the first case, I thought it would be strange to remove information
from the UI by supplying an error instead of container information. So
that stays as it is.

The second case is the best point, since it occurs just before a
lasting change (git commit) is made. The third case, failing to apply
the change, is really a backstop -- a change has been committed by
this point, which shouldn't have happened -- but it means we can
remove a bunch of code dealing with `kubectl rolling-upgrade`.

There is a choice between failing to update _just this service_, and
failing a release altogether. I have opted for the latter, because it
means we can put a big help message in front of the user.

(A more subtle (but maybe confusing?) approach is to fail just the
service, if it's one of many, or fail the whole thing if the release
is specifically for this service.)

### DIFF
--- a/platform/kubernetes/errors.go
+++ b/platform/kubernetes/errors.go
@@ -1,0 +1,43 @@
+package kubernetes
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/weaveworks/flux"
+)
+
+var ErrReplicationControllersDeprecated = flux.UserConfigProblem{
+	&flux.BaseError{
+		Err: errors.New("updating replication controllers is deprecated"),
+		Help: `Using Flux to update replication controllers is deprecated.
+
+ReplicationController resources are difficult to update, and it is
+almost certainly better to use a Deployment resource instead. Please
+see
+
+    https://kubernetes.io/docs/user-guide/replication-controller/#deployment-recommended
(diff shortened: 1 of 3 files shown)


awh:
This is quite different to what I imagined - I was expecting all references to `ReplicationController` to be removed from the code, `podController` struct to be removed etc. Is there a reason we can't do that? Even if so, can we strip RC support from export/`fluxctl save`?

squaremo:
> I was expecting all references to ReplicationController to be removed from the code

People (all those people) will be expecting flux to work with ReplicationControllers. So it has to _at least_ appear in code to warn that's not the case.

I can rewrite some code to remove the `podController` "polymorphism", which will simplify things a bit. But in general, I have tried to minimise the things you can no longer do. It is still possible to see which images are being used in a service that uses RCs, for instance, since that's still useful.

squaremo:
There may be further simplification (i.e., code that can be removed). But probably not easy while we want to keep the `list-services` bits and pieces.

PR #504, opened 2017-03-31
Omit change-cause annotation when saving resources

awh:
Don't version control `kubernetes.io/change-cause`


paulbellamy:
Would a white-list be more robust?

awh:
Those annotations are free-form - I don't think we can do it with a whitelist...

PR #506, opened 2017-04-03
Warn and crashloop on daemon registration endpoint deprecation

awh:
Facilitate deprecation of old daemon registration endpoints by having fluxd log that an upgrade is required and crashloop to attract attention from the user.

### DIFF
--- a/http/daemon.go
+++ b/http/daemon.go
@@ -3,6 +3,7 @@ package http
 import (
 	"net/http"
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


http/daemon.go
```diff
@@ -32,7 +33,8 @@ type Daemon struct {
 }
 
 var (
-	connectionDuration = prometheus.NewGaugeFrom(stdprometheus.GaugeOpts{
+	ErrEndpointDeprecated = errors.New("Your fluxd version is deprecated - please upgrade")
```
awh: Could use `ErrorDeprecated` from `http/errors.go` here, but it is multiline (IMO not great for printing to a daemon log) and mentions fluxctl specifically - thoughts?

squaremo: This is fine -- I agree with your rationale.

squaremo:
I suspect this has upset a mock implementation somewhere (i.e., the CircleCI failure is not a flake).

http/daemon.go
paulbellamy: instructions on how to upgrade?

awh:
> instructions on how to upgrade?

PR #507, opened 2017-04-03
WIP React to git changes (watch/unwatch, hooks)

squaremo:
A rebase of `463-react-to-git`, which I am leaving in place for now.


squaremo:
Removing this in favour of moving sync into the daemon per #509. Some bits from here will end up there, no doubt ..

PR #508, opened 2017-04-03
Fix trivial typo

nomisbeme:


squaremo:
Thanks @nomisbeme :star:

PR #515, opened 2017-04-05
add before, limit, and simple=1 query params to history api

### LINKED ISSUE #503: History API response too big for integrations
At Kubecon, loading the history on socks shop took forever, repsonse sizes of >1MB.
For most timelines, the `[{Stamp, Message},...]` is enough.

paulbellamy:
Fixes #503 

Called e.g. `http://192.168.99.100:30081/api/app/local-test/api/flux/v3/history?service=%3Call%3E&simple=1&before=2017-04-05T14:48:44.88041415Z&limit=10`

New flags are:
```
simple=1                             # Do not include event data in the items.
before=2017-04-05T14:48:44.000000Z   # Only include events older than X
limit=10                             # Only return 10 events at a time
```

So to paginate you'd first do a request with `?limit=10`, then find the earliest "before" value of the messages, then do a request with `?limit=10&before=<earliest-before-value>`, and repeat for subsequent pages.

The reason to not just do `page=5&perPage=10`, is that new events might accumulate in between requests, which would cause the page-alignment of events to change. `before` does not have this issue.

@davkal does this work for you?

### DIFF
--- a/cmd/fluxctl/history_cmd.go
+++ b/cmd/fluxctl/history_cmd.go
@@ -40,7 +40,7 @@ func (opts *serviceHistoryOpts) RunE(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	events, err := opts.API.History(noInstanceID, service, time.Now().UTC(), -1)
+	events, err := opts.API.History(noInstanceID, service, time.Time{}, -1)
 	if err != nil {
 		return err
 	}
--- a/cmd/fluxctl/history_cmd.go
+++ b/cmd/fluxctl/history_cmd.go
@@ -40,7 +40,7 @@ func (opts *serviceHistoryOpts) RunE(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	events, err := opts.API.History(noInstanceID, service)
+	events, err := opts.API.History(noInstanceID, service, time.Now().UTC(), -1)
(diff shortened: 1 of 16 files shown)


foot:
This looks great!

cmd/fluxctl/history_cmd.go
```diff
@@ -40,7 +40,7 @@ func (opts *serviceHistoryOpts) RunE(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	events, err := opts.API.History(noInstanceID, service)
+	events, err := opts.API.History(noInstanceID, service, time.Now().UTC(), -1)
```
squaremo: This is the wall-clock time at the client, which may be quite different from the time at the service. Could we use a sentinel value to mean "just the latest" instead?

http/server/server.go
```diff
@@ -246,12 +246,35 @@ func (s HTTPService) History(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+	h, err := s.service.History(inst, spec, before, limit)
 	if err != nil {
 		errorResponse(w, r, err)
 		return
 	}
 
+	if r.FormValue("simple") == "1" {
```
squaremo: How come "1" rather than, e.g., "true"?

PR #517, opened 2017-04-10
Encode test invocation into Makefile

squaremo:
This moves the invocation of `go test` into the Makefile, for
convenience of users and the CircleCI script; and removes the need for
`glide` to be installed.

NB drops the `go build ...` step from the CircleCI test stage -- it's
not doing anything that `go test` and `make all` won't do, and it adds
significantly to the build time.


squaremo:
This has a subtle problem when it comes to CircleCI (it passes, but it detects way fewer tests than it should). Investigating.

PR #518, opened 2017-04-10
Implemented `ServiceFilter` to filter defined or running services.

### LINKED ISSUE #492: locked services should only be included in output if they would have been changed
Note below `default/ui-server` shows as skipped, but it doesn't use that image, so wouldn't have normally been affected. Seems odd to include it.

```
$ fluxctl release --all --update-image=quay.io/weaveworks/scope:master-2fe3216e
Submitting release job...
Release job submitted, ID 6e45282d-e199-ca22-693e-2c60d1ac1389
Status: Complete.

Here's what happened:
SERVICE            STATUS   UPDATES
default/ui-server  skipped  locked
extra/demo         success  demo: quay.io/weaveworks/scope:master-456ac0bf -> master-2fe3216e
kube-system/scope  success  scope-app: quay.io/weaveworks/scope:master-456ac0bf -> master-2fe3216e
scope/collection   success  collection: quay.io/weaveworks/scope:master-456ac0bf -> master-2fe3216e
scope/control      success  control: quay.io/weaveworks/scope:master-456ac0bf -> master-2fe3216e
scope/pipe         success  pipe: quay.io/weaveworks/scope:master-456ac0bf -> master-2fe3216e
scope/query        success  query: quay.io/weaveworks/scope:master-456ac0bf -> master-2fe3216e
Took 42.864772s
```

philwinder:
The service is ignored if: excluded OR not included OR not correct image.
The service is skipped if: not ignored AND (locked or not found in cluster)
Else the service is pending.
The filters must be run across both the services to update and the services that are not running in the cluster. This is because the services not running in the cluster might have been ignored anyway.

Fixes #492

### DIFF
--- a/automator/automator.go
+++ b/automator/automator.go
@@ -118,7 +118,8 @@ func (a *Automator) handleAutomatedInstanceJob(logger log.Logger, job *jobs.Job,
 	}
 
 	// Get the list of services that are automated, in the repo, and in the running service.
-	updates, err := rc.SelectServices(automatedServiceIDs, flux.ServiceIDSet{}, flux.ServiceIDSet{}, results, logInJob)
+	incFilt := &release.IncludeFilter{automatedServiceIDs}
+	updates, err := rc.SelectServices(results, logInJob, incFilt)
 	if err != nil {
 		logInJob("error finding services: %s", err)
 		return followUps, err
--- a/automator/automator.go
+++ b/automator/automator.go
@@ -118,8 +118,13 @@ func (a *Automator) handleAutomatedInstanceJob(logger log.Logger, job *jobs.Job,
 	}
 
 	// Get the list of services that are automated, in the repo, and in the running service.
-	incFilt := &release.IncludeFilter{automatedServiceIDs}
-	updates, err := rc.SelectServices(results, logInJob, incFilt)
+	updates, err := rc.SelectServices(
+		results,
+		logInJob,
+		&release.IncludeFilter{
+			IDs: automatedServiceIDs,
+		},
+	)
 	if err != nil {
 		logInJob("error finding services: %s", err)
 		return followUps, err
--- a/release/context.go
+++ b/release/context.go
@@ -121,11 +121,11 @@ func (rc *ReleaseContext) SelectServices(results flux.ReleaseResult, logStatus s
 	var filteredUpdates []*ServiceUpdate
 	for _, s := range updates {
 		fr := s.filter(filters...)
-		if fr.ServiceResult.Error != "" {
-			logStatus(fr.String())
+		if fr.Error != "" {
+			logStatus(fr.Msg(s.ServiceID))
 		}
-		results[s.ServiceID] = fr.ServiceResult
-		if fr.ServiceResult.Status == flux.ReleaseStatusPending || fr.ServiceResult.Status == flux.ReleaseStatusSuccess || fr.ServiceResult.Status == "" {
+		results[s.ServiceID] = fr
+		if fr.Status == flux.ReleaseStatusPending || fr.Status == flux.ReleaseStatusSuccess || fr.Status == "" {
 			filteredUpdates = append(filteredUpdates, s)
 		}
 	}
@@ -134,11 +134,11 @@ func (rc *ReleaseContext) SelectServices(results flux.ReleaseResult, logStatus s
 	filteredDefined := map[flux.ServiceID]*ServiceUpdate{}
 	for k, s := range definedMap {
 		fr := s.filter(filters...)
-		if fr.ServiceResult.Error != "" {
-			logStatus(fr.String())
+		if fr.Error != "" {
+			logStatus(fr.Msg(s.ServiceID))
 		}
-		results[s.ServiceID] = fr.ServiceResult
-		if fr.ServiceResult.Status != flux.ReleaseStatusIgnored {
+		results[s.ServiceID] = fr
+		if fr.Status != flux.ReleaseStatusIgnored {
 			filteredDefined[k] = s
 		}
 	}
@@ -155,14 +155,14 @@ func (rc *ReleaseContext) SelectServices(results flux.ReleaseResult, logStatus s
 	return filteredUpdates, nil
 }
 
-func (s *ServiceUpdate) filter(filters ...ServiceFilter) FilterResult {
+func (s *ServiceUpdate) filter(filters ...ServiceFilter) flux.ServiceResult {
 	for _, f := range filters {
 		fr := f.Filter(*s)
-		if fr.ServiceResult.Error != "" {
+		if fr.Error != "" {
 			return fr
 		}
 	}
-	return FilterResult{}
+	return flux.ServiceResult{}
 }
 
 func (rc *ReleaseContext) FindDefinedServices() ([]*ServiceUpdate, error) {
@@ -191,32 +191,20 @@ func (rc *ReleaseContext) FindDefinedServices() ([]*ServiceUpdate, error) {
 	return defined, nil
 }
 
-type FilterResult struct {
-	ServiceResult flux.ServiceResult
-	ID            flux.ServiceID
-}
-
-func (fr *FilterResult) String() string {
-	return fmt.Sprintf("%s service %s as it is %s", fr.ServiceResult.Status, fr.ID, fr.ServiceResult.Error)
-}
-
 type ServiceFilter interface {
-	Filter(ServiceUpdate) FilterResult
+	Filter(ServiceUpdate) flux.ServiceResult
 }
 
 type SpecificImageFilter struct {
 	Img flux.ImageID
 }
 
-func (f *SpecificImageFilter) Filter(u ServiceUpdate) FilterResult {
+func (f *SpecificImageFilter) Filter(u ServiceUpdate) flux.ServiceResult {
 	// If there are no containers, then we can't check the image.
 	if len(u.Service.Containers.Containers) == 0 {
-		return FilterResult{
-			ID: u.ServiceID,
-			ServiceResult: flux.ServiceResult{
-				Status: flux.ReleaseStatusIgnored,
-				Error:  NotInCluster,
-			},
+		return flux.ServiceResult{
+			Status: flux.ReleaseStatusIgnored,
+			Error:  NotInCluster,
 		}
 	}
 	// For each container in update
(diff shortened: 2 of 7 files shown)


release/releaser_test.go
```diff
@@ -189,8 +233,137 @@ func TestUpdateOne(t *testing.T) {
 	if result.Status != flux.ReleaseStatusSuccess {
 		t.Errorf("expected entry to be success, but was %s", result.Status)
 	}
+}
+
+func Test_492_OnUpdateImage_OnlyShowServicePertainingToImage(t *testing.T) {
```
paulbellamy: We don't need the issue number in tests/comments. It'll be in the commit msg so still findable from git-blame.

release/releaser.go
```diff
@@ -154,6 +154,16 @@ func (r *Releaser) release(instanceID flux.InstanceID, job *jobs.Job, logStatus
...
 
+	// #492: If updating all services, but requested a specific image, only result in services that pertain to that image
+	if spec.ServiceSpec == flux.ServiceSpecAll && spec.ImageSpec != flux.ImageSpecNone {
+		for k, v := range results {
+			if v.Status == flux.ReleaseStatusSkipped || v.Status == flux.ReleaseStatusIgnored {
+				delete(results, k)
+			}
+		}
+
+	}
```
paulbellamy: afaict, this is saying
```Go
if --all && --update-image=xxx {
  hide all skipped/ignored services
}
```
Which is not quite right. Skipped/ignored services should still be included, but only if they would have been modified (i.e. if they use the image in question).

philwinder: This actually brought up another issue.
What happens when `--all` and `--latest`? I'm assuming you want to see everything then.

automator/automator.go
```diff
@@ -118,7 +118,8 @@ func (a *Automator) handleAutomatedInstanceJob(logger log.Logger, job *jobs.Job,
 	}
 
 	// Get the list of services that are automated, in the repo, and in the running service.
-	updates, err := rc.SelectServices(automatedServiceIDs, flux.ServiceIDSet{}, flux.ServiceIDSet{}, results, logInJob)
+	incFilt := &release.IncludeFilter{automatedServiceIDs}
```
paulbellamy: ah, just inline this below.

release/context.go
```diff
@@ -76,90 +86,86 @@ func (rc *ReleaseContext) Clean() {
...
-// running. Services in the locked and excluded sets are omitted (and
-// recorded as so). The return value is a set of potentially
-// updateable services.
-func (rc *ReleaseContext) SelectServices(included []flux.ServiceID, locked flux.ServiceIDSet, excluded flux.ServiceIDSet, results flux.ReleaseResult, logStatus statusFn) ([]*ServiceUpdate, error) {
-	// Figure out all services that are defined in the repo and should
-	// be selected for upgrading/applying.
+// files and the running platform.
+// ServiceFilter's can be provided to filter the found services.
+// Be careful about the ordering of the filters. Filters that are earlier
+// in the slice will have higher priority (they are run first).
```
paulbellamy: formatting of this paragraph seems off.

release/context.go
```diff
@@ -76,90 +86,86 @@ func (rc *ReleaseContext) Clean() {
...
-			status = flux.ReleaseStatusSkipped
+
+	// Filter both updates ...
+	var filteredUpdates []*ServiceUpdate
+	for _, s := range updates {
+		fr := s.filter(filters...)
+		if fr.Result.Error != "" {
+			logStatus(fr.String())
+		}
+		results[s.ServiceID] = fr.Result
```
paulbellamy: Wait, so filters always returns a result? Doesn't that mean all service will be output?

paulbellamy: Ooooh. If it is filtered out, Result is the null-type. That's a bit subtle, maybe there's a way to make it more obvious?

paulbellamy: Wait, no. what? I don't understand how this filter system works...

release/context.go
```diff
@@ -187,3 +193,106 @@ func (rc *ReleaseContext) FindDefinedServices() ([]*ServiceUpdate, error) {
 	}
 	return defined, nil
 }
+
+type FilterResult struct {
+	Result flux.ServiceResult
```
paulbellamy: resultresultresultresultresult

release/releaser.go
```diff
@@ -259,10 +259,35 @@ func selectServices(rc *ReleaseContext, spec *flux.ReleaseSpec, results flux.Rel
...
+	if len(spec.Excludes) > 0 {
+		exFilt := &ExcludeFilter{spec.Excludes}
+		filtList = append(filtList, exFilt)
+	}
+
+	// Get locked services
+	lockedSet := LockedServices(conf)
+	lockFilt := &LockedFilter{lockedSet.ToSlice()}
+	filtList = append(filtList, lockFilt)
```
paulbellamy: `func filters(spec ReleaseSpec) []ServiceFilter` ?

release/releaser.go
```diff
@@ -271,23 +296,18 @@ func selectServices(rc *ReleaseContext, spec *flux.ReleaseSpec, results flux.Rel
...
 			if err != nil {
 				return nil, err
 			}
 			ids = append(ids, id)
 		}
-		return rc.SelectServices(ids, lockedSet, excludedSet, results, logStatus)
-	case flux.ServiceSpecAll:
-		return rc.SelectServices(nil, lockedSet, excludedSet, results, logStatus)
+		incFilt := &IncludeFilter{ids}
+		return rc.SelectServices(results, logStatus, append([]ServiceFilter{incFilt}, filtList...)...)
```
paulbellamy: maybe
```Go
return rc.SelectServices(
  results,
  logStatus,
  append(filtList, &IncludeFilter{ids})...,
)
```

release/releaser_test.go
```diff
@@ -40,157 +110,282 @@ func setup(t *testing.T, mocks instance.Instance) (*Releaser, func()) {
...
+						flux.ContainerUpdate{
+							Container: "helloworld",
+							Current:   oldImageID,
+							Target:    newImageID,
+						},
+					},
+				},
+				flux.ServiceID("default/locked-service"): flux.ServiceResult{
+					Status: flux.ReleaseStatusIgnored,
+					Error:  DifferentImage,
```
paulbellamy: Isn't this test saying that for `--all` locked services for different images will still be output?

release/context.go
philwinder: Yes, there's always a `Result`. A `Result` is the thing that is shown to the user. This is separate from a `ServiceUpdate` that actually does the updating.

All defined services have a result. The ones that have the status type "ignored" will be ignored by the Printer unless verbose is selected: https://github.com/weaveworks/flux/blob/master/release/print.go#L19

release/context.go
philwinder: Due to top level package naming. ;-)
Renamed to ServiceResult, but not much more I can do.

release/releaser.go
philwinder: We need to make sure the include filter runs first. Filters run in order. 

So for example, we need to make sure the include filter is before the lock filter because we want to result in "ignore", even if it is locked.

release/releaser_test.go
philwinder: So this test is: `--all --image-update=some/image:123`
So the result of locked-service should be:
"ignored": "a different image".
Since locked-service does not run that image. So the fact that it is locked makes no difference. So yes, the test is as intended. Whether that is right or not is open to debate.

So to answer your question, yes, if locked and they do not have the correct image, they will output "ignored": "different image".

Remember all "ignored" items are ignored when Printing.

paulbellamy: Right, so this doesn't fix https://github.com/weaveworks/flux/issues/492, then...

philwinder: @paulbellamy Yes it does. It's ignored.
This is the output for `--all --image=...hello-world...` when verbose:
```
SERVICE                 STATUS   UPDATES
default/helloworld      success  helloworld: quay.io/weaveworks/helloworld:master-a000001 -> master-a000002
default/locked-service  ignored  not included
default/test-service    ignored  not included
```

This is the normal output: 
```
default/helloworld  success  helloworld: quay.io/weaveworks/helloworld:master-a000001 -> master-a000002
```

Note how the locked service was ignored.

paulbellamy: ah verbose vs non-verbose output. right-o.

release/context.go
```diff
@@ -195,12 +192,12 @@ func (rc *ReleaseContext) FindDefinedServices() ([]*ServiceUpdate, error) {
 }
 
 type FilterResult struct {
-	Result flux.ServiceResult
-	ID     flux.ServiceID
+	ServiceResult flux.ServiceResult
```
paulbellamy: this still stutters and reads awkwardly, but whatever.

release/releaser.go
```diff
@@ -261,8 +261,44 @@ func selectServices(rc *ReleaseContext, spec *flux.ReleaseSpec, results flux.Rel
...
 	// Build list of filters
-	var filtList []ServiceFilter
+	filtList, err := filters(spec, conf)
+	if err != nil {
+		return nil, err
+	}
+
+	// For backwards-compatibility, there's two fields: ServiceSpec
+	// and ServiceSpecs. An entry in ServiceSpec takes precedence.
+	switch spec.ServiceSpec {
```
paulbellamy: It might make sense to do this as part of the `filters` function, so all the service spec -> filter conversion happens in one place.

release/context.go
philwinder: I've got an idea...

release/releaser.go
philwinder: I removed the legacy `ServiceSpec` from `ReleaseSpec` since it wasn't used anymore. This simplified the code dramatically. (All methods now pass a `[]ServiceSpec`)

release/context.go
philwinder: Removed FilterResult completely and just used ServiceResult.

release/releaser.go
```diff
@@ -307,20 +281,38 @@ func filters(spec *flux.ReleaseSpec, conf instance.Config) ([]ServiceFilter, err
...
+		}
+		id, err := flux.ParseServiceID(string(s))
 		if err != nil {
 			return nil, err
 		}
-		incFilt := &IncludeFilter{[]flux.ServiceID{svcID}}
+		ids = append(ids, id)
+	}
+	if len(ids) > 0 {
+		incFilt := &IncludeFilter{ids}
```
paulbellamy: afaict, in the instance of: `spec.ServiceSpecs = []string{"some-id", "<all>"}`, this will scope to just `some-id`, instead of letting `<all>` override that (like currently).
Not sure it is a bug, or if we should fix it, just checking that is what you meant.

philwinder: Ah yes, you're right. I'll add a test, then make sure "<all>" overrides everything.

PR #523, opened 2017-04-11
Do fingerprinting and secret masking server-side so we can use it via the UI

### LINKED ISSUE #510: Support `get config --fingerprint` in the browser API
Its tricky to generate the fingerprint of an ssh public key in the browser, but it would be great to be able to show it in the UI.

E.g. when we push to github, the github UI shows the md5 fingerprint of the key which the user could then compare against the one flux-ui is showing to aid certain debugging cases.

Something like `/api/flux/v4/config?secrets=false&fingerprint=md5`

Supports https://github.com/weaveworks/service-ui/issues/385

paulbellamy:
Fixes #510 

Code is super messy. I'm really unhappy with how it turned out. Suggestions for improvement welcome.

The http server seems like the right place to be doing the masking, as it is a formatting decision, but this makes the Safe/UnsafeInstanceConfig distinction not really used anymore, and a bit awkward.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -19,7 +19,7 @@ type ClientService interface {
 	Lock(flux.InstanceID, flux.ServiceID) error
 	Unlock(flux.InstanceID, flux.ServiceID) error
 	History(flux.InstanceID, flux.ServiceSpec, time.Time, int64) ([]flux.HistoryEntry, error)
-	GetConfig(_ flux.InstanceID, secrets bool, fingerprint string) (flux.InstanceConfig, error)
+	GetConfig(_ flux.InstanceID, fingerprint string) (flux.InstanceConfig, error)
 	SetConfig(flux.InstanceID, flux.UnsafeInstanceConfig) error
 	PatchConfig(flux.InstanceID, flux.ConfigPatch) error
 	GenerateDeployKey(flux.InstanceID) error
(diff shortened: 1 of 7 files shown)


api/api.go
```diff
@@ -19,7 +19,7 @@ type ClientService interface {
 	Lock(flux.InstanceID, flux.ServiceID) error
 	Unlock(flux.InstanceID, flux.ServiceID) error
 	History(flux.InstanceID, flux.ServiceSpec, time.Time, int64) ([]flux.HistoryEntry, error)
-	GetConfig(_ flux.InstanceID) (flux.InstanceConfig, error)
+	GetConfig(_ flux.InstanceID, secrets bool, fingerprint string) (flux.InstanceConfig, error)
```
squaremo: Is there a requirement to send the secrets across the API?

paulbellamy: no, good point.

http/server/server.go
```diff
@@ -281,13 +285,38 @@ func (s HTTPService) History(w http.ResponseWriter, r *http.Request) {
...
-	config, err := s.service.GetConfig(inst)
+	fingerprint := r.FormValue("fingerprint")
+	config, err := s.service.GetConfig(inst, fingerprint)
 	if err != nil {
 		errorResponse(w, r, err)
 		return
 	}
 
-	jsonResponse(w, r, config)
+	safeConfig := config.HideSecrets()
```
squaremo: Worth a comment saying why this is necessary (I immediately thought: "hang on, it didn't used to do this!")

PR #526, opened 2017-04-13
WIP WIP WIP Move sync into daemon

squaremo:


squaremo:
Retiring in favour of #585.

PR #532, opened 2017-04-17
Don't log ignored services.

### LINKED ISSUE #527: Verbose history messages
The history messages are a big text blob: 
screen shot 2017-04-13 at 2 27 04 pm

API request:
https://frontend.dev.weave.works/api/app/proud-wind-05/api/flux/v3/history?service=%3Call%3E

philwinder:
Added test for `logEvent` method in `releaser.go`.
Ignore "ignored" services in event log.
Fixes #527

### DIFF
--- a/release/releaser.go
+++ b/release/releaser.go
@@ -203,8 +203,10 @@ func logEvent(inst *instance.Instance, executeErr error, release flux.Release) e
 	}
 
 	var serviceIDs []flux.ServiceID
-	for _, id := range release.Result.ServiceIDs() {
-		serviceIDs = append(serviceIDs, flux.ServiceID(id))
+	for k, v := range release.Result {
+		if v.Status != flux.ReleaseStatusIgnored {
+			serviceIDs = append(serviceIDs, flux.ServiceID(k))
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


release/releaser.go
```diff
@@ -203,8 +203,10 @@ func logEvent(inst *instance.Instance, executeErr error, release flux.Release) e
 	}
 
 	var serviceIDs []flux.ServiceID
-	for _, id := range release.Result.ServiceIDs() {
-		serviceIDs = append(serviceIDs, flux.ServiceID(id))
+	for k, v := range release.Result {
```
squaremo: What about services that were skipped?

philwinder: I initially had that in, but then took it out, because the previous functionality was to include skipped services in the history message.

If you think we should be ignoring skipped services too, let me know and I'll change.

squaremo: Gotcha, ta. If things are skipped it's because they were _supposed_ to be updated (e.g., they were specifically requested) but weren't in fact updated. It's then a question of reporting what the user expects to see (the things they mentioned) vs being accurate (what was *actually* updated).

Let's preserve the current behaviour for now. When we include the result in the notification this will be resolved.

PR #534, opened 2017-04-18
use go list so `make test` works on OSX

paulbellamy:


squaremo:
`go list` doesn't seem to find platform/kubernetes/testdata, which has tests to run.

squaremo:
It is .. quite strange .. there's nothing about that package that makes it special, except that it's only imported by other test files. I thought the build flag `-tags 'test'` might help, but doesn't appear to.

paulbellamy:
https://github.com/golang/go/issues/14715#issuecomment-193924969

> Directory and file names that begin with "." or "_" are ignored
by the go tool, as are directories named "testdata".

PR #535, opened 2017-04-18
If service specifically requested and not in repo, add skipped "not in repo"

### LINKED ISSUE #473: fluxctl release --service does not report on undefined services
If you ask for a service that is not defined in the git repo, flux will skip over it without even recording it as `ignored`.
It'd be better if it set it to `skipped` (since it was specifically requested; otherwise there's still just an empty response).

philwinder:
Added test.
Added logic to test if all specified `ServiceSpec`'s are found. If not, manually add a result entry.

Fixes #473

### DIFF
--- a/release/releaser.go
+++ b/release/releaser.go
@@ -112,6 +112,24 @@ func (r *Releaser) release(instanceID flux.InstanceID, job *jobs.Job, logStatus 
 		return nil, err
 	}
 	logStatus("Found %d services.", len(updates))
+
+	// If the request was for a specific service and the service was
+	// not in the cluster, then add a result with a skipped status
+	for _, v := range spec.ServiceSpecs {
+		if v == flux.ServiceSpecAll {
+			continue
+		}
+		id, err := v.AsID()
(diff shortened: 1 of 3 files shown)


release/releaser.go
```diff
@@ -112,6 +112,24 @@ func (r *Releaser) release(instanceID flux.InstanceID, job *jobs.Job, logStatus
...
+	// not in the cluster, then add a result with a skipped status
+	for _, v := range spec.ServiceSpecs {
+		if v == flux.ServiceSpecAll {
+			continue
+		}
+		id, err := v.AsID()
+		if err != nil {
+			continue
+		}
+		if _, ok := results[id]; ok == false {
```
paulbellamy: this is `!ok`

squaremo: Ooops did not even spot that!

philwinder:
@paulbellamy Inside select services everything is just a Filter. There's no contextual information about `ServiceSpec`'s and the like. So I had to do it at this level.

Will fix !ok.

PR #537, opened 2017-04-18
Store policy in annotations

paulbellamy:
Part of: https://github.com/weaveworks/flux/issues/283
Based on: https://github.com/weaveworks/flux/pull/526

@squaremo wdyt so far? worth pursuing? wait for base branch to stabilize?

Edit: have rebased it onto master.
Edit: have rebased it back onto branch.

## TODO

- [x] Make the code changes less horrible
  - Maybe extract some git repo helpers?
  - Leaving kubernetes helpers where they are, as they will move with `move-sync-...`, anyway
- [x] Fix up tests
  - A lot more tests now depend on git repos being set up in specific states.
- [x] Make sure `---` in yaml is preserved
- [x] Make sure it doesn't re-sort all yaml keys.
- [x] Make sure it doesn't nuke comments when locking/unlocking.
- [x] Investigate other yaml packages if they have a `RawMessage` equivalent.
  - They do not.


squaremo:
> wdyt so far? worth pursuing? wait for base branch to stabilize?

API and commands look sound. Worth basing on master for now -- it should be portable to `move-sync-...`.

paulbellamy:
This does make the automator harder to scale in the current implementation, as it will have to clone all the repos every time. With the automation done in fluxd it is easier, as you can just sync the repo there.

squaremo:
> This does make the automator harder to scale in the current implementation, as it will have to clone all the repos every time

Good point, it rules out the shortcut of not bothering to clone if there's no automated services. Ah well, we know how to scale that up if we need to.

paulbellamy:
I think that is all the places updated which write/read policies.

paulbellamy:
Note, generated git commit messages are either like:
```
commit 3be9c3863b1a134a5acf3803fedc72320cc310cc
Author: Weave Flux <support@weave.works>
Date:   Thu Apr 20 13:32:12 2017 +0000

    Locked: default/helloworld
```
or, for multiple changes in the same commit:
```
commit 3be9c3863b1a134a5acf3803fedc72320cc310cc
Author: Weave Flux <support@weave.works>
Date:   Thu Apr 20 13:32:12 2017 +0000

    Updated service policies:

    - Automated: default/helloworld
    - Locked: default/nginx
```

paulbellamy:
API is backwards compatible, but the new API endpoint should be used. Is a:
```
PATCH /v4/policies
{
  "default/helloworld": {
    "add": ["locked"],
    "remove": ["automated"]
  }
}
```
Note, you can apply policy changes to a bunch of a different services at the same time (in a single commit), and if we later want to do more complex policies, the strings could be replaced by an object.

squaremo:
We should annotate all our own manifests before this gets merged. (It's safe if we don't, but may be surprising when everything stops)

paulbellamy:
I was debating if we need a migration which takes the in-db settings and tries to store them in the repo. wdyt?

platform/kubernetes/policies.go
```diff
@@ -0,0 +1,153 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	yaml "gopkg.in/yaml.v2"
+
+	"github.com/weaveworks/flux"
+)
+
+const (
+	policyPrefix = "flux.policy."
```
squaremo: Actually, would be good to change this to conform ..

server/server.go
```diff
@@ -217,106 +232,54 @@ func (s *Server) History(inst flux.InstanceID, spec flux.ServiceSpec, before tim
...
-		LogLevel:   flux.LogLevelInfo,
-	}); err != nil {
-		return err
-	}
-	return recordAutomated(inst, service, true)
-}
-
-func (s *Server) Deautomate(instID flux.InstanceID, service flux.ServiceID) error {
-	inst, err := s.instancer.Get(instID)
+	// TODO: Should all this be encapsulated in kubernetes package?
```
squaremo: (should remove the comment, I suppose)

platform/standalone.go
```diff
@@ -44,7 +44,7 @@ func (s *StandaloneMessageBus) Connect(inst flux.InstanceID) (Platform, error) {
 // trying to use it is the only way to tell if it's closed -- the
 // error representing the cause will be sent to the channel supplied.
 func (s *StandaloneMessageBus) Subscribe(inst flux.InstanceID, p Platform, complete chan<- error) {
-	s.Lock()
+	s.RWMutex.Lock()
```
paulbellamy: Ah ha. I had plumbed the `(Un)Lock` methods through, which meant I had to rename this. but not anymore!

release/context.go
```diff
@@ -112,7 +112,11 @@ func (rc *ReleaseContext) SelectServices(results flux.ReleaseResult, logStatus s
 	var updates []*ServiceUpdate
 	for _, s := range services {
 		logStatus("Found service %s", s.ID)
-		update := definedMap[s.ID]
+		update, ok := definedMap[s.ID]
```
paulbellamy: We take the ids from the defined services, then look up only those in the running system, so the running services we found *should* only ever be a subset. It was panicking in a test with mocks returning slightly odd results. Figured no harm in leaving in the check.

paulbellamy:
Per discussion in slack, we've opted to delay merging this until it can be released along with the git-sync stuff.

paulbellamy:
An example commit is: https://github.com/paulbellamy/flux-example/commit/85122c97b5acd7c4f75d20a7b7e60c1ca22508ae

git/repo.go
```diff
@@ -39,6 +39,7 @@ func (r Repo) Clone() (path string, err error) {
 
 	// Hack, while it's not possible to mount a secret with a
 	// particular mode in Kubernetes
+	// FIXME this fails if the key is mounted as read-only.
```
squaremo: Absolutely true --  if it's mounted as read-only, we can't proceed. We could at least test whether the permissions are correct first, and provide a reasonable error if they aren't and we can't do anything about it.

squaremo:
This is real nice :star2:

PR #538, opened 2017-04-18
Increase registry timeout to 300s to allow for large numbers of tags

### LINKED ISSUE #533: Timeouts on calls that reach out to docker hub
Using docker containers that have a large number of tags causes the `get images/tags/manifests` calls to time out. E.g.
```
fluxctl list-images
Error: 500 Internal Server Error getting images for services: fetching image metadata for quay.io/weaveworks/fluxd: Get https://quay.io/v2/weaveworks/fluxd/manifests/master-f94fd3e: context deadline exceeded
```
This occurs after 10s, which indicates that it is due to a default timeout somewhere.

It could be in Weave Cloud's authentication layer, or it could be in Flux.

philwinder:
Fixes #533

(Partially, there still might be an issue with authfe timeouts, but we shall see when this gets deployed).

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	requestTimeout = 10 * time.Second
+	requestTimeout = 300 * time.Second
 	maxConcurrency = 10 // Chosen arbitrarily
 )
 


paulbellamy:
How will we not just get rate-limited?

squaremo:
> How will we not just get rate-limited?

We probably will, as soon as we go to fetch with a cold cache (i.e., a new instance running sock shop). We'll bounce off it (backoff.go) which is to the good.

However if there's a few of these running, we may get blacklisted, since they will all be bouncing off the rate limit, while looking like the same incoming IP. We would need to make all requests back off, perhaps by using the same HTTP client everywhere (is that a possibility?)

philwinder:
All I know is that before this change it wouldn't work at all. After this change it worked fine. So I'd recommend putting it in anyway, even if we do get rate limited, because that's a separate problem.

squaremo:
> even if we do get rate limited

The problem isn't rate limiting as much as being blacklisted; the consequence is that we have to rebuild entire hosts, so they get new IP addresses (i.e., it's pretty drastic).

squaremo:
Closing in favour of a hypothetical fix which rate-limits requests, or otherwise avoids throttling and blacklisting.

PR #542, opened 2017-04-20
Registry auth keys to accept more formats

### LINKED ISSUE #453: Should allow URL schema in auth configs, like dockerconfig does.
This results in auth error:
```YAML
registry:
  auths:
    "https://gcr.io":
      auth: "..."
```
...but this works:
```YAML
registry:
  auths:
    "gcr.io":
      auth: "..."
```

philwinder:
Added url parse with a bit of jiggery pokery to parse more formats.
Added test to ensure all formats are parsed properly.

Fixes #453

### DIFF
--- a/registry/remote_factory_test.go
+++ b/registry/remote_factory_test.go
@@ -123,7 +123,88 @@ func TestRemoteFactory_CredentialsFromConfigHTTPSHosts(t *testing.T) {
 	if pass != c.password {
 		t.Fatalf("Expected %q, got %q.", pass, c.password)
 	}
-	if len(creds.Hosts()) != 1 || httpsHost != creds.Hosts()[0] {
+	if len(creds.Hosts()) != 1 || host != creds.Hosts()[0] {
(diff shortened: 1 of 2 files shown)


registry/remote_factory_test.go
```diff
@@ -123,7 +123,59 @@ func TestRemoteFactory_CredentialsFromConfigHTTPSHosts(t *testing.T) {
...
+		{
+			host:        "https://gcr.io",
+			imagePrefix: "gcr.io",
+		},
+		{
+			host:        "https://gcr.io/v1",
+			imagePrefix: "gcr.io",
+		},
+		{
+			host:        "https://gcr.io/v1/",
```
paulbellamy: maybe another scheme as well for completeness, e.g. `telnet://gcr.io/v1`

paulbellamy: Should probably also test some malformed ones, and what happens...
`"https://"`
`""`
`"/usr/local"`

PR #544, opened 2017-04-20
Report 200 `{"connected": false}` when fluxd is not connected

### LINKED ISSUE #440: `/api/ping` returning 500s when fluxd not connected to instance.
We're expected 404s! See

https://github.com/weaveworks/service-ui/issues/256.

We'll change the ui so any error code === no connection too.

philwinder:
Altered `/v4/ping` endpoint to check for specific errors that occur when fluxd isn't connected.
Added check for standalone.
Added check for nats.

Fixes #440


paulbellamy:
Should the nats timeout for ping be turned way down, too?

philwinder:
It's 5 seconds:
https://github.com/weaveworks/flux/blob/master/platform/rpc/nats/bus.go#L22
Seems about right?

PR #547, opened 2017-04-21
Docker swarm platform integration

vlal:
* Integrate Docker Swarm platform
* Added namespace field in config
* Try K8s and fallback to swarm else fail
* Added docker to fluxd dockerfile

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/cluster"
+	"github.com/weaveworks/flux/cluster/docker"
 	"github.com/weaveworks/flux/cluster/kubernetes"
 	"github.com/weaveworks/flux/daemon"
 	"github.com/weaveworks/flux/git"
@@ -104,6 +105,9 @@ func main() {
 		upstreamURL = fs.String("connect", "", "Connect to an upstream service e.g., Weave Cloud, at this base address")
 		token       = fs.String("token", "", "Authentication token for upstream service")
 
+		//docker
+		namespace = fs.String("namespace", "", "Docker swarm stack name")
+
 		// Deprecated
 		_ = fs.String("docker-config", "", "path to a docker config to use for credentials")
 	)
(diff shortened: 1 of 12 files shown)


squaremo:
Wow, thank you for this PR!

We are undergoing some seismic rearrangement in the branch https://github.com/weaveworks/flux/tree/move-sync-into-daemon at the minute. While things are unstable, I'm going to hold off making suggestions here. It should be done in the next couple of weeks.

lukemarsden:
How's this going @squaremo?

squaremo:
>  squaremo commented 26 days ago 
> It should be done in the next couple of weeks.

Ha, never trust estimates.

The branch `move-sync-into-daemon` is fairly stable (i.e., it builds, passes tests, and actually does what it's supposed to for the most part). But we'll be knocking off rough edges for a while, and it may go through some rebasing and rearrangement.

squaremo:
The "gitsync" stuff is merged into master now. The main change from this PR's point of view is that most of the work previously done by the service is now done by the daemon. `platform.Platform` is just an RPC layer from the service to the daemon now.

Instead, you want to implement (to begin with) `cluster.Cluster` which is the interface that represents stuff running; and,`cluster.Manifests` which is the interface that represents a directory of config files.

vlal:
@squaremo ready for review!

squaremo:
@vlal, would you mind doing `git rebase -i master` such that the various "merge lukes branch" and fixups so on are elided. Alternatively, since that may end up being a rather arduous, you could `git reset master` and assemble the changes into a small number of self-contained commits.

vlal:
@squaremo done

vlal:
ping @squaremo

cmd/fluxd/main.go
```diff
@@ -142,90 +146,106 @@ func main() {
...
-	{
-		restClientConfig, err := rest.InClusterConfig()
-		if err != nil {
-			logger.Log("err", err)
-			os.Exit(1)
-		}
 
-		restClientConfig.QPS = 50.0
-		restClientConfig.Burst = 100
+	if *namespace == "" {
```
squaremo: This seems like a strange way to determine if Kubernetes is the intended target rather than Docker Swarm. I would prefer something more explicit, against which the arguments can be sanity-checked. That is, if you say you're using Swarm, it breaks when you supply arguments to do with Kubernetes, and vice-versa.

image.go
```diff
@@ -34,6 +34,7 @@ func ParseImageID(s string) (ImageID, error) {
 		return ImageID{}, ErrBlankImageID
 	}
 	var img ImageID
+	s = strings.Split(s, "@")[0]
```
squaremo: I guess this is to do with image digests -- we should address this separately (since Kubernetes also accepts digests).

jasonrichardsmith:
So at this point if @stefanprodan is taking over this merge, should @vlal leave it alone and hand it over? Or should he continue hacking away at it and when merged then hand it over?
@mewzherder

squaremo:
Closing this to remove it from the radar screen until someone can carry it on. It's worth keeping the branch around in case we want to revivify this effort.

PR #549, opened 2017-04-24
Avoid using 500s for application-level errors from the daemon

squaremo:
We distinguish between "fatal" errors from the daemon, which indicate
that it is most likely not responsive (and therefore, please
disconnect it) and errors which simply indicate that the daemon could
not carry out the request.

However, both are returned up to the HTTP server as unadorned errors,
rather than wrapped in an `HelpfulError` from flux/errors.go; and as a
result, the server will send a HTTP 500 Internal Server Error status
to the client, which is misleading, and also a false alarm for monitoring.

This commit wraps non-fatal errors in a UserConfigProblem so that it
gets translated into a 422 Precondition Failed (which is the closest
we have to "application error").

It also makes the "upgrade needed" error (used by deprecated daemon
methods) a `UserConfigProblem`, for a similar reason.

It _also_ makes asking for a namespace that doesn't exist (in `fluxctl
list-services`) an error, rather than returning nothing.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -9,6 +9,13 @@ type HelpfulError interface {
 	Base() *BaseError
 }
 
+func UnderlyingError(err error) error {
+	if helpful, ok := err.(HelpfulError); ok {
+		return helpful.Base().Err
+	}
+	return err
+}
+
 // Representation of errors in the API. These are divided into a small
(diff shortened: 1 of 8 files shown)


squaremo:
> Would be nice to have a test, if possible, but.

Ugh, yes, sorry. Next time ..

PR #552, opened 2017-04-27
Fix and document a bunch of bugs in the image updating regexes.

paulbellamy:
This was prompted by flux failing to release an image, because I had the `name` and `image` attributes of the map in a different order.

There are still a slew of bugs remaining in this (and policy annotations), but I think they are mostly documented now.

PRed into the git-sync branch because these files have all moved and changed.

### DIFF
--- a/cluster/kubernetes/policies.go
+++ b/cluster/kubernetes/policies.go
@@ -28,11 +28,11 @@ func (c *Cluster) UpdatePolicies(in []byte, update flux.PolicyUpdate) ([]byte, e
 }
 
 func updateAnnotations(def []byte, f func(map[string]string) map[string]string) ([]byte, error) {
-	ann, err := parseAnnotations(def)
+	manifest, err := parseManifest(def)
 	if err != nil {
 		return nil, err
 	}
-	newAnnotations := f(ann)
+	newAnnotations := f(manifest.Metadata.AnnotationsOrNil())
 
 	// Write the new annotations back into the manifest
 	// Generate a fragment of the new annotations.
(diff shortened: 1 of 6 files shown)


cluster/kubernetes/policies.go
```diff
@@ -89,21 +89,40 @@ func updateAnnotations(def []byte, f func(map[string]string) map[string]string)
...
+			} `yaml:"spec"`
+		} `yaml:"template"`
+	} `yaml:"spec"`
 }
 
-func parseAnnotations(def []byte) (map[string]string, error) {
-	var p annotationsParser
-	if err := yaml.Unmarshal(def, &p); err != nil {
-		return nil, errors.Wrap(err, "decoding annotations")
+func (m Metadata) AnnotationsOrNil() map[string]string {
```
squaremo: There's already some code for parsing manifests in `cluster/kubernetes/resources`, could it be used here _mutatis mutandis_?

cluster/kubernetes/policies_test.go
```diff
@@ -99,7 +99,7 @@ func TestUpdatePolicies(t *testing.T) {
 	} {
 		caseIn := templToString(t, annotationsTemplate, c.in)
 		caseOut := templToString(t, annotationsTemplate, c.out)
-		out, err := UpdatePolicies([]byte(caseIn), c.update)
+		out, err := (&Cluster{}).UpdatePolicies([]byte(caseIn), c.update)
```
squaremo: Good evidence that we should split out `Cluster` and um, `ManifestRepo` (or something). But doesn't need to be done here and now.

cluster/kubernetes/update.go
```diff
@@ -81,89 +81,95 @@ func updatePodController(def []byte, newImageID flux.ImageID) ([]byte, error) {
...
+	// Detect how indented the "containers" block is.
+	newDef := string(def)
+	matches := regexp.MustCompile(`( +)containers:.*`).FindStringSubmatch(newDef)
+	if len(matches) != 2 {
+		return fmt.Errorf("could not find container specs")
+	}
+	indent := matches[1]
+
+	// Replace the container images
+	containersRE := regexp.MustCompile(`(?m:^` + indent + `containers:\s*(?:#.*)*$(?:\n(?:` + indent + `[-\s].*)?)*)`)
```
squaremo: Some comments saying what these regexps are for (and more distinct names) would be good here.

(Yes, I am the original offender .. I have sought forgiveness at the altar of Parsec and will continue to do so til my dying day)

cluster/kubernetes/policies.go
squaremo: (Can be tidied later)

paulbellamy: Probably yeah.

PR #555, opened 2017-05-03
Add event logging from the daemon

paulbellamy:
I've only added one example event from the policy update (to avoid merge conflicts).

Are `commit`, and `sync` the only events we want? I was debating if we should keep the `locked`, `automated` events (and send both). They are more descriptive of the users' intent. We could also just make the Commit event metadata more descriptive of *why* the commit was made, and then sort it out on the UI side.


squaremo:
> We could also just make the Commit event metadata more descriptive of why the commit was made, and then sort it out on the UI side.

Yeah I like this idea. We could put the e.g., `ReleaseSpec` (or `PolicyUpdates`) in there.

api/api.go
```diff
@@ -17,6 +17,7 @@ type ClientService interface {
 	SyncNotify(flux.InstanceID) error
 	SyncStatus(flux.InstanceID, string) ([]string, error)
 	UpdatePolicies(flux.InstanceID, flux.PolicyUpdates) (job.ID, error)
+	LogEvent(inst flux.InstanceID, e flux.Event) error
```
squaremo: I think this should go into the `DaemonService`, since it's for daemons to call. (It doesn't make much difference, it's just layout, but good to be consistent).

paulbellamy: I agree, and had put it there, but the method is implemented as part of the api client. Will have a think about it. We could add a `fluxctl add-event -m "incident with the frobulator"` for user-custom events. :sparkles:

squaremo: I guess we could have a daemonClient; or just include it in the DaemonService API but have a very thin wrapper around the transport, for daemon.LogEvent to use. Which may amount to the same thing.

paulbellamy:
re: putting the `updates.Spec` in the `CommitEventMetadata`, I agree it would be nice. I don't see a way to do it without creating an import loop.

paulbellamy:
@squaremo I've cracked the re-org to let me add the update.Spec into the CommitEventMetadata. It added ~700 lines to the diff of mechanical `gofmt -r` changes. Happy to explain the changes if you need.

New history looks like:
```json
[
  {
    "Data": "Commit: 5a846bf6b5f901cd98ab7171cd35c0e42cd8bdc8, flux-example/nginx",
    "Event": {
      "endedAt": "2017-05-03T16:19:02.618089Z",
      "id": 2,
      "logLevel": "info",
      "metadata": {
        "revision": "5a846bf6b5f901cd98ab7171cd35c0e42cd8bdc8",
        "spec": {
          "spec": {
            "flux-example/nginx": {
              "add": [
                "automated"
              ],
              "remove": null
            }
          },
          "type": "policy"
        }
      },
      "serviceIDs": [
        "flux-example/nginx"
      ],
      "startedAt": "2017-05-03T16:19:02.618089Z",
      "type": "commit"
    },
    "Stamp": "2017-05-03T16:19:02.618089Z",
    "Type": "v0"
  }
]
```

PR #559, opened 2017-05-08
Separate Manifests from Cluster

squaremo:
We've been speculating that it makes sense to have operations on config files be (further) distinguished from operations on a cluster (formerly platform). This PR separates those out by introducing an interface `cluster.Manifests`, and changing procedures to use that _or_ `cluster.Cluster` as appropriate.

(It also adds a test for `sync.Sync`, as a sweetener)


paulbellamy:
:+1: for idea, will need rebase tho

PR #560, opened 2017-05-08
Make fluxctl wait for release to finish again

paulbellamy:
Based on #555 

Also:
- emitting events when releasing, and syncing
- moving the tag when syncing
- add lock around the repo, to avoid a pull race condition
- Fix bug in revlist splitting
- sync all files, but only emit changed ones in the history even

## TODO

- [x] Make sure I've not made the tests worse.
- [x] Get rid of "Pending" status on serviceResults, it doesn't make sense anymore.
- [x] Check it works without an existing flux-sync tag
- [x] Hide scary log messages when notes and flux-sync tag don't exist.
- [x] Figure out how to make resource.ServiceIDs better.
- [x] Expand sync loop test


git/repo.go
```diff
@@ -181,37 +187,67 @@ func (c Checkout) CommitAndPush(commitMessage, note string) error {
...
 	}
-	// this fetches and updates the local ref, so we'll see the new
-	// notes; but it's possible that the upstream doesn't have this
-	// ref.
-	if err := fetch(c.repo.Key, c.Dir, c.repo.URL, c.realNotesRef+":"+c.realNotesRef); err != nil &&
-		!strings.Contains(err.Error(), "Couldn't find remote ref") {
-		return err
+	for _, ref := range []string{
+		c.realNotesRef + ":" + c.realNotesRef,
+		c.SyncTag,
```
squaremo: Ah! Whoops, me.

cluster/kubernetes/resource/resource.go
```diff
@@ -36,6 +37,19 @@ func (o *baseObject) debyte() {
 	o.bytes = nil
 }
 
+// TODO: Implement this for more service types
+func (o baseObject) ServiceIDs() []flux.ServiceID {
```
paulbellamy: > though we have gone to lengths elsewhere to avoid that assumption.

sounds sensible

> Can we co-opt FindDefinedServices here to get the list of services affected?

yes and no. figuring out which service a deployment belongs to (same for configmaps, etc), means we need access to all the files, so we can scan the service's selectors for ones which match. Wiring it through is quite a large change.

cluster/kubernetes/resource/deployment.go
```diff
@@ -1,10 +1,41 @@
...
 type Deployment struct {
 	baseObject
 	Spec DeploymentSpec
 }
 
+func (o Deployment) ServiceIDs(all map[string]resource.Resource) []flux.ServiceID {
+	found := flux.ServiceIDSet{}
+
+	// Add the base service for this deployment
+	// TODO: Is this even the right thing to do?
```
paulbellamy: Yeah, let's back this out and do it at once.

cluster/kubernetes/resource/service.go
```diff
@@ -12,12 +17,31 @@ type Service struct {
...
+}
+
 type ServiceSpec struct {
 	Type     string            `yaml:"type"`
 	Ports    []ServicePort     `yaml:"ports"`
 	Selector map[string]string `yaml:"selector"`
 }
 
+func (s ServiceSpec) Matches(l labels.Labels) bool {
+	return labels.SelectorFromSet(labels.Set(s.Selector)).Matches(l)
```
paulbellamy: Unknown, there's also the `omitempty` issue with yaml parsing which will play into this.

cmd/fluxctl/automate_cmd.go
```diff
@@ -12,6 +10,7 @@ import (
 type serviceAutomateOpts struct {
 	*serviceOpts
 	service string
+	verbose bool
```
paulbellamy: We should refactor this out into a `resultPrintingOptions`

daemon/daemon.go
```diff
@@ -123,66 +127,114 @@ func (d *Daemon) queueJob(do JobFunc) job.ID {
...
+				Spec:   &spec,
+				Result: &update.Result{},
+			}
+			for serviceID, u := range s {
 				// find the service manifest
 				err := d.Cluster.UpdateManifest(working.ManifestDir(), string(serviceID), func(def []byte) ([]byte, error) {
-					return d.Cluster.UpdatePolicies(def, update)
+					newDef, err := d.Cluster.UpdatePolicies(def, u)
+					if err != nil {
+						(*metadata.Result)[serviceID] = update.ServiceResult{
```
paulbellamy: Might not be needed. I wasn't sure on `omitempty` and go type aliases to maps interaction.

daemon/loop.go
```diff
@@ -70,11 +75,89 @@ func (d *Daemon) askForSync() {
...
+	// Get a map of all resources defined in the repo
+	allResources, err := d.Cluster.LoadManifests(working.ManifestDir())
+	if err != nil {
+		logger.Log("err", errors.Wrap(err, "loading resources from repo"))
+		return
+	}
+
 	// TODO supply deletes argument from somewhere (command-line?)
-	if err := sync.Sync(d.Checkout.ManifestDir(), d.Cluster, false); err != nil {
+	if err := sync.Sync(allResources, d.Cluster, false); err != nil {
```
paulbellamy: Yes, we could, but this way if you change the kubernetes cluster "live", and update the git repo it will "fix everything"

PR #561, opened 2017-05-09
Move SSH keygen into daemon

awh:
https://docs.google.com/document/d/1CvePrBuCGKz-NP8U_0IeifQkaYpN12CI90IP4Z0zvIY


ssh/keygen.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (ktv *KeyTypeValue) Type() string {
+	return "string"
+}
+
+func (ktv *KeyTypeValue) Specified() bool {
+	return ktv.specified
+}
+
```
philwinder: This lot above seems a little verbose.

ssh/keygen.go
```diff
@@ -0,0 +1,119 @@
...
+		return "", nil, nil, err
+	}
+
+	return privateKeyPath, privateKey, publicKey, nil
+}
+
+// ExtractPublicKey extracts and returns the public key from the specified
+// private key.
+func ExtractPublicKey(privateKeyPath string) (publicKey []byte, err error) {
+	return exec.Command("ssh-keygen", "-y", "-f", privateKeyPath).Output()
```
philwinder: So the private key is written to the current dir in the local filesystem, right? (i.e. in the docker container or wherever it is running...)

cluster/kubernetes/sshkeyring.go
```diff
@@ -0,0 +1,130 @@
...
+	sshKeyRingConfig
+	publicKey      []byte
+	privateKeyPath string
+}
+
+// newSSHKeyRing constructs an sshKeyRing backed by a kubernetes secret
+// resource. The keyring is initialised with the key that was previously stored
+// in the secret (either by regenerate() or an administrator), or a freshly
+// generated key if none was found.
+func newSSHKeyRing(config sshKeyRingConfig) (*sshKeyRing, error) {
```
philwinder: I thought the key should come from k8s secrets, not the os?

philwinder:
No tests. Possibly easy to test generation and access code. More difficult to test the main git workflow.

ssh/keygen.go
philwinder: Oh just seen bug comment earlier on. Yes, would be better to generate the key in a more secure way. Good thinking.

ssh/keygen.go
awh: I didn't have this originally, but the code elsewhere was much uglier as a result...

cluster/kubernetes/sshkeyring.go
awh: I don't follow - it does come from the k8s secret

philwinder: After running the sync->daemon branch I understand why you're accessing through files now. It is k8s that mounts the secret onto the filesystem.

I was expecting us to use the k8s Go API to access the secrets directly.

ssh/keygen.go
awh: Aye, kubernetes mounts secrets in a tmpfs volume inside the container; we re-use that tmpfs to capture the output of ssh-keygen

cluster/kubernetes/sshkeyring.go
awh: > I was expecting us to use the k8s Go API to access the secrets directly.

I didn't want to read the private key into memory any more than necessary, particularly in a long lived process where we have little control over the memory management (other than mclockall future which is a sledgehammer). This way, under normal operation k8s puts the secret in a tmpfs where it it accessed directly by `ssh`, which is written to handle the private key in memory in a secure fashion.

awh:
Deployment instructions for interested parties. Checkout this branch and deploy fluxd to the default namespace:
1. Edit `deploy/flux-deployment.yaml`:
    1. Change `--git-url` as you like
    1. Set `--connect=ws://authfe.default.svc.cluster.local/api/flux`
    1. Set `--token=local-test-token`
1. `kubectl apply -f deploy/memcache-dep.yaml -f deploy/memcache-svc.yaml`
1. `kubectl apply -f deploy/flux-secret.yaml -f deploy/flux-deployment.yaml`

Now checkout the `move-sync-into-daemon` branch of `service-conf` and deploy fluxsvc to the fluxy namespace:
1. `kubectl apply -f service-conf/k8s/local/default`
1. `kubectl apply -f service-conf/k8s/local/fluxy-ns.yaml`
1. `kubectl apply -f service-conf/k8s/local/fluxy`

At this point your fluxd will be crashlooping because it has generated an SSH key on startup and you need to configure that as a GH deploy key. Use `kubectl logs` to inspect the flux pod and you'll find the public key:

```
$ kubectl logs flux-3514388910-f3n7g 
ts=2017-05-23T16:32:19Z caller=main.go:136 component=platform identity=/etc/fluxd/ssh/identity
ts=2017-05-23T16:32:19Z caller=main.go:137 component=platform identity.pub="ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDqvC94ujw5ptzcCdS2mLbceUxWeC30KxJnWvjHIAlbr\n"
...
```

Configure that as a deploy key for your `--git-url` repo and it should come to life. Now you can play with the `fluxctl identity` command:

```
$ fluxctl --token local-test-token --url http://$(minikube ip):30081/api/flux identity
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDqvC94ujw5ptzcCdS2mLbceUxWeC30KxJnWvjHIAlbr
$ fluxctl --token local-test-token --url http://$(minikube ip):30081/api/flux identity --regenerate
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIp5w+uwz5CFEPdV/lrDNpZYit0aO27lq06MieQ3KI2z root@flux-3514388910-f3n7g
```
```
```

awh:
GH is listing the commits out of order because it doesn't cope with rebasing properly - the actual order according to git log is
```
ef004cd Implement ssh-keygen wrapper functions
b500135 SSH keyring backed by k8s secret resource
a7814ab Instantiate keyring and plumb into git operations
20acaf1 Replace GenerateDeployKey with PublicSSHKey
892a236 Update kubernetes deployment manifests
15200a6 Tag built images with quay.io/ prefix
6793824 Retry initial clone in background daemon loop
```

squaremo:
Memo from IRL discussion: the last commit here 6793824 is to account for the extra `NotReady` state this all implies. We think a better way to handle this is as a distinct state of the daemon; but, since this PR is self-contained and gives a working system, it can be reviewed as it stands.

remote/rpc/nats/bus.go
```diff
@@ -285,6 +292,17 @@ func (r *natsPlatform) SyncStatus(ref string) ([]string, error) {
 	return response.Result, extractError(response.ErrorResponse)
 }
 
+func (r *natsPlatform) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
+	var response PublicSSHKeyResponse
+	if err := r.conn.Request(r.instance+methodPublicSSHKey, regenerate, &response, timeout); err != nil {
+		if err == nats.ErrTimeout {
```
philwinder: The SSH regeneration can take a long time depending on the algorithm. Is a 10s timeout long enough? ( I think so, just. I've seen it take about 5s).

remote/metrics.go
```diff
@@ -122,6 +123,16 @@ func (i *instrumentedPlatform) SyncStatus(cursor string) (_ []string, err error)
 	return i.p.SyncStatus(cursor)
 }
 
+func (i *instrumentedPlatform) PublicSSHKey(regenerate bool) (_ ssh.PublicKey, err error) {
```
philwinder: This is mostly measuring (when regenerating keys) the time it takes for the remote fluxd to create a new key.
Is this what we want?
(I suppose there's no harm in collecting, I just think that the duration won't be very useful).

The count might be more interesting though.

daemon/loop.go
```diff
@@ -21,8 +21,24 @@ const (
...
 // Loop for potentially long-running stuff. This includes running
 // jobs, and looking for new commits.
-
 func (d *Daemon) Loop(stop chan struct{}, logger log.Logger) {
+	logger.Log("user", d.GitConfig.UserName,
+		"email", d.GitConfig.UserEmail,
+		"sync-tag", d.GitConfig.SyncTag,
+		"notes-ref", d.GitConfig.NotesRef)
+
+	for d.Checkout == nil {
```
philwinder: Doing the checkout in a while loop here feels a bit wierd.
> Temporary bodge.

Probably best to put a TODO in to look at this again.

PR #564, opened 2017-05-12
Ignore `remote ref` errors when calling getNotesRef

philwinder:


git/repo.go
```diff
@@ -86,7 +86,7 @@ func (r Repo) Clone(c Config) (*Checkout, error) {
 	}
 
 	notesRef, err := getNotesRef(repoDir, c.NotesRef)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "Couldn't find remote ref") {
```
squaremo: `getNotesRef` doesn't try to fetch anything, so I'd be very surprised if it could result in that error (string). Does this definitely avoid a spurious error?

philwinder: getNotesRef uses `execGitCmd` to get the notes. It's this that is erroring out:
https://github.com/weaveworks/flux/blob/8283d84b2b8ec26f72ce5154fdd380c4b9933286/git/operations.go#L91
Yes, it definitely removed the warning.

I'm wondering whether we should do this check on the main `execGitCmd` instead. It seems like the check should apply to every git error command. "If it does error because of this, we don't care".

squaremo: Yeah, but that git command doesn't do any remote operations, it just tells you what the full ref would be for the shortcut name you give it, and it doesn't even have to exist locally. So I can do, for example,

```
$ git notes --ref=madeupjustnow get-ref
refs/notes/madeupjustnow
```

philwinder: After creating a new blank repo I can't get it to create this error anymore. It must have been something bad in my previous repo. Closing for now.

PR #567, opened 2017-05-12
Add git hosts to known_hosts so we don't have to disable StrictHostCh…

philwinder:
…ecking

### DIFF
--- a/docker/Dockerfile.fluxd
+++ b/docker/Dockerfile.fluxd
@@ -2,5 +2,12 @@ FROM alpine:3.5
 WORKDIR /home/flux
 ENTRYPOINT [ "/sbin/tini", "--", "fluxd" ]
 RUN apk add --no-cache openssh ca-certificates tini 'git>=2.3.0'
+
+# Add git hosts to known hosts file so when git ssh's using the deploy
+# key we don't get an unknown host warning.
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.fluxd
```diff
@@ -2,5 +2,14 @@ FROM alpine:3.5
 WORKDIR /home/flux
 ENTRYPOINT [ "/sbin/tini", "--", "fluxd" ]
 RUN apk add --no-cache openssh ca-certificates tini 'git>=2.3.0'
+
+# Add git hosts to known hosts file so when git ssh's using the deploy
+# key we don't get an unknown host warning.
+RUN mkdir ~/.ssh && touch ~/.ssh/known_hosts && \
+    ssh-keyscan -H github.com >> ~/.ssh/known_hosts && \
```
squaremo: Can this be one invocation of `ssh-keyscan`?

squaremo: Is `-H` necessary?

philwinder:
@squaremo The simplest way would be to mount a volume over the top of `/.ssh` with their known hosts.

ssh-keyscan simplified. The -H is to has the ip addresses of the servers in the known hosts. But since they're public hosts anyway, we don't need to hash them.

PR #568, opened 2017-05-16
Use time.Time rather than *time.Time, and conflate Image and ImageDescription

squaremo:
Use a time.Time value rather than a pointer, and fall back to time.IsZero() to detect missing/uninitialised values. Although it's nice to have a nullable value and rely on the JSON encoder to miss it out or serialise it to `null`, it is also a repeating source of confusion, since it's unusual to use a pointer to a value type.

The custom JSON marshalling is not strictly necessary (a zero time value will roundtrip), it's tidier and matches the expectation from clients of the API.

The two types Image and ImageDescription have the same fields and are put to the same use, so
just use one of them.

(The existing HTTP API is also the reason for naming the ID field instead of embedding it: in the API it is serialised as `ID`, so it is simpler to just call it that everywhere.)

### DIFF
--- a/cmd/fluxctl/list_images_cmd.go
+++ b/cmd/fluxctl/list_images_cmd.go
@@ -90,7 +90,7 @@ func (opts *serviceShowOpts) RunE(_ *cobra.Command, args []string) error {
 				}
 				if printLine {
 					createdAt := ""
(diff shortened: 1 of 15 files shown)


squaremo:
Thanks Adam!

PR #576, opened 2017-05-23
Reinstate slack notifications to git-sync branch

philwinder:
To test, create a working setup of flux with both fluxd and fluxsvc with a valid flux helloworld repo.

Add a nodeport to the service manifest so we can access the API.

Make sure you can access the api with `curl -XGET http://192.168.99.100:30303/v4/config`

Now push your slack settings to fluxsvc: `curl -XPOST http://192.168.99.100:30303/v4/config -d @flux.json`

Where flux.json looks like:
```
{"git":{"URL":"","path":"","branch":"","key":""},"slack":{"hookURL":"https://hooks.slack.com/services/JFIODSA/432feas/specialr53wfds","username":"phils-test-flux","releaseTemplate":""},"registry":{"auths":null}}
```
Make sure you change the `hookURL` to point at your slack service URL (available from slack).

Now perform some sort of release and watch it ping your slack:
`fluxctl -u http://192.168.99.100:30303 release --service=default/helloworld --user=phil --message="Because I had to" --update-image=quay.io/weaveworks/helloworld:master-a000001`

### DIFF
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -97,16 +97,6 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
 	}
 	defer working.Clean()
 
-	// update notes and emit events for applied commits
-	revisions, err := working.RevisionsBetween(working.SyncTag+"~1", "HEAD")
-	if isUnknownRevision(err) {
-		// No sync tag, grab all revisions
-		revisions, err = working.RevisionsBefore("HEAD")
-	}
-	if err != nil {
-		logger.Log("err", err)
-	}
-
 	// TODO logging, metrics?
 	// Get a map of all resources defined in the repo
 	allResources, err := d.Manifests.LoadManifests(working.ManifestDir())
@@ -142,6 +132,16 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
 		serviceIDs.Add(r.ServiceIDs(allResources))
 	}
 
+	// update notes and emit events for applied commits
+	revisions, err := working.RevisionsBetween(working.SyncTag, "HEAD")
+	if isUnknownRevision(err) {
+		// No sync tag, grab all revisions
+		revisions, err = working.RevisionsBefore("HEAD")
+	}
+	if err != nil {
+		logger.Log("err", err)
+	}
+
 	// Emit an event
 	if len(revisions) > 0 {
 		if err := d.LogEvent(history.Event{
@@ -156,11 +156,7 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
 		}
 
 		// Find notes in revisions.
-		for i, rev := range revisions {
-			// Skip last revision (since we requested HEAD~1 above)
-			if i == len(revisions)-1 {
-				continue
-			}
+		for _, rev := range revisions {
 			n, err := working.GetNote(rev)
 			if err != nil {
 				logger.Log("err", errors.Wrap(err, "loading notes from repo; possibly no notes"))
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -164,6 +164,7 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
 			n, err := working.GetNote(rev)
 			if err != nil {
 				logger.Log("err", errors.Wrap(err, "loading notes from repo; possibly no notes"))
+				// TODO: We're ignoring all errors here, not just the "no notes" error. Parse error to report proper errors.
 				continue
 			}
 
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -13,6 +13,7 @@ import (
 	"github.com/weaveworks/flux/git"
 	"github.com/weaveworks/flux/history"
 	"github.com/weaveworks/flux/resource"
+	"github.com/weaveworks/flux/update"
 	fluxsync "github.com/weaveworks/flux/sync"
 )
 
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -137,12 +139,40 @@ func (s *Server) SyncStatus(instID flux.InstanceID, ref string) (res []string, e
 }
 
 // LogEvent stores an event in the instance's history.
-func (s *Server) LogEvent(instID flux.InstanceID, event history.Event) error {
+func (s *Server) LogEvent(instID flux.InstanceID, e history.Event) error {
```
squaremo: Comment is out of date (except taken abstractly I suppose ..)

server/server.go
```diff
@@ -137,12 +139,40 @@ func (s *Server) SyncStatus(instID flux.InstanceID, ref string) (res []string, e
...
+	// Log event in history first. This is less likely to fail
+	err = helper.LogEvent(e)
+	if err != nil {
+		return errors.Wrapf(err, "logging event")
+	}
+
+	// If this is a release
+	if e.Type == history.EventRelease {
+		// Casting to ReleaseEventMetadata
+		bytes, err := json.Marshal(e.Metadata)
```
squaremo: What was the problem with just making UnmarshalJSON unmarshal to the specific metadata type?

philwinder: Because there are other types of metadata

daemon/loop.go
```diff
@@ -149,6 +150,52 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
 		}); err != nil {
 			logger.Log("err", err)
 		}
+
+		// Find notes in revisions.
+		for i, rev := range revisions {
+			// Skip last revision (since we requested HEAD~1 above)
+			if i == len(revisions)-1 {
```
philwinder: See this line: https://github.com/weaveworks/flux/blob/a532aac4e14410abfb1d7c1186000ff601a4b599/daemon/loop.go#L97
It requests two revisions, HEAD and HEAD~1. We need to skip HEAD~1 to avoid duplicates.

daemon/loop.go
```diff
@@ -149,6 +150,52 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
...
+				continue
+			}
+			n, err := working.GetNote(rev)
+			if err != nil {
+				logger.Log("err", errors.Wrap(err, "loading notes from repo; possibly no notes"))
```
philwinder: Yes, usually the error is that there are no notes.

I think it's better to just log this error and move on, rather than producing a proper error, because this is quite normal when users first start using flux.

daemon/loop.go
```diff
@@ -149,6 +150,52 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
...
+			if n.Spec.Type == update.Images {
+				// Map new note.Spec into ReleaseSpec
+				spec := n.Spec.Spec.(update.ReleaseSpec)
+				status := update.ReleaseStatusSuccess
+				if n.Result.Error() != "" {
+					status = update.ReleaseStatusFailed
+				}
+				// And create a release event
+				releaseEvent := update.Release{
+					StartedAt: started,
```
philwinder: I don't think I have access to that at this point. It's recorded in the original release request, but not in the notes.

daemon/loop.go
squaremo: Oh I see, it can get an error if the note ref doesn't exist? Fair enough; maybe we can catch that earlier and present it as "no note and no error". Not necessarily in this PR.

philwinder: Ok will add TODO.

daemon/loop.go
squaremo: Are you saying that because it asks for `synctag~1..HEAD`, it will get _at least_ `HEAD~1..HEAD`? If so, and the code can't be more explicit about what's happening, I think it's worth having an explanation at the line mentioned -- i.e., why do we even ask for `synctag~1..HEAD` in the first place?

daemon/loop.go
squaremo: I think this is a good candidate for field tweaking then :) Can you either make sure it's in the right place, or file an issue so we don't assume it's all fine because it has a value ..

server/server.go
squaremo: So unmarshal those as well?

I suppose we could in principle get any type here; but we can at least unmarshal those we know about, and leave the others as `map[string]interface{}`, to be interpreted however.

daemon/loop.go
philwinder: You can tell me why we ask for both! :-) I don't know.

server/server.go
philwinder: The only other type of metadata is raw `map[string]interface{}` that is being written to postgres. This is the only one that needs converting to a struct.

So the only question is where to do it. I did it here to avoid changes to the http handler.

daemon/loop.go
squaremo: In fact since it's a release event, we could have `committed` and `applied`, say. (these are wall-clock times, so :shrug:)

server/server.go
squaremo: > Because there are other types of metadata
> The only other type of metadata is raw map[string]interface{}

?

philwinder: Prior to this code, the only use of the `Event.Metadata` was when it was converted to `map[string]interface{}` by the HTTP handler, then stored in postgres (after re-marshalling again).

I'm the first person to unmarshall the data because I need to to get the info for the notification.

So there are no other things to unmarshall.

squaremo: OK I think I see. This is the only place having something other than `map[string]interface{}` is necessary.

But neither would it hurt to implement `func (e *Event) UnmarshalJSON` so that it constructs the real types, since they would marshal back to JSON for the database. And that would avoid this extra trip through encoding/json.

philwinder: Ok, fair enough. Will see about converting the other one too.

philwinder:
Ready for re-review.

daemon/loop.go
squaremo: My point is that this comment is misleading. We don't request HEAD\~1 (we may end up with it, but equally we might end up with some arbitrary other commit <=HEAD~1).

history/event.go
```diff
@@ -180,3 +182,43 @@ type ReleaseEventMetadata struct {
 	// Message of the error if there was one.
 	Error string `json:"error,omitempty"`
 }
+
+func (e *Event) UnmarshalJSON(in []byte) error {
+	type alias Event
```
squaremo: :100:

daemon/loop.go
squaremo: Also: if the earliest revision is the last in the list, doesn't that mean the events are fired in the wrong order?

daemon/loop.go
```diff
@@ -150,6 +151,49 @@ func (d *Daemon) pullAndSync(logger log.Logger) {
 		}); err != nil {
 			logger.Log("err", err)
 		}
+
+		// Find notes in revisions.
+		for _, rev := range revisions {
```
squaremo: Will these be in the right order? In general we'd want notifications from earliest to latest.

PR #581, opened 2017-05-26
Add daemon tests

philwinder:
Coverage is at about 80%. The parts that are not covered are error cases
(e.g. job failed, etc.).

The git repo being used during the test is real. Hence, any tests that
alter manifests really will alter files.

The k8s cluster used in the daemon is a mock. Hence, this doesn't
actually test whether a new update is applied to the cluster.

The daemon is quite complex to set up. The key thing is stopping the
jobs queue loop and the daemon loop in a specific order. If the queue
stops first, the daemon will receive an invalid nil job.

### DIFF
--- a/daemon/daemon_test.go
+++ b/daemon/daemon_test.go
@@ -1,0 +1,510 @@
+package daemon
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/weaveworks/flux"
+	"github.com/weaveworks/flux/cluster"
+	"github.com/weaveworks/flux/cluster/kubernetes"
+	kresource "github.com/weaveworks/flux/cluster/kubernetes/resource"
+	"github.com/weaveworks/flux/cluster/kubernetes/testfiles"
+	"github.com/weaveworks/flux/git"
+	"github.com/weaveworks/flux/git/gittest"
+	"github.com/weaveworks/flux/history"
+	"github.com/weaveworks/flux/job"
+	"github.com/weaveworks/flux/policy"
+	"github.com/weaveworks/flux/registry"
+	"github.com/weaveworks/flux/resource"
+	"github.com/weaveworks/flux/update"
+)
+
+const (
+	svc               = "default/helloworld"
+	ns                = "default"
+	invalidNS         = "adsajkfldsa"
+	testVersion       = "test"
+	newHelloImage     = "quay.io/weaveworks/helloworld:2"
+	currentHelloImage = "quay.io/weaveworks/helloworld:master-a000001"
+)
+
+var (
+	testBytes = []byte(`{}`)
+	timeout   = 5 * time.Second
+)
+
+// When I ping, I should get a response
+func TestDaemon_Ping(t *testing.T) {
+	d, clean, _, _ := mockDaemon(t)
+	defer clean()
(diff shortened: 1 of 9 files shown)


daemon/daemon_test.go
```diff
@@ -0,0 +1,513 @@
...
+		close(shutdownDaemon)
+		daemonWg.Wait() // Wait for it to close, it might take a while
+		close(shutdownQueue)
+		jobsWg.Wait()
+		repoCleanup()
+	}, k8s, events
+}
+
+// DAEMON TEST HELPERS
+type wait struct {
```
squaremo: This is a neat way to make the tests read well!

git/gittest/repo.go
```diff
@@ -38,6 +38,10 @@ func Repo(t *testing.T) (git.Repo, func()) {
 		cleanup()
 		t.Fatal(err)
 	}
+	if err = execCommand("git", "-C", filesDir, "tag", "flux-test"); err != nil {
```
squaremo: This will mean that the case of a fresh repo will not be tested. If you want to avoid having to deal with that when testing aspects of the daemon, perhaps add a method here for initialising the tag and use it from there.

job/job.go
```diff
@@ -89,7 +89,8 @@ func (q *Queue) ForEach(fn func(int, *Job) bool) {
 	}
 }
 
-func (q *Queue) Loop(stop chan struct{}) {
+func (q *Queue) Loop(stop chan struct{}, wg *sync.WaitGroup) {
+	defer wg.Done()
```
squaremo: Similar to above, I think better to do `Add()` here as well.

daemon/daemon_test.go
```diff
@@ -0,0 +1,513 @@
+package daemon
+
+import (
```
philwinder: a.k.a. goimports

git/gittest/repo.go
philwinder: Yes, the issue with this is that I think a few of the tests require a tag to be set. Although I haven't tracked them down yet. When I take this out something stops working.

daemon/loop.go
```diff
@@ -22,7 +23,8 @@ const (
 // Loop for potentially long-running stuff. This includes running
 // jobs, and looking for new commits.
 
-func (d *Daemon) Loop(stop chan struct{}, logger log.Logger) {
+func (d *Daemon) Loop(stop chan struct{}, wg *sync2.WaitGroup, logger log.Logger) {
+	defer wg.Done()
```
philwinder: Can't do this, this would case a race. I.e. no guarantee that `wg.Add()` inside the goroutine would run be for the outer `wg.wait()`.

(I only realised this by trying it out! :-) )

job/job.go
philwinder: See https://github.com/weaveworks/flux/pull/581#discussion_r119036992

git/gittest/repo.go
philwinder: Ah, not quite true.
I think SyncStatus requires a flux tag. Since it does a revision search from `flux-tag..$HASH`.

So I'll need to add the tag for this one.

daemon/loop.go
squaremo: Oh! Yes very good point :blush: . That suggests to me that `WaitGroup`s aren't suitable for things that get called as `go Loop(...)`*. But, I think we _do_ want a `WaitGroup` here (so that one can be used for both the queue and the daemon, after removing the shutdown order dependency), so let's stick with it.

PR #584, opened 2017-05-30
Deal only in quay.io/ image names

squaremo:
quay.io is where we push images, so just build images with that in the
name, rather than requiring people (i.e., me) to retag.


squaremo:
NB I have also removed the pushing of `:latest` images. `:latest` images are to make something you shouldn't do more convenient to do.

PR #586, opened 2017-05-31
Explicitly model "not ready" daemon state

awh:
To facilitate configuration of a generated keypair as a deploy key via
the UI, it is necessary to move the initial git clone operation after
the upstream connection, and to retry it periodically until it is
successful. This commit introduces a NotReadyDaemon which is capable of
supporting specific operations critical to the task (e.g. getting the
public key) but not others which depend on the cloned config repository;
once the full daemon is available, it is substituted in its place.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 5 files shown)


cmd/fluxd/main.go
```diff
@@ -211,18 +265,21 @@ func main() {
...
-		working, err := repo.Clone(gitConfig)
-		if err != nil {
-			logger.Log("component", "git", "err", err.Error())
-			os.Exit(1)
+		for checkout == nil {
+			working, err := repo.Clone(gitConfig)
+			if err != nil {
+				logger.Log("component", "git", "err", err.Error())
+				notReadyDaemon.UpdateReason(err)
+				time.Sleep(time.Minute)
```
philwinder: I think a minute is a bit too high. Recommend 10s?

PR #589, opened 2017-06-02
Automator tag filtering

### LINKED ISSUE #260: More specific automator policies
As container solutions, I'd like more specific control over what the automator is doing, so that I can push images willy-nilly without having them all be deployed.

In more detail, they are pushing images to quay for feature-branches, which then the automator just deploys. They move a tag to indicate the latest master. Because of this, just deploying the "newest" available image is not good enough. It might be good if the automator had a "track tag" mode.

sambooo:
This adds image tag filtering to automation so that we don't always just deploy the latest pushed tag. Currently just does string globbing as implemented by [go-glob](https://github.com/ryanuber/go-glob).


update/release.go
```diff
@@ -54,10 +54,11 @@ type Release struct {
 // NB: these get sent from fluxctl, so we have to maintain the json format of
 // this. Eugh.
 type ReleaseSpec struct {
-	ServiceSpecs []ServiceSpec
-	ImageSpec    ImageSpec
-	Kind         ReleaseKind
-	Excludes     []flux.ServiceID
+	ServiceSpecs  []ServiceSpec
+	ContainerName string
```
squaremo: Could we have some commentary on the meaning and use of this field please? E.g., what happens if the ServiceSpec is `<all>`, and the ImageSpec is `<latest>` (or whatever those literals are).

daemon/images.go
```diff
@@ -13,65 +14,112 @@ import (
...
+	}
+	if p, ok := patterns[serviceID][container]; ok {
+		return p
+	}
+	return "*"
 }
 
-func (d *Daemon) NewImage(imageID flux.ImageID) error {
+func (d *Daemon) ReleaseImage(serviceID flux.ServiceID, container string, imageID flux.ImageID) error {
 	// Try to update any automated services using this image
```
squaremo: This comment is out of date; but it also suggests why things worked in this way -- the idea was to release, in a single go, all the services that used a particular image. This way, the notifications are kept concise.

Now that we are filtering (and specifying particular containers), it's not necessarily the case that a single (new) image will be released to all services that use that image repo. But it would be good to preserve as far as possible the minimal notifications.

cluster/manifests.go
```diff
@@ -27,6 +27,8 @@ type Manifests interface {
 	UpdatePolicies([]byte, policy.Update) ([]byte, error)
 	// ServicesWithPolicy finds the services which have a particular policy set on them.
 	ServicesWithPolicy(path string, p policy.Policy) (flux.ServiceIDSet, error)
+	// ServicesMetadata gets metadata for all services
+	ServicesMetadata(path string) (map[flux.ServiceID]map[string]string, error)
```
sambooo: You're right, these could be combined with a change to the return type of ServicesWithPolicy.

sambooo:
Following on from @squaremo's comments above, I need to add more than the internal logic to this before it can be merged:
- Refactor Manifest as suggested
- Update CLI in all relevant places - automate, release, etc.
- Document edge cases and produce useful errors/warnings in both fluxctl and fluxd

Additional changes that are out of scope for this PR:
- Drastically reduce image repo requests with tag-filters **and heavily suggest users do this**
- Add (or modify) a CLI command to add a tag-filter without automating or releasing

squaremo:
> Following on from @squaremo's comments above, I need to add more than the internal logic to this before it can be merged:
> 
>     Refactor Manifest as suggested
>     Update CLI in all relevant places - automate, release, etc.

We will get an opportunity to try it out with just the annotations, in our dev environment. If that would help with designing the CLI changes, I think you can defer this line item. If you already have a good idea of how it should work, though ... proceed :)

>     Document edge cases and produce useful errors/warnings in both fluxctl and fluxd
>     Tests

Good idea about corner cases; this may help flush out any surprises in behaviour too.

> Additional changes that are out of scope for this PR:
> 
>     Drastically reduce image repo requests with tag-filters and heavily suggest users do this
>     Add (or modify) a CLI command to add a tag-filter without automating or releasing

:white_check_mark: I agree these are out of scope.

release/releaser.go
```diff
@@ -35,26 +34,57 @@ func Release(rc *ReleaseContext, spec update.ReleaseSpec, cause update.Cause, lo
...
+	timer.ObserveDuration()
+	if err != nil {
+		return nil, nil, err
 	}
+	return updates, results, nil
+}
 
-	// If the request was for a specific service and the service was
-	// not in the cluster, then add a result with a skipped status
+// marks any service skipped if it's not in the cluster
```
squaremo: This comment does not match the error `NotInRepo` below

rade:
What is left to do here?

sambooo:
@rade 
> What is left to do here?

Assuming @squaremo likes the actual implementation,
- User-facing stuff (commit messages, slack notifications, CLI output, etc.)
- Testing

daemon/daemon.go
```diff
@@ -186,35 +185,8 @@ func (d *Daemon) UpdateManifests(spec update.Spec) (job.ID, error) {
...
-				}
-				defer d.askForSync()
-			}
-			return &history.CommitEventMetadata{
-				Revision: revision,
-				Spec:     &spec,
-				Result:   result,
-			}, nil
-		}), nil
+	case release.Changes:
```
squaremo: I don't understand this change. `http/daemon/server.go` constructs an `update.Spec` with a `release.Spec` in it to give to this method, doesn't it?

release/releaser.go
```diff
@@ -1,279 +1,51 @@
...
-	}(started)
+type Observer interface {
+	Observe(time.Time, error)
+}
 
-	logger = log.NewContext(logger).With("type", "release")
-	// We time each stage of this process, and expose as metrics.
-	var timer *metrics.Timer
+func Release(rc *ReleaseContext, changes Changes, logger log.Logger) (results update.Result, err error) {
+	if o, ok := changes.(Observer); ok {
```
squaremo: This seems destined to cause confusion -- why would changes observe anything? Better to be explicit, either by wrapping this function, or having another argument.

update/automated.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return updates, result, err
+}
+
+func (a *Automated) ReleaseKind() ReleaseKind {
+	return ReleaseKindExecute
+}
+
+func (a *Automated) CommitMessage() string {
+	return "(TODO) automated release commit message"
```
squaremo: I think this oughta be todone :)

daemon/daemon.go
squaremo: Ah, because this is an interface, which both `Automated` and `ReleaseSpec` implement. Cool.

cluster/kubernetes/update.go
```diff
@@ -34,7 +34,7 @@ func updatePodController(def []byte, newImageID flux.ImageID) ([]byte, error) {
 	}
 
 	var buf bytes.Buffer
-	err = tryUpdate(def, newImageID, &buf)
+	err = tryUpdate(def, stripNamespace(container), newImageID, &buf)
```
squaremo: (We figured this out IRL -- it was a bandaid over a problem with the test fixture)

policy/policy.go
```diff
@@ -17,56 +17,88 @@ const (
...
+	case Locked, Automated, Ignore:
+		return true
 	}
-	return None
+	return false
 }
 
 type Updates map[flux.ServiceID]Update
 
 type Update struct {
```
squaremo: Changing this type also changes the effective argument type in the API, which breaks the backwards-compatible routes (which, to be fair, were put in place long after this PR was started). It is not a huge deal since we will shortly be removing them; just noting FTR.

daemon/images.go
```diff
@@ -13,76 +13,80 @@ import (
...
+	if err != nil {
+		return nil, err
 	}
+	return automatedServices.Without(lockedServices), nil
+}
 
-	_, err := d.UpdateManifests(update.Spec{Type: update.Images, Cause: cause, Spec: spec})
+func (d *Daemon) ReleaseChanges(changes release.Changes) {
+	cause := update.Cause{}
+	_, err := d.UpdateManifests(update.Spec{Type: update.Images, Cause: cause, Spec: changes})
```
sambooo: Wow this one is really bad, good catch

update/release.go
```diff
@@ -59,6 +73,240 @@ func (s ReleaseSpec) ReleaseType() string {
...
+				Status: ReleaseStatusSkipped,
+				Error:  ImageNotFound,
+			}
+		}
+	}
+
+	return updates, nil
+}
+
+func (s ReleaseSpec) Observe(start time.Time, err error) {
```
sambooo: Yes, an artifact of that bad `Observer` interface

service.go
```diff
@@ -53,60 +53,60 @@ func (id ServiceID) Components() (namespace, service string) {
 	return toks[0], toks[1]
 }
 
-type ServiceIDSet map[ServiceID]struct{}
+type ServiceIDMap map[ServiceID]map[string]string
```
sambooo: You aren't, this is a remnant of a former design that is no longer with us

update/release.go
sambooo: I didn't realise that was an empty method. Must be more careful with my commits.

update/release.go
```diff
@@ -59,6 +72,237 @@ func (s ReleaseSpec) ReleaseType() string {
...
+	// Service filter
+	ids := []flux.ServiceID{}
+specs:
+	for _, s := range s.ServiceSpecs {
+		switch s {
+		case ServiceSpecAll:
+			// "<all>" Overrides any other filters
+			ids = []flux.ServiceID{}
+			break specs
+		case ServiceSpecAutomated:
```
squaremo: Is this still used?

sambooo: It is not

PR #592, opened 2017-06-02
Compare repos on image release

### LINKED ISSUE #579: [gitsync] Releasing same image gives "skipped: cannot find image"
```
$ fluxctl $DAEMON list-services
SERVICE                    CONTAINER          IMAGE                                        RELEASE  POLICY
...
default/helloworld         helloworld         quay.io/squaremo/hello_world:master-a000002  ready    
                           sidecar            quay.io/weaveworks/sidecar:master-a000002             
...
$ fluxctl $DAEMON release --service=default/helloworld --update-image=quay.io/squaremo/hello_world:master-a000002
Requesting release ...
default/helloworld  skipped  cannot find one or more images
```

The result -- do nothing -- is correct, but the description is misleading.

sambooo:
The misleading description in @squaremo's particular case arose because we try to update all containers within the service with a subset of all images.

We currently call ExactImages with the 'update-image' which gives us an ImageMap containing only a key for that repository, so calling LatestImage for any other repository will always return nil.

Fixes #579

### DIFF
--- a/release/context.go
+++ b/release/context.go
@@ -15,14 +15,15 @@ import (
 )
 
 const (
-	Locked         = "locked"
-	NotIncluded    = "not included"
-	Excluded       = "excluded"
-	DifferentImage = "a different image"
-	NotInCluster   = "not running in cluster"
-	NotInRepo      = "not found in repository"
-	ImageNotFound  = "cannot find one or more images"
-	ImageUpToDate  = "image(s) up to date"
+	Locked          = "locked"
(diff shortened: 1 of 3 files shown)


sambooo:
Could do with a nice explanatory refactor perhaps. Is that worth its own issue just so we don't forget it?

squaremo:
> Is that worth its own issue just so we don't forget it?

https://github.com/weaveworks/flux/issues/599

PR #597, opened 2017-06-05
Bump alpine base image to 3.6

awh:
quay.io security scanner reports a security vulnerability in zlib which is addressed in alpine 3.6


sambooo:
LGTM

PR #603, opened 2017-06-07
Enforce event Metadata field to be a pointer

awh:
Nice idea @squaremo!

### DIFF
--- a/history/event.go
+++ b/history/event.go
@@ -59,7 +59,7 @@ type Event struct {
 
 	// Metadata is Event.Type-specific metadata. If an event has no metadata,
 	// this will be nil.
-	Metadata interface{} `json:"metadata,omitempty"`
+	Metadata EventMetadata `json:"metadata,omitempty"`
 }
 
 func (e Event) ServiceIDStrings() []string {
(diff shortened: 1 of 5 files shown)


history/event.go
```diff
@@ -207,18 +209,43 @@ func (e *Event) UnmarshalJSON(in []byte) error {
...
 		break
 	default:
 		if len(wireEvent.MetadataBytes) > 0 {
-			var metadata interface{}
+			var metadata UnknownEventMetadata
 			if err := json.Unmarshal(wireEvent.MetadataBytes, &metadata); err != nil {
 				return err
 			}
-			e.Metadata = metadata
+			e.Metadata = &metadata
```
squaremo: Does this need to be a pointer? (it's a bit unusual to be passing around pointers to maps, unless you really want someone to swap the whole thing)

history/event.go
```diff
@@ -207,18 +209,43 @@ func (e *Event) UnmarshalJSON(in []byte) error {
...
+
+func (cem *SyncEventMetadata) Type() string {
+	return EventSync
+}
+
+func (rem *ReleaseEventMetadata) Type() string {
+	return EventRelease
+}
+
+func (uem *UnknownEventMetadata) Type() string {
```
squaremo: See above re pointers to maps. I guess it looks odd if you don't uniformly use pointer receivers?

awh:
Converted to value receiver with explanatory comment.

PR #604, opened 2017-06-07
Fix inference of upstream API client endpoint

awh:
Currently we infer `http://` from `wss://`, causing authfe to respond to event posting with a redirect. The client doesn't follow this, so events are dropped on the floor. The patch also makes `--connect` more liberal; it will accept either ws or http forms and infer the other as appropriate.

### DIFF
--- a/http/daemon/upstream_test.go
+++ b/http/daemon/upstream_test.go
@@ -1,0 +1,39 @@
+package daemon
+
+import (
+	"testing"
+)
+
+func TestEndpointInference(t *testing.T) {
+	wsEndpoint := "ws://cloud.weave.works/api/flux"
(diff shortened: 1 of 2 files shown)


http/daemon/upstream_test.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func assertEquals(t *testing.T, expected, actual string) {
+	if expected != actual {
+		t.Errorf("Expected [%s], actual [%s]", expected, actual)
+	}
+}
+
+func TestUnsupportedEndpoint(t *testing.T) {
+	_, _, err := inferEndpoints("mailto://cloud.weave.works/api/flux")
```
squaremo: Pls file a bug for supporting mailto as a scheme :jk:

PR #616, opened 2017-06-15
Fix cause of release events

squaremo:
Release events in history and notifications were not showing the user (e.g., "automation") or message (when supplied).

While rewriting for gitsyncness, the release event structure got a bit overcomplicated. This PR simplifies it, and makes sure that the cause (user and message) is faithfully transcribed to the event sent upstream to the service.

It also fixes a couple of minor bugs and makes a small refactor, as detailed in the first commit message.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"sort"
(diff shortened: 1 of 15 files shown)


sambooo:
Pretty straightforward, LGTM

PR #618, opened 2017-06-15
Mark ignored resources

squaremo:
* treat "ignore" like other policy
 * show it in `fluxctl list-services` (and in the corresponding API)

(NB I have tried to keep out of the way of #589 :)

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -9,11 +9,6 @@ import (
 	"github.com/weaveworks/flux/resource"
 )
 
-const (
-	ResultDelete = "delete"
-	ResultApply  = "apply"
-)
-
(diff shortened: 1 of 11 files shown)


sync/sync.go
```diff
@@ -1,21 +1,21 @@
...
 import (
+	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
 
 	"github.com/weaveworks/flux/cluster"
+	"github.com/weaveworks/flux/policy"
 	"github.com/weaveworks/flux/resource"
 )
 
 const (
```
sambooo: These constants are unused, clean the campground

Issue #595, opened 2017-06-05
[gitsync] Failure to push release/policy change means waiting for next sync


squaremo:
If someone has pushed a commit to the config repo from outside flux, it can take a while for that to be noticed (up to five minutes, which is the period of the automatic sync). In the meantime, everything that would require a change to the repo will fail; and, there is no way to overcome this other than waiting.

Possibly the simplest way to improve this would be to notice when an operation has failed specifically because of a push conflict, and in that case schedule a sync, and tell the user to try again. The sync will most likely run before the user can resubmit the job, so the latter is likely to succeed.

An alternative would be to give the user a means to trigger a sync; there's no way to do this synchronously at present (i.e., to trigger it then wait for it to complete); but it still has the same property as above, that the sync will likely happen before the operation is retried.

squaremo:
#623 implemented the first option: if a job fails to push its commit, it will pull from the upstream repo before returning. That means the next thing that runs will at least have an up-to-date repo.

Issue #591, opened 2017-06-02
Schedule an automation run when automation is enabled for a service


awh:
Currently when you enable automation for a service you have to wait up to five minutes to see any images updates applied - scheduling an immediate automation run will result in a better user experience.

squaremo:
#623 made policy update jobs ask for an image poll if something was newly automated.
However, since it separated the git and image polling loops, it was only partially successful. If you ask for an image poll inside the job, chances are that will run straight away (i.e., before the commit has been pulled from upstream), and if so, it will not see the newly automated service as automated!

sambooo:
@squaremo did #661 resolve this?

sambooo:
Flux will now do an automation run after any commit is synced, which happens when automation in enabled.

PR #623, opened 2017-06-19
Resync on push fail, and poll images straight after automating something

squaremo:
Rewrites the loop so that

 1. if a job fails when pushing new commits, a pull will be done straight after so that a subsequent attempt has more chance of succeeding (issue #595)
 2. if something has been automated, run an image poll ASAP so that the something will be updated (issue #591)

Also improves a couple of minor things: error messages from git are slightly better (looks for git output beginning with `error: ` as well as `fatal: `); and adapts a use of `logrus` (how did it get there?) to `go-kit/kit/log` which we use everywhere else.

### DIFF
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -21,7 +21,15 @@ type LoopVars struct {
 	GitPollInterval      time.Duration
 	RegistryPollInterval time.Duration
 	syncSoon             chan struct{}
-	initSyncSoon         sync.Once
+	pollImagesSoon       chan struct{}
+	initOnce             sync.Once
+}
+
+func (loop *LoopVars) ensureInit() {
+	loop.initOnce.Do(func() {
+		loop.syncSoon = make(chan struct{}, 1)
+		loop.pollImagesSoon = make(chan struct{}, 1)
+	})
 }
 
 // Loop for potentially long-running stuff. This includes running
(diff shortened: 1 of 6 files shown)


daemon/loop.go
```diff
@@ -21,40 +21,74 @@ type LoopVars struct {
...
 // jobs, and looking for new commits.
 
 func (d *Daemon) Loop(stop chan struct{}, wg *sync.WaitGroup, logger log.Logger) {
 	defer wg.Done()
-	pollGit := time.NewTimer(d.GitPollInterval)
+	gitPollTimer := time.NewTimer(d.GitPollInterval)
+	// This can be called any time to push the next git poll out by
+	// d.GitPollInterval. It must _at least_ be called when the timer
+	// fires, so that the polling continues.
 	resetGitPoll := func() {
```
philwinder: There are some functions defined as methods on the daemon (e.g. `doSync`) and others that are inlined

Why have you inlined this function and the two below? Personally, I think it bloats the Loop method.

I would prefer separate methods / class for the git timer stuff, and move `pullThen` as a proper Daemon method (because of it's use of `d.Checkout...`).

But obviously, this is just a style issue nit. No biggie.

daemon/loop.go
```diff
@@ -21,40 +21,74 @@ type LoopVars struct {
...
+			imagePollTimer = time.NewTimer(d.RegistryPollInterval)
 		}
 	}
 
-	pollImages := time.Tick(d.RegistryPollInterval)
-	// Ask for a sync straight away
+	// Ask for a sync, and to poll images, straight away
 	d.askForSync()
+	d.askForImagePoll()
 	for {
```
philwinder: I think it might be worth spltting the git loop and the image poll loop. Then you could reduce the number of cases in each loop and reduce duplication. For example, you could just do one `resetGitPull()` at the bottom of the loop, rather than repeating it in every case statement.

Another nit.

squaremo:
Thanks for the review Phil -- I was not overly confident in how I'd done this so it's good to have your feedback. I'll give it another go.

daemon/loop.go
squaremo: Yeahp, style difference :) Let me explain my thinking a bit, for posterity.

It is mainly a matter of what's convenient.

`doSync` is long*, though reasonably self-contained, and aside from its receiver, only needs a logger. It's convenient to be able to use it as a value.

`pullThen` is much shorter, and used like a local macro (i.e., not relevant anywhere else). It's convenient to close over the logger and daemon rather than supplying them as arguments. But it's not _toooo_ much different as a method, and if it were useful to call from elsewhere, it could easily become one.

*we should break it down into its parts! later ...

daemon/loop.go
squaremo: Yeah it's really two loops, tangled together, and that makes it hard to understand. I'll see if I can detangle them, and if that makes the code clearer.

PR #624, opened 2017-06-19
Don't treat "no note for commit" as an error

squaremo:
`git note show` will exit with 1 if there is no note for a given
commit (including the case of the notes ref being unknown). We want to
distinguish between this and there being a genuine error reading a
note: so, return `(nil, error)` if there was a problem, and `(nil,
nil)` if there was simply no note.

This needed better git error message parsing, so I cherry-picked a
commit doing just that from `resync-on-push-fail`.

### DIFF
--- a/git/operations.go
+++ b/git/operations.go
@@ -86,7 +86,7 @@ func refExists(workingDir, ref string) (bool, error) {
 	return true, nil
 }
 
-// Get the full ref for a shorthand notes ref
+// Get the full ref for a shorthand notes ref.
 func getNotesRef(workingDir, ref string) (string, error) {
 	out := &bytes.Buffer{}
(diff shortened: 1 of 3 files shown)


git/operations.go
```diff
@@ -103,9 +103,13 @@ func addNote(workingDir, rev, notesRef string, note *Note) error {
 	return execGitCmd(workingDir, nil, nil, "notes", "--ref", notesRef, "add", "-m", string(b), rev)
 }
 
+// NB return values (*Note, nil), (nil, error), (nil, nil)
 func getNote(workingDir, notesRef, rev string) (*Note, error) {
 	out := &bytes.Buffer{}
 	if err := execGitCmd(workingDir, nil, out, "notes", "--ref", notesRef, "show", rev); err != nil {
+		if strings.Contains(err.Error(), "No note found for object") {
```
sambooo: This needs to match both `N` and `n` for different git versions. I'll make this change.

PR #628, opened 2017-06-20
Add 'after' parameter to history API

leth:
I'm not sure this is the best way to add this parameter, I've simply followed 'before' through the code, and added 'after' wherever it complained. That probably breaks compatibility somewhere...


leth:
I think `time.Time{}` is used in a test when against the mock history backend, which doesn't try to use the time objects.

PR #633, opened 2017-06-30
Feature/registry rate limiting

philwinder:
Hi everyone.
This is two things.
1) Implementation of the "How to better look for new images": https://docs.google.com/document/d/1z3CqJaj8oiatIdNnxW2qbSg4WBf9lSHElUjyg9KVGP4/edit
2) Refactor of the registry package.

**New features**
- Super fast `fluxctl list-images` output. All info is now cached, so there is no secondary call out to a remote registry to wait for.
- No more `you have been rate limited` or `timeout exception` user messages, because the data is either in the cache, or it's not.
- Refactoring of the registry code; much easier to understand.
- A cache warmer, to make sure the registry information is up-to-date as quickly as possible.

**Knock on effects**
- Memcache is now a requirement (we could possibly implement a filestore-based cache?)

**Dev Notes**
There are a lot of changes. It's up to you how you review it, but I might recommend reviewing commit-by-commit. I've made the commits fairly linear (apart from some bugs). But you might find that some of the code might have been deleted later on.

The test coverage is pretty good, and relies a lot on the integration test (which only runs when you do `-tags integration`, because it requires memcache).

I've stripped down the memcache interface to make it simpler. I've consolidated a lot of the registry/remote_factory/remote stuff as it didn't make sense any more.

The design now looks like (from the top): Registry->ClientFactory->Client. A call to get tags or images from a `Registry` creates a new `Client` from a `ClientFactory` based upon the host (each host needs different settings). The client can either be a `Cache` or a `Remote`.

`warming.go` holds all the cache warming stuff. It has a `Queue` to provide new images to warm and a `Warmer` to reach out to the `Remote` and insert into the `Cache`.

The rate limiting is done as a `RoundTripper` middleware for the `http.Client`. This works great, except that if it does rate limit, or timeout, you get a hard cancel which can spout `context cancelled errors` (see todo).

`monitoring`, `mocks`, `credentials` and the other middlewares have been left as is. I might remove the `backoff` middleware as we don't need to backoff if we correctly rate limit.

**Todos**

There are some things left to do:
- [x] Consolidate use of `flux.ImageID, tag string` in signatures, because the tag is part of ImageID. (legacy thing)
- [x] Improve the errors shown to users. At the moment we show "err cache miss"
 - Add error wrapper to provide better errors
 - Alter result of list-images to prevent erroring when an image isn't found. Instead, list "waiting to be cached" or whatever.
- [x] Better queing algorithm. At the moment I just randomly pick a container.
- [x] Better scheduling (maybe). At the moment we just tick every x ms for new containers.
- [x] Get rid of "connection disconnected" error in logs due to rate limiting cancel.
- [x] Perform a proper, monitored rate limit test to see how much we can push docker hub (and whether the rate limiting works in real life)
- [x] Set better defaults for rate limits, number of threads, timeouts etc.
- [x] Remove backoff if tests prove fruitful.
- [ ] Refactor warming
- [ ] improve coverage on warming
---

## Errors shown to users

To test: start everything as usual. Kill the memcache container. Now keep doing a `list-images` over time the list will change from `waiting for cache` to information about containers.

```
$ fluxctl -u http://192.168.99.100:30303 list-images
SERVICE             CONTAINER   IMAGE                              CREATED
default/flux        flux        quay.io/weaveworks/flux            waiting for cache
                    fluxsvc     quay.io/weaveworks/flux-service    waiting for cache
default/helloworld  helloworld  quay.io/weaveworks/helloworld      waiting for cache
                    sidecar     quay.io/weaveworks/sidecar         waiting for cache
default/kubernetes                                                 
default/memcached   memcached   index.docker.io/library/memcached  waiting for cache
default/nats        nats        index.docker.io/library/nats       waiting for cache

$ fluxctl -u http://192.168.99.100:30303 list-images
SERVICE             CONTAINER   IMAGE                              CREATED
default/flux        flux        quay.io/weaveworks/flux            waiting for cache
                    fluxsvc     quay.io/weaveworks/flux-service    waiting for cache
default/helloworld  helloworld  quay.io/weaveworks/helloworld      waiting for cache
                    sidecar     quay.io/weaveworks/sidecar         
                                '-> master-a000002                 23 Aug 16 10:05 UTC
                                    master-a000001                 23 Aug 16 09:53 UTC
default/kubernetes                                                 
default/memcached   memcached   index.docker.io/library/memcached  waiting for cache
default/nats        nats        index.docker.io/library/nats       waiting for cache

$ fluxctl -u http://192.168.99.100:30303 list-images
SERVICE             CONTAINER   IMAGE                              CREATED
default/flux        flux        quay.io/weaveworks/flux            
                                |   master-db05f0f                 10 Jul 17 15:33 UTC
                                |   master-2cd55ca                 10 Jul 17 14:20 UTC
                                |   master-1a6088c                 10 Jul 17 14:15 UTC
                                |   master-0024666                 10 Jul 17 10:30 UTC
                                |   master-73a088b                 07 Jul 17 09:26 UTC
                                |   master-a26a2f6                 05 Jul 17 13:34 UTC
                                |   master-cb9fd4c                 04 Jul 17 16:04 UTC
                                |   master-e13450f                 30 Jun 17 17:04 UTC
                                |   master-52db12c                 28 Jun 17 10:58 UTC
                                |   master-2c83861                 27 Jun 17 14:03 UTC
                    fluxsvc     quay.io/weaveworks/flux-service    waiting for cache
default/helloworld  helloworld  quay.io/weaveworks/helloworld      
                                '-> master-9a16ff945b9e            20 Jul 16 13:19 UTC
                                    master-b31c617a0fe3            20 Jul 16 13:19 UTC
                                    master-a000002                 12 Jul 16 17:17 UTC
                                    master-a000001                 12 Jul 16 17:16 UTC
                    sidecar     quay.io/weaveworks/sidecar         
                                '-> master-a000002                 23 Aug 16 10:05 UTC
                                    master-a000001                 23 Aug 16 09:53 UTC
default/kubernetes                                                 
default/memcached   memcached   index.docker.io/library/memcached  waiting for cache
default/nats        nats        index.docker.io/library/nats       waiting for cache
...
```


squaremo:
Thanks for the great PR write-up :star:

> Super fast fluxctl list-images output. All info is now cached, so there is no secondary call out to a remote registry to wait for.
> No more you have been rate limited or timeout exception user messages, because the data is either in the cache, or it's not.
> Refactoring of the registry code; much easier to understand.
> A cache warmer, to make sure the registry information is up-to-date as quickly as possible.

:heart:

OK now I'll look at the actual changes ...

squaremo:
Arrangement of the pieces looks good to me :+1:

I'll reserve specific comments for later, since I think you're still changing things -- I've made some markings on paper.

> Improve the errors shown to users. At the moment we show "err cache miss"
> Add error wrapper to provide better errors
> Alter result of list-images to prevent erroring when an image isn't found. Instead, list "waiting to be cached" or whatever.

On these three: yes I like the idea of providing partial results, i.e., answering with what we know and marking things as missing otherwise.

>  Better queing algorithm. At the moment I just randomly pick a container.

Here is a sketchy suggestion: use a priority queue, with high priority for things that just got requested, medium priority for things that just aren't known yet, and lower for things that are coming up for expiry (or just old).

May be tricky to combine this with a receive loop, as I found with the job queue, though. And other trickinesses no doubt.

>  Better scheduling (maybe). At the moment we just tick every x ms for new containers.

Yep. May as well do this as fast as you can find things for it to do.

>  Perform a proper, monitored rate limit test to see how much we can push docker hub (and whether the rate limiting works in real life)
>  Set better defaults for rate limits, number of threads, timeouts etc.
>  Remove backoff if tests prove fruitful.

All :+1:

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+		warmQueue:            make(chan flux.ImageID, 100), // Don't close this. It will be GC'ed when this instance is destroyed.
+	}
+}
+
+// Queue loop to maintain the queue and periodically add a random
+// repository that is running in the cluster.
+func (w *Queue) Loop(stop chan struct{}, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	if w.RunningContainers == nil || w.Logger == nil || w.RegistryPollInterval == 0 {
```
awh: Makes more sense for this to go in the constructor IMO

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+
+	pollImages := time.Tick(w.RegistryPollInterval)
+
+	for {
+		select {
+		case <-stop:
+			w.Logger.Log("stopping", "true")
+			return
+		case <-pollImages:
+			if len(w.warmQueue) == 0 { // Only add to queue if queue is empty
```
awh: Why?

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+				containers := w.RunningContainers() // Just add containers in order for now
+				w.queueLock.Lock()
```
awh: Why are we taking this lock?

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+					}
+				}
+				w.queueLock.Unlock()
+			}
+		}
+	}
+}
+
+func (w *Queue) Queue() chan flux.ImageID {
+	w.queueLock.Lock()
```
awh: Why are we taking this lock?

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+		select {
+		case <-stop:
+			w.Logger.Log("stopping", "true")
+			return
+		case <-pollImages:
+			if len(w.warmQueue) == 0 { // Only add to queue if queue is empty
+				containers := w.RunningContainers() // Just add containers in order for now
+				w.queueLock.Lock()
+				for _, c := range containers {
+					// if we can't write it immediately, drop it and move on
```
awh: Not sure this is the right thing to do

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+	warmQueue            chan flux.ImageID
+	queueLock            sync.Mutex
+}
+
+func NewQueue(runningContainersFunc func() []flux.ImageID, l log.Logger, emptyQueueTick time.Duration) Queue {
+	return Queue{
+		RunningContainers:    runningContainersFunc,
+		Logger:               l,
+		RegistryPollInterval: emptyQueueTick,
+		warmQueue:            make(chan flux.ImageID, 100), // Don't close this. It will be GC'ed when this instance is destroyed.
```
awh: Why 100?

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+		go func() {
+			for i := range toFetch {
+				// Get the image from the remote
+				img, err := client.Manifest(i)
+				if err != nil {
+					if !strings.Contains(err.Error(), context.DeadlineExceeded.Error()) && !strings.Contains(err.Error(), "net/http: request canceled") {
+						w.Logger.Log("err", errors.Wrap(err, "requesting manifests"))
+					}
+				}
+				fetched <- img // Always return an image, otherwise the for loop below will never finish.
```
awh: Then just range over the channel below, instead of assuming it always has `cap` elements in it. You could also then `continue` early in this loop on error, remove the comment, and the `if img.ID.String() == "" continue` guard.

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+		case <-stop:
+			w.Logger.Log("stopping", "true")
+			return
+		case r := <-warm:
+			w.warm(r)
+		}
+	}
+}
+
+func (w *Warmer) warm(id flux.ImageID) {
```
awh: A general observation about this function: it is bloated because the cache API is too low level - IMO we should push the JSON marshalling and key extraction responsibilities down into the cache, since every user wants those things. Then you'd just have single operations here (e.g. 'put this in the cache') rather than making the key/marshalling the json/setting the thing

cmd/fluxd/main.go
```diff
@@ -185,31 +198,61 @@ func main() {
...
+			RPS:   *registryRPS,
+			Burst: *registryBurst,
+		})
+
+		// Warmer
+		queueLogger := log.NewContext(logger).With("component", "warmer_queue")
+		warmerQueue = registry.NewQueue(
+			servicesToRepositories(k8s, queueLogger),
+			queueLogger,
+			100*time.Millisecond,
```
awh: Is this a mistake? If I understand it correctly this means the warmer is hitting k8s API ten times a second to ask it for all namespaces/deployments/replication controllers/services (and then probably throwing most of that work away because it doesn't enqueue anything into the channel unless it is empty)

registry/warming.go
```diff
@@ -0,0 +1,234 @@
...
+		RunningContainers:    runningContainersFunc,
+		Logger:               l,
+		RegistryPollInterval: emptyQueueTick,
+		warmQueue:            make(chan flux.ImageID, 100), // Don't close this. It will be GC'ed when this instance is destroyed.
+	}
+}
+
+// Queue loop to maintain the queue and periodically add a random
+// repository that is running in the cluster.
+func (w *Queue) Loop(stop chan struct{}, wg *sync.WaitGroup) {
```
awh: IMO the way this should work is that periodically (say once a minute) we enumerate all the image IDs and feed them into the warm queue. If the queue becomes full, we should block, rather than hammering k8s asking it to enumerate things we can't use anyway.

registry/cache/memcached.go
```diff
@@ -0,0 +1,257 @@
+package cache
+
+import (
```
squaremo: :white_check_mark:

registry/client.go
```diff
@@ -0,0 +1,129 @@
+package registry
+
+import (
```
squaremo: :white_check_mark:

registry/client_factory.go
```diff
@@ -0,0 +1,121 @@
+package registry
+
+import (
```
squaremo: :white_check_mark:

registry/middleware/rate_limiter.go
```diff
@@ -0,0 +1,50 @@
+package middleware
+
+import (
```
squaremo: :white_check_mark:

registry/registry.go
```diff
@@ -1,36 +1,64 @@
...
+// has been cached. A separate process is responsible for ensuring the
+// cache is up to date. The benefit of this is that we can rate limit
+// the requests to prevent rate limiting on the remote side without
+// affecting the UX. To the user, repository information will appear to
+// be returned "quickly"
+//
+// This means that the cache is now a flux requirement.
 package registry
 
 import (
```
squaremo: :white_check_mark:

registry/registry.go
```diff
@@ -117,25 +155,35 @@ func (reg *registry) tagsToRepository(remote Remote, repository Repository, tags
...
+type HerokuRegistryLibrary interface {
+	Tags(repository string) (tags []string, err error)
+	Manifest(repository, reference string) ([]schema1.History, error)
+}
+
+// ---
+
+// Convert between types. dockerregistry returns the *same* type but from a
+// vendored library. Because golang doesn't like to apply interfaces to a
+// vendored type, we have to provide an adaptor to isolate it.
```
squaremo: I suspect this is to do with having the same type returned by two different vendored packages. Let's see what Phil says, and hence or otherwise investigate in another thread.

registry/warming.go
```diff
@@ -0,0 +1,234 @@
+// Runs a daemon to continuously warm the registry cache.
+package registry
+
+import (
```
squaremo: :white_check_mark:

registry/warming.go
squaremo: It could do; that would make the interface lopsided (accept `interface{}`, return `[]byte`) but maybe that doesn't matter. I guess it could go further and have an interface of `SetTags`, `SetManifest`. Leaving for now.

registry/registry.go
philwinder: This is left over from when Paul decided to use the heroku library.

The library owners have commited the vendor directory. We have conflicts between our own vendored library and theirs. I can't quite remember why, but there was an issue returning their implementation of the interface, so it was necessary to create another interface that didn't conflict.

PR #638, opened 2017-07-05
Give container images labels

squaremo:
Appropriate labelling per http://label-schema.org/rc1/. Most of it is static, other than the revision, which is supplied when building.


squaremo:
Curses, this needs a far newer version of Docker than Circle 1.0 provides. Would need to upgrade to Circle 2.0 most likely.

squaremo:
Closing to take off the radar, as I don't yet want to go through the pain of updating our CI.

leth:
@awh solved the pain, is this worth reopening?

squaremo:
> @awh solved the pain, is this worth reopening?

Yas! Might need some rebass, mind you.

PR #640, opened 2017-07-06
Parse serialisations of []Policy as policy.Set

squaremo:
When extending policy to include tag filters, the type representing a
set of policies changed from `[]Policy` (a slice of policies to
consider as "on") to `map[Policy]string` (since policies may now have
values other than on and .. absent).

For most purposes this is internal, however it appears in public APIs in two places:

 1. when calling `UpdatePolicy` you provide a policy.Updates
 2. historical events and commit notes include a `update.Spec` which
 can contain `policy.Updates`

.. and `policy.Updates` has fields of the kind described above.

To smooth this over, adapt []Policy to map[Policy]string when
deserialising.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -1,6 +1,7 @@
 package policy
 
 import (
+	"encoding/json"
 	"strings"
 
 	"github.com/weaveworks/flux"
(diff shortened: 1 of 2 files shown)


squaremo:
Thanks Adam! :star:

PR #642, opened 2017-07-07
Distinguish automated release events

### LINKED ISSUE #641: Notifications for automated releases not firing
Notifications from our own deployment of Flux stopped appearing in Slack, some time after 2pm yesterday. NB there have been _only_ automated releases since then.

Looking at the notes pushed to the upstream repo, I see they have no cause or spec recorded:

```
{"jobID":"70094831-cf19-97ea-1080-2be8d8d0befe","spec":{"type":"image","cause":{"Message":"","User":""},"spec":{}}, ...}
```

In the service there's at least one place where a release event is checked to see if it is an "execute" (i.e., not a dry run). This is redundant, since there should not have been a commit or commit note if it wasn't, and therefore no event posted. But it doesn't hurt (what if something else was posting events?).

squaremo:
Previously when we ran automation the changes to be made were
collected into a set of releases of individual images. This was
largely to fit with the existing release mechanism.

The recent tag filtering implementation changed how releases could be
specified, allowing you to supply an implementation of
`update.Changes` which would then be used to calculate the changes to
manifests.

However, the commit notes and events still accommodate only the
original kind of release spec. The upshot: fields in commit notes were
empty, and notifications missing.

One quick fix would be to finesse automated release specs to look like
the original kind. However, since the former can mention multiple
images, it is easier said than done. Instead I decided to bite the
bullet and create a new kind of event (and hook it into notification).

Also removes the custom Slack template, which we didn't use.

(Heads-up: this creates a new kind of Event with the type "autorelease", which will need a view in the UI.)

Fixes #641.

### DIFF
--- a/update/automated.go
+++ b/update/automated.go
@@ -10,17 +10,17 @@ import (
 )
 
 type Automated struct {
-	changes []change
+	Changes []Change
 }
 
-type change struct {
-	service   flux.ServiceID
-	container cluster.Container
-	imageID   flux.ImageID
+type Change struct {
+	ServiceID flux.ServiceID
+	Container cluster.Container
+	ImageID   flux.ImageID
 }
 
 func (a *Automated) Add(service flux.ServiceID, container cluster.Container, image flux.ImageID) {
-	a.changes = append(a.changes, change{service, container, image})
+	a.Changes = append(a.Changes, Change{service, container, image})
 }
 
 func (a *Automated) CalculateRelease(rc ReleaseContext, logger log.Logger) ([]*ServiceUpdate, Result, error) {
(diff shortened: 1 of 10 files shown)


update/automated.go
```diff
@@ -10,17 +10,17 @@ import (
...
 }
 
-type change struct {
-	service   flux.ServiceID
-	container cluster.Container
-	imageID   flux.ImageID
+type Change struct {
+	ServiceID flux.ServiceID
+	Container cluster.Container
+	ImageID   flux.ImageID
```
sambooo: Do these types need to be exported for (un)marshaling purposes? I see no new usage outside of this package.

squaremo: Yes, exactly. Otherwise you get an empty struct `{}`

PR #645, opened 2017-07-10
Better notifications

squaremo:
release_02


aaron7:
Awesome! It would be great to also sort the service names in the list.

squaremo:
NB I like Aaron's idea of putting images (if >1) in a sorted list, so I may do another PR for that soon.

PR #647, opened 2017-07-10
Tag filter CLI

sambooo:
Currently working towards implementing this command:

```
$ fluxctl policy --service=foo \
    [--(de)automate] \
    [--(un)lock] \
    [--container=bar --tag=master-*]
```

Progress:

- [x] `--(de)automate`
- [x] `--(un)lock`
- [x] `--container --tag` (with arbitrarily many pairs of these)
- [x] commit messages must make sense
- [x] history

### DIFF
--- a/cmd/fluxctl/policy.go
+++ b/cmd/fluxctl/policy.go
@@ -10,9 +10,10 @@ import (
 type servicePolicyOpts struct {
 	*rootOpts
 	outputOpts
-	service   string
-	container string
-	tag       string
+
+	service    string
+	containers []string
+	tags       []string
 
 	automate, deautomate bool
 	lock, unlock         bool
@@ -40,8 +41,8 @@ func (opts *servicePolicyOpts) Command() *cobra.Command {
 	AddCauseFlags(cmd, &opts.cause)
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.service, "service", "s", "", "Service to modify")
-	flags.StringVar(&opts.container, "container", "", "Container to set tag filter")
-	flags.StringVar(&opts.tag, "tag", "", "Tag filter pattern")
+	flags.StringSliceVar(&opts.containers, "container", nil, "Containers to set tag filter")
+	flags.StringSliceVar(&opts.tags, "tag", nil, "Tag filter patterns")
 	flags.BoolVar(&opts.automate, "automate", false, "Automate service")
 	flags.BoolVar(&opts.deautomate, "deautomate", false, "Deautomate for service")
 	flags.BoolVar(&opts.lock, "lock", false, "Lock service")
@@ -63,11 +64,8 @@ func (opts *servicePolicyOpts) RunE(cmd *cobra.Command, args []string) error {
 	if opts.lock && opts.unlock {
 		return newUsageError("lock and unlock both specified")
 	}
-	if opts.container != "" && opts.tag == "" {
-		return newUsageError("container specified without a tag pattern")
-	}
-	if opts.tag != "" && opts.container == "" {
-		return newUsageError("tag pattern specified without a container")
+	if len(opts.containers) != len(opts.tags) {
+		return newUsageError("mismatched container name and tag filter count")
 	}
 
 	serviceID, err := flux.ParseServiceID(opts.service)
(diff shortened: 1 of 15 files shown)


sambooo:
Does it make sense to instead make subcommands under policy? ie:

```
$ fluxctl policy automate --service=foo
$ fluxctl policy filter --service=foo --container=bar --tag=baz
$ fluxctl policy lock --service=foo
```

squaremo:
> Does it make sense to instead make subcommands under policy

Not if you want to do more than one at a time, e.g., automate with a particular tag. You'd have to have a (more of) a model in your head about how to get where you want to go -- e.g., don't automate before supplying the filter.

sambooo:
Good point. A user that knows they want to both automate and specify a tag shouldn't need to figure out which order to do those actions in.

squaremo:
> --container --tag

Will these come in pairs? E.g., if you want to set the filter on more than one container at a time.

If you're automating a service that hasn't been automated before, you might want to provide filters for each container, so that you don't accidentally let your sidecar image roam free.

sambooo:
Good point about sidecar! Without allowing multiple pairs to be specified we'd end up with pretty much the same issue as above.

I shall look into how to make this work

cmd/fluxctl/policy.go
```diff
@@ -0,0 +1,121 @@
...
+	if opts.deautomate {
+		remove = remove.Add(policy.Automated)
+	}
+	if opts.unlock {
+		remove = remove.Add(policy.Locked)
+	}
+
+	for i, container := range opts.containers {
+		tag := opts.tags[i]
+		if tag != "" && tag != "*" {
```
philwinder: Is `--tag=""` some special case? I don't think that would be allowed by the cli argument parser.

cmd/fluxctl/policy.go
```diff
@@ -0,0 +1,121 @@
...
+			add = add.Set(policy.TagPrefix(container), "glob:"+tag)
+		} else if tag == "*" {
+			remove = remove.Add(policy.TagPrefix(container))
+		}
+	}
+
+	update := policy.Update{}
+	if len(add) > 0 {
+		update.Add = add
```
philwinder: `update` will have an empty `Add` and `Remove`. So there's no need for the check. It's safe to set `update.Add` to another empty add (same for remove).

daemon/daemon.go
```diff
@@ -455,20 +455,24 @@ func policyCommitMessage(us policy.Updates, cause update.Cause) string {
 func policyEventTypes(u policy.Update) []string {
 	types := map[string]struct{}{}
```
philwinder: I don't see the point of using an empty struct here. Why not just an array?

Just looked at the rest of the code. It eventually returns an array anyway, so just swap this out to be an array.

sambooo: This is a common Go pattern used to simulate a Set. Without this we'd end up with duplicates.

sambooo:
@philwinder Thanks for the review!


The globbing just uses the go-glob library which allows arbitrarily many asterisks in its patterns.

Do you think that we need tests to check that the library we use does what it should? I'm not sure whether that's necessary but if you think so then I'll write some.

sambooo:
After some discussion with Michael, the command is being changed to this:

```
$ fluxctl policy --service=foo \
    [--(de)automate] \
    [--(un)lock] \
    [--tag-all='*'] `
    [--tag='container=pattern']+
```

Where the `--tag` flags are applied after `--tag-all`, so that a default can be specified and overridden in a single command.

cmd/fluxctl/policy.go
```diff
@@ -0,0 +1,122 @@
...
+		remove = remove.Add(policy.Automated)
+	}
+	if opts.unlock {
+		remove = remove.Add(policy.Locked)
+	}
+
+	for _, tagPair := range opts.tags {
+		parts := strings.Split(tagPair, "=")
+		if len(parts) != 2 {
+			return policy.Update{}, fmt.Errorf("invalid container/tag pair: %q", tagPair)
```
squaremo: Perhaps spell out what the expected format is, here.

history/event.go
```diff
@@ -23,6 +23,7 @@ const (
 	EventDeautomate  = "deautomate"
 	EventLock        = "lock"
 	EventUnlock      = "unlock"
+	EventUpdateTag   = "update_tag"
```
squaremo: I foresee there being more kinds of policy, and we could get bogged down in individual events for each. How about we have a PolicyChange event which subsumes the others, and has the previous and new policies. We can attach labels, if necessary, to make it easier to tell whether it represents a transition e.g., locked, automated.

squaremo: (Yes we'll have to make compensating changes in the UI)

PR #657, opened 2017-07-13
Notify on sync

lukemarsden:
~Don't merge!~ (EDIT(squaremo): can be merged)

Support notifying slack about "sync" events, as well as "release" events. So that you can see when your config changes have been applied.


lukemarsden:
@foot This could do with a UI, maybe some checkboxes? Something like this?

Configure Slack notifications:
- [ ] New images applied to config
- [ ] Config applied to cluster

PR #661, opened 2017-07-14
Merge the image and git loops (again)

squaremo:
To accomplish #591 it is a lot easier if the image polling and git
polling run in the same loop.

When they are in different loops, is it tricky to synchronise them
such that a service being automated will result in the new commit
being pulled _then_ the images examined. If they happen in the other
order, the commit showing the newly automated service won't be present
and that service will be passed over.

This adds a bit more fuss in the one loop but is simpler to
understand, I reckon.

### DIFF
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -32,28 +32,6 @@ func (loop *LoopVars) ensureInit() {
 	})
 }
 
-// Loop for potentially long-running stuff. This includes running
-// jobs, and looking for new commits.
-
-func (d *Daemon) ImagePollLoop(stop chan struct{}, wg *sync.WaitGroup, logger log.Logger) {
-	defer wg.Done()
-
-	imagePollTicker := time.NewTicker(d.RegistryPollInterval)
-	d.askForImagePoll()
-	for {
-		select {
-		case <-stop:
-			logger.Log("stopping", "true")
-			return
-		case <-d.pollImagesSoon:
-			d.pollForNewImages(logger)
-		case <-imagePollTicker.C:
-			// Time to poll for new images
-			d.askForImagePoll()
-		}
-	}
-}
-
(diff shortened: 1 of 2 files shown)


daemon/loop.go
```diff
@@ -72,13 +50,22 @@ func (d *Daemon) GitPollLoop(stop chan struct{}, wg *sync.WaitGroup, logger log.
...
+	d.askForImagePoll()
 	for {
 		select {
 		case <-stop:
 			logger.Log("stopping", "true")
 			return
+		case <-d.pollImagesSoon:
+			d.pollForNewImages(logger)
+			imagePollTimer.Stop()
+			imagePollTimer = time.NewTimer(d.RegistryPollInterval)
```
philwinder: Why use a timer here and not a ticker? (`time.NewTicker`)

squaremo:
> So effectively I think you have exactly what you had before, just in on loop.

When a job completes, the repo is pulled before returning control to the loop. Since that's now in the same loop as image polling, if the job also asked for an image poll to happen, the latter cannot happen until after the commit containing the automation change has been pulled into the local repo. If they are in different loops, they race (and the image poll usually wins).

daemon/loop.go
squaremo: Because it can be renewed either by the timer firing or by something asking explicitly for an image poll, and in both those cases we want to reset the timer.

philwinder: Ah ok. So you don't want it to run at a scheduled rate. Cool.

philwinder:
Understood LGTM.

daemon/loop.go
squaremo: Yeah, and in fact in the previous incarnation it should have been a timer too, since it had the same treatment -- I must have had Ticker/Timer blindness.

PR #662, opened 2017-07-14
[WIP] move service things into service/ and remove dead wood

squaremo:
Move things into `service/`

 - [x] Message bus
 - [ ] history and notifications (but not events)
 - [ ] server/server and http/server
 - [ ] remove InstanceID from api interfaces
 - [ ] move service-only api interface to service/

Remove dead wood:

 - [x] instance/standalone (kind of a mock instance DB)
 - [x] remote/standalone (message bus)


squaremo:
This seems to have made the flakes flakier. Maybe it's a good thing ..

squaremo:
Out of date, and some things have been accomplished already. Will keep the branch around for reference, since it does some of the painful bits.

PR #667, opened 2017-07-21
More information in Sync notifications

squaremo:
In particular, gives a recapitulation of `git log --oneline`.

image

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -323,7 +323,7 @@ func (d *Daemon) JobStatus(jobID job.ID) (job.Status, error) {
 	if err := d.Checkout.Pull(); err != nil {
 		return job.Status{}, errors.Wrap(err, "updating repo for status")
 	}
-	refs, err := d.Checkout.RevisionsBefore("HEAD")
+	refs, _, err := d.Checkout.RevisionsBefore("HEAD")
 	if err != nil {
 		return job.Status{}, errors.Wrap(err, "checking revisions for status")
 	}
@@ -350,7 +350,10 @@ func (d *Daemon) JobStatus(jobID job.ID) (job.Status, error) {
 // you'll get all the commits yet to be applied. If you send a hash
 // and it's applied _past_ it, you'll get an empty list.
 func (d *Daemon) SyncStatus(commitRef string) ([]string, error) {
-	return d.Checkout.RevisionsBetween(d.Checkout.SyncTag, commitRef)
+	revs, _, err := d.Checkout.RevisionsBetween(d.Checkout.SyncTag, commitRef)
+	// NB we could use the messages too if we decide to change the
+	// signature of the API to include it.
+	return revs, err
 }
 
 func (d *Daemon) GitRepoConfig(regenerate bool) (flux.GitConfig, error) {
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -323,17 +323,17 @@ func (d *Daemon) JobStatus(jobID job.ID) (job.Status, error) {
 	if err := d.Checkout.Pull(); err != nil {
 		return job.Status{}, errors.Wrap(err, "updating repo for status")
 	}
-	refs, _, err := d.Checkout.RevisionsBefore("HEAD")
+	commits, err := d.Checkout.CommitsBefore("HEAD")
 	if err != nil {
 		return job.Status{}, errors.Wrap(err, "checking revisions for status")
 	}
-	for _, ref := range refs {
-		note, _ := d.Checkout.GetNote(ref)
+	for _, commit := range commits {
+		note, _ := d.Checkout.GetNote(commit.Revision)
 		if note != nil && note.JobID == jobID {
 			return job.Status{
 				StatusString: job.StatusSucceeded,
 				Result: history.CommitEventMetadata{
-					Revision: ref,
+					Revision: commit.Revision,
 					Spec:     &note.Spec,
 					Result:   note.Result,
 				},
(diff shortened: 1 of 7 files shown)


squaremo:
> It feels like you're forking information there. IMO it might be a little nicer to either pass back revisions, err :=... where revisions are []struct{ref, message string} or continue to just pass back the ref, then get the messages in a second call.

Yeah, this was a bit knotty because I didn't want to break the treatment of existing events, or use more heavy machinery like a new event type. Maybe I could finesse the parsing (as we did with `policy.Set`). Let me have a quick look at that.

squaremo:
> this was a bit knotty because I didn't want to break the treatment of existing events

Actually there are two concerns: the return type of the git operations, and the types of the events (SyncEventMetadata) and API (for e.g., SyncStatus). The return types can be tidier without upsetting the other things.

squaremo:
@foot From your point of view, is it better to have (paraphrasing slightly)

```
type SyncEvent struct {
  Revisions []string
  Messages []string // this is a new field
}
```
or
```
type SyncEvent struct {
  Commits []struct{ // this is a replacement
    Revision string
    Message string
  }
}
```

The first one is an additive change -- existing UI etc., can be ignorant of the `Messages` field and still work, but it's a bit awkward to use, because you have to run through the arrays in parallel. The second one is not additive -- it replaces the old field -- but it is less weird. We can paper over old, serialised events to make them look like this instead.

foot:
Yes I vote for number 2! As a zig away from all the backwards-compat support we've done in other areas :).

squaremo:
If you do a release, it gets a bit busy:

image

Sections top to bottom: Sync message text, sync message attachment with commit log; Release message text, release message attachment with author (fluxctl caller) and user-supplied message (also used as the commit message), release message attachment with release result.

@foot  Any ideas for tidying these up? Bearing in mind that they can appear in isolation as well (e.g., a sync can be not related to any release).

foot:
+1 on not over-spamming!

I guess a Release is actually the announcement of a sync of a certain type of commit already... we get this duplicated info as our notification levels have a bit of overlap. Our current notifications settings kind of work like this:

Notify me on:
 - [ ] Release Syncs
 - [ ] Auto Release Syncs
 - [ ] All Syncs

Perhaps we could move to a model like this instead:

Notify me on:
 - [ ] Release Syncs
 - [ ] Auto Release Syncs
 - [ ] *Other* Syncs

And include the commit id in the Release slack message and leave it at that. Later on we could split that out a bit more too:

Notify me on:
 - [ ] Release
 - [ ] Auto Release
 - [ ] Lock, Automate and Tag changes.
 - [ ] Other commits not made by flux.

squaremo:
> I guess a Release is actually the announcement of a sync of a certain type of commit already... we get this duplicated info as our notification levels have a bit of overlap.

Yes, good insight. People will in general care about 1. releases happening; and, 2. PR merges (or commits to master!) being applied.

Typically, when we sync, there are `N` commits involved. Even if we're moving from one PR merge to another, there's the merge commit as well as the commit with the actual changes; and often there is also stuff that happened in the meantime, like automated releases.

This means each sync is one event itself, but often refers to a range of commits, and in principle includes an arbitrary number of releases or autoreleases (or policy changes). At present we run through these and send them as separate events; they _could_ be bundled up in one event representing the sync and all it entails.

(A corner case is the first sync, which implicitly includes _all_ commits, releases, and so on. Possibly we can mark that as a special case and _not_ send all the accompanying notifications and other data).

squaremo:
The above (two comments) would entail compensating design and UI changes (e.g., to change how history is displayed -- I suppose we could keep that the same and just process notification separately).

A simpler solution would be to mark sync events as having included releases, autoreleases, or policy changes. Then, they can just be discarded depending on how we interpret the user's preferences.

squaremo:
> A simpler solution would be to mark sync events as having only included releases, autoreleases, or policy changes.

We don't report policy changes in the sync loop; they get reported as soon as committed. We don't notifications for them, but could similarly drop sync notifications that were *only* these, if we did.

foot:
Yes! Sorry to derail this PR. For now either leaving it is as or merging the sync and release slack messages into 1 when notification settings allow sounds good.

--- 


> People will in general care about 1. releases happening; and, 2. PR merges 

Yep!

I just wrote an essay but realised all I was pointing out was that the fairly contextless sync notifications we have now are not ideal. It would be cool instead to have notifications for `updatepolicy` and `external-git-change` as we currently have for `releases`.

An `external-git-change` event might include PRs merges... - Will flux be able to read a potentially squashed PR and extract "events" like release/policyupdate out of it?

squaremo:
> It would be cool instead to have notifications for updatepolicy and external-git-change as we currently have for releases.

We send things saying "Locked", "Unlocked", and "PolicyChange" I think; but we do this when the commit has been made (since that's when they take effect), rather than when syncing. We can send notifications for them, and we can elide sync notifications that include (only) updates, if necessary.

> An external-git-change event might include PRs merges... - Will flux be able to read a potentially squashed PR and extract "events" like release/policyupdate out of it?

Tricky one. In general, we won't be able to track git notes through arbitrary history rewriting, and that may include squash-merges. Let's burn this bridge when we come to it!

But we can surely look sync'd commits up to see whether they are a PR merge or whatever via the github API.

squaremo:
The state of this is:

 - sync events include a formatting display of the commits included in the sync, as the first screenshot shows
 - sync events that _only_ include releases won't result in a sync notification, so we won't see the doubling-up shown in the second screenshot.

foot:
> I think; but we do this when the commit has been made (since that's when they take effect), rather than when syncing. We can send notifications for them, and we can elide sync notifications that include (only) updates, if necessary.

Yes! That would be cool.

> sync events include a formatting display of the commits included in the sync, as the first screenshot shows
> sync events that only include releases won't result in a sync notification, so we won't see the doubling-up shown in the second screenshot.

👍 . So we only get sync notifications when an external change to the repo has been made?

squaremo:
> So we only get sync notifications when an external change to the repo has been made?

Not quite; either commits we didn't make ourselves ("external changes"), or commits that correspond to things we don't send notifications for (i.e., policy changes). Can change that to what you said pretty easily, I think (i.e., only notify on external changes).

squaremo:
@foot Just to call this out: once this is merged, Sync events (in the history output) look like this

```json
{
  "Stamp": "2017-07-31T13:06:50.983157Z",
  "Type": "v0",
  "Data": "Sync: 4b7bbdf, hello/helloworld",
  "Event": {
    "id": 48,
    "serviceIDs": [
      "hello/helloworld"
    ],
    "type": "sync",
    "startedAt": "2017-07-31T13:06:50.983157Z",
    "endedAt": "2017-07-31T13:06:50.983157Z",
    "logLevel": "info",
    "metadata": {
      "commits": [
        {
          "revision": "4b7bbdfe2db6bf00b29eb1efdbfb7326a243330b",
          "message": "Release quay.io/squaremo/hello_world:master-a000002 to hello/helloworld"
        }
      ],
      "includes": {
        "release": true
      }
    }
  }
}
```

The two changes:

 1. instead of `revisions` you get `commits`, as in comment https://github.com/weaveworks/flux/pull/667#issuecomment-318016545
 2. there's also an `includes` field, a `map[string]bool`, which say what kinds of commit were included in the sync. The values come from `history/events.go` and can be `"release"`, `"autorelease"`, `"update_policy"`, and `"other"` (aka `history.NoneOfTheAbove`).

philwinder:
LGTM

PR #670, opened 2017-07-24
Fix git diff command to allow absolute git paths

### LINKED ISSUE #655: Better `git-path` validation: "No such file or directory"
Some parts of flux work if you pass `--git-path=/my/directory`, but others don't.

For example it correctly applies manifests, but you get the error:
```
ts=2017-07-13T09:28:57Z caller=loop.go:123 component=sync-loop err="fatal: Invalid path '/deploy': No such file or directory"
```

If you strip the leading `/` it works. `--git-path=my/directory`

philwinder:
Fixes #655

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -69,7 +69,7 @@ func main() {
 		// Git repo & key etc.
 		gitURL          = fs.String("git-url", "", "URL of git repo with Kubernetes manifests; e.g., git@github.com:weaveworks/flux-example")
 		gitBranch       = fs.String("git-branch", "master", "branch of git repo to use for Kubernetes manifests")
-		gitPath         = fs.String("git-path", "", "path within git repo to locate Kubernetes manifests")
+		gitPath         = fs.String("git-path", "", "path within git repo to locate Kubernetes manifests (relative path)")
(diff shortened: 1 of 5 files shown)


squaremo:
In terms of safety, this absolutely _should_ break if the subpath starts with a slash -- it shouldn't be supplied such a subpath. Silently correcting them will mean problems in the caller will go unnoticed for longer.

philwinder:
Ah OK. All other git commands work correctly with a leading slash. So I
assumed it was OK.

I will change this to and error and add some validation to the flux main.go

squaremo:
> All other git commands work correctly with a leading slash. So I
> assumed it was OK.

Fair enough :) We (I) got those wrong too!

squaremo:
> This does go against the idea: "just make it work instead of giving the user some warning" though.

Well kind of. That idea is fine if there is a safe and mechanical way to get from what the user told you to something that works. If the mistake is unambiguous, there's no point pushing the remediation back on the user.

But if the mistake is ambiguous -- maybe the person is trying to access something they shouldn't by supplying an absolute path -- then it's a case for validation (and behind that, enforcement).

philwinder:
Yes. By definition, it's hard to decide what is ambiguous. :-)

I've updated the code to error out now, instead of silently work around.

Thanks.

PR #674, opened 2017-07-24
Link to releases rather than releases/latest URL

squaremo:
The `github.com/weaveworks/flux/releases/latest` URL points at the
most recent GA (i.e., non-pre) release. Since the one we want people
using the instructions to use is a pre-release, point at `.../release`
instead.

Adjust the quick setup instructions, which no longer apply since you
now need to edit the config.

### DIFF
--- a/site/standalone/installing.md
+++ b/site/standalone/installing.md
@@ -15,17 +15,17 @@ how.
 
 # Quick Install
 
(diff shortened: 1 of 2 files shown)


squaremo:
Thanks Phil!

Issue #652, opened 2017-07-12
Leverage pod template imagePullSecrets


awh:
Any pod templates that refer to images hosted in private registries _must_ have `imagePullSecrets` specified, or Kubernetes wouldn't be able to pull them. Flux should be able to read these secrets itself, and therefore does not have to be independently configured with a registry secret.

awh:
This seems to be relatively straightforward when built on top of #633 - we would extend the `servicesToRepositories` function to extract credentials from k8s alongside the image IDs, and have the warmer pass them down into the registry client. The `Cluster` interface will need to be changed to support getting the secrets. I would estimate this is only a few days effort @rade.

awh:
In GKE environments it's possible for fluxd to gain access to the underlying instance's service account that authenticates it to GCR using a combination of https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#applications and https://cloud.google.com/container-registry/docs/advanced-authentication#using_an_access_token

rade:
To clarify, we need a special approach for GKE+GCR because GKE has some magic sauce that allows it to pull images from GCR; it's not using `imagePullSecrets`. Correct?

awh:
Yes - Kubernetes has native support for GCR when running on GCE as documented here https://kubernetes.io/docs/concepts/containers/images/#using-google-container-registry - no `imagePullSecrets` required.

philwinder:
GCR integration was implemented in #663

PR #678, opened 2017-07-26
Parse docker credentials from imagePullRequests (not .dockerconfig)

squaremo:
This PR moves the method that returns a list of images to warm into
the kubernetes package. This is because the method is specific to k8s
anyway and the new code to parse the secrets needed access to the k8s
cluster.

ImageCreds is a new struct to colocate the ImageID and the credentials
required to pull them.

Credentials are now added to the client via the client factory for each
new client. This is because each image might have different credentials.

Minor fix to ImageID parsing because it wasn't able to parse id's in the format: `localhost:5000/my/image:tag`

GCR support was implemented in #663

# Testing

Push an image to a private repository. For example:

```
docker pull quay.io/weaveworks/helloworld:master-9a16ff945b9e
docker tag quay.io/weaveworks/helloworld:master-9a16ff945b9e quay.io/weaveworks/helloyou:master-9a16ff945b9e
docker login quay.io
docker push quay.io/weaveworks/helloyou:master-9a16ff945b9e
```

Create a secret on your cluster that provides your login credentials:

```
kubectl create secret docker-registry myregistrykey --docker-server=quay.io --docker-username=myuser --docker-password=mypass --docker-email=foo@bar.com
```

And finally edit your k8s deployment manifest to include the private image, and the secret in an `imagePullSecret`. For example:

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: helloworld
spec:
  minReadySeconds: 5
  replicas: 1
  template:
    metadata:
      labels:
        name: helloworld
    spec:
      containers:
      - name: helloworld
        image: quay.io/weaveworks/helloyou:master-9a16ff945b9e
        args:
        - -msg=Ahoy
        ports:
        - containerPort: 80
      - name: sidecar
        image: quay.io/weaveworks/sidecar:master-a000002
        args:
        - -addr=:8080
        ports:
        - containerPort: 8080
      imagePullSecrets:
        - name: myregistrykey
```

### DIFF
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -288,6 +288,16 @@ type podController struct {
 	Deployment            *apiext.Deployment
 }
 
+func (p podController) secrets() (rawSecrets []v1.LocalObjectReference) {
+	// If deployment doesn't contain any secrets, just return empty secret
+	if p.Deployment != nil && p.Deployment.Spec.Template.Spec.ImagePullSecrets != nil {
+		rawSecrets = p.Deployment.Spec.Template.Spec.ImagePullSecrets
+	} else if p.ReplicationController != nil && p.ReplicationController.Spec.Template.Spec.ImagePullSecrets != nil {
+		rawSecrets = p.ReplicationController.Spec.Template.Spec.ImagePullSecrets
+	}
+	return
+}
+
 func (p podController) templateContainers() (res []cluster.Container) {
 	var apiContainers []v1.Container
 	if p.Deployment != nil {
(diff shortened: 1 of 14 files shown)


squaremo:
Yep, beaut :gem: 

One thing: I reckon we can get away with _not_ putting another method in `cluster.Cluster`, since we "know" we're using Kubernetes (or whatever else) when we refer to the `servicesToRepositories`.

In other words, that could just become a method of `kubernetes.Cluster`, to be assigned to a `var refreshImages WarmerRefreshFunc` and thereafter supplied to the warmer loop.

cluster/kubernetes/kubernetes.go
```diff
@@ -483,6 +484,120 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
...
+				continue
+			}
+			serviceControllers = append(serviceControllers, servicePod{s: service, pc: pc})
+		}
+	}
+	return
+}
+
+// ImagesToFetch is a k8s specific method to get a list of images to update along with their credentials
+func (c *Cluster) ImagesToFetch() func() []registry.ImageCreds {
```
squaremo: You can safely make this return `[]registry.ImageCreds` and supply it as `k8s.ImagesToFetch` elsewhere, rather than returning the `func()`.

cluster/kubernetes/kubernetes.go
```diff
@@ -483,6 +484,120 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
...
+					c.logger.Log("err", err.Error())
+					continue
+				}
+
+				// Merge into the credentials for this PodSpec
+				creds.Merge(crd)
+			}
+
+			// Now create the service and attach the credentials
+			svc := c.makeService(service.Namespace, &service, []podController{controller})
```
squaremo: This seems to repeat work that's already been done in `allServices`.

cluster/kubernetes/kubernetes.go
```diff
@@ -483,6 +484,120 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
...
+			return
+		}
+
+		// Foreach service-controller combo
+		for _, servicePod := range serviceControllers {
+			service := servicePod.s
+			controller := servicePod.pc
+			var rawSecrets []v1.LocalObjectReference
+			// If deployment doesn't contain any secrets, just return empty
+			if controller.Deployment != nil && controller.Deployment.Spec.Template.Spec.ImagePullSecrets != nil {
```
squaremo: Having carefully used `podController` so that we're not committing to the particular resource being a Deployment, we now assuming that it's a Deployment ..
Could you add a method for getting the imagePullSecrets from a podController, instead?

cluster/kubernetes/kubernetes.go
```diff
@@ -483,6 +484,120 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
...
+
+			// Now create the service and attach the credentials
+			svc := c.makeService(service.Namespace, &service, []podController{controller})
+			for _, ctn := range svc.Containers.Containers {
+				r, err := flux.ParseImageID(ctn.Image)
+				if err != nil {
+					c.logger.Log("err", err.Error())
+					continue
+				}
+				imageCreds = append(imageCreds, registry.ImageCreds{ID: r, Creds: creds})
```
squaremo: If an image is used more than once, it'll appear here more than once. That is not incorrect, but it may mean doing more work than is necessary. We don't really care which credentials we use to fetch images, so long as they work; so I think you could deduplicate images here. (I think the entries will include tags, so you'd have to drop those and make sure the warmer is still happy).

cmd/fluxd/main.go
```diff
@@ -134,6 +133,8 @@ func main() {
 			logger.Log("err", err)
 			os.Exit(1)
 		}
+		clientset.Core()
```
squaremo: What's this do? Some side-effect, presumably. Please document in a comment, or remove.

cmd/fluxd/main.go
```diff
@@ -118,6 +116,7 @@ func main() {
 	var clusterVersion string
 	var sshKeyRing ssh.KeyRing
 	var k8s cluster.Cluster
+	var k8s_inst *kubernetes.Cluster
```
squaremo: Might be a bit more inclusive of hypothetical alternate cluster implementations to have a variable of type `() -> []ImageCreds` to be assigned. A nit.

cluster/kubernetes/kubernetes.go
philwinder: 👍

cluster/kubernetes/kubernetes.go
philwinder: 👍

cluster/kubernetes/kubernetes.go
philwinder: Similar, but not exactly the same. I don't want to refactor AllServices (a core part of the API) until we have tests in place to prove I don't break it.

squaremo: OK. But can't you just look at the containers in the single podController that you have already in front of you?

cluster/kubernetes/kubernetes.go
philwinder: At the moment I don't think this is a performance issue and I think it simplifies the code.

First, the only performance hit is memory usage. You're duplicating items in the map. This is very minor.

When the creds are used, the warmer gets the Image-Cred pair and uses them to create a new client. If we weren't using an Image-Cred pair, I.e. some Image->mapToCred type structure, it would still have to create a new client.

So there is no performance or client benefit to not-duplicating.

Finally, I think de-duplicating, and the additional code required in the warmer would significantly increase the complexity of what is very simple. Get a list of image-credential pairs.

cmd/fluxd/main.go
philwinder: 👍

cmd/fluxd/main.go
philwinder: Looks like some dummy code I left in when exploring the k8s api! Oops!

_aside_: How on earth does this not error!

:+1:

cluster/kubernetes/kubernetes.go
squaremo: I think I have not explained well what I was suggesting. If you just use a `map[ImageID]Creds` here, you won't post duplicate entries, and the warmer won't do unnecessary work*, and you won't have extra code. If it's more complicated than that, I agree that it wouldn't be worth it.

*Logically, the warmer is constantly looking for new images to fetch, and each will eventually be examined, so it doesn't matter whether there are duplicate entries. In practice, it'll mean that some images are polled more often than others, if they appear in the list more often. Not so much a problem for correctness as for fairness.

cmd/fluxd/main.go
squaremo: Ha, OK. I wondered what it did!

> aside: How on earth does this not error!

Because the return value(s), error or otherwise, are discarded.

cluster/kubernetes/kubernetes.go
philwinder: Ah I see what you mean. Yes, I can. Thanks.

:+1:

cluster/kubernetes/kubernetes.go
philwinder: Ah, right. Ok. Let me try returning a map instead.

cluster/kubernetes/kubernetes.go
```diff
@@ -483,6 +494,114 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
 	return publicKey, nil
 }
 
+type servicePod struct {
+	s  v1.Service
+	pc podController
+}
+
+// Internal function to get a collection of service-controller pairs for a given namespace
+func (c *Cluster) allServices(ns string) (serviceControllers []servicePod, _ error) {
```
squaremo: I reckon we could change `AllServices` to use this (and get rid of `makeService`, which would just repeat the matchControllers step). But can be done in a subsequent refactor.

cluster/kubernetes/kubernetes.go
philwinder: :+1:

cluster/kubernetes/kubernetes.go
squaremo: Oh, SomeServices also uses `makeService`, so not as simple as that. Still ...

philwinder:
Fixes #652

PR #683, opened 2017-08-01
Support dockerconfigjson imagePullSecrets

squaremo:
You can create imagePullSecrets in either `.dockercfg` or `~/.docker/config.json` format. We use the latter ourselves, so it's important to support both.

(Also: deprecate the --docker-config argument, rather than removing it, since removing it will make existing deployments crashloop.)

### DIFF
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -288,14 +288,14 @@ type podController struct {
 	Deployment            *apiext.Deployment
 }
 
-func (p podController) secrets() (rawSecrets []v1.LocalObjectReference) {
+func (p podController) secrets() []v1.LocalObjectReference {
 	// If deployment doesn't contain any secrets, just return empty secret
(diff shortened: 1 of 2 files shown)


cluster/kubernetes/kubernetes.go
```diff
@@ -569,10 +569,21 @@ func (c *Cluster) ImagesToFetch() (imageCreds registry.ImageCreds) {
 				c.logger.Log("err", errors.Wrapf(err, "getting secret %q from namespace %q", secName.Name, service.Namespace))
 				continue
 			}
-			if sec.Type != v1.SecretTypeDockercfg {
+
+			var decoded []byte
+			var ok bool
+			// These differ in format; but, ParseCredentials will
+			// handle either.
+			switch api.SecretType(sec.Type) {
```
squaremo: NB this cast: the type of the field is v1.SecretType, but the value(s) we want to test against are type `api.SecretType` (both aliases of string). There is some k8s API versioning tomfoolery I'm missing here, since you can definitely have both kinds of secret.

PR #684, opened 2017-08-01
Add service policies to ListServices result

sambooo:
This will allow @foot to retrieve tag filter policies on the frontend

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -77,6 +77,10 @@ func (d *Daemon) ListServices(namespace string) ([]flux.ServiceStatus, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "checking service policies")
 	}
+	allServices, err := d.Manifests.ServicesWithPolicy(d.Checkout.ManifestDir(), policy.None)
+	if err != nil {
+		return nil, errors.Wrap(err, "checking service policies")
+	}
 
 	for _, service := range services {
 		res = append(res, flux.ServiceStatus{
@@ -86,6 +90,7 @@ func (d *Daemon) ListServices(namespace string) ([]flux.ServiceStatus, error) {
 			Automated:  automatedServices.Contains(service.ID),
 			Locked:     lockedServices.Contains(service.ID),
 			Ignore:     ignoredServices.Contains(service.ID),
+			Policies:   allServices[service.ID].ToStringMap(),
 		})
 	}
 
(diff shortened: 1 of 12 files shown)


daemon/daemon.go
```diff
@@ -77,6 +77,10 @@ func (d *Daemon) ListServices(namespace string) ([]flux.ServiceStatus, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "checking service policies")
 	}
+	allServices, err := d.Manifests.ServicesWithPolicy(d.Checkout.ManifestDir(), policy.None)
```
squaremo: This seems like a round-about way of getting the result you want. To some extent it's an accident of history (automated and locked started as properties from somewhere else, blah etc.).

But now you always want to get all the services, with their policy, from the manifests. Why not do just it in one call? Then you don't have to have the special case of `policy.None`.

policy/policy.go
```diff
@@ -103,6 +103,14 @@ func (s Set) Get(p Policy) (string, bool) {
 	return v, ok
 }
 
+func (s Set) ToStringMap() map[string]string {
```
squaremo: Does `Set` not already marshal to JSON without intervention? The underlying type is `map[string]string` after all.

foot:
Fixes https://github.com/weaveworks/flux/issues/679

Functionally works nicely though 👍

policy/policy.go
sambooo: This was added so that the root `flux` package doesn't have a dependency on `policy`, since this introduces a cycle

daemon/daemon.go
sambooo: Yep, I'll revisit this

policy/policy.go
squaremo: Ah right, I see. Probably those types used as return types should be in `api`, but .. way bigger change. Let's put that on the refactor backlog.

daemon/daemon.go
```diff
@@ -57,40 +57,32 @@ func (d *Daemon) Export() ([]byte, error) {
...
+	ignoredServices := services.OnlyWithPolicy(policy.Ignore)
 
-	for _, service := range services {
+	var res []flux.ServiceStatus
+	for _, service := range clusterServices {
 		res = append(res, flux.ServiceStatus{
 			ID:         service.ID,
 			Containers: containers2containers(service.ContainersOrNil()),
 			Status:     service.Status,
 			Automated:  automatedServices.Contains(service.ID),
```
squaremo: `services[service.ID].Contains(policy.Automated)`?

PR #685, opened 2017-08-02
Refactor kubernetes.go

philwinder:
The reason for this refactor is to allow testing k8s methods.
This is achieved by introducing two levels of interface. A low-level
KubeAPI, which encapsulates all the required kubernetes API calls and
a mid-level KubeFluxAdapter interface which signifies the change from
the kubernetes domain (v1.*) to the flux domain (flux.Service).

- Tests added by mocking out these interfaces
- Code moved around to make the interfaces more prominent
- Main altered to adhere to new interfaces
- AllServices and SomeServices (core api methods) have been altered
to use the new interfaces. Tests have been added.
- Sync, Export and Ping haven't been altered yet

A quick note on the SomeServices change. I decided to simply list all
services then filter for the ones requested to allow testing, simplify
the code and avoid adding more methods to the KubeAPI interface.
Previously it hit the raw k8s API to get each namespace, get a requested
service for that ns then find all `podController`s for the service. This
change results in a few more calls than necessary going to the k8s api,
but I deemed this acceptable to reduce the API and add tests.

### DIFF
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -24,6 +24,7 @@ import (
 	"github.com/weaveworks/flux/cluster"
 	"github.com/weaveworks/flux/registry"
 	"github.com/weaveworks/flux/ssh"
+	"k8s.io/client-go/1.5/pkg/labels"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


squaremo:
My first impression is that this is a lot of extra abstraction for not much gain. I mean, tests are good -- what are we guarding against with these tests, though, and what extra hiccoughs might result? If the thing under test is "Given these k8s objects are in the cluster, I get these this result when I ask the a `kubernetes.Cluster` for services" then there might be stubs in the go-client library we can put to use. Maybe I'm misunderstanding the intention here.

I don't think the payoff for implementing SomeServices in terms of AllServices is worthwhile. Typically, SomeServices is used to get a single service, so this is the equivalent of doing a table scan to get an row already identified by a key. OK, a table of O(100) rows, but each is an API call. (Using a caching client would make this less of an issue, but this detail belongs in the implementation, not the interface).

philwinder:
Fair comments. The goal was to better test the k8s code. I did this through liberal use of abstractions/interfaces which I thought made sense. And the goal was to test current functionality, so it maybe doesn't cover "extra hiccups".

If you think that the change is more risky than no test coverage, then that's cool.

Second point is a simpler code vs optimised code question. I agree with your comments but still prefer simpler code. :-)

I won't have time to change all this in the future, so it's entirely up to you.

Thanks.

PR #687, opened 2017-08-04
Set number of http and memcache connections to be equal

### LINKED ISSUE #675: Memcache I/O and connection timeouts
I see strings of log entries from the memcache client:

```
630626-ts=2017-07-24T14:03:17Z caller=warming.go:161 component=warmer err="storing manifests in cache: memcache: connect timeout to 172.17.0.24:11211"
630770-ts=2017-07-24T14:03:17Z caller=memcached.go:179 component=memcached err="storing in memcache: memcache: connect timeout to 172.17.0.24:11211"
630912-ts=2017-07-24T14:03:17Z caller=warming.go:161 component=warmer err="storing manifests in cache: memcache: connect timeout to 172.17.0.24:11211"
631056:ts=2017-07-24T14:03:18Z caller=memcached.go:179 component=memcached err="storing in memcache: read tcp 172.17.0.21:47202->172.17.0.24:11211: i/o timeout"
631210:ts=2017-07-24T14:03:18Z caller=warming.go:161 component=warmer err="storing manifests in cache: read tcp 172.17.0.21:47202->172.17.0.24:11211: i/o timeout"
631366-ts=2017-07-24T14:03:19Z caller=memcached.go:179 component=memcached err="storing in memcache: memcache: connect timeout to 172.17.0.24:11211"
631508-ts=2017-07-24T14:03:19Z caller=warming.go:161 component=warmer err="storing manifests in cache: memcache: connect timeout to 172.17.0.24:11211"
631652-ts=2017-07-24T14:03:19Z caller=memcached.go:179 component=memcached err="storing in memcache: memcache: connect timeout to 172.17.0.24:11211"
```

This also affects us in dev, and leads to automated services not being updated, and the UI showing no entries for services (because it either doesn't have a full complement of images, or cannot complete a query for all the images).

philwinder:
- Use MaxIdleCons parameter in memcache library. Ensures that this many connections are available to use.
- Set the number of http connections and memcache connections to be equal in the warmer. This is required because the number http fetches cannot exceed the number of memcache connections. If they did, it would return a connection error to memcache.
- Create a separate memecache client for warming and for a user reading images from the registry.

Fixes #675

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -38,9 +38,7 @@ import (
 var version string
 
 const (
-	// The number of connections chosen for memcache and remote GETs must match.
-	// If they do not match then the remote either creates too many memcache connections
-	// or not enough to be efficient.
+	// The number of connections chosen for memcache and remote GETs should match for best performance (hence the single hardcoded value)
 	// Value chosen through performance tests on sock-shop. I was unable to get higher performance than this.
 	defaultRemoteConnections   = 125 // Chosen performance tests on sock-shop. Unable to get higher performance than this.
 	defaultMemcacheConnections = 10  // This doesn't need to be high. The user is only requesting one tag/image at a time.
@@ -79,11 +77,11 @@ func main() {
 		memcachedHostname    = fs.String("memcached-hostname", "", "Hostname for memcached service to use when caching chunks. If empty, no memcached will be used.")
 		memcachedTimeout     = fs.Duration("memcached-timeout", time.Second, "Maximum time to wait before giving up on memcached requests.")
 		memcachedService     = fs.String("memcached-service", "memcached", "SRV service used to discover memcache servers.")
-		memcachedConnections = fs.Int("memcached-connections", defaultMemcacheConnections, "maximum number of registry connections to memcache")
 		registryCacheExpiry  = fs.Duration("registry-cache-expiry", 20*time.Minute, "Duration to keep cached registry tag info. Must be < 1 month.")
 		registryPollInterval = fs.Duration("registry-poll-interval", 5*time.Minute, "period at which to poll registry for new images")
 		registryRPS          = fs.Int("registry-rps", 200, "maximum registry requests per second per host")
 		registryBurst        = fs.Int("registry-burst", defaultRemoteConnections, "maximum number of warmer connections to remote and memcache")
+
 		// k8s-secret backed ssh keyring configuration
 		k8sSecretName            = fs.String("k8s-secret-name", "flux-git-deploy", "Name of the k8s secret used to store the private SSH key")
 		k8sSecretVolumeMountPath = fs.String("k8s-secret-volume-mount-path", "/etc/fluxd/ssh", "Mount location of the k8s secret storing the private SSH key")
@@ -219,7 +217,7 @@ func main() {
 				Timeout:        *memcachedTimeout,
 				UpdateInterval: 1 * time.Minute,
 				Logger:         log.NewContext(logger).With("component", "memcached"),
-				MaxConnections: *memcachedConnections,
+				MaxIdleConns:   defaultMemcacheConnections,
 			})
 			memcacheRegistry = registryMemcache.InstrumentMemcacheClient(memcacheRegistry)
 			defer memcacheRegistry.Stop()
@@ -232,7 +230,7 @@ func main() {
 				Timeout:        *memcachedTimeout,
 				UpdateInterval: 1 * time.Minute,
 				Logger:         log.NewContext(logger).With("component", "memcached"),
-				MaxConnections: *registryBurst,
+				MaxIdleConns:   *registryBurst,
 			})
 			memcacheWarmer = registryMemcache.InstrumentMemcacheClient(memcacheWarmer)
 			defer memcacheWarmer.Stop()
@@ -242,7 +240,7 @@ func main() {
 		cache = registry.NewRegistry(
 			registry.NewCacheClientFactory(cacheLogger, memcacheRegistry, *registryCacheExpiry),
 			cacheLogger,
-			*memcachedConnections,
+			defaultMemcacheConnections,
 		)
 		cache = registry.NewInstrumentedRegistry(cache)
 
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -38,11 +38,12 @@ import (
 var version string
 
 const (
-	defaultRemoteConnections = 125 // Chosen performance tests on sock-shop. Unable to get higher performance than this.
-	// Memcache: There is only one thread to accept new client connections. If you are cycling connections very quickly, you can overwhelm the thread. Use persistent connections or UDP in this case.
-	// See https://github.com/memcached/memcached/wiki/Performance
-	// Hence, keep the number of new connections low.
-	defaultMemcacheConnections = 10
+	// The number of connections chosen for memcache and remote GETs must match.
+	// If they do not match then the remote either creates too many memcache connections
+	// or not enough to be efficient.
+	// Value chosen through performance tests on sock-shop. I was unable to get higher performance than this.
+	defaultRemoteConnections   = 125 // Chosen performance tests on sock-shop. Unable to get higher performance than this.
+	defaultMemcacheConnections = 10  // This doesn't need to be high. The user is only requesting one tag/image at a time.
 )
 
 func optionalVar(fs *pflag.FlagSet, value ssh.OptionalValue, name, usage string) ssh.OptionalValue {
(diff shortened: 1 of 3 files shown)


squaremo:
We mitigated the problem we were having ourselves by giving memcache operations a generous timeout. The rationale is that we care less about doing these quickly than we care about succeeding at operations. If memcached is under load, it's better to treat that as backpressure, and slow down, than to repeatedly fail (though quickly).

However: there is merit in this PR I think, since we can still try to make things run quickly when they can. It seems reasonable to me that letting the client keep a larger number of idle connections around, as the only user of the memcached in question, is an improvement.

I'm less convinced that we understand how the fetching and storing perform (i.e., the argument that the request burst number and the max idle connections should be the same), or that this is the end of the story. So I think some more digging is warranted here.

philwinder:
There are a few improvements in this PR.

1. We update the memcache version to use the MaxIdleCons setting. This setting initially creates x number of connections to memcache and holds them open.

This is better than requesting new connections because memcache is slow to open new connections (it handles connection requests on the same thread).

2. We create a new memcache client for the warmer and the registry. The idea behind this is that the warmer needs lots of connections to do lots of reading and writing. The registry connection only needs a small number because a user/the web UI is unlikely to make concurrent registry requests.

3. We set the number of http goroutines to match the number of warmer memcache client connnections to be the same.

This is because if there are more HTTP goroutines, there will not be enough memcache connections to service each remote GET. I.e. remember that the code first does a memcache GET to see if it has the item in the cache. If not, it then does a memcache PUT to update it.

Hence, if we're getting lots of manifests all at the same time in the HTTP goroutines, we need the corresponding number of memcache connections to check whether we have already cached it and set it.

So, I think the issues you were talking about are orthogonal to this PR. The issues you are talking about refer to the total timeout for getting a set of manifests for a single image. Currently we're at a 10s timeout. If it takes longer than that, all the other requests will get cancelled. Also see #713.

squaremo:
> There are a few improvements in this PR.
> 
> We update the memcache version to use the MaxIdleCons setting. This setting initially creates x number of connections to memcache and holds them open.

I thought at first `MaxIdleConns` might put a cap on the number of open connections, and/or create them ahead of time; but it doesn't. The client just closes any connections _in excess of_ `MaxIdleConns` (per memcached address) after using them, and it doesn't open any ahead of time.

> This is better than requesting new connections because memcache is slow to open new connections (it handles connection requests on the same thread).

Did we get evidence that that was the specific problem we're having? I remember seeing both connection timeouts and I/O timeouts in logs (more so the latter, when stressing it locally).

> We create a new memcache client for the warmer and the registry. The idea behind this is that the warmer needs lots of connections to do lots of reading and writing. The registry connection only needs a small number because a user/the web UI is unlikely to make concurrent registry requests.

There's also the automator checking for new versions, though that's only once every minute (unless people decide they need it to be more often). I think it's a reasonable change to separate the memcache clients.

> We set the number of http goroutines to match the number of warmer memcache client connnections to be the same.
> 
> This is because if there are more HTTP goroutines, there will not be enough memcache connections to service each remote GET. I.e. remember that the code first does a memcache GET to see if it has the item in the cache. If not, it then does a memcache PUT to update it.
>
> Hence, if we're getting lots of manifests all at the same time in the HTTP goroutines, we need the corresponding number of memcache connections to check whether we have already cached it and set it.

I don't think things are quite as absolute as you've described here: there are always memcached connections available, since the client will just create them. But I suppose we could expect less churn of connections if we usually have as many connections idle as there are operations to do.

> So, I think the issues you were talking about are orthogonal to this PR. The issues you are talking about refer to the total timeout for getting a set of manifests for a single image. Currently we're at a 10s timeout. If it takes longer than that, all the other requests will get cancelled. Also see #713.

Nope, I was definitely talking about memcache operations timing out.

philwinder:
> Did we get evidence that that was the specific problem we're having? I remember seeing both connection timeouts and I/O timeouts in logs (more so the latter, when stressing it locally).

The evidence was that without the `MaxIdleCons` setting, we were getting memcache timeouts. With it, we are not.

> there are always memcached connections available, since the client will just create them.

The client will create them, but it creates them too slowly. Imagine you make 1000 HTTP requests. This causes 1000 new memcache connections. Memcache is very slow to create new connections. See documentation here: https://github.com/memcached/memcached/wiki/Performance#max-clients

And there are other complaints around the internet. Their recommendation is:
> Use persistent connections ...

This is what maxidlecons does.

> Nope, I was definitely talking about memcache operations timing out.

My misunderstanding. The current default is 1s. This can be altered at the command line.

cmd/fluxd/main.go
```diff
@@ -38,11 +38,12 @@ import (
...
 
 const (
-	defaultRemoteConnections = 125 // Chosen performance tests on sock-shop. Unable to get higher performance than this.
-	// Memcache: There is only one thread to accept new client connections. If you are cycling connections very quickly, you can overwhelm the thread. Use persistent connections or UDP in this case.
-	// See https://github.com/memcached/memcached/wiki/Performance
-	// Hence, keep the number of new connections low.
-	defaultMemcacheConnections = 10
+	// The number of connections chosen for memcache and remote GETs must match.
+	// If they do not match then the remote either creates too many memcache connections
+	// or not enough to be efficient.
```
squaremo: I think this comment, in light of what `MaxIdleConns` actually does, is a bit strongly worded! It's more like a rule of thumb ...

cmd/fluxd/main.go
```diff
@@ -78,11 +79,11 @@ func main() {
 		memcachedHostname    = fs.String("memcached-hostname", "", "Hostname for memcached service to use when caching chunks. If empty, no memcached will be used.")
 		memcachedTimeout     = fs.Duration("memcached-timeout", time.Second, "Maximum time to wait before giving up on memcached requests.")
 		memcachedService     = fs.String("memcached-service", "memcached", "SRV service used to discover memcache servers.")
-		memcachedConnections = fs.Int("memcached-connections", defaultMemcacheConnections, "maximum number of connections to memcache")
+		memcachedConnections = fs.Int("memcached-connections", defaultMemcacheConnections, "maximum number of registry connections to memcache")
```
squaremo: Consider renaming this, or at least describing it differently -- it's really the target for the number of operations done for any one image data request (can these occur concurrently?). Maybe we shouldn't even have this option.

cmd/fluxd/main.go
```diff
@@ -210,22 +211,36 @@ func main() {
...
+		var memcacheRegistry registryMemcache.Client
 		if *memcachedHostname != "" {
-			memcacheClient = registryMemcache.NewMemcacheClient(registryMemcache.MemcacheConfig{
+			memcacheRegistry = registryMemcache.NewMemcacheClient(registryMemcache.MemcacheConfig{
 				Host:           *memcachedHostname,
 				Service:        *memcachedService,
 				Timeout:        *memcachedTimeout,
 				UpdateInterval: 1 * time.Minute,
 				Logger:         log.NewContext(logger).With("component", "memcached"),
+				MaxConnections: *memcachedConnections,
```
squaremo: Ought to just name this after the client setting. This name is misleading ..

cmd/fluxd/main.go
```diff
@@ -210,22 +211,36 @@ func main() {
...
+		}
+		var memcacheWarmer registryMemcache.Client
+		if *memcachedHostname != "" {
+			memcacheWarmer = registryMemcache.NewMemcacheClient(registryMemcache.MemcacheConfig{
+				Host:           *memcachedHostname,
+				Service:        *memcachedService,
+				Timeout:        *memcachedTimeout,
+				UpdateInterval: 1 * time.Minute,
+				Logger:         log.NewContext(logger).With("component", "memcached"),
+				MaxConnections: *registryBurst,
```
squaremo: As above

squaremo:
> The client will create them, but it creates them too slowly. Imagine you make 1000 HTTP requests. This causes 1000 new memcache connections. Memcache is very slow to create new connections. See documentation here: https://github.com/memcached/memcached/wiki/Performance#max-clients

Sure, but we're not doing 1000 HTTP requests, we're globally rate limiting them to about a hundred requests at a time.

I do take your point about connection churn -- having more idle connections around is going to help with that. I just don't want to oversell it. Or rather, I don't want to be claiming that the memcache client was being _limited_ to a number of connections -- that's not the case, and therefore not something that's being alleviated here.

philwinder:
> I don't want to be claiming that the memcache client was being limited to a number of connections
Sure thing. But the point stands. We were getting timeout errors because memcache couldn't create connections quickly enough.

I'll fix the review comments.

cmd/fluxd/main.go
philwinder: ✔️

cmd/fluxd/main.go
philwinder: ✔️

cmd/fluxd/main.go
philwinder: ✔️

cmd/fluxd/main.go
philwinder: Agreed. Shouldn't ever need to alter this value. There are always going to be a small (usually 1) number of concurrent cache requests from the user.

✔️

PR #688, opened 2017-08-04
don't log errors that were caused by a context timeout

philwinder:
For each image fetch (including tags and all manifests) is limited by a context deadline of 10s. After this all remaining http requests will return an `i/o timeout` type error.

We wish to allow time for the other images to go and fetch their stuff, so it was recommended that we just suppress this error and re-request the unfetched manifests on a subsequent warm.

### DIFF
--- a/registry/warming.go
+++ b/registry/warming.go
@@ -4,13 +4,13 @@ package registry
 import (
 	"context"
 	"encoding/json"
+	"net"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
-
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/registry/cache"
 )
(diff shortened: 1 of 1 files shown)


registry/warming.go
```diff
@@ -141,9 +141,11 @@ func (w *Warmer) warm(id flux.ImageID, creds Credentials) {
 			// Get the image from the remote
 			img, err := client.Manifest(imageID)
 			if err != nil {
-				if !strings.Contains(err.Error(), context.DeadlineExceeded.Error()) && !strings.Contains(err.Error(), "net/http: request canceled") {
-					w.Logger.Log("err", errors.Wrap(err, "requesting manifests"))
+				if err, ok := errors.Cause(err).(net.Error); ok && err.Timeout() {
```
squaremo: Does this amount to a better test for the same thing?

squaremo: A: Nope -- it's testing for a timeout rather than a deadline. (Also it looks cleaner!)

squaremo:
This PR is an advance on what we have -- it tests for the correct kind of error to suppress. I have logged the issue #713 for replacing our use of context timeouts with a deadline.

PR #690, opened 2017-08-07
Log failure instead of returning on CollectAvailableImages

sambooo:


update/release.go
```diff
@@ -196,7 +196,7 @@ func (s ReleaseSpec) calculateImageUpdates(rc ReleaseContext, candidates []*Serv
 
 	switch s.ImageSpec {
 	case ImageSpecLatest:
-		images, err = collectUpdateImages(rc.Registry(), candidates)
+		images, err = collectUpdateImages(rc.Registry(), candidates, logger)
```
squaremo: Are you sure we want to fail silently in this case?

sambooo: If the repo update fails then that image won't be included in the result and the failure will be logged. What alternative would you suggest?

squaremo: What I'm afraid of the situation in which you ask to upgrade a particular service to `--update-all-images`, and you are told `Nothing to do`, which you take to mean it is up to date.

```
$ kubectl -n hello delete pod memcached-3275692475-7hbpt
pod "memcached-3275692475-7hbpt" deleted
mikeb@rufus in ~/space/fluxy-gopath/fluxy(s:log-on-repo-update-faili)
$ fluxctl $LOCAL release --service=hello/helloworld --update-all-images
Submitting release ...
Nothing to do
```

But it turns out we _already_ do that and this is not making it worse.

For the record, the alternative I had in mind was that any services for which the cache does not have a complete picture are reported as a failure in the above; e.g.,

```
$ kubectl -n hello delete pod memcached-3275692475-7hbpt
pod "memcached-3275692475-7hbpt" deleted
mikeb@rufus in ~/space/fluxy-gopath/fluxy(s:log-on-repo-update-faili)
$ fluxctl $LOCAL release --service=hello/helloworld --update-all-images
Submitting release ...
SERVICE           STATUS   UPDATES
hello/helloworld  failed   error: incomplete image data
```

squaremo: (I've logged the problem mentioned as https://github.com/weaveworks/flux/issues/691)

PR #693, opened 2017-08-08
Log every case in GitPollLoop

sambooo:
To be reverted after I'm done with it


daemon/loop.go
```diff
@@ -51,28 +51,35 @@ func (d *Daemon) GitPollLoop(stop chan struct{}, wg *sync.WaitGroup, logger log.
 	}
 
 	imagePollTimer := time.NewTimer(d.RegistryPollInterval)
+	loopLogger := log.NewContext(logger).With("thing", "git-poll-loop")
```
squaremo: structured logging ftw :)

sambooo: gotta know which thing it is

PR #697, opened 2017-08-09
[WIP] An edgy test case with Istio example

errordeveloper:
While playing with Istio, I've encotered an issue with Flux failing to parse a manifest that contains 1 service and 3 deployments.

The reviews service is defind as part of a larger [multi-doc manifest]( https://github.com/istio/istio/blob/release-0.1/samples/apps/bookinfo/bookinfo.yaml#L85-L156). In order to enable it with Istion sidecars, one has to use `istioctl kube-inject`, and only after that it becomes consumable by Flux. I've checked in the output of `istioctl kube-inject` into a repo of my own and split the app into smaller [multi-docs](https://github.com/errordeveloper/istio-gitops/blob/bc1f39c739c9b5676cf252eeada9f79042a727bd/app/reviews.yaml), and hooked it up to Flux (using v1.0.0-beta).

Flux tripped over this with an obscure error message:
```
ts=2017-08-09T14:58:11Z caller=loop.go:132 component=sync-loop err="Service default/reviews: running kubectl: error: error validating \"STDIN\": error validating data: field spec.template.spec.containers[0].name for v1.Container is required; if you choose to ignore these errors, turn validation off with --validate=false; Deployment default/reviews-v1: yaml: line 1: mapping values are not allowed in this context"
```
I've test the manifest using `kubectl apply` and it worked just fine.

I then broke the manifest into 4 (1 for the service, and one for each of the deployments), and that worked fine with Flux (see [`reviews*.yaml`](https://github.com/errordeveloper/istio-gitops/tree/9890d9a666a0e9061f9888e2bd0ece02f0b0bd3a/app)).
I've then combined deployments into one file, and that gave me a new error:
```
ts=2017-08-09T16:34:53Z caller=loop.go:132 component=sync-loop err="Deployment default/reviews-v1: running kubectl: The Deployment \"reviews-v3\" is invalid: \n* spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`\n* spec.template.spec.containers: Required value"
```
So I realised something must be a litte odd here.

In this PR I've added a test case with large multi-doc and after I've made some trivial updates in the tests, I've discovered that we end-up with a bit of a mess:

- [a fragment of `PodSpec` for one of the `review-v?` deployments](https://gist.github.com/errordeveloper/e8851a23e87951e26ddbb0692f6a7ae2#file-flux-istio-example-test-L95-L130)
- [a top fragment of `reviews-3` deployment](https://gist.github.com/errordeveloper/e8851a23e87951e26ddbb0692f6a7ae2#file-flux-istio-example-test-L132-L154) (there is also separate [full-copy of it](https://gist.github.com/errordeveloper/e8851a23e87951e26ddbb0692f6a7ae2#file-flux-istio-example-test-L217-L276))
- [a top fragment of `reviews-2` deployment](https://gist.github.com/errordeveloper/e8851a23e87951e26ddbb0692f6a7ae2#file-flux-istio-example-test-L310-L332)

### DIFF
--- a/cluster/kubernetes/files.go
+++ b/cluster/kubernetes/files.go
@@ -37,7 +37,7 @@ func (c *Manifests) FindDefinedServices(path string) (map[flux.ServiceID][]strin
 			for _, template := range templates {
 				if res.Meta.Namespace == template.namespace && matches(res, template.PodTemplate) {
 					sid := res.ServiceID()
-					result[sid] = append(result[sid], template.source)
+					result[sid] = appendIfMissing(result[sid], template.source)
 				}
 			}
 		case *resource.Deployment:
@@ -46,7 +46,7 @@ func (c *Manifests) FindDefinedServices(path string) (map[flux.ServiceID][]strin
 			for _, service := range services {
 				if res.Meta.Namespace == service.Meta.Namespace && matches(service, &res.Spec.Template) {
 					sid := service.ServiceID()
-					result[sid] = append(result[sid], source)
+					result[sid] = appendIfMissing(result[sid], source)
 				}
 			}
 		}
@@ -70,4 +70,13 @@ func matches(s *resource.Service, t *resource.PodTemplate) bool {
 		}
 	}
 	return true
+}
+
+func appendIfMissing(slice []string, i string) []string {
+	for _, v := range slice {
+		if v == i {
+			return slice
+		}
+	}
+	return append(slice, i)
 }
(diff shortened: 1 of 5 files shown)


squaremo:
Good to have this extra test case. Can you reduce it to just the thing(s) that upset the parsing? (And I guess we don't need it to be istio-specific, which would remove the licensing concern)

errordeveloper:
I'd like to understand what exactly causes this behaviour, and I don't know
enough yet. I agree, however I think it's possible that a significantly
large manifest may be the key to this behaviour.

squaremo:
> a significantly
> large manifest may be the key to this behaviour.

#826 increased the buffer used for scanning multidocument YAML files, if that was indeed the problem here.

I'm going to close this PR, but please do raise an issue if you can still reproduce the problem, @errordeveloper.

PR #699, opened 2017-08-10
Show either the public key _or_ fingerprint

squaremo:
You don't usually need both, and it's convenient for copying and pasting if by default it just shows the base64-encoded key.

### DIFF
--- a/cmd/fluxctl/identity_cmd.go
+++ b/cmd/fluxctl/identity_cmd.go
@@ -8,7 +8,9 @@ import (
 
 type identityOpts struct {
 	*rootOpts
-	regenerate bool
+	regenerate  bool
+	fingerprint bool
+	visual      bool
 }
 
 func newIdentity(parent *rootOpts) *identityOpts {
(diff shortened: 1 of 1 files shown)


cmd/fluxctl/identity_cmd.go
```diff
@@ -22,6 +24,8 @@ func (opts *identityOpts) Command() *cobra.Command {
 		RunE:  opts.RunE,
 	}
 	cmd.Flags().BoolVarP(&opts.regenerate, "regenerate", "r", false, `Generate a new identity`)
+	cmd.Flags().BoolVarP(&opts.fingerprint, "fingerprint", "l", false, `Show fingerprint of public key`)
```
sambooo: Why `l` for this flag?

squaremo:
The flags mirror those of `ssh-keygen`, with a minor adaption: `-v` in ssh-keygen is for `verbose`, and used in conjunction with `-l` (show fingerprint) prints the asciiiart. I've just called it "visual" and made it imply `-l`.

cmd/fluxctl/identity_cmd.go
squaremo: That's what `ssh-keygen` uses for the same purpose. I don't know why; but, `-f` would be bad because it usually means "file".

PR #701, opened 2017-08-10
Replace docker.io with index.docker.io in ImageID

### LINKED ISSUE #692: shouldn't fail on images with docker.io
It looks like we treat image that begin with `docker.io/` incorrectly.

When I've tried using `docker.io/errordeveloper/prom-nodejs-demo-load-test:slow`, I got this in the Weave Cloud UI:

fluxdockerhub404

So what we are seeing is the pretty HTML displaying a 404 message, as a result of fetching https://www.docker.com/v2/errordeveloper/prom-nodejs-demo-load-test/tags/list.

It looks like we should somehow go to a different domain, but I have no idea which. This smells like a special case that exists for historic reasons. Namely, docker.io predates Docker Hub, its API and the docker.com website.

sambooo:
Requests to `docker.io` redirect to `www.docker.com` rather than `index.docker.io`, so this change will rewrite image ids to use the latter.

Fixes #692

### DIFF
--- a/image.go
+++ b/image.go
@@ -12,6 +12,8 @@ import (
 const (
 	dockerHubHost    = "index.docker.io"
 	dockerHubLibrary = "library"
+
+	oldDockerHubHost = "docker.io"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


squaremo:
It's fine and good this got merged; it made me wonder if there's not a better overall approach to dealing with image names vs where to fetch image manifests.

At the minute we essentially mutate the image name to its canonical form. But this means, for example, that the image names we display in `list-images` aren't always as supplied by the user, and that can be confusing and might even lead to mistaken copypasta.

So I think it'd be better to keep image names as-is in results, and just use the canonical form where it's needed for fetching things.

squaremo:
Created https://github.com/weaveworks/flux/issues/704

PR #705, opened 2017-08-11
Propagate helpful errors through RPC

squaremo:
Sending errors over RPC is delicate. The scheme in `net/rpc` is that
an error returned by the server is stringified, and made into a new error at
the client end.

Since we want to preserve the structure of errors returned by the
daemon, we embed application-level errors in the response field. This
is a lot easier to do if there's a single type -- so one big change is
that different kinds of application error are discriminated using a
field of a single Error type, rather than being a type each.

They also need transmission over the bus, so there's a compensating
change to the bus code.

Because this fundamentally changes the data sent over RPC, it needs a
new protocol version. (These still share version numbers with the API,
since the connection endpoint -- RegisterDaemon -- is part of that
API).


squaremo:
> Renames and moves look just as important as new code and they hide the most important changes.

I tried to keep those isolated in just one of the commits.

squaremo:
Thanks @philwinder. @sambooo Want to provide a second opinion?

remote/rpc/doc.go
```diff
@@ -0,0 +1,28 @@
...
+for `flux/remote.Platform`.
+
+The purpose is to be able to access a daemon from an upstream
+service. The daemon makes an outbound connection (over, say,
+websockets), then the service can make RPC calls over that connection.
+
+On errors:
+
+Errors from the daemon can come in two varieties: application errors
+(i.e., a `*flux/errors.Error`), and internal errors (any other
```
sambooo: Nitpick: should probably be `*(flux/errors).Error` like how Go printifies em.

service/bus/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/prometheus"
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
+
+	"github.com/weaveworks/flux/service"
+)
+
+// BusMetrics has metrics for messages buses.
+type BusMetrics struct {
```
sambooo: Should be renamed `Metrics`

squaremo: Yeah fair call.

remote/rpc/doc.go
squaremo: Makes sense. I wonder if it is ambiguous otherwise, or just considered more clear.

PR #720, opened 2017-08-24
Noisily fail manifest updates

squaremo:
Currently we not only fail to update certain manifests, we fail to notice that we've failed. This adds a small test for the failing case, makes sure updates report an error when they don't do all the expected replacements, and fixes the original problem (which was failing to account for having comments between container specs).


squaremo:
> Let's just create and maintain a comment round-tripping yamel library instead.

OMG yes

Issue #714, opened 2017-08-23
fluxd can overload and become unresponsive


squaremo:
In our own dev environment, we have noticed from time to time that fluxd has stopped automatically deploying images. When we go to look at the pod, it shows up as using gigabytes of RAM and having hundreds of goroutines -- normally we'd expect maybe 50MB and tens of goroutines.

Luckily, I managed to capture a stack dump from fluxd this most recent time. Notable bit:

```
goroutine 18627648 [semacquire]:
sync.runtime_Semacquire(0x1faef64)
	/usr/local/go/src/runtime/sema.go:47 +0x30
sync.(*Mutex).Lock(0x1faef60)
	/usr/local/go/src/sync/mutex.go:85 +0xd0
sync.(*RWMutex).Lock(0x1faef60)
	/usr/local/go/src/sync/rwmutex.go:86 +0x31
syscall.forkExec(0xc44f06df40, 0xc, 0xc457fc2000, 0x6, 0x6, 0xc45d781310, 0x0, 0xc41fee25b8, 0xc44d288ea0)
	/usr/local/go/src/syscall/exec_unix.go:185 +0x1fd
syscall.StartProcess(0xc44f06df40, 0xc, 0xc457fc2000, 0x6, 0x6, 0xc45d781310, 0x2, 0x4, 0x20, 0x20)
	/usr/local/go/src/syscall/exec_unix.go:240 +0x64
os.startProcess(0xc44f06df40, 0xc, 0xc457fc2000, 0x6, 0x6, 0xc45d7814b8, 0x0, 0x15e6fe0, 0xc420dafd50)
	/usr/local/go/src/os/exec_posix.go:45 +0x1a3
os.StartProcess(0xc44f06df40, 0xc, 0xc457fc2000, 0x6, 0x6, 0xc45d7814b8, 0x0, 0x0, 0x6)
	/usr/local/go/src/os/doc.go:28 +0x64
os/exec.(*Cmd).Start(0xc42716edc0, 0xc457fc2001, 0xc420dafd50)
	/usr/local/go/src/os/exec/exec.go:358 +0x3c9
os/exec.(*Cmd).Run(0xc42716edc0, 0xc420dafd50, 0xc44f06df60)
	/usr/local/go/src/os/exec/exec.go:276 +0x2b
[...]

PR #721, opened 2017-08-28
Optimisation - reduce git note gets

philwinder:
This is an optimisation to reduce the number of `gitExecCommand` calls due to repeated requests to see if a commit has a note.

First a method to get the notes was added and the results are placed in a map. I used a map to make it easier to do "is this commid id in the map" type queries later.

Next I added a check in the main loop to see if the current commit has a note. If not, continue.

An optimisation for #714.

### DIFF
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -176,7 +176,7 @@ func (d *Daemon) doSync(logger log.Logger) {
 		serviceIDs.Add(r.ServiceIDs(allResources))
 	}
 
-	notes, err := working.NoteRevList()
+	notes, err := working.NoteRevList(ctx)
 	if err != nil {
 		logger.Log("err", errors.Wrap(err, "loading notes from repo"))
 		return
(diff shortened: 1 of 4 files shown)


squaremo:
> First a method to get the notes was added and the results are placed in a map.

If you're looking at the commits in the same order you get the notes (I think this is the case, but I'd have to check), it'd be better to keep them in a slice and step through with the commits. Then you know, for example, when there are no more notes.

philwinder:
> If you're looking at the commits in the same order you get the notes

Unfortunately during my tests the notes did not come back in a useful order. They appeared to be ordered by the commit hash. So I can't loop through them because they are not in the same order as the commits.

squaremo:
> Unfortunately during my tests the notes did not come back in a useful order. They appeared to be ordered by the commit hash.

Ach that's a shame -- thinking about it, why would they be in order. Not to worry.

PR #725, opened 2017-08-30
682 sync notification on irrelevant changes

tamarakaufler:


git/operations.go
```diff
@@ -137,12 +137,26 @@ func revlist(path, ref string) ([]string, error) {
...
-func onelinelog(path, refspec string) ([]Commit, error) {
+func onelinelog(path, refspec, gitpath string) ([]Commit, error) {
 	out := &bytes.Buffer{}
+
+	// we need to distinguish whether gitpath is populated or not,
+	// because supplying an empty string to execGitCmd results in git complaining about
+	// >> ambiguous argument '' <<
+	if gitpath != "" {
+		if err := execGitCmd(path, nil, out, "log", "--oneline", "--no-abbrev-commit", refspec, gitpath); err != nil {
+			fmt.Printf("\n onelinelog ERROR: %v\n\n", err)
```
squaremo: Debug printf?

git/operations.go
```diff
@@ -137,12 +137,26 @@ func revlist(path, ref string) ([]string, error) {
...
+			return nil, err
+		}
+		return splitLog(out.String())
+	}
+
 	if err := execGitCmd(path, nil, out, "log", "--oneline", "--no-abbrev-commit", refspec); err != nil {
+		fmt.Printf("\n onelinelog ERROR: %v\n\n", err)
```
squaremo: Debug printf?

git/operations_test.go
```diff
@@ -1,11 +1,14 @@
 package git
 
 import (
-	"github.com/weaveworks/flux/cluster/kubernetes/testfiles"
+	"fmt"
```
squaremo: Thanks for reordering these, someone somewhere is not using gofmt!

git/operations_test.go
```diff
@@ -59,24 +62,89 @@ func TestChangedFiles_NoPath(t *testing.T) {
...
 	}
-	if err = execCommand("git", "-C", dir, "add", "--all"); err != nil {
-		return err
+
+	commits, err := onelinelog(newDir, "HEAD~2..HEAD", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fmt.Printf(">>> commits: %+v\n", commits)
```
squaremo: `go test -v` will include this output, which can add to the noise. Consider only printing diagnostic messages when things have gone wrong.

git/operations.go
```diff
@@ -137,12 +137,26 @@ func revlist(path, ref string) ([]string, error) {
 }
 
 // Return the revisions and one-line log commit messages
-func onelinelog(path, refspec string) ([]Commit, error) {
+func onelinelog(path, refspec, gitpath string) ([]Commit, error) {
```
squaremo: I would interpret `gitpath` as meaning "path to git", and hence or otherwise not associate it with a subdirectory of the repo. Perhaps something like `subdir` would be better?

squaremo: We have been inconsistent elsewhere, which doesn't help :/

git/repo.go
```diff
@@ -206,13 +206,13 @@ func (c *Checkout) TagRevision(tag string) (string, error) {
 func (c *Checkout) CommitsBetween(ref1, ref2 string) ([]Commit, error) {
 	c.RLock()
 	defer c.RUnlock()
-	return onelinelog(c.Dir, ref1+".."+ref2)
+	return onelinelog(c.Dir, ref1+".."+ref2, c.repo.GitRemoteConfig.Path)
```
squaremo: This expression and the one below don't need `.GitRemoteConfig`, because the GitRemoteConfig struct is embedded, so its fields are accessible directly.

(There may still be a good reason to spell out the path, in which case just comment here :)

git/operations.go
```diff
@@ -191,7 +205,8 @@ func changedFiles(path, subPath, ref string) ([]string, error) {
 }
 
 func execGitCmd(dir string, keyRing ssh.KeyRing, out io.Writer, args ...string) error {
-	//	println("git", strings.Join(args, " "))
+	// println("COMMAND: git", strings.Join(args, " "))
```
squaremo: Since you've touched this line: it's probably not worth keeping this -- if we decide we need to see the arguments, we should log them.

git/operations.go
tamarakaufler: Yes, I chose "gitpath" for the association with the gitpath flag. Will rename it to subdir and put a comment.

git/operations.go
tamarakaufler: Sorry about the debug printouts. I thought I cleaned up - obviously now.

git/operations.go
tamarakaufler: Sorry about the debug printouts. I thought I cleaned up - obviously now.

git/operations.go
tamarakaufler: will remove

git/operations_test.go
tamarakaufler: any time :)

git/operations_test.go
tamarakaufler: A forgotten debug :(

git/repo.go
tamarakaufler: Forgot about this. Will change.

squaremo:
Splendid, thank you @tamarakaufler !

PR #732, opened 2017-09-05
Wait for platform presence in fluxsvc test

squaremo:
One source of flakes appears to be the race between the platform being registered and seeing if the platform is registered (both `TestFluxsvc_Register` and `TestFluxsvc_Ping`).

This change forces those tests to block until the platform has demonstrably subscribed, or fail if it does not happen within five seconds.

I've run this a few times in circleci, and although you never know with flakes, it appears to improve the situation. Still a bit of a shot in dim light.


squaremo:
If this flakes, I swear .......

PR #733, opened 2017-09-06
695 include user in git commit

tamarakaufler:


daemon/daemon.go
```diff
@@ -252,6 +253,8 @@ func (d *Daemon) updatePolicy(spec update.Spec, updates policy.Updates) DaemonJo
 			return metadata, nil
 		}
 
+		ctx = context.WithValue(ctx, "user", spec.Cause.User)
```
squaremo: It's perhaps arguable, but I think this falls foul of the advice "Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions." In other words, the author would be better as an argument to the specific git operations, rather than data smuggled through the context.

git/operations.go
```diff
@@ -42,6 +43,21 @@ func clone(ctx context.Context, workingDir string, keyRing ssh.KeyRing, repoURL,
 }
 
 func commit(ctx context.Context, workingDir, commitMessage string) error {
+	user := ctx.Value("user").(string)
+
+	// The user in the update.Cause.User must be a valid github user, otherwise the commit fails
```
squaremo: Not quite accurate. `git` (not github) will reject a commit if you supply _only_ an email address rather than e.g., `"Michael Bridgen <michael@weave.works>"`, and that email address has not been used in a prior revision.

This is certainly a wrinkle though. Some alternatives:

 - assume that commits will have been made by a particular user, and let git reject those that aren't
 - make people supply a git-author-like value as the username, e.g., `fluxctl release --user="Michael Bridgen <michael@weave.works>" ...`. Not so easy if you are the web UI and you simply don't have all of that information.
 - let people toggle this behaviour, and then do one of the above in the case that it's switched on

daemon/daemon.go
tamarakaufler: In my mind this was request specific, which is why I used the context. I shall refactor.

squaremo:
> In my mind this was request specific

If it were an authorised user that had been injected by some request-processing middleware (which ideally, it would be!) then we'd probably fish it out of the context in some or other handler and supply it to the lower level operations. But as it stands, it's a parameter to the API call, so I think it should be treated like the other parameters.

tamarakaufler:
I have created a new branch (imaginatively called 695b-include-user-in-git-commit) with totally new implementation. I am therefore closing this PR and will create a new one there.

PR #735, opened 2017-09-07
695b include user in git commit

tamarakaufler:


git/operations.go
```diff
@@ -41,11 +42,31 @@ func clone(ctx context.Context, workingDir string, keyRing ssh.KeyRing, repoURL,
 	return repoPath, nil
 }
 
-func commit(ctx context.Context, workingDir, commitMessage string) error {
+func commit(ctx context.Context, workingDir string, commitAction *CommitAction) error {
+	commitAuthor := commitAction.Author
+	// The user in the update.Cause.User must be a valid github user, otherwise the commit fails
```
squaremo: This is not quite accurate. git (not github) will reject a commit if you supply only an email address rather than e.g., "Michael Bridgen <michael@weave.works>", and that email address has not been used in a prior revision.

This is certainly a wrinkle though. Some alternatives:

assume that commits will have been made by a particular user, and let git reject those that aren't
make people supply a git-author-like value as the username, e.g., fluxctl release --user="Michael Bridgen <michael@weave.works>" .... Not so easy if you are the web UI and you simply don't have all of that information.
let people toggle this behaviour, and then do one of the above in the case that it's switched on

git/operations.go
```diff
@@ -41,11 +42,31 @@ func clone(ctx context.Context, workingDir string, keyRing ssh.KeyRing, repoURL,
...
+	// Hardcoding of the user to ignore is not ideal. If there is only one to ignore, ie test@test,
+	// then this is just about ok. Otherwise there could be config file with stored users to ignore etc.
+	// Currently the committer is hardcoded to flux-bot for user requested actions.
+	// TODO (potentially): allow the users to set the committer, ie have an additional
+	// update.Cause.Committer field.
```
squaremo: The committer is already supplied as the arguments to the daemon -- `--git-user` and `--git-email` -- and written into the repo config when it's cloned. Unless there's some reason to vary it per commit, I don't think this new code needs to do anything relating to the committer.

git/operations.go
```diff
@@ -41,11 +42,31 @@ func clone(ctx context.Context, workingDir string, keyRing ssh.KeyRing, repoURL,
...
+	if commitAuthor != "" && commitAuthor != "test@test" {
```
squaremo: Any solution to the wrinkle above should obviate this special case.

git/operations.go
```diff
@@ -41,11 +42,31 @@ func clone(ctx context.Context, workingDir string, keyRing ssh.KeyRing, repoURL,
...
+		if err := config(ctx, workingDir, "flux-bot", "flux-bot@weave.works"); err != nil {
+			return errors.Wrap(err, "git commit config")
+		}
+		if err := execGitCmd(ctx,
+			workingDir, nil, nil,
```
squaremo: Might be easier just to add the author to a slice of arguments, and exec in one place.

daemon/daemon.go
```diff
@@ -96,7 +97,7 @@ func (d *Daemon) ListServices(namespace string) ([]flux.ServiceStatus, error) {
 	return res, nil
 }
 
-// List the images available for set of services
+// ListImages - List the images available for set of services
```
squaremo: Are these a result of linting or something?

git/operations.go
tamarakaufler: These three versions worked:

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="Michael Bridgen <michael@weave.works>" lock --service=default/helloworld
default/helloworld  success  
Commit pushed:	af4ed19

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="Michael Bridgen" unlock --service=default/helloworld
default/helloworld  success  
Commit pushed:	40f30d4

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="michael@weave.works" lock --service=default/helloworld
default/helloworld  success  
Commit pushed:	036e2a5

git/operations.go
tamarakaufler: a)
I  misunderstood our discussion about the custom committer a couple of days. I thought you were suggesting an additional fluxctl option. Understanding corrected.
b)
I  always had the default git-user, Weave Flux, as a commit author and a committer prior to this implementation.  (This poses a question (to me) what git-user and git-email is used elsewhere when starting the daemon. ) My understanding is that non user requested commits are using that (whether the default value or not). The issue 695 talks about having a different commit author/committer for user requested actions. This I interpret that different commits, depending on whether through actions requested by the user, can have a different committer.

git/operations.go
tamarakaufler: I thought I was adding the --author etc to the argument slice. Will revisit.

git/operations.go
squaremo: Yup; the first gives the full author format, so it succeeds (for the record: there are no commits using that name _and_ email as the author in `weaveworks/flux-example`). The second and third succeed because git can find a commit with the username, and email, respectively. If you look at the commit log, it has the full author user and email recorded.

tamarakaufler: This does not work:

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="test@test" unlock --service=default/helloworld
Error: git commit: fatal: --author 'test@test' is not 'Name <email>' and matches no existing author

and this does:

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="Tamara Kaufler <test@test>" unlock --service=default/helloworld
default/helloworld  success

git/operations.go
squaremo: OK, sorry about leading you down the garden path on the git committer.

> The issue 695 talks about having a different commit author/committer for user requested actions

I have asked for clarification in that issue.

git/operations.go
squaremo: The potential "unhappy path" is if you supply an email or a user (but not both) that hasn't been used before. For example,

    fluxctl --url=http://$(minikube ip):30081/api/flux --token=local-test-token --user="Brichael Midgen" unlock --service=default/helloworld

tamarakaufler: These work as well:

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="Test <test@test>" lock --service=default/helloworld

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="Test <test>" unlock --service=default/helloworld

fluxctl --url=http://$(minikube ip):30081/api/flux  --token=local-test-token --user="test@test <test@test>" lock --service=default/helloworld


The custom author will be accepted if the format is:
some_text <another_text> 
some_text<some_text>

daemon/daemon.go
squaremo: (this is ignorable, I just find those changes a bit distracting so I wondered what produced them)

tamarakaufler: It's recommended that public methods/functions are commented and the comment starts with the method/function name. I am using Visual Studio and it is >helpfully<  highlighting all that does not have the recommended documentation. So, sometimes, I update the comment.

sambooo: FWIW I think this should be `ListImages lists the images ...`

tamarakaufler: I think both options are acceptable. A matter of personal preference I would say.
FYI, the implementation is being refactored.

squaremo:
image

cmd/fluxctl/args.go
```diff
@@ -16,11 +20,70 @@ func parseServiceOption(s string) (update.ServiceSpec, error) {
...
+		return userGitconfig
+	}
+	res := out.String()
+	return userGitconfigMap(res)
+}
+
+func userGitconfigMap(s string) map[string]string {
+	c := make(map[string]string)
+	lines := splitList(s)
+	if len(lines) == 0 {
```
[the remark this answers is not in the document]
tamarakaufler: true, removed

cmd/fluxd/main.go
```diff
@@ -71,12 +72,13 @@ func main() {
...
-		gitPath   = fs.String("git-path", "", "path within git repo to locate Kubernetes manifests (relative path)")
-		gitUser   = fs.String("git-user", "Weave Flux", "username to use as git committer")
-		gitEmail  = fs.String("git-email", "support@weave.works", "email to use as git committer")
-		gitLabel  = fs.String("git-label", "", "label to keep track of sync progress; overrides both --git-sync-tag and --git-notes-ref")
+		gitURL       = fs.String("git-url", "", "URL of git repo with Kubernetes manifests; e.g., git@github.com:weaveworks/flux-example")
+		gitBranch    = fs.String("git-branch", "master", "branch of git repo to use for Kubernetes manifests")
+		gitPath      = fs.String("git-path", "", "path within git repo to locate Kubernetes manifests (relative path)")
+		gitUser      = fs.String("git-user", "Weave Flux", "username to use as git committer")
+		gitEmail     = fs.String("git-email", "support@weave.works", "email to use as git committer")
+		gitSetAuthor = fs.Bool("git-set-author", false, "boolean to set a git author different to git committer")
```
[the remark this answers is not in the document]
tamarakaufler: it's a bit tricky this one as it should be short enough but self explanatory too. What about:
If set, the author of git commits will reflect the user who initiated the commit and will differ from the git committer.

daemon/daemon.go
```diff
@@ -511,3 +525,10 @@ func policyEventTypes(u policy.Update) []string {
 	sort.Strings(result)
 	return result
 }
+
+func getCommitAuthor(setAuthor bool, author string) string {
```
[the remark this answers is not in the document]
tamarakaufler: It's used twice, hence the function. I shall remove it if you still think it's the best

cmd/fluxctl/args.go
```diff
@@ -16,11 +20,70 @@ func parseServiceOption(s string) (update.ServiceSpec, error) {
...
-	username := ""
-	user, err := user.Current()
-	if err == nil {
-		username = user.Username
-	}
+	username := getCommitAuthor()
+
 	cmd.Flags().StringVarP(&opts.Message, "message", "m", "", "attach a message to the update")
-	cmd.Flags().StringVar(&opts.User, "user", username, "override the user reported as initating the update")
+	cmd.Flags().StringVar(&opts.User, "user", username, "override the user reported as initiating the update")
```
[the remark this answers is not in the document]
tamarakaufler: :D

daemon/daemon.go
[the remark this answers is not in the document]
tamarakaufler: done

cmd/fluxctl/args.go
```diff
@@ -16,11 +19,64 @@ func parseServiceOption(s string) (update.ServiceSpec, error) {
...
+}
+
+func getCommitAuthor(authorInfo map[string]string) string {
+	userName := authorInfo["user.name"]
+	userEmail := authorInfo["user.email"]
+
+	switch userName {
```
squaremo: This is the equivalent of a nested if. I think the table of results is better expressed as

```
switch {
case userEmail != "" && userName != "":
  return fmt.Sprintf("%s <%s>", userName, userEmail)
case userEmail != "":
  return userEmail
case userName != "":
  return userName
}
return ""
```

cmd/fluxctl/args_test.go
```diff
@@ -0,0 +1,169 @@
...
+	d := `push.default=simple
+	merge.conflictstyle=diff3
+	pull.ff=only
+	core.repositoryformatversion=0
+	core.filemode=true
+	core.bare=false`
+
+	userGitconfigInfo := userGitconfigMap(d)
+	if len(userGitconfigInfo) != 6 {
+		t.Fatal("expected no content")
```
squaremo: Holdover from previous formulation, I think

cmd/fluxd/main.go
```diff
@@ -71,12 +72,13 @@ func main() {
...
-		gitPath   = fs.String("git-path", "", "path within git repo to locate Kubernetes manifests (relative path)")
-		gitUser   = fs.String("git-user", "Weave Flux", "username to use as git committer")
-		gitEmail  = fs.String("git-email", "support@weave.works", "email to use as git committer")
-		gitLabel  = fs.String("git-label", "", "label to keep track of sync progress; overrides both --git-sync-tag and --git-notes-ref")
+		gitURL       = fs.String("git-url", "", "URL of git repo with Kubernetes manifests; e.g., git@github.com:weaveworks/flux-example")
+		gitBranch    = fs.String("git-branch", "master", "branch of git repo to use for Kubernetes manifests")
+		gitPath      = fs.String("git-path", "", "path within git repo to locate Kubernetes manifests (relative path)")
+		gitUser      = fs.String("git-user", "Weave Flux", "username to use as git committer")
+		gitEmail     = fs.String("git-email", "support@weave.works", "email to use as git committer")
+		gitSetAuthor = fs.Bool("git-set-author", false, "If set, the author of git commits will reflect the user who initiated the commit and will differ from the git committer.")
```
squaremo: :100: This is a clear formulation

daemon/daemon.go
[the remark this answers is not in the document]
squaremo: Are you sure?

squaremo:
(Please `rebase -i` or squash before/while merging)

daemon/daemon.go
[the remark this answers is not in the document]
tamarakaufler: What have I missed?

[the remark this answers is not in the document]
squaremo: If you had removed it I would not expect to see it here :)

PR #736, opened 2017-09-12
Use 1.0.0 image in example deployment

sambooo:
Also adds a step in the release doc to remind us to do this in future.


squaremo:
Thanks Sam

PR #744, opened 2017-09-15
WIP Store (instance, giturl) pairs

sambooo:
Paving the way for mapping webhooks to instances later on


sambooo:
Going to try a different approach to this

PR #754, opened 2017-09-26
Fluxctl policy aliases

### LINKED ISSUE #672: Make fluxctl [(de)automate|(un)lock] commands aliases of fluxctl policy

sambooo:
Fixes #672


cmd/fluxctl/automate_cmd.go
```diff
@@ -35,23 +33,12 @@ func (opts *serviceAutomateOpts) Command() *cobra.Command {
...
-	}
-
-	jobID, err := opts.API.UpdatePolicies(noInstanceID, policy.Updates{
-		serviceID: policy.Update{Add: policy.Set{policy.Automated: "true"}},
-	}, opts.cause)
-	if err != nil {
-		return err
-	}
-	return await(cmd.OutOrStdout(), cmd.OutOrStderr(), opts.API, jobID, false, opts.verbose)
+	return policyOpts.RunE(policyOpts.Command(), args)
```
squaremo: This makes me a bit nervous -- I don't know if I trust `.Command()` to not mutate the option fields from immediately above. Does it definitely not do that? Can we be sure?

sambooo: Good point. I'll figure it out after rebass.

sambooo: FWIW, flags are only parsed once so this approach didn't break anything. I've changed the code to just use `cmd` anyway, since a future change to `spf13/cobra` could break the CLI.

PR #757, opened 2017-09-27
Skip unsupported kinds

awh:
Some resource kinds known to flux may not be supported or could have been administratively disabled; skip over them silently.


squaremo:
> Some resource kinds known to kured

kured?

awh:
> kured ?

Fixed!

PR #759, opened 2017-09-28
Get rid of instance IDs in the API, and use `context.Context`

squaremo:
We have InstanceID as the first argument to all the api interface
methods, but it's only the upstream service that cares about
instance(ID)s. Nothing else cares, and they all have to fake it.

They _do_ however care about contexts, both on the client side (for
supplying contexts to requests, even if it's `context.Background()`)
and server side (for threading contexts through from requests).

So, make the first argument of all api methods a `context.Context`
rather than an InstanceID. In the case of the service, the instance ID
is fished out of the request and supplied to the context (the _other_
use of contexts ..).

NB old and new flux{ctl,svc,d} can all still interact, since the context change is in code and not in the API or RPC protocol.

### DIFF
--- a/cmd/fluxctl/await.go
+++ b/cmd/fluxctl/await.go
@@ -18,8 +18,8 @@ var ErrTimeout = errors.New("timeout")
 
 // await polls for a job to complete, then for the resulting commit to
 // be applied
-func await(stdout, stderr io.Writer, client api.Client, jobID job.ID, apply, verbose bool) error {
-	metadata, err := awaitJob(client, jobID)
+func await(ctx context.Context, stdout, stderr io.Writer, client api.Client, jobID job.ID, apply, verbose bool) error {
+	metadata, err := awaitJob(ctx, client, jobID)
 	if err != nil && err.Error() != git.ErrNoChanges.Error() {
 		return err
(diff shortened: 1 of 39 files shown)


squaremo:
(Back-filling test fixups)

cmd/fluxctl/await.go
```diff
@@ -69,24 +70,24 @@ func awaitJob(client api.ClientService, jobID job.ID) (history.CommitEventMetada
...
+		refs, err := client.SyncStatus(ctx, revision)
 		return err == nil && len(refs) == 0, err
 	})
 }
 
 // backoff polls for f() to have been completed, with exponential backoff.
 func backoff(initialDelay, factor, maxFactor, timeout time.Duration, f func() (bool, error)) error {
 	maxDelay := initialDelay * maxFactor
-	finish := time.Now().UTC().Add(timeout)
-	for delay := initialDelay; time.Now().UTC().Before(finish); delay = min(delay*factor, maxDelay) {
```
squaremo: _en passant_ -- I don't know why these included a call to UTC() -- paranoia perhaps. But: unnecessary and thereby mildly misleading.

PR #762, opened 2017-09-28
Release and sync fixes

awh:
Fix the spurious 'unsupported kind' errors and `fluxctl save`. See commit messages for a detailed explanation.

### DIFF
--- a/cluster/kubernetes/files.go
+++ b/cluster/kubernetes/files.go
@@ -21,7 +21,10 @@ func (c *Manifests) FindDefinedServices(path string) (map[flux.ResourceID][]stri
 	var result = map[flux.ResourceID][]string{}
 	for _, obj := range objects {
 		id := obj.ResourceID()
-		result[id] = append(result[id], obj.Source())
+		_, kind, _ := id.Components()
+		if _, ok := resourceKinds[kind]; ok {
+			result[id] = append(result[id], obj.Source())
+		}
 	}
 	return result, nil
 }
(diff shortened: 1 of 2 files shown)


cluster/kubernetes/resourcekinds.go
```diff
@@ -86,13 +86,13 @@ func (dk *deploymentKind) getPodControllers(c *Cluster, namespace string) ([]pod
 
 	var podControllers []podController
 	for _, deployment := range deployments.Items {
-		podControllers = append(podControllers, makeDeploymentPodController(&deployment))
+		podControllers = append(podControllers, makeDeploymentPodController(deployment))
```
squaremo: Or, ftr, take the address of the slice item rather than the variable.

cluster/kubernetes/files.go
```diff
@@ -21,7 +21,10 @@ func (c *Manifests) FindDefinedServices(path string) (map[flux.ResourceID][]stri
 	var result = map[flux.ResourceID][]string{}
 	for _, obj := range objects {
 		id := obj.ResourceID()
-		result[id] = append(result[id], obj.Source())
+		_, kind, _ := id.Components()
+		if _, ok := resourceKinds[kind]; ok {
```
squaremo: `resourceKinds` strikes me as misnamed. ConfigMaps are resources, aren't they?

cluster/kubernetes/resourcekinds.go
awh: That's a nicer fix, I'll do that.

cluster/kubernetes/files.go
awh: Certainly, we can rename that (and anything else, chosen names were intended as placeholders anyway)

squaremo: How about `controllerKinds`?

PR #764, opened 2017-10-02
Deprecate service nomenclature

### LINKED ISSUE #776: support namespace:kind/name syntax in fluxctl
fluxctl should support the same `namespace:kind/name` syntax as is used in the UI, so that's easier for users to switch between and c&p between CLI and UI.

awh:
### Mechanical gorenames of types and constants

See the commit message for a summary of the renames. You will note that in some cases 'Service' is replaced by 'Resource' and in other cases by 'Controller' - this split is because sometimes (starting with `flux.ResourceID` itself) these types can be used to refer to _any_ kind of cluster resource, but in other cases (such as `flux.ControllerStatus`) they obviously refer specifically to pod controllers only. 

### Changes to the `fluxctl` UI

* The `list-services` command is no longer supported, and now returns a deprecation error (not a warning!) advising the use of `list-controllers` instead
* Similarly, commands previously accepting `--service` will now return a deprecation error if this flag is passed and advise the use of `--controller` instead

The rationale for this decision (rather than a deprecation warning) is that the 'new' fluxd can't provide meaningful behaviour WRT the old `<namespace>/<servicename>` identifier (other than perhaps guessing that there is also a deployment in `<namespace>` called `<servicename>`, and using that) so it seemed safest to generate a hard error.

Fixes #776.

### DIFF
--- a/cmd/fluxctl/list_services_cmd.go
+++ b/cmd/fluxctl/list_services_cmd.go
@@ -63,7 +63,7 @@ func (opts *serviceListOpts) RunE(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-type serviceStatusByName []flux.ServiceStatus
+type serviceStatusByName []flux.ControllerStatus
 
 func (s serviceStatusByName) Len() int {
 	return len(s)
@@ -77,7 +77,7 @@ func (s serviceStatusByName) Swap(a, b int) {
 	s[a], s[b] = s[b], s[a]
 }
 
-func policies(s flux.ServiceStatus) string {
+func policies(s flux.ControllerStatus) string {
 	var ps []string
 	if s.Automated {
 		ps = append(ps, string(policy.Automated))
--- a/cmd/fluxctl/list_services_cmd.go
+++ b/cmd/fluxctl/list_services_cmd.go
@@ -1,15 +1,9 @@
 package main
 
 import (
-	"context"
-	"fmt"
-	"sort"
-	"strings"
+	"errors"
 
 	"github.com/spf13/cobra"
-
-	"github.com/weaveworks/flux"
-	"github.com/weaveworks/flux/policy"
 )
 
 type serviceListOpts struct {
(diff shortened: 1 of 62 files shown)


squaremo:
The rationale is :100: I reckon.

remote/errors.go
```diff
@@ -56,14 +56,13 @@ func UnsupportedResourceKind(err error) error {
...
-Important - releasing by service name is no longer supported - if
-you're using an old version of fluxctl that does not let you specify
-a kind other than 'service', you will need to upgrade it.
+The version of the agent running in your cluster (fluxd) can only release
+updates to pod controllers such as Deployments, DaemonSets, StatefulSets and
+CronJobs.
+
+Releasing by service name is no longer supported - if you're using an old
+version of fluxctl that does not let you specify a kind other than 'service',
+you will need to upgrade it.
```
squaremo: Yeah, nice.

cmd/fluxctl/list_services_cmd.go
```diff
@@ -23,71 +17,15 @@ func newServiceList(parent *rootOpts) *serviceListOpts {
...
 func (opts *serviceListOpts) Command() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "list-services",
-		Short:   "List services currently running on the platform.",
-		Example: makeExample("fluxctl list-services"),
-		RunE:    opts.RunE,
+		Use:    "list-services",
+		Short:  "Deprecated - use list-controllers instead",
+		Hidden: true,
+		RunE:   opts.RunE,
```
squaremo: :white_check_mark:

cmd/fluxctl/lock_cmd.go
```diff
@@ -6,37 +6,52 @@ import (
...
 		RunE: opts.RunE,
 	}
 	AddOutputFlags(cmd, &opts.outputOpts)
 	AddCauseFlags(cmd, &opts.cause)
+	cmd.Flags().StringVarP(&opts.namespace, "namespace", "n", "default", "Controller namespace")
+	cmd.Flags().StringVarP(&opts.controller, "controller", "c", "", "Controller to lock")
+
+	// Deprecated
 	cmd.Flags().StringVarP(&opts.service, "service", "s", "", "Service to lock")
+	cmd.Flags().MarkHidden("service")
```
squaremo: Isn't there a `MarkDeprecated`?

awh: There is, but it just prints a warning message and then continues as normal. This meant I had to manually test to see if the arg was supplied and then error out, so I just used `MarkHidden` + the normal error return path to deliver the message.

squaremo: Got it :+1:

remote/errors.go
squaremo: It occurs to me that this could be generated from the value in resourcekinds.go. Gilding the lily perhaps ...

PR #765, opened 2017-10-02
Bug fix to skip empty resource yaml configuration (eg due to a comment)

tamarakaufler:


cluster/kubernetes/resource/resource.go
```diff
@@ -105,6 +105,9 @@ func unmarshalObject(source string, bytes []byte) (resource.Resource, error) {
 		// The remainder are things we have to care about, but not
 		// treat specially
 	default:
+		if base.Kind == "" {
```
awh: Since we're already switching on `base.kind`, this is neater IMO:
```
    case "":
        return nil, nil
        // The remainder are things we have to care about, but not
        // treat specially
    default:
        return &base, nil
```
You could go further and add a descriptive error constant:
```
    case "":
        // Kind is missing, this isn't a valid kubernetes resource
        return nil, errorInvalidResource
```
and `continue` on that in `ParseMultiDoc` - then it'd be clear in both sites what is going on.

tamarakaufler: A good suggestion about the "" case. Will use it.
I was thinking about introducing a special error, like your errorInvalidResource, then decided against it on the grounds that when the resource is empty due to an existence of a comment, it is not really an error in my mind. I shall introduce it if you prefer returning errorInvalidResource rather than nil in this scenario.

squaremo: This will also catch documents that are yamels but not kubernetes manifests (or at least, lack a mandatory field of Kubernetes manifests). Whether we should treat those the same as non-documents is another question. Fine for now at least.

PR #771, opened 2017-10-04
Use paulbellamy's docker-registry-client

sambooo:
This pinned dependency was missed in the gvt/dep migration.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -18,10 +18,15 @@
   name = "github.com/mattes/migrate"
   revision = "ce1b59bb819416ca67422929b1cb48bb342720b6"
 
+[[constraint]]
+  name = "github.com/gorilla/mux"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -18,10 +18,15 @@
   name = "github.com/mattes/migrate"
   revision = "ce1b59bb819416ca67422929b1cb48bb342720b6"
 
+[[constraint]]
```
squaremo: Where did this come from?

sambooo: The override was added in [this commit](https://github.com/weaveworks/flux/commit/8d84d8d4b55ec84f399d808a72a89750414b8039), presumably because of a build issue, but I've just removed the constraint and the repo still builds. I changed it to a constraint instead of an override just as a cleanup, but I can go one better and just remove it altogether.

PR #772, opened 2017-10-04
Makes possible git clone failure reasons stand out

tamarakaufler:


squaremo:
Does the error in question only ever get logged, or can it be returned through the API?

If you want to provide a long-form error message for users, you can construct a `fluxerr.Error{...}`. In the case that it gets logged, just the error message (which could be this wrapped error) will be output; but if it makes its way back through the API to the user, it will be displayed in full.
For an example of these in use, see https://github.com/weaveworks/flux/blob/master/remote/errors.go.

If it will not be returned via the API ... ignore me.

git/operations.go
```diff
@@ -37,7 +37,8 @@ func clone(ctx context.Context, workingDir string, keyRing ssh.KeyRing, repoURL,
 	}
 	args = append(args, repoURL, repoPath)
 	if err := execGitCmd(ctx, workingDir, keyRing, nil, args...); err != nil {
-		return "", errors.Wrap(err, "git clone")
+		clmsg := fmt.Sprintf("%s (%s)", "git clone", "requested repo does not exist, no permission to access it (is deploy key set up?) or timeout")
```
squaremo: Why the format string with two literal strings inserted?

tamarakaufler:
This branch can be deleted: please see https://github.com/weaveworks/flux/issues/722

PR #773, opened 2017-10-04
Instructions for upgrading to 1.0(.1)

squaremo:
This adds documentation for upgrading OSS Flux; i.e., if you have not used the instructions in this repo to set it up, or you have used a generated config but customised it, or otherwise know what Flux is at all.

### DIFF
--- a/site/upgrading-to-1.0.md
+++ b/site/upgrading-to-1.0.md
@@ -144,17 +144,18 @@ key used by Flux with
 
 To run the new Flux with Weave Cloud: 
 
- * Go to your instance settings (the cog icon) and click “Deploy Config”
+ * Go to your instance settings (the cog icon) and click the “Config”
+   then “Deploy” menu items
  * Enter the git URL, path and branch values from the saved config
-   (old-config.yaml)
+   (in `old-config.yaml`)
  * Run the `kubectl` command shown below the form.
- * Wait for the 
- * Follow the instructions below that to push the deploy key (this is
-   also a good opportunity to delete any old keys, if you didn’t do
-   that above).
+ * Following the instructions underneath the command, to install the
+   deploy key (this is also a good opportunity to delete any old keys,
+   if you didn’t do that above).
+ * Wait for the big red "Agent not configured" message to clear.
 
-You should now be able to click the “Deploy” tab and see your running
-system (again), with the updated Flux daemon.
+You should now be able to click the “Deploy” tab at the top and see
+your running system (again), with the updated Flux daemon.
 
 ## If you are running Flux in "standalone" mode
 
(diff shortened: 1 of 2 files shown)


site/upgrading-to-1.0.md
```diff
@@ -0,0 +1,257 @@
...
+> other words, it tries to keep the cluster running whatever is
+> represented in the repo. (Though it doesn’t delete things, yet.)
+
+To run the new Flux with Weave Cloud: 
+
+ * Go to your instance settings (the cog icon) and click “Deploy Config”
+ * Enter the git URL, path and branch values from the saved config
+   (old-config.yaml)
+ * Run the `kubectl` command shown below the form.
+ * Wait for the
```
tamarakaufler: something missing?

site/upgrading-to-1.0.md
```diff
@@ -0,0 +1,257 @@
...
+> represented in the repo. (Though it doesn’t delete things, yet.)
+
+To run the new Flux with Weave Cloud: 
+
+ * Go to your instance settings (the cog icon) and click “Deploy Config”
+ * Enter the git URL, path and branch values from the saved config
+   (old-config.yaml)
+ * Run the `kubectl` command shown below the form.
+ * Wait for the 
+ * Follow the instructions below that to push the deploy key (this is
```
tamarakaufler: That likely redundant

site/upgrading-to-1.0.md
```diff
@@ -0,0 +1,257 @@
...
+
+Set the environment variable FLUX_URL to point to the Flux service you
+are running, as described in
+[the old deployment docs](https://github.com/weaveworks/flux/blob/0.3.0/site/using.md#fluxctl-setup). The
+particular URL will differ, depending on how you have told Kubernetes
+to expose the Flux service.
+
+Before making any changes, get the config so that it can be consulted later:
+
+    ./fluxctl_030 get-config --fingerprint=md5 | tee old-config.yaml
```
tamarakaufler: md5 is hardcoded in flux. If a user uses  --fingerprint=SHA1, he will get the same result, which may be a bit confusing. Probably nitpicking.

site/upgrading-to-1.0.md
squaremo: Oops. Done in a hurry is my only defense.

site/upgrading-to-1.0.md
squaremo: Old `fluxctl` should probably have checked that the value supplied is `md5`, yeah. But I have to supply some value here, otherwise the output is the whole public key.

site/upgrading-to-1.0.md
squaremo: Oh yes: I write these in as stubs because I went away to check what you had to wait for, and didn't come back.

site/upgrading-to-1.0.md
squaremo: Why so?

Issue #778, opened 2017-10-05
Cache warming exceeds GCR rate limiting


jpellizzari:
Reported by a customer.

Logs for the `weave-flux-agent` show that the agent is making way too many requests in a short period of time:
![screen shot 2017-10-05 at 1 26 39 pm](https://user-images.githubusercontent.com/2802257/31251857-35caf764-a9d4-11e7-9c18-c64db9f11192.png)

Caller shows as `warming.go`: https://github.com/weaveworks/flux/blob/master/registry/warming.go

This seems to have started happening since the upgrade to 1.0.1

jpellizzari:
After bouncing the agent, it spammed a little more, then terminated:
![screen shot 2017-10-05 at 1 55 25 pm](https://user-images.githubusercontent.com/2802257/31252163-27a7b6d0-a9d5-11e7-9bfc-2186303b9446.png)

jr-qordoba:
This appears to be the result of Google deprecating V1:
`textPayload: "ts=2017-10-05T21:14:19Z caller=warming.go:145 component=warmer err="requesting manifests: getting remote manifest: Get https://gcr.io/v2/qordoba-devel/segmentation/manifests/latest: http: non-successful response (status=404 body=\"{\\\"errors\\\":[{\\\"code\\\":\\\"MANIFEST_UNKNOWN\\\",\\\"message\\\":\\\"Manifest with tag 'latest' has media type 'application/vnd.docker.distribution.manifest.v2+json', but client accepts 'application/vnd.docker.distribution.manifest.v1+json'.\\\"}]}\")"
"  `

squaremo:
Mitigation: tune the daemon arguments `--registry-rps` and `--registry-burst` way down (10 and 1). These govern the rate limiting for outgoing image registry requests so it doesn't spam even if it's getting errors.

The errors in question are

```
textPayload: "ts=2017-10-05T21:14:19Z caller=warming.go:145 component=warmer err="requesting manifests: getting remote manifest: Get https://gcr.io/v2/path/to/repo/manifests/latest: http: non-successful response (status=404 body=\"{\\\"errors\\\":[{\\\"code\\\":\\\"MANIFEST_UNKNOWN\\\",\\\"message\\\":\\\"Manifest with tag 'latest' has media type 'application/vnd.docker.distribution.manifest.v2+json', but client accepts 'application/vnd.docker.distribution.manifest.v1+json'.\\\"}]}\")"
```

[...]

PR #780, opened 2017-10-06
Use schema2 manifest when available

squaremo:
Version 2 of the Docker Registry API has two different schemas for
image manifests: schema1 and schema2.

Different registries deal with this differently. Google Container
Registry (gcr.io) will serve only schema2, at least for some
images. In this case we get an HTTP 404 response -- the
docker-registry-client library sets an `Accept` header with the schema1
media type, which the registry then refuses to supply.

Thus, in this commit we ask for a schema2 image first, which is then
processed slightly differently (the data we want is in a layer, which
we have to download subsequently). If we get a 404 from that, fall
back to asking for a schema1 manifest.

It's possible to get an empty schema2 result from some registries: the
registry ignores the Accept header and sends a schema1 manifest
anyway (possibly). In this case, fall back to asking for schema1.

Left to resolve:

For some images, gcr has a manifest of media type `application/vnd.docker.distribution.manifest.v1+prettyjws` (a signed schema1 manifest), but the docker-registry-client uses an Accept header of `application/vnd.docker.distribution.manifest.v1+json` (an unsigned schema1 manifest). This leads to a 404 which we might avoid otherwise.

Addresses #778.

### DIFF
--- a/registry/client.go
+++ b/registry/client.go
@@ -3,9 +3,12 @@ package registry
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/go-kit/kit/log"
+	dockerregistry "github.com/heroku/docker-registry-client/registry"
 	"github.com/pkg/errors"
 
 	"github.com/weaveworks/flux"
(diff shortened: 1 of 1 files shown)


squaremo:
Removed WIP, since I think this is an incremental improvement at least.

PR #783, opened 2017-10-12
Add indexes for common queries

leth:
This adds some indexes covering the top queries. Analysis of these queries is in a related Issue.

I've added indexes for the jobs table since they coexist in the same DB right now, and thus they may have a performance impact on each other.

### DIFF
--- a/db/migrations/postgres/012_query_index.up.sql
+++ b/db/migrations/postgres/012_query_index.up.sql
@@ -1,5 +1,7 @@
+-- Optimise per-instance history queries
 CREATE INDEX events_instance_id_started_at_desc_idx ON events USING btree(instance_id, started_at DESC);
 
+-- These optimise queries performed by flux 0.3.x
 CREATE INDEX jobs_instance_id_idx ON jobs USING btree(instance_id);
 CREATE INDEX jobs_busy_instance_id_idx ON jobs USING btree(instance_id) WHERE claimed_at IS NOT NULL AND finished_at IS NULL;
 CREATE INDEX jobs_queue_order_idx ON jobs USING btree((-1 * priority), scheduled_at, submitted_at) WHERE finished_at IS NULL AND claimed_at IS NULL;
--- a/db/migrations/postgres/012_query_index.up.sql
+++ b/db/migrations/postgres/012_query_index.up.sql
@@ -1,0 +1,6 @@
+CREATE INDEX events_instance_id_started_at_idx ON events USING btree(instance_id, started_at);
+
+CREATE INDEX jobs_instance_id_idx ON jobs USING btree(instance_id);
+CREATE INDEX jobs_busy_instance_id_idx ON jobs USING btree(instance_id) WHERE claimed_at IS NOT NULL AND finished_at IS NULL;
+CREATE INDEX jobs_queue_order_idx ON jobs USING btree((-1 * priority), scheduled_at, submitted_at) WHERE finished_at IS NULL AND claimed_at IS NULL;
+CREATE INDEX jobs_instance_id_unfinished_idx ON jobs USING btree(instance_id) WHERE finished_at IS NULL;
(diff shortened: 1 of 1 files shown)


aaron7:
It will be interesting to see the affect this has on flux api calls.

db/migrations/postgres/012_query_index.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE INDEX events_instance_id_started_at_idx ON events USING btree(instance_id, started_at);
+
+CREATE INDEX jobs_instance_id_idx ON jobs USING btree(instance_id);
+CREATE INDEX jobs_busy_instance_id_idx ON jobs USING btree(instance_id) WHERE claimed_at IS NOT NULL AND finished_at IS NULL;
+CREATE INDEX jobs_queue_order_idx ON jobs USING btree((-1 * priority), scheduled_at, submitted_at) WHERE finished_at IS NULL AND claimed_at IS NULL;
+CREATE INDEX jobs_instance_id_unfinished_idx ON jobs USING btree(instance_id) WHERE finished_at IS NULL;
```
aaron7: How much of an advantage do we get with these partial indexes `jobs_busy_instance_id_idx` and `jobs_instance_id_unfinished_idx` on `instance_id` when there is also a complete index?

I'm not familiar with how these work in postgres. Does it use only the partial index and ignore the complete index if it recognises it can do so?

leth: >I'm not familiar with how these work in postgres. Does it use only the partial index and ignore the complete index if it recognises it can do so?

In theory the query planner picks the index most appropriate to the current query.
The partial clauses match things from the WHERE condition of the query.

> How much of an advantage do we get with these partial indexes jobs_busy_instance_id_idx and jobs_instance_id_unfinished_idx on instance_id when there is also a complete index?

The more specific the index, the better the query performance. (However maintaining each extra index has a performance impact).


I've extracted them from the common queries:
- jobs_instance_id_idx matches a specific instance status query
- jobs_busy_instance_id_idx matches part of a nested query
- jobs_queue_order_idx matches the sorting constraints on a large query
- jobs_instance_id_unfinished_idx matches part of a query looking for what is next

I'm not certain these are the perfect indexes, but they should help.
We can run some analysis after deploying to see which the query planner decides to use.

db/migrations/postgres/012_query_index.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE INDEX events_instance_id_started_at_idx ON events USING btree(instance_id, started_at);
```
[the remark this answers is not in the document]
leth: Just found this: https://www.postgresql.org/docs/9.5/static/indexes-ordering.html
Sounds like it is needed!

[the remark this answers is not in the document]
leth: FIWI on this one, I was focusing on the WHERE clause, but it makes sense to cover the ordering part too!

db/migrations/postgres/012_query_index.up.sql
aaron7: thanks

leth:
Roughly, these are the queries these indexes target:

```sql
SELECT id, service_ids, type, started_at, ended_at, log_level, message, metadata FROM events WHERE instance_id = $1 AND started_at < $2 AND started_at > $3 ORDER BY started_at desc
```

```sql
SELECT id, service_ids, type, started_at, ended_at, log_level, message, metadata FROM events WHERE instance_id = $1 AND started_at < $2 AND started_at > $3 ORDER BY started_at desc LIMIT 250
```

```sql
SELECT instance_id, id, queue, method, params,
			 scheduled_at, priority, key, submitted_at,
			 claimed_at, heartbeat_at, finished_at, log, status,
			 done, success
FROM jobs

-- Scope it to our selected queues
WHERE queue IN ($1)

-- Only unclaimed/unfinished jobs are available
AND claimed_at IS NULL
AND finished_at IS NULL

-- Don't make jobs available until after they are scheduled
AND scheduled_at <= $2

-- Only one job at a time per instance * queue
AND instance_id NOT IN (
	SELECT instance_id
	FROM jobs
	WHERE queue IN ($3)
	AND claimed_at IS NOT NULL
	AND finished_at IS NULL
	GROUP BY instance_id
)

-- subtraction is to work around for ql, not being able to sort
-- multiple columns in different ways.
ORDER BY (-1 * priority), scheduled_at, submitted_at
LIMIT 1
```

```sql
SELECT count ( 1 ) FROM jobs WHERE instance_id = '$1' AND KEY = '$2' AND finished_at IS NULL;
```

PR #786, opened 2017-10-13
Make yaml parsing failure a User error

sambooo:
This allows us to detect when fluxd returns an error due to user
fault, and use a 4xx instead of 5xx status code.

### DIFF
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/pkg/errors"
 	"github.com/weaveworks/flux/resource"
 )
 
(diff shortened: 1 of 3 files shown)


cluster/kubernetes/resource/load.go
```diff
@@ -19,16 +20,16 @@ func Load(roots ...string) (map[string]resource.Resource, error) {
 	for _, root := range roots {
 		err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
 			if err != nil {
-				return fmt.Errorf(`walking %q for yamels: %s`, path, err.Error())
+				return errors.Wrap(err, fmt.Sprintf("walking %q for yamels", path))
```
rndstr: FWIW there is `errors.Wrapf()`

sambooo: Oh so there is. I'll fix that, cheers.

cluster/kubernetes/resource/resource.go
```diff
@@ -70,7 +71,17 @@ func unmarshalObject(source string, bytes []byte) (resource.Resource, error) {
 	if err := yaml.Unmarshal(bytes, &base); err != nil {
 		return nil, err
 	}
+	r, err := unmarshalKind(base, bytes)
+	if err != nil {
+		return nil, &fluxerr.Error{
+			Type: fluxerr.User,
+			Err:  err,
+		}
```
squaremo: Needs help text I reckon. Something like 

> Could not parse "howdoes.yaml". This most likely means it is malformed YAML, possibly because you read the YAML spec and decided the universe has no inherent order or meaning.

PR #787, opened 2017-10-13
Add godoc link to README

lelenanam:


lelenanam:
There was a question about API docs in Weave Users slack.

PR #791, opened 2017-10-17
Add `dep prune` to building instructions

sambooo:


squaremo:
What difference does this make?

PR #792, opened 2017-10-17
Add events-url flag for internal notifications service

jpellizzari:
Fixes https://github.com/weaveworks/notification/issues/51

Adds a flag to tell `fluxsvc` to override instance-specific Slack settings and apply a default config instead.

Also allows for a tokenized URL to be used to populate the user's instanceID and the eventType (all of which are set to deploy for now).

This represents the shortest path for us to get Flux converted to the new notifications service.

Related: https://github.com/weaveworks/service-conf/pull/1415

### DIFF
--- a/cmd/fluxsvc/main.go
+++ b/cmd/fluxsvc/main.go
@@ -13,6 +13,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
 
+	"github.com/weaveworks/flux/event"
+	"github.com/weaveworks/flux/service"
 	"github.com/weaveworks/flux/service/bus"
 	"github.com/weaveworks/flux/service/bus/nats"
 	"github.com/weaveworks/flux/service/db"
@@ -45,6 +47,7 @@ func main() {
 		databaseMigrationsDir = fs.String("database-migrations", "./service/db/migrations", "Path to database migration scripts, which are in subdirectories named for each driver")
 		natsURL               = fs.String("nats-url", "", `URL on which to connect to NATS, or empty to use the standalone message bus (e.g., "nats://user:pass@nats:4222")`)
 		versionFlag           = fs.Bool("version", false, "Get version number")
+		eventsURL             = fs.String("events-url", "", "URL to which events will be sent, or empty to use instance-specific Slack settings")
 	)
 	fs.Parse(os.Args)
 
@@ -54,6 +57,24 @@ func main() {
 	if *versionFlag {
 		fmt.Println(version)
 		os.Exit(0)
+	}
+
+	// If the events-url flag is present, ignore instance specific Slack settings.
+	var defaultEventsConfig *instance.Config
+	if *eventsURL != "" {
+		defaultEventsConfig = &instance.Config{
+			Settings: service.InstanceConfig{
+				Slack: service.NotifierConfig{
+					HookURL:         *eventsURL,
+					ReleaseTemplate: "",
+					NotifyEvents: []string{
+						event.EventRelease,
+						event.EventAutoRelease,
+						event.EventSync,
+					},
+				},
+			},
+		}
 	}
 
 	// Logger component.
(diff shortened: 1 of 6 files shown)


cmd/fluxsvc/main.go
```diff
@@ -56,6 +58,25 @@ func main() {
 		os.Exit(0)
 	}
 
+	// If the events-url flag is present, ignore instance specific Slack settings.
+	var defaultEventsConfig *instance.Config
+	if eventsURL != nil {
+		defaultEventsConfig = &instance.Config{
+			Settings: service.InstanceConfig{
+				Slack: service.NotifierConfig{
+					Username:        "Weave Cloud",
```
bboreham: This name shouldn't be hard-coded.
Maybe this whole section should be read from a file which could be configured via ConfigMap?

squaremo:
Thanks Jordan :star: Looks sensible, at a scan.

I don't think we should merge this until it's going to do the right thing when activated (by supplying `--events-url`) -- see my comments in weaveworks/service-conf#1415. I'll mark it as WIP for now.

service/notifications/notifications.go
```diff
@@ -2,23 +2,24 @@ package notifications
...
 import (
 	"github.com/weaveworks/flux/history"
+	"github.com/weaveworks/flux/service"
 	"github.com/weaveworks/flux/service/instance"
 )
 
 var DefaultNotifyEvents = []string{"release", "autorelease"}
 
-func Event(cfg instance.Config, e history.Event) error {
+func Event(cfg instance.Config, e history.Event, instanceID service.InstanceID) error {
```
bboreham: I think it would be a small improvement here to put this `InstanceID` inside the `instance.Config`.
(but see also later comments about including it in HookURL)

service/server/server.go
```diff
@@ -220,11 +223,18 @@ func (s *Server) LogEvent(ctx context.Context, e history.Event) error {
 		return errors.Wrapf(err, "logging event")
 	}
 
-	cfg, err := helper.Config.Get()
-	if err != nil {
-		return errors.Wrapf(err, "getting config")
+	// Override the users's slack settings if an events-url flag is provided.
+	var cfg instance.Config
+	if s.defaultEventsConfig != nil {
+		cfg = *s.defaultEventsConfig
```
bboreham: If you put the instance ID in here you would avoid having to pass it down through umpteen functions.
Though it's a bit more messy to do the event type replacement later.

cmd/fluxsvc/main.go
jpellizzari: Notifications service ignores this field anyway. Removed.

jpellizzari:
Updated with feedback. @bboreham Please take another look.

The ConfigMap has been updated to reflect the types of events that flux produces: https://github.com/weaveworks/service-conf/pull/1415

jpellizzari:
> it would make rollout easier to have individual config take priority over the blanket setting

@bboreham Won't our migration in https://github.com/weaveworks/notification/issues/64 cover this?

sambooo:
`fluxy.db` should not be included. I can't comment on it directly apparently.

cmd/fluxsvc/main.go
```diff
@@ -56,6 +59,24 @@ func main() {
 		os.Exit(0)
 	}
 
+	// If the events-url flag is present, ignore instance specific Slack settings.
+	var defaultEventsConfig *instance.Config
+	if eventsURL != nil {
```
squaremo: I don't think this will ever be nil (since pflags will give it the default, by now). But `*eventsURL` it may be an empty string.

squaremo: I don't think this will ever be nil (since pflags will give it the default, by now). But `*eventsURL` may be an empty string.

jpellizzari:
@squaremo Fixed

bboreham:
> @bboreham Won't our migration in weaveworks/notification#64 cover this?

It allows for a big-bang migration with a small window to have user surprise.
I was suggesting to allow existing config to take priority so we could roll it out to individual users, then wipe everyone's old config once we are happy and let the blanket setting take over.

PR #793, opened 2017-10-18
Dependency changes to facilitate fluxsvc migration

sambooo:
In preparation for moving fluxsvc to [service](https://github.com/weaveworks/service), these dependencies need to be updated:

1. `github.com/mattes/migrate`: Needs to use v1.3.2 from `gopkg.in` in order to not cause a conflict. Fortunately the API didn't change in the ~1 year between our old version and v1.3.2.
2. `github.com/paulbellamy/docker-registry-client`: Had to be forked to remove the (false) dependency on `github.com/Sirupsen/logrus` because of that wonderful S-to-s change.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -20,12 +20,6 @@
   revision = "de5bf2ad457846296e2031421a34e2568e304e35"
 
 [[projects]]
-  name = "github.com/Sirupsen/logrus"
-  packages = ["."]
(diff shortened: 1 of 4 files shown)


Gopkg.lock
```diff
@@ -272,8 +266,8 @@
   branch = "handle-relative-nextlinks"
```
squaremo: If we own it, can't this be master?

sambooo: Yeah I suppose. I'll change this and then merge.

PR #796, opened 2017-10-18
Remove unnecessary fluff from docker-registry-client constraint

sambooo:
I didn't realise `https://` and `.git` were unneeded when I added this constraint. Best get it removed before it stays around forever.


sambooo:
LGTM

Issue #602, opened 2017-06-07
Apply kubernetes namespaces before all other objects


squaremo:
At the minute we load all the manifests and apply them one by one. If you have, say, resources in a namespace that is new, they can fail if the namespace is not created before the files are.

Ultimately, this will succeed on the next sync. But it would have faster turnaround, and look less worrying, if it succeeded the first time around. As an approximation, just running namespaces before resources would be a start.

tamarakaufler:
The implementation is in incorrectly named branch

_756-include-errors-in-sync-events_

marccarre:
Since:

1. Flux (alpha) now supports Helm, and 
2. Helm exposes a [`dependencies.yml`](https://github.com/kubernetes/helm/blob/master/docs/helm/helm_dependency.md) file listing explicitly all dependencies,

does that mean we effectively get this automatically for Helm users?
If not, we might want to try to integrate Flux with this feature of Helm.

squaremo:
> does that mean we effectively get this automatically for Helm users?

It looks like Helm does a crude but effective sort, such that it applies kinds that can be dependencies before other kinds.

We could do that too, since we stage everything before applying.

sambooo:
Note that the original issue as it was originally filed was fixed by #812.

sambooo:
@marccarre "dependencies" in this issue specifically referred to the dependency a namespaced kubernetes object has on the namespace object itself. Applying resources used to fail depending on the order in which it was done because we could attempt to create an object in a namespace which hasn't yet been created.

Helm dependencies are entirely separate from this aspect of Flux. If you think it's something we should be paying attention to, please file a new issue.

PR #802, opened 2017-10-19
Sync namespaces before resources, to avoid spurious errors

tamarakaufler:


squaremo:
This doesn't appear to include errors in the sync event, as the title suggests.

squaremo:
Suggestion: "re-target" this at https://github.com/weaveworks/flux/issues/602, which it goes a long way toward fixing. By "re-target" I mean "change the title (and maybe the branch name)" :)

tamarakaufler:
You are absolutely right. Got carried away and in the end forgot about sending the errors.

tamarakaufler:
> Suggestion: "re-target" this at #602, which it goes a long way toward fixing. By "re-target" I mean "change the title (and maybe the branch name)" :)


I shall have a look how much more it will take to send the errors upstream. It would be nice to get this done.
If only the current implementation is the result of this issue, then I shall rename it

tamarakaufler:
> This doesn't appear to include errors in the sync event, as the title suggests.
> (Although it certainly improves the syncing by avoiding a common cause of errors)

You are absolutely right. Got carried away and in the end forgot about sending the errors.

sync/mock.go
```diff
@@ -0,0 +1,77 @@
...
+type rsc struct {
+	bytes []byte
+	Kind  string
+	Meta  struct {
+		Namespace string
+		Name      string
+	}
+}
+
+type rscIgnorePolicy struct {
```
squaremo: Can you get the same effect by embedding `rsc` in this struct and overriding only the method you need, rather than recapitulating its fields?

sync/sync.go
```diff
@@ -26,38 +27,84 @@ func Sync(m cluster.Manifests, repoResources map[string]resource.Resource, clus
...
+
+func separateResourcesByType(resources map[string]resource.Resource) (map[string]resource.Resource, map[string]resource.Resource) {
+	if len(resources) == 0 {
+		return nil, nil
+	}
+	nsResources := make(map[string]resource.Resource)
+	otherResources := make(map[string]resource.Resource)
+	for id, res := range resources {
+		_, kind, _ := res.ResourceID().Components()
+		if kind == "namespace" {
```
squaremo: It occurs to me that this test is rather particular to Kubernetes -- if it compared the `Kind` to a constant, that constant would be defined in `cluster/kubernetes`. Indeed, you could argue that the question of what-depends-on-what is particular to the cluster implementation.

But: I will file it as a follow-up issue, since we don't have implementations other than Kubernetes now anyway (i.e., nothing to be done here)

squaremo:
(also, either `git rebase -i master` before merging, or squash-and-merge.

sync/mock.go
tamarakaufler: I cannot confirm at the moment as my local env is broken. Will hopefully sort out tomorrow.

tamarakaufler: My local env is broken, so will do it as soon as it's fixed.

sync/sync.go
tamarakaufler: You are right. I did not consider other than Kubernetes, I assume because, as you said, we have only deal with Kubernetes at the moment.

PR #809, opened 2017-10-23
Update docs to refer to controllers instead of services

awh:


site/using.md
```diff
@@ -13,188 +13,166 @@ the command line. The `--help` for `fluxctl` is described below.
...
-  save          save service definitions to local files in platform-native format
-
-With side effect
-
-  automate      Turn on automatic deployment for a service.
-  deautomate    Turn off automatic deployment for a service.
-  lock          Lock a service, so it cannot be deployed.
-  release       Release a new version of a service.
-  unlock        Unlock a service, so it can be deployed.
+  automate         Turn on automatic deployment for a controller.
```
sambooo: Might be better to group related commands instead of sorting them lexicographically. Specifically lock/unlock and automate/deautomate.

site/using.md
```diff
@@ -13,188 +13,166 @@ the command line. The `--help` for `fluxctl` is described below.
...
-# Unlocking a Service
+# Unlocking a Controller
 
-Unlocking a service allows it to have manual or automated releases
+Unlocking a controller allows it to have manual or automated releases
 (again).
 
 ```sh
-$ fluxctl unlock --service=default/helloworld
+$ fluxctl unlock --controller=default/helloworld
```
sambooo: Should this be `deployment/helloworld`?

site/using.md
awh: This is the order they're printed out by `spf13/cobra`.

site/using.md
awh: Yus

site/using.md
sambooo: Fair enough.

PR #816, opened 2017-10-27
Remove fluxsvc

sambooo:
This shall no longer live here


Makefile
```diff
@@ -50,16 +34,11 @@ build/.%.done: docker/Dockerfile.%
 	touch $@
 
 build/.flux.done: build/fluxd build/kubectl
-build/.flux-service.done: build/fluxsvc build/migrations.tar
 
-build/fluxd: $(FLUXD_DEPS)
+build/fluxd: vendor/
```
squaremo: What was wrong with `$(FLUXD_DEPS)`?

sambooo: It called godeps: `FLUXD_DEPS:=$(call godeps,./cmd/fluxd)`. Something I didn't notice when I did that particular migration.

squaremo: Oh is `godeps` not a thing any more

sambooo: Wait, yep I'm wrong. Derp.

squaremo: Do we vendor the other bits and pieces into vendor/ now?

sambooo: Which other bits and pieces are you referring to?

squaremo:
OMG OMG OMG :100:

PR #829, opened 2017-11-07
Update gorilla/mux vendorage and adjust

### LINKED ISSUE #828: Fix Client dependency on old gorilla/mux behaviour
Due to [changes](https://github.com/gorilla/mux/compare/780415097119f6f61c55475fe59b66f3c3e9ea53...24fca303ac6da784b9e8269f724ddeb0b2eea5e7#diff-576803eb92caa3eefb60a6957a2e11d5R516) in gorilla/mux between the version pinned in this repo and the version pinned in weaveworks/service, when a route has [query parameters defined](https://github.com/weaveworks/flux/blob/22027863ebf466f38860217e7bf10c03d9324f9c/http/transport.go#L29) they are no longer optional.

We should remove any optional query parameters from the router(s) so we can use the latest mux when required.

squaremo:
To bring things in line with weaveworks/service, where these packages
are themselves vendored, update gorilla mux.

There is one hiccough: `MakeURL` no longer works as it is, since
Route.URL() now complains if it doesn't get all the parameters given
in the route, including query parameters. To work around this for the
time being, use Route.URLPath() instead, relying on the coincidence
that our routes don't contain parameters in the path.

Fixes #828, though not as described there. The problem is that `transport.MakeURL(...)` accepts both mandatory and optional query parameters alike, with no way to distinguish which to give to `Route.URL(...)`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -172,7 +172,8 @@
 [[projects]]
   name = "github.com/gorilla/mux"
   packages = ["."]
-  revision = "780415097119f6f61c55475fe59b66f3c3e9ea53"
+  revision = "7f08801859139f86dfafd1c296e2cba9a80d292e"
+  version = "v1.6.0"
 
 [[projects]]
   name = "github.com/gorilla/websocket"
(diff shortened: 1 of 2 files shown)


Gopkg.lock
```diff
@@ -32,6 +32,60 @@
   revision = "1952afaa557dc08e8e0d89eafab110fb501c1a2b"
 
 [[projects]]
+  branch = "master"
```
sambooo: This large block of additions looks a lot like fluxsvc dependencies. Did you run `dep ensure` with some service code still around by accident?

PR #830, opened 2017-11-09
Rewrite some FAQ answers and mention Flux Classic

squaremo:
Someone asked on twitter what happened to what we call Flux Classic (a
prototype we did of staged deployments with charts and so on). This
provides an answer in the FAQ.

I also tidied up some of the other answers.


squaremo:
Thanks @abuehrle :star:

PR #835, opened 2017-11-13
Image metadata lookup and automation for push hooks

squaremo:
1. To process image push hooks, we need to be able to pre-empt the cache
warming loop.

This commit introduces a priority channel, as a field of the warmer,
and changes the loop such that it looks up image data from the
priority channel in preference to refreshing image data from the
cluster.

It is possible to be notified about an image for which the loop does
not have credentials. This means the priority lookup will be skipped,
and is otherwise harmless. The credentials are only requested at the
start, and when the regular refresh is run. A refinement which
improves on this would be to plumb through a means of getting a
specific image's credentials from the cluster.

2. Add a "new image notification" to the warmer, and plumb it through to the daemon so that it prompts the daemon to do an automation run.

This is done as a callback, and the daemon method (newly exported) `AskForImagePoll` supplied. This is the simplest way, given that we want to keep the image polling (and other things) running on a ticker as well.

The code that attempts to figure out if there's a new image comes with a couple of caveats:
 * it only checks tags as strings, so it won't notice if a tag has moved (but in general, we don't really deal with that anyway)
 * if there's no prior record of the image (no tags value in memcached), it'll fire the new image notification; if memcached has restarted, that means it'll be called for each image. This doesn't necessarily result in as many automation runs, though, since the `AskForImagePoll` will discard duplicate requests.

### DIFF
--- a/registry/warming.go
+++ b/registry/warming.go
@@ -26,6 +26,7 @@ type Warmer struct {
 	Writer        cache.Writer
 	Reader        cache.Reader
 	Burst         int
+	Priority      chan image.Name
 }
 
 type ImageCreds map[image.Name]Credentials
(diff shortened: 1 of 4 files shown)


registry/warming.go
```diff
@@ -169,6 +226,36 @@ func (w *Warmer) warm(id image.Name, creds Credentials) {
...
+		// If there's more tags than there used to be, there must be
+		// at least one new tag.
+		if len(cacheTags) < len(tags) {
+			w.Notify()
+			return
+		}
+		// Otherwise, check whether there are any entries in the
+		// fetched tags that aren't in the cached tags, ignoring any
+		// in the cached tags that aren't in fetched tags.
+		sort.Strings(cacheTags)
```
sambooo: It's not super obvious from just glancing at the code why exactly this section works. I figured it out because I have context but maybe it could do with comments in the switch statement and the last if statement.

squaremo: That is a fair point! I can't say I agonised over this but you are right, it does deserve explanation.

PR #842, opened 2017-11-16
Add NotifyChange to api.Upstream

sambooo:
Missed this in #841


squaremo:
(Note to future: because of an outage at CircleCI, this appears to have been run in CI *after* being merged, and failed because the branch was deleted. The merge commit passed CI, happily.)

PR #861, opened 2017-12-01
Supply a reason for "available images" being blank

squaremo:
When flux has not managed to fetch the image metadata, it will supply
a null in the "available images" field. But usually, we can figure out
why there's no result -- it could be because we didn't have
credentials for the registry, or perhaps because we just haven't tried
to fetch them yet.

This commit effectively adds a field to the API return value to hold
that reason, and gives a placeholder reason in the case that available
images weren't found.

More work is needed to plumb through the _actual_ reasons from the
registry. But it's probably worth waiting for #851 to be merged before
doing that.

### DIFF
--- a/cmd/fluxctl/list_images_cmd.go
+++ b/cmd/fluxctl/list_images_cmd.go
@@ -9,6 +9,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/weaveworks/flux"
+	"github.com/weaveworks/flux/registry"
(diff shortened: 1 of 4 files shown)


sambooo:
> it's probably worth waiting for #851 to be merged

anything else worth waiting for?

squaremo:
> it's probably worth waiting for #851 to be merged
>> anything else worth waiting for?

Nope; but, after this is merged, we can try to plumb through some more informative errors.

PR #865, opened 2017-12-04
Check git repo is writable before proceeding

squaremo:
Our "gitsync" mechanism relies on being able to push a tag back to the
upstream repository. But in the daemon, we consider it ready to go
once we've managed to clone the repository. This results in confusing
errors: if the repo is read-only (either because the deploy key is, or
because it's an `https` URL) then syncing will succeed in effect, but
report problems.

This commit adds a check that we can write to the upstream repo,
before considering the daemon ready to go. The check is that we can
push an otherwise-meaningless tag (then delete it) -- this being a
fairly innocuous write operation.

I have made this generate a "helpful error" (i.e., with extended help text), so that it will be returned to `fluxctl` if the daemon has not got past cloning the repo.

### DIFF
--- a/git/repo.go
+++ b/git/repo.go
@@ -148,7 +148,10 @@ func (c *Checkout) ManifestDir() string {
 func (c *Checkout) CheckOriginWritable(ctx context.Context) error {
 	c.Lock()
 	defer c.Unlock()
-	return checkPush(ctx, c.repo.KeyRing, c.Dir, c.repo.URL)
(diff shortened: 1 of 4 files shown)


git/repo.go
```diff
@@ -16,6 +16,7 @@ import (
 
 const (
 	DefaultCloneTimeout = 2 * time.Minute
+	CheckPushTag        = "flux-write-check"
```
sambooo: Does this need to be exported?

squaremo: As things stand, no, but since it's something we push to an external system, I thought it was better to expose it.

PR #866, opened 2017-12-07
Report the status of the git repo when asked

squaremo:
It's helpful to be able to tell where the git clone got up to, since
there's a bunch of ways it can fail. This commit adds a git status
field in the return value of `GitRepoConfig`.

There's a couple of other tidies:

 - If there's no git URL given, cloning can never succeed, so just go
   and wait quietly to be shut down.

 - The loop attempting to clone the repo was not interruptable, so
   typically, rolling out a daemon with updated config would have to
   wait for Kubernetes to run out of patience (30s) and kill the old
   pod. Instead, check if we've had a shutdown signal, after each
   attempt.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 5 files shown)


cmd/fluxd/main.go
```diff
@@ -389,13 +409,20 @@ func main() {
 					"notes-ref", *gitNotesRef,
 					"set-author", *gitSetAuthor)
 				checkout = working
+				break
+			}
+
+			tryAgain := time.NewTimer(10 * time.Second)
```
sambooo: I believe you're leaking timers here

squaremo: By not receiving on the channel and/or calling `Stop()`?

squaremo: I'm not sure that's true; what is leaking here? I only reassign `tryAgain` in the case that we received a value on its channel, meaning the timer has served its purpose. I don't see anything in the documentation to suggest that it needs to be explicitly `Stop`ped in order to be gc'ed.

squaremo: (of `time.Ticker`s it is true that you have to `Stop()` them to release resources)

squaremo:
Rebassed on #865 and rewritten slightly to account for the extra stage of git readiness.

PR #869, opened 2017-12-15
Separate filtering into pre- and post- phases

squaremo:
When selecting controllers to be updated, we find all those defined in
the repo, then ask the cluster for each one of those, before
subjecting any of them to filtering.

In the case of updating a single controller, this means we do a lot of
calculation that will be redundant, since we don't care about anything
but that controller. It also means any problem with some other
controller in the cluster will prevent that controller being updated.

Instead, use a prefilter to select only the controllers we're going to
ask the cluster about, *then* apply filters on the information we got
from the cluster*.

An exeption is the treatment of Locked resources -- we have this
information from the repo, but look at it _after_ everything
else. I've kept this, somewhat artificially, to be consistent for now.

*Technically we shouldn't be asking the cluster anything, just
updating the controllers defined in the repo. But one thing at a time.

### DIFF
--- a/release/context.go
+++ b/release/context.go
@@ -59,74 +59,68 @@ func (rc *ReleaseContext) WriteUpdates(updates []*update.ControllerUpdate) error
 // ---
 
 // SelectServices finds the services that exist both in the definition
-// files and the running platform.
-//
-// `ServiceFilter`s can be provided to filter the found services.
-// Be careful about the ordering of the filters. Filters that are earlier
-// in the slice will have higher priority (they are run first).
-func (rc *ReleaseContext) SelectServices(results update.Result, filters ...update.ControllerFilter) ([]*update.ControllerUpdate, error) {
-	defined, err := rc.FindDefinedServices()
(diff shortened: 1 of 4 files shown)


release/context.go
```diff
@@ -59,94 +59,88 @@ func (rc *ReleaseContext) WriteUpdates(updates []*update.ControllerUpdate) error
...
 
 // SelectServices finds the services that exist both in the definition
-// files and the running platform.
-//
-// `ServiceFilter`s can be provided to filter the found services.
-// Be careful about the ordering of the filters. Filters that are earlier
-// in the slice will have higher priority (they are run first).
-func (rc *ReleaseContext) SelectServices(results update.Result, filters ...update.ControllerFilter) ([]*update.ControllerUpdate, error) {
-	defined, err := rc.FindDefinedServices()
+// files and the running platform. `ContollerFilter`s can be provided
```
sambooo: s/Contoller/Controller/

PR #872, opened 2017-12-17
Apply all resources in a single kubectl call

sambooo:
In an attempt to make full-syncs as fast as possible, this change combines all yamels into a big multidoc and attempts to apply that with a single kubectl call.

In the event that the mega-apply fails, fallback to applying yamels one at a time.

Obligatory random cleanups included at no extra cost.


rade:
Does this make flux less tolerant to
- syntax errors in individual yamls
- failures in applying a particular yaml
?

sambooo:
Yes (to both), fallback yet to be implemented. I'm trying to figure out if there's some way to get the best of both worlds (single `kubectl apply` that doesn't stop at the first failure), but if that doesn't work we can always just attempt the mega-apply and go back to applying yamels one at a time if that fails.

squaremo:
Shout-out for the great commit messages :star:

squaremo:
Does this make syncs appreciably faster? It's one exec vs ~100, of course, but I would expect most of the latency to be from the network round-trips kubectl does on each apply.

I'm also worried that there's much more chance of running into the limitation that caused #826 (which will be built into kubectl). It'd be quite easy to accumulate 64k of manifests. Granted, it will fall back to doing things one by one.

rade:
> Does this make syncs appreciably faster? 

Some measurements would be nice.

squaremo:
Probably the best place to get measurements is by running in dev :) We can do that by `ignore`ing the fluxd deployment, applying a config to use an image built from this branch, and looking at the metrics.

cluster/kubernetes/kubernetes.go
```diff
@@ -42,19 +43,16 @@ type extendedClient struct {
...
 		Namespace string `yaml:"namespace"`
 	} `yaml:"metadata"`
 }
 
-func (obj *apiObject) namespaceOrDefault() string {
-	if obj.Metadata.Namespace == "" {
-		return "default"
-	}
-	return obj.Metadata.Namespace
+func (o *apiObject) hasDefaultNamespace() bool {
```
squaremo: I thought the problem was with manifests that did not give a namespace. If "default" is supplied, that should be fine, as with any other namespace. In other words, what you _really_ care about here is whether it's supplied or not.

cluster/kubernetes/kubernetes.go
```diff
@@ -89,55 +87,42 @@ func isAddon(obj namespacedLabeled) bool {
...
 // (Typically, this code is deployed into the same cluster.)
 type Cluster struct {
 	client     extendedClient
 	applier    Applier
-	actionc    chan func()
 	version    string // string response for the version command.
 	logger     log.Logger
 	sshKeyRing ssh.KeyRing
+
+	sync.Mutex
```
squaremo: If this isn't be to be used outside the methods of Cluster, it should be given an unexported name.

cluster/kubernetes/release.go
```diff
@@ -49,16 +58,63 @@ func (c *Kubectl) connectArgs() []string {
 	return args
 }
 
-func (c *Kubectl) kubectlCommand(args ...string) *exec.Cmd {
-	cmd := exec.Command(c.exe, append(c.connectArgs(), args...)...)
-	cmd.Stdout = c.stdout
-	cmd.Stderr = c.stderr
-	return cmd
+func (c *Kubectl) execute(logger log.Logger, errs cluster.SyncError) {
+	defer c.changeSet.clear()
```
squaremo: This implicit statefulness makes me feel queasy -- it feels like you would really want to lock the applier while you staged things, then execute; or, you'd construct a set of staged actions, and supply that to execute. I guess it doesn't matter too much if the methods aren't exported.

cluster/kubernetes/release.go
```diff
@@ -49,16 +58,63 @@ func (c *Kubectl) connectArgs() []string {
...
+		for _, obj := range s.objs {
+			r := bytes.NewReader(obj.bytes)
+			if err := c.doCommand(logger, r, s.cmd...); err != nil {
+				errs[obj.id] = err
+			}
+		}
+	}
+}
+
+type executeSet struct {
```
squaremo: Is this layered `exec -> stage -> exec` structure necessary? Can you not just keep two lists when staging, and do two execs here?

cmd/fluxd/main.go
```diff
@@ -146,7 +146,7 @@ func main() {
 	var clusterVersion string
 	var sshKeyRing ssh.KeyRing
 	var k8s cluster.Cluster
-	var image_creds func() registry.ImageCreds
+	var imageCreds func() registry.ImageCreds
```
squaremo: :100:

cluster/kubernetes/release.go
```diff
@@ -49,16 +58,63 @@ func (c *Kubectl) connectArgs() []string {
...
+		c.exec(logger, otherSet, cmd, errs)
+	}
+}
+
+func (c *Kubectl) exec(logger log.Logger, s *executeSet, cmd string, errs cluster.SyncError) {
+	if len(s.objs) == 0 {
+		return
+	}
+	// Attempt to apply everything in s at once, else fallback to applying one at a time.
+	if err := c.doCommand(logger, s.rw, s.cmd...); err != nil {
```
squaremo: I anticipate that it'd be useful to keep count of syncs that have to fall back -- either a separate counter, or as a label of the sync counter.

cluster/kubernetes/release.go
```diff
@@ -49,16 +58,63 @@ func (c *Kubectl) connectArgs() []string {
...
+
+		for _, obj := range c.objs[cmd] {
+			set := defaultSet
+			if !obj.hasDefaultNamespace() {
+				set = otherSet
+			}
+			set.stage(obj)
+		}
+
+		c.exec(logger, defaultSet, cmd, errs)
```
squaremo: If you always apply things without namespaces first, maybe you can get rid of the sorting of namespace resources to the front, in `sync/sync.go`?

squaremo: (in fact I would prefer that, because it is particular to Kubernetes, which that package is not supposed to be ..)

cluster/kubernetes/kubernetes.go
sambooo: Correctamundo. Addressed in an unpushed commit.

cluster/kubernetes/kubernetes.go
sambooo: Fixed.

cluster/kubernetes/release.go
sambooo: Removed. The sifting is now done in `changeSet.stage(...)`

cmd/fluxd/main.go
sambooo: 😄

cluster/kubernetes/release.go
sambooo: Done.

cluster/kubernetes/release.go
sambooo: Yep you're totally right here. Why did I stash this data in the Kubectl struct again? 🤷‍♂️

cluster/kubernetes/kubernetes.go
```diff
@@ -207,40 +215,39 @@ func (c *Cluster) AllControllers(namespace string) (res []cluster.Controller, er
...
 			}
 		}
-		if len(errs) > 0 {
-			errc <- errs
-		} else {
-			errc <- nil
-		}
 	}
-	return <-errc
+	c.applier.apply(logger, cs, errs)
```
squaremo: You don't need the lock until here, I believe

sambooo: Right you are

PR #882, opened 2017-12-22
Check for a .gcr.io suffix too.

### LINKED ISSUE #878: Support gcr hostnames other than gcr.io
GCR has regional hostnames (eg. `eu.gcr.io`) for a number of regions, but only `gcr.io` is hardcoded in https://github.com/weaveworks/flux/blob/5245a0cf9e9d2c8f760726fd0a4ceb3c9d6f186f/registry/credentials.go#L93

mithrandi:
Closes #878.

There doesn't seem to be unit test coverage of this code path, but I have tested the fix manually in my cluster and it works. (Open to suggestions about how I could add tests, if relevant)

### DIFF
--- a/registry/credentials.go
+++ b/registry/credentials.go
@@ -90,7 +90,7 @@ func (cs Credentials) credsFor(host string) creds {
 	if cred, found := cs.m[host]; found {
 		return cred
 	}
-	if host == "gcr.io" {
+	if host == "gcr.io" || strings.HasSuffix(host, ".gcr.io") {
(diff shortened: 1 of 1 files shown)


squaremo:
> There doesn't seem to be unit test coverage of this code path, but I have tested the fix manually in my cluster and it works.

Yeah, tricky to get test coverage on this one. We'd have to fake the environment so much, it'd be more of a test of our ability to do that, than the code itself.

PR #886, opened 2017-12-22
Copy the bytes kept in resources

### LINKED ISSUE #883: Flux doesn't like some of my manifests
I see this logged by Flux:
```
ts=2017-12-22T15:15:57.130121631Z caller=loop.go:164 component=sync-loop err="default:clusterrolebinding/prometheus: yaml: line 1: mapping values are not allowed in this context; default:prometheus/prometheus: yaml: line 2: mapping values
are not allowed in this context; default:service/prometheus-global: yaml: line 2: mapping values are not allowed in this context; default:serviceaccount/prometheus: yaml: line 1: mapping values are not allowed in this context; default:servicemonitor/prometheus-global: running kubectl: error: unable to decode \"STDIN\": Object 'Kind' is missing in '{\".rules\":\"groups:\\n- name: isaacranks.rules\\n  rules:\\n  - record: service_version:isaacranks_ballot_generation_seconds_count:irate\\n    expr: sum(irate(isaacranks_ballot_generation_seconds_cou\\n\"}'"
```

The resource definitions it complains about seem to be these:
https://github.com/mithrandi/my-kube/blob/5daf8f0755ff66f54ef4a63c8b59afea2e6904d5/manifests/prometheus/monitoring.yaml

I think possibly the other errors are caused by the last error which seems to be caused by the ConfigMap, but I'm not sure exactly where to begin debugging.

squaremo:
Scanners do not promise to leave byte buffers alone --
https://golang.org/pkg/bufio/#Scanner.Bytes -- so make a copy when
constructing resources, since they keep the bytes given to them.

Fixes #883.

### DIFF
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -59,7 +59,13 @@ func ParseMultidoc(multidoc []byte, source string) (map[string]resource.Resource
 	var obj resource.Resource
 	var err error
 	for chunks.Scan() {
-		if obj, err = unmarshalObject(source, chunks.Bytes()); err != nil {
+		// It's not guaranteed that the return value of Bytes() will not be mutated later:
+		// https://golang.org/pkg/bufio/#Scanner.Bytes
+		// But we will be snaffling it away, so make a copy.
+		bytes := chunks.Bytes()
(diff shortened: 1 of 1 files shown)


squaremo:
(merged without review, on the basis that it's small and harmless and useful)

PR #887, opened 2017-12-27
Make release output quieter

squaremo:
One doesn't usually care about skipped or ignored things, so exclude
them from output by default. Change the `--verbose` flag (shortcut
`-v`) to be a level of verbosity, s.t. `-vv` gives all output (skipped
and ignored things as well as successes and errors).

### DIFF
--- a/update/print.go
+++ b/update/print.go
@@ -8,14 +8,23 @@ import (
 	"github.com/weaveworks/flux"
 )
 
-func PrintResults(out io.Writer, results Result, verbose bool) {
+// PrintResults outputs a result set to the `io.Writer` provided, at
+// the given level of verbosity:
+//  - 2 = include skipped and ignored resources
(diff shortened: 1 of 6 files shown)


update/print.go
```diff
@@ -8,14 +8,23 @@ import (
 	"github.com/weaveworks/flux"
 )
 
-func PrintResults(out io.Writer, results Result, verbose bool) {
+// PrintResults outputs a result set to the `io.Writer` provided, at
+// the given level of verbosity:
+//  - 2 = include skipped and ignored resources
+//  - 1 = include skipped resources, exclude ignored resources
+//  - 0 = exclude skipped and ignored resources
+func PrintResults(out io.Writer, results Result, verbose int) {
```
sambooo: `verbose` should be `verbosity` for consistency with other changes

PR #892, opened 2018-01-08
Write to SSH config instead of using env

squaremo:
This commit remove the SSH key handling from the git package, instead
writing an SSH config that will be used by git+ssh.

This accomplishes two things:
 1. it decouples the git operations from SSH key handling; and,
 2. it makes using a third-party git library easier, since that won't
 need special SSH key handling either.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -188,12 +188,6 @@
   revision = "5b9ff866471762aa2ab2dced63c9fb6f53921342"
 
 [[projects]]
-  name = "github.com/kevinburke/ssh_config"
-  packages = ["."]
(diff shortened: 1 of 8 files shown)


squaremo:
I write the config to /etc/ssh/ssh_config, which makes it a global config; the rationale was that this leaves ~/.ssh/config available to specialise the deployment (e.g., by mounting a configmap). However it does mean that this cannot really be run outside of a container, since it'll stomp on that config.

squaremo:
Thanks Tamara :star:

squaremo:
> However it does mean that this cannot really be run outside of a container, since it'll stomp on that config.

Second commit: remove the bit that writes to a file (possibly stomping on system config!), and copy that file into the container image when building it, instead.

PR #897, opened 2018-01-12
Use the exact credentials collected for the image domain

### LINKED ISSUE #896: Unable to fetch image tags for Dockerhub private repo
As of flux 1.2.0, the cache warmer is failing to fetch image tags from dockerhub private repos.

Example log output:
```
ts=2018-01-11T23:30:04.828130984Z caller=warming.go:154 component=warmer err="requesting tags: errors:\ndenied: requested access to the resource is denied\nunauthorized: authentication required\n"
```

The logs should also make it clear which image we're failing to fetch tags for.

squaremo:
1. add an argument for enabling request tracing in the registry client
2. put a bit of extra logging in to report the nature of credentials being used in the registry client
3. ask for credentials using the image's name, rather than the auth domain, since those can differ (e.g., index.docker.io vs auth.docker.io).

Fixes #896.

### DIFF
--- a/registry/credentials.go
+++ b/registry/credentials.go
@@ -4,14 +4,23 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
 	"net/url"
 	"strings"
+
+	"github.com/pkg/errors"
 )
 
 // Registry Credentials
 type creds struct {
-	username, password string
+	username, password   string
+	registry, provenance string
+}
+
+func (c creds) String() string {
+	if (creds{}) == c {
+		return "<zero creds>"
+	}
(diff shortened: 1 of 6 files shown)


registry/credentials.go
```diff
@@ -4,14 +4,23 @@ import (
...
 
 // Registry Credentials
 type creds struct {
-	username, password string
+	username, password   string
+	registry, provenance string
+}
+
+func (c creds) String() string {
+	if c.username == "" && c.password == "" && c.registry == "" && c.provenance == "" {
```
sambooo: This can be `if c == creds{}` and that's better because

squaremo: Oh yeah!

squaremo:
Thanks @tamarakaufler and @sambooo

PR #898, opened 2018-01-12
Log more details on cache warming errors

### LINKED ISSUE #895: Improve logging in order to help diagnosing issues
**Problem**:
The following log statement does not say enough to be able to debug the actual issue.
```
ts=2018-01-10T21:16:55.42468337Z caller=warming.go:154 component=warmer err="requesting tags: errors:\ndenied: requested access to the resource is denied\nunauthorized: authentication required\n"
```

**Solution**:
Log details like:

- what image was Flux requesting tags for,
- from which repository,
- with which credentials,
- etc.

**Where the issue occurred**:
- Slack: https://weave-community.slack.com/archives/C4U5ATZ9S/p1515685389000781
- Full logs: https://weave-community.slack.com/files/U8S9JQP2S/F8QQX4MAL/logs-from-agent-in-weave-flux-agent-c65ccd47b-v99lh.txt

marccarre:
Fixes #895.

With this PR, Flux now logs:
```
ts=2018-01-25T11:39:40.147644595Z \
  auth="{map[index.docker.io:<registry creds for marccarrecicd@index.docker.io, from default:secret/dockerhub-secret>]}" \
  caller=warming.go:154 \
  component=warmer \
  canonical_name=index.docker.io/marccarre/go-hello-world \
  err="requesting tags: Get https://index.docker.io/v2/marccarre/go-hello-world/tags/list: unauthorized: incorrect username or password"
```

Note that password is completely masked.

Not sure if we still want this PR though, given [this comment](https://github.com/weaveworks/flux/issues/895#issuecomment-357019949) and #897.

### DIFF
--- a/registry/cache/warming.go
+++ b/registry/cache/warming.go
@@ -111,9 +111,10 @@ func imageCredsToBacklog(imageCreds registry.ImageCreds) []backlogItem {
 }
 
 func (w *Warmer) warm(ctx context.Context, logger log.Logger, id image.Name, creds registry.Credentials) {
+	errorLogger := log.With(logger, "canonical_name", id.CanonicalName(), "auth", creds)
 	client, err := w.clientFactory.ClientFor(id.CanonicalName(), creds)
 	if err != nil {
-		logger.Log("err", err.Error())
+		errorLogger.Log("err", err.Error())
 		return
 	}
 
@@ -128,10 +129,9 @@ func (w *Warmer) warm(ctx context.Context, logger log.Logger, id image.Name, cre
 	}
 
 	if err != nil {
-		logger.Log("err", errors.Wrap(err, "fetching previous result from cache"))
+		errorLogger.Log("err", errors.Wrap(err, "fetching previous result from cache"))
 		return
 	}
-
 	// Save for comparison later
 	oldImages := repo.Images
 
@@ -144,14 +144,14 @@ func (w *Warmer) warm(ctx context.Context, logger log.Logger, id image.Name, cre
 			err = w.cache.SetKey(repoKey, bytes)
 		}
 		if err != nil {
-			logger.Log("err", errors.Wrap(err, "writing result to cache"))
+			errorLogger.Log("err", errors.Wrap(err, "writing result to cache"))
 		}
 	}()
 
 	tags, err := client.Tags(ctx)
 	if err != nil {
 		if !strings.Contains(err.Error(), context.DeadlineExceeded.Error()) && !strings.Contains(err.Error(), "net/http: request canceled") {
-			logger.Log("err", errors.Wrap(err, "requesting tags"))
+			errorLogger.Log("err", errors.Wrap(err, "requesting tags"))
 			repo.LastError = err.Error()
 		}
 		return
@@ -212,7 +212,7 @@ func (w *Warmer) warm(ctx context.Context, logger log.Logger, id image.Name, cre
 						// This was due to a context timeout, don't bother logging
 						return
 					}
-					logger.Log("err", errors.Wrap(err, "requesting manifests"))
+					errorLogger.Log("err", errors.Wrap(err, "requesting manifests"))
 					return
 				}
 
@@ -220,12 +220,12 @@ func (w *Warmer) warm(ctx context.Context, logger log.Logger, id image.Name, cre
 				// Write back to memcached
 				val, err := json.Marshal(img)
 				if err != nil {
-					logger.Log("err", errors.Wrap(err, "serializing tag to store in cache"))
+					errorLogger.Log("err", errors.Wrap(err, "serializing tag to store in cache"))
 					return
 				}
 				err = w.cache.SetKey(key, val)
 				if err != nil {
-					logger.Log("err", errors.Wrap(err, "storing manifests in cache"))
+					errorLogger.Log("err", errors.Wrap(err, "storing manifests in cache"))
 					return
 				}
 				successMx.Lock()
--- a/registry/cache/warming.go
+++ b/registry/cache/warming.go
@@ -46,8 +46,8 @@ type backlogItem struct {
 	registry.Credentials
 }
 
-// Continuously get the images to populate the cache with, and
-// populate the cache with them.
+// Loop continuously gets the images to populate the cache with,
+// and populate the cache with them.
 func (w *Warmer) Loop(logger log.Logger, stop <-chan struct{}, wg *sync.WaitGroup, imagesToFetchFunc func() registry.ImageCreds) {
 	defer wg.Done()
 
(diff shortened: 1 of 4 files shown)


squaremo:
#897 puts the logging behind a `--registry-trace` flag, and I think there's still value in also logging when there's been an error.

However, I'd request that it just omits passwords entirely -- I think it's more important to say where the credentials came from (`struct cred` now has a `String` method which summarises that) so people can investigate for themselves. Emitting any part of a password (as well as the length) is not sound.

registry/cache/warming.go
```diff
@@ -46,8 +46,8 @@ type backlogItem struct {
 	registry.Credentials
 }
 
-// Continuously get the images to populate the cache with, and
-// populate the cache with them.
+// Loop continuously get the images to populate the cache with,
```
squaremo: Better fix the verbs before the grammar police spot this ..

registry/cache/warming.go
```diff
@@ -151,7 +150,7 @@ func (w *Warmer) warm(ctx context.Context, logger log.Logger, id image.Name, cre
 	tags, err := client.Tags(ctx)
 	if err != nil {
 		if !strings.Contains(err.Error(), context.DeadlineExceeded.Error()) && !strings.Contains(err.Error(), "net/http: request canceled") {
-			logger.Log("err", errors.Wrap(err, "requesting tags"))
+			logger.Log("err", errors.Wrapf(err, "requesting tags for %v from %v using %v", id.CanonicalName(), creds.Hosts(), creds))
```
squaremo: Since we're using structured logging (e.g., the pair of `"err"` and the error), it's better to include the image name and credentials used as fields of the log message rather than as part of the error.

marccarre:
Thanks a lot for the feedback @squaremo, I'll rebase this branch on #897's one and make the suggested changes :+1:

registry/cache/warming.go
marccarre: I was thinking we could also have:
```go
logger = log.With(logger, "canonical_name", id.CanonicalName(), "creds", creds)
```
at the top of the method so that these details are added for all log statements.
Would this be preferable?

registry/cache/warming.go
marccarre: Sorry. Fixed with `Loop continuously gets` (as I guess that's what you were referring to).

registry/cache/warming.go
squaremo: That's a great idea

marccarre: OK. Done. And [example statement updated](https://github.com/weaveworks/flux/pull/898#issue-288221611).

marccarre:
Also note that we now log additional details even when there is no error, e.g. when the registry is made public and the credentials are removed from the manifest:
```
ts=2018-01-25T11:58:41.349125409Z \
  caller=warming.go:190 \
  component=warmer \
  canonical_name=index.docker.io/marccarre/go-hello-world \
  creds={map[]} \
  fetching=marccarre/go-hello-world \
  total=2 \
  expired=2 \
  missing=0

ts=2018-01-25T11:58:41.481967446Z \
  caller=warming.go:238 \
  component=warmer \
  canonical_name=index.docker.io/marccarre/go-hello-world \
  creds={map[]} \
  updated=marccarre/go-hello-world \
  count=2
```

registry/credentials.go
```diff
@@ -123,3 +123,7 @@ func (cs Credentials) Merge(c Credentials) {
 		cs.m[k] = v
 	}
 }
+
+func (cs Credentials) String() string {
+	return fmt.Sprintf("{%v}", cs.m)
```
squaremo: What will this output if there's some credentials present?

marccarre: Something like:
```
"{map[localhost:5000:<registry creds for testuser@localhost:5000, from test>]}"
```
I've added a test to ensure this is the case, see also: `registry/credentials_test.go#TestStringShouldNotLeakPasswords`.
If you remove this implementation of `Credentials#String`, and let Go use the default one, you end up with something which does print the password:
```
"{map[localhost:5000:{testuser testpassword localhost:5000 test}]}"
```

squaremo: That does end up being rather chatty:

```
ts=2018-01-25T17:35:40.862006068Z caller=warming.go:117 \
 component=warmer \
 canonical_name=kube-registry.kube-system.svc.cluster.local:5000/mikeb/hello \
 creds="{map[kube-registry.kube-system.svc.cluster.local:5000:<registry creds for mikeb@kube-registry.kube-system.svc.cluster.local:5000, from hello:secret/kube-registry-docker-secret>]}" \
 err="Get https://kube-registry.kube-system.svc.cluster.local:5000/v2/: http: server gave HTTP response to HTTPS client"
```

(Note in particular the repetition of the image domain x 3)

marccarre: Well, since the original goal of this PR was to help investigate the issue that #897 fixed, I wanted it to be chatty to better see what was happening. 

Also, aren't the domains from the image and the credentials unrelated? (as in, coming from different sources?)

What would you rather have?

squaremo: >  I wanted it to be chatty to better see what was happening.

Yeah fair enough. Having full info in a trace is good for that; otherwise I think we'll mostly care only if there's a problem. But it's always a fine balance isn't it.

> Also, aren't the domains from the image and the credentials unrelated?

They are exactly the same. The key of a map entry is the domain given in the docker-config secret, which is also given to the `cred{}` value to report as its domain; the `cred{}` value to use for an image is picked by determining an image's canonical domain and indexing into the map at that key.

The problem fixed by #897 was that the domain of the authentication URL, rather than the domain of the canonical name of the image, was being used to determine which credentials to use. I noticed that was the case by observing that the value was `<zero creds>` rather than creds from the secret.

I think it'd be fine to log the map on errors or on entry to the method, and just use the creds specific to the image otherwise (i.e., in the client itself). The factory is an interstitial structure -- not much interesting goes on there anyway.

marccarre: Thanks a lot for the explanations and suggestions @squaremo.

> But it's always a fine balance isn't it.

Indeed.

> They are exactly the same.

OK, I'll drop `"canonical_name", id.CanonicalName()` then.

> log the map on errors

OK, I'll do that, i.e. if I understood well:
```go
logger.Log("auth", creds, "err", err)
```
> log the map [...] on entry to the method

I believe we [do almost this already](https://github.com/weaveworks/flux/blob/master/registry/client_factory.go#L79-L81) post #897, in the factory, when creating the client, though only when traces are enabled and only with the creds for the current domain, but that's probably enough?

> just use the creds specific to the image otherwise (i.e., in the client itself)

I guess that's covered by the previous point?

squaremo: I think where I've ended up is that your "fallback" suggestion -- log the canonical name, and creds, but only on error -- is the right thing. Sorry for the run-around!

squaremo: NB the test for not leaking passwords is certainly worth keeping whatever else happens.

marccarre: > log the canonical name, and creds, but only on error

OK, adding the canonical name back then :+1:

PR #900, opened 2018-01-15
Log on startup

sambooo:
This log message should be printed almost instantly after fluxd starts, allowing
us to debug slow startup issues.


cmd/fluxd/main.go
```diff
@@ -128,6 +128,7 @@ func main() {
 		logger = log.With(logger, "ts", log.DefaultTimestampUTC)
 		logger = log.With(logger, "caller", log.DefaultCaller)
 	}
+	logger.Log("started", true)
```
squaremo: Structured logging, it's the way of the future ..

sambooo: *whispers* logrus

rade:
I did suggest it should log the version. Oh well.

PR #911, opened 2018-01-24
Unify remote.Platform and api.Server interfaces

sambooo:


http/transport.go
```diff
@@ -32,13 +32,11 @@ func NewAPIRouter() *mux.Router {
 	r.NewRoute().Name("ListServices").Methods("GET").Path("/v6/services").Queries("namespace", "{namespace}") // optional namespace!
 	r.NewRoute().Name("ListImages").Methods("GET").Path("/v6/images").Queries("service", "{service}")
 
-	r.NewRoute().Name("UpdateImages").Methods("POST").Path("/v6/update-images").Queries("service", "{service}", "image", "{image}", "kind", "{kind}")
```
squaremo: We'll have to take care to reinstate these routes in the service code, since our contract there is different*.

*Here we expect people to download a `fluxctl` compatible with their daemon, though perhaps we shouldn't. In the service, we have to support `fluxctls` in a range of versions -- so while we can modify the api interfaces post-hoc, we oughtn't remove routes that will still be called.

sambooo: Fortunately if this route doesn't exist after revendoring in the service code it'll panic when building the handler :)

deploy/flux-deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
         # - --memcached-service=memcached
 
         # replace (at least) the following URL
-        - --git-url=git@github.com:weaveworks/flux-example
+        - --git-url=git@github.com:sambooo/flux-secret-test
```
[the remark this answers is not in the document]
sambooo: Yar

api/api.go
```diff
@@ -23,11 +23,3 @@ type Client interface {
 	Export(context.Context) ([]byte, error)
 	PublicSSHKey(ctx context.Context, regenerate bool) (ssh.PublicKey, error)
 }
-
-// API for daemons connecting to an upstream service
-type Upstream interface {
```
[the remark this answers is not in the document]
sambooo: The methods belong somewhere in the service because iirc there's neither a reference to nor an implementation of this interface anywhere in this repo.

cmd/fluxctl/release_cmd.go
```diff
@@ -124,16 +124,21 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
...
+	}
+	jobID, err := opts.API.UpdateManifests(ctx, update.Spec{
+		Type:  update.Images,
+		Cause: opts.cause,
+		Spec:  spec,
+	})
 	if err != nil {
 		return err
 	}
+	fmt.Fprintln(cmd.OutOrStderr(), "jobID:", jobID)
```
sambooo: Oops

http/transport.go
squaremo: Panic or fail to compile?

squaremo: Yes it'll probably panic. We'd have to change this around a fair bit to induce a compilation error (types, what are those).

Perhaps ... perhaps the router definitions should be versioned as well? We've opened a can of worms here ..

sambooo: We should probably version everything about the API sooner rather than later.

I'm not sure we can introduce a compile error at all, how would you go about it?

http/client/client.go
```diff
@@ -101,19 +80,24 @@ func (c *Client) Export(ctx context.Context) ([]byte, error) {
...
 
+func (c *Client) Ping(context.Context) error {
+	return errNotImplemented
+}
+
+func (c *Client) Version(context.Context) (string, error) {
+	return "", errNotImplemented
+}
+
+func (c *Client) NotifyChange(context.Context, api.Change) error {
```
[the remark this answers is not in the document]
sambooo: `errNotImplemented` as in _oh this? err... not implemented_

http/transport.go
squaremo: I figured out why this bothers me -- people shouldn't need to download a new `fluxctl` after we've refactored. I suggest we reinstate the `UpdateImages` and `UpdatePolicy` endpoints, and implement them in terms of `UpdateManifests`.

NB they don't need to appear in the `Server` interface, since old `fluxctl` will have been built with the old interface, and new `fluxctl` will just use `UpdateManifests`.

PR #918, opened 2018-01-25
Support basic authentication for registries

### LINKED ISSUE #915: Try flux against an image registry that uses basic auth
It's possible to deploy your own image registry and secure it with basic auth. I _think_ that flux assumes you are using token-based auth, and will fail to fetch image metadata from a registry using basic auth. So we should try this out by running a local registry and turning `--registry-trace` on to see what happens.

squaremo:
The image registry code assumed
 - all image registries use HTTPS
 - all image registries use token authentication

..but if you run your own registry in your cluster, neither of these things is likely to be true.

To support basic authentication, all we need to do is add a handler in when constructing the registry client; it will get used if the authentication challenge indicates so.

Supporting HTTP is (oddly) a bit trickier, since there's no indication from an image name (which is all we have) whether the registry uses HTTP or HTTPS. We want to use HTTPS in general, so make the user tell us which hosts to use HTTP for (so that it's at least possible to use it), and otherwise use HTTPS.

Fixes #915.

EDIT: describe alternate solution, from comment below, to HTTP vs HTTPS.

### DIFF
--- a/registry/client_factory.go
+++ b/registry/client_factory.go
@@ -15,11 +15,13 @@ import (
 )
 
 type RemoteClientFactory struct {
-	Logger           log.Logger
-	Limiters         *middleware.RateLimiters
-	Trace            bool
+	Logger        log.Logger
+	Limiters      *middleware.RateLimiters
+	Trace         bool
+	InsecureHosts []string
+
+	mu               sync.Mutex
 	challengeManager challenge.Manager
-	mx               sync.Mutex
 }
 
 type logging struct {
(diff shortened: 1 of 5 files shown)


squaremo:
One caveat: since we try HTTP first (in the expectation that we'll be redirected if HTTPS is required), if a registry accepts HTTP _as well as_ HTTPS we might end up sending auth headers over an insecure connection. All of {`index.docker.io`, `quay.io`, `gcr.io`} require HTTPS, but I it's possible that somebody could set up a private registry that's not in their cluster and doesn't use HTTPS. Suggestions welcome!

squaremo:
> One caveat: since we try HTTP first

An alternative would be to require people using insecure registries to tell us about them explicitly, e.g., as a command-line argument something like

```
--registry-insecure-host=kube-registry.kube-system.svc.cluster.local
```

registry/client_factory.go
```diff
@@ -43,47 +45,66 @@ func (f *RemoteClientFactory) ClientFor(repo image.CanonicalName, creds Credenti
...
+	f.mu.Lock()
 	if f.challengeManager == nil {
 		f.challengeManager = challenge.NewSimpleManager()
 	}
-	f.mx.Unlock()
 	manager := f.challengeManager
+	f.mu.Unlock()
+
+	scheme := "https"
+	for _, h := range f.InsecureHosts {
```
sambooo: I can't not comment about this not being a map/set lookup, but it's incredibly unlikely to matter...

sambooo:
Capturing thoughts:
In general we would benefit from integration tests for this sort of thing, but I don't think it's necessary to require it here. Perhaps we should capture "integration test all the things" in its own issue since it's likely not a small task.
For now let's leave the bar at "I ran it a few times locally and it seemed to work and it also looks sane"

registry/client_factory.go
squaremo: Right -- I wouldn't expect enough entries here for O(n) to matter.

PR #921, opened 2018-01-26
[WIP] Derot integration tests

squaremo:

PR #922, opened 2018-01-30
[WIP] Add support for kubernetes yaml List objects

jpellizzari:


jpellizzari:
@rade 

> The baseObject->BaseObject change really should be in a separate commit.

Why?

rade:
Because it's a distinct change from the others. Furthermore it is scattered across the code base and doesn't alter behaviour.

rade:
What's left to do here?

jpellizzari:
updating annotations and make sure the write to file is working correctly

cluster/kubernetes/resource/resource.go
```diff
@@ -124,6 +154,37 @@ func unmarshalKind(base baseObject, bytes []byte) (resource.Resource, error) {
...
+
+	if err != nil {
+		return err
+	}
+
+	for _, i := range list.Items {
+		// Re-marshal this snippet. We need this item represented in byte form.
+		// This will eventually be used to generate an update here:
+		// https://github.com/weaveworks/flux/blob/master/cluster/sync.go#L9
+		itemBytes, _ := yaml.Marshal(i)
```
jpellizzari: catch this error

cluster/kubernetes/update.go
```diff
@@ -34,6 +37,20 @@ func updatePodController(def []byte, container string, newImageID image.Ref) ([]
...
+	var containers []resource.Container
+
+	for _, item := range l.Items {
+		containers = append(containers, getContainersFromManifest(item)...)
+	}
+	return containers
+}
+
+func getContainersFromManifest(manifest resource.BaseObject) []resource.Container {
+	// return append(manifest.Spec.Template.Spec.Containers, manifest.Spec.JobTemplate.Spec.Template.Spec.Containers...)
```
jpellizzari: remove

squaremo:
Superficial observation: it'd be good to have more informative commit messages. The summaries are fine, but there's no explanation of what motivates the particular code changes. I am thinking of things like "baseObject is must now be exported, because ....".

Second superficial observation: there's a stray commit sandwiched in the middle. `git rebase -i` to elide it.

sambooo:
I'm having a hard time reviewing this due to the significant number of changes in the first commit. 

Could you separate out the pre-requisite changes from the actual addition of List support? It'd be much easier to:
1. Review some (potentially significant) code changes that don't add or remove Flux behaviour, then
2. Review a minimal change which specifically only adds List object support.

Elaborating in the pre-requisite commit messages will help us understand how the changes tie together. See #911 for what I think is a good example* of doing this.

_*no bias, obviously_

jpellizzari:
@sambooo should be more manageable now. PTAL.

test/flux-deploy-all.yaml
```diff
@@ -84,12 +84,9 @@ spec:
         - name: git-key
           mountPath: /etc/fluxd/ssh
         - name: ssh-known-hosts
-          mountPath: /root/.ssh
+          mountPath: /root/.ssh/known_hosts
+          subPath: known_hosts
```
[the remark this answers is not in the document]
squaremo: It's already in #921, but I haven't finished that. I don't think this commit belongs here in any case.

squaremo:
One thing I am rather unsure about is exporting `baseObject` -- used to get common fields and methods in resources, via embedding -- as a replacement for `Manifest`, though I see why the temptation is there. (In particular I don't like how it requires all the extra `Spec` guff in baseObject, which is superfluous for its other use). Maybe it's possible to push the bit of parsing you need for updating annotations down into `kubernetes/resource`.

But if it works, it works. The last thing I'd want is for anyone to have to spend more time with regular expressions. Imma try it out.

squaremo:
I added this in a file to my git repo:

```
apiVersion: v1
kind: List
items:
- apiVersion: extensions/v1beta1
  kind: Deployment
  metadata:
    name: olleh
    namespace: hello
  spec:
    minReadySeconds: 1
    replicas: 5
    template:
      metadata:
        labels:
          name: ollehworld
      spec:
        containers:
        - name: ollehworld
          image: quay.io/weaveworks/helloworld:master-a000004
          args:
            - -msg=OLLEH
          ports:
          - containerPort: 80
        - name: sidecar
          image: quay.io/weaveworks/sidecar:master-a000002
          args:
          - -addr=:8080
          ports:
          - containerPort: 8080
- apiVersion: v1
  kind: Service
  metadata:
    name: olleh
    namespace: hello
  spec:
    ports:
      - port: 80
    selector:
      name: olleh
```

 - fluxd synced the two resources happily :white_check_mark: 
 - I could do `flux release --namespace hello --controller deployment/olleh ...` to update the image in the deployment :white_check_mark: 
 - but, `fluxctl automate --namespace hello --controller deployment/olleh` failed with `Error: Could not update resource annotations` :red_circle:

jpellizzari:
@squaremo Thanks for giving it a go. Probably a problem with the regex stuff (☹️). @sambooo Any further feedback on the code before I do another iteration?

sambooo:
@squaremo do we have an ETA on shelling out yaml-changing to python? Given the (relatively) low turnaround rate both on changes and reviews here, it might be a better idea to just not add the complexity only to remove it shortly after.

If it's a while away then let's continue as we are.

jpellizzari:
@squaremo Your YAML is invalid. The `items` list should be indented.

jpellizzari:
@sambooo I did my best to de-dupe the regular expressions. Unfortunately, the metadata REs are different enough to make de-duping them difficult, since we need to capture different things in the different situations. I will fixup https://github.com/weaveworks/flux/pull/922/commits/84de103b218added2af9219347e594114d540d48 if you think it is more understandable.

squaremo:
> Your YAML is invalid. The items list should be indented (or at least the regular expressions assume that).

It's valid YAML (as evidenced by it being synced successfully). I adapted it from [here](https://github.com/kubernetes/kubernetes/blob/master/hack/testdata/list.yaml).

sambooo:
Does this PR support yamels that look like @squaremo's yet? We should avoid forcing users to change their yaml just so that flux supports it.

PR #927, opened 2018-02-03
Enable pprof

### LINKED ISSUE #907: Consider exposing the pprof HTTP endpoints for in-situ profiling
It would be nice to be able to answer performance issues seen in the wild by "please run kubectl port-forward + go tool pprof $url" to obtain a CPU profile we can examine.

stefanprodan:
Fix #907

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -26,6 +26,7 @@ _testmain.go
 # Tool related configurations
 .ackrc
 .envrc
+.idea
```
sambooo: This should go in your `~/.gitignore` instead of in this project. The above files probably shouldn't even be in here.

PR #928, opened 2018-02-03
Fix cache warming rapid loop

### LINKED ISSUE #926: Flux using >50% of one CPU
Noticing multiple `kubectl apply` every second.

This occurs with v1.2.0 and v1.2.1.

stefanprodan:
Fix #926

Tested with `stefanprodan/flux:cpufix-090b064-wip3` on GKE:

screen shot 2018-02-03 at 16 46 05

### DIFF
--- a/registry/cache/warming.go
+++ b/registry/cache/warming.go
@@ -60,6 +60,18 @@ func (w *Warmer) Loop(logger log.Logger, stop <-chan struct{}, wg *sync.WaitGrou
 	// just rattle through them one by one, however long they take.
 	ctx := context.Background()
 
+	// NB the implicit contract here is that the prioritised
+	// image has to have been running the last time we
+	// requested the credentials.
+	priorityWarm := func (name image.Name) {
+		logger.Log("priority", name.String())
+		if creds, ok := imageCreds[name]; ok {
+			w.warm(ctx, logger, name, creds)
+		} else {
+			logger.Log("priority", name.String(), "err", "no creds available")
+		}
+	}
+
 	// This loop acts keeps a kind of priority queue, whereby image
 	// names coming in on the `Priority` channel are looked up first.
 	// If there are none, images used in the cluster are refreshed;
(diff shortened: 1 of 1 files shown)


rade:
FWIW, the spin loop was introduced in #835.

registry/cache/warming.go
```diff
@@ -91,10 +91,18 @@ func (w *Warmer) Loop(logger log.Logger, stop <-chan struct{}, wg *sync.WaitGrou
 			w.warm(ctx, logger, im.Name, im.Credentials)
 		} else {
 			select {
+			case <-stop:
+				logger.Log("stopping", "true")
```
rade: there's a `return` missing here.

registry/cache/warming.go
```diff
@@ -91,10 +91,18 @@ func (w *Warmer) Loop(logger log.Logger, stop <-chan struct{}, wg *sync.WaitGrou
...
 				imageCreds = imagesToFetchFunc()
 				backlog = imageCredsToBacklog(imageCreds)
-			default:
+			case name := <-w.Priority:
+				logger.Log("priority", name.String())
+				if creds, ok := imageCreds[name]; ok {
+					w.warm(ctx, logger, name, creds)
+				} else {
+					logger.Log("priority", name.String(), "err", "no creds available")
+				}
```
rade: I'd be tempted to lift these six lines and their copies (plus comment) in the earlier `select` into a
```golang
// ...
priorityWarm := func (name string) {
  ...
}
```
before the loop.

rade:
Why is this not consuming an entire core all the time? Is that because every `askForNewImagesInterval` (1min) it does some real work, which involves i/o and hence releases the CPU?

stefanprodan:
Yes every time the refresh timer hits it does some work for a couple of seconds. The rest of the time just sits in the default loop.

screen shot 2018-02-03 at 20 10 45

squaremo:
Looks good in sum @stefanprodan, thank you. Thanks @rade for helping knock it into shape.

Please rebase into a single commit -- or two, if you want to do the factoring of `priorityWarm` separately -- and write a short commit message explaining what the problem was and how this fixes it.

stefanprodan:
@squaremo I've squashed the commits into one. 
@rade thanks for guiding me through this.

stefanprodan:
@squaremo let me know if it's ok for me to merge this PR.

squaremo:
> I've squashed the commits into one.

Thank you. I am a bit baffled by the commit message though -- did you really add those things? (I don't think you did.) Something like this would describe what's being fixed:

> To implement the loop as described in the comment above it, when the backlog is empty we need to block until the refresh tick fires (or a priority image comes in). By having a default case, it's possible to spin.

(the commit summary can be a bit shorter too -- just "Avoid spin in cache refresh loop" would be fine IMO).

PR #936, opened 2018-02-07
AskForSync only when GitChange matches what we have

sambooo:
This requires that the upstream tell Flux both the repo and branch that got updated so that we can avoid constantly syncing on events that will have no effect.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -342,14 +342,19 @@ func (d *Daemon) release(spec update.Spec, c release.Changes) DaemonJobFunc {
 func (d *Daemon) NotifyChange(ctx context.Context, change api.Change) error {
 	switch change.Kind {
 	case api.GitChange:
-		// TODO: check if it's actually our repo
+		gitUpdate := change.Source.(api.GitUpdate)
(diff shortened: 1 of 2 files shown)


daemon/daemon.go
```diff
@@ -342,14 +342,19 @@ func (d *Daemon) release(spec update.Spec, c release.Changes) DaemonJobFunc {
 func (d *Daemon) NotifyChange(ctx context.Context, change api.Change) error {
 	switch change.Kind {
 	case api.GitChange:
-		// TODO: check if it's actually our repo
+		gitUpdate := change.Source.(api.GitUpdate)
+		if gitUpdate.URL != d.Repo.URL && gitUpdate.Branch != d.Repo.Branch {
+			// It isn't strictly an _error_ to be notified about a repo/branch pair
+			// that isn't ours, but it's worth logging anyway for debugging.
+			d.Logger.Log("msg", "notified about unrelated change",
+				"sshurl", gitUpdate.URL,
```
squaremo: Why `"sshurl"` and not just `"url"`?

sambooo: No good reason

PR #938, opened 2018-02-09
Move Upstream interface out of api.Server

sambooo:
Some methods on api.Server (Ping, Version, NotifyChange) only make sense in the
case that a Flux is connected to an Upstream (Weave Cloud). The result of this
was that the HTTP client ended up with 'not implemented' implementations of
methods which are out of place.

Since api.UpstreamServer is a superset of api.Server and struct embedding
exists, there luckily isn't a huge amount of change the the boilerplatey stuff.

Type assertions of the form `var _ api.Server = Foo{}` have been updated to make
sure things that should implement UpstreamServer do so.

### DIFF
--- a/http/daemon/server.go
+++ b/http/daemon/server.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/weaveworks/flux/api"
 	transport "github.com/weaveworks/flux/http"
-	"github.com/weaveworks/flux/image"
 	"github.com/weaveworks/flux/job"
 	fluxmetrics "github.com/weaveworks/flux/metrics"
 	"github.com/weaveworks/flux/policy"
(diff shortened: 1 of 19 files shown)


sambooo:
Note that this distinction already existed before the merging of #911. I combined a little _too_ much in that PR which I didn't realise until revendoring in service.

sambooo:
Also note that the first commit overlaps #936, so I'll have to fix whichever is merged second.

http/daemon/server.go
```diff
@@ -32,9 +31,6 @@ var (
 func NewRouter() *mux.Router {
 	r := transport.NewAPIRouter()
 
-	r.NewRoute().Methods("POST").Name("GitPushHook").Path("/v9/hook/git").Queries("repo", "{repo}")
```
squaremo: These are useful! At least, the git one -- you can provoke a sync by `curl`ing it, and do so from e.g., a git hook.

PR #940, opened 2018-02-13
Move repo syncing into repo

squaremo:
To ease the transition into supporting zero git repos, and later supporting more than one repo, this PR moves the periodic `git fetch` loop into the git package.

The effect is to make a git repo a state machine; it starts uncloned, then is mirrored, and finally tested to see if it's writable. Treating it as a state machine has a couple of benefits:

 1. We will be better able to cope with intermediate states -- for example, to give results that don't need write access when we don't have write access. This will matter the more so when we're dealing with more than one repo.
 2. It decouples the git repo polling from the daemon's "main loop", so not as many things are rolled up in the one place.

This PR doesn't change how the daemon works, in the main -- it will still wait for the repo to be ready before proceeding. To support zero git repos, we can remove that gate and respond to each API method by checking whether there's a git repo ready or not, and acting accordingly.

Tasting notes:
 * there's still a need for the git mirror to tell the daemon that there are fresh commits; and for the daemon to tell the git mirror that it has pushed something upstream. These are accomplished with a non-blocking channel send in each direction.
 * there's also a synchronous `*Repo.Refresh(...)` for when the daemon needs to wait for the repo to be up to date before proceeding.
 * the `Repo` is a mirror, which means it doesn't have a working directory to look around in; the downside is you have to clone it every time you want to look at the files (a cache could be done in the daemon loop, if necessary). The upside (aside from some convenience in the git operations) is that there's no confusion over which things you can write to -- if you want to make a commit, you have to clone first.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -19,8 +19,6 @@ import (
 	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
-	"context"
-
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/cluster"
 	"github.com/weaveworks/flux/cluster/kubernetes"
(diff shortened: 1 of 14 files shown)


squaremo:
Ops, accidental review requests. My apologues.

PR #945, opened 2018-02-15
CLI help cleanup

rndstr:
This PR cleans up the CLI help text to use the new namespace:kind/name format. It also uppercases `release` help descriptions for consistency. Additionally, it fixes the type of the `ReleaseKindExecute` constant.

### DIFF
--- a/cmd/fluxctl/automate_cmd.go
+++ b/cmd/fluxctl/automate_cmd.go
@@ -26,7 +26,7 @@ func (opts *controllerAutomateOpts) Command() *cobra.Command {
 		Use:   "automate",
 		Short: "Turn on automatic deployment for a controller.",
 		Example: makeExample(
-			"fluxctl automate --controller=deployment/helloworld",
+			"fluxctl automate --controller=default:deployment/helloworld",
 		),
 		RunE: opts.RunE,
 	}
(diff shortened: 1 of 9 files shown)


cmd/fluxctl/list_images_cmd.go
```diff
@@ -31,7 +31,7 @@ func (opts *controllerShowOpts) Command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "list-images",
 		Short:   "Show the deployed and available images for a controller.",
-		Example: makeExample("fluxctl list-images --namespace default --controller=deployment/foo"),
+		Example: makeExample("fluxctl list-images --namespace default --controller=default:deployment/foo"),
```
squaremo: The idea in general is that you provide `--namespace` to give the namespace, and `--controller` to name the resource within that namespace, to be a bit more like kubectl.

The `namespace:kind/name` syntax is mainly because we refer to things elsewhere in that way -- perhaps we shouldn't -- and it's convenient. In any case, examples ought not use both the namespace argument and the namespace-specifying syntax.

squaremo: (How this works may need rethinking; nonetheless, it's how it works :)

rndstr: Thanks, I overlooked these namespace args, was replacing blindly.

PR #947, opened 2018-02-16
Flux helm operator alpha

tamarakaufler:


squaremo:
```
$ helm-operator --help
DESCRIPTION
  helm-operator is a Kubernetes operator for Helm integration into flux.

FLAGS
      --charts-sync-interval duration         Interval at which to check for changed charts (default 3m0s)
      --charts-sync-timeout duration          Timeout when checking for changed charts (default 1m0s)
      --connect string                        Connect to an upstream service e.g., Weave Cloud, at this base address
      --event-handler-workers uint            Number of workers processing events for Flux-Helm custom resources (default 2)
      --git-branch string                     branch of git repo to use for Kubernetes manifests (default "master")
      --git-charts-path string                path within git repo to locate Helm Charts (relative path) (default "charts")
      --git-url string                        URL of git repo with Kubernetes manifests; e.g., git@github.com:weaveworks/flux-example
      --k8s-secret-data-key string            Data key holding the private SSH key within the k8s secret (default "identity")
      --k8s-secret-name string                Name of the k8s secret used to store the private SSH key (default "flux-git-deploy")
      --k8s-secret-volume-mount-path string   Mount location of the k8s secret storing the private SSH key (default "/etc/fluxd/ssh")
      --kubeconfig string                     Path to a kubeconfig. Only required if out-of-cluster.
      --kubernetes-kubectl string             Optional, explicit path to kubectl tool
      --master string                         The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.
      --queue-worker-count int                Number of workers to process queue with Chart release jobs. Two by default (default 2)
      --ssh-keygen-bitsintegrations/ uint     -b argument to ssh-keygen (default unspecified)
      --ssh-keygen-type string                -t argument to ssh-keygen (default unspecified)
      --tiller-ip string                      Tiller IP address. Only required if out-of-cluster.
      --tiller-namespace string               Tiller namespace. If not provided, the default is kube-system. (default "kube-system")
      --tiller-port string                    Tiller port. Only required if out-of-cluster.
      --token string                          Authentication token for upstream service
```

What are these for:

 * `--connect`, `--token` (we don't connect upstream or post events from here, so far as I know?)
 * `--master` (is it genuinely possible to run this outside of a cluster? have you needed to override this?)
 * `--ssh-keygen-bitsintegrations/` (suspected typo; also, are we generating SSH keys, how does that work with fluxd also doing so)
 * `--tiller-*` -- can these be combined into one hostname
 * `--git-branch string                     branch of git repo to use for Kubernetes manifests (default "master") -- are we looking at Kubernetes manifests, or this help message copypasta?

deploy/helm-operator-deployment.yaml
```diff
@@ -0,0 +1,54 @@
+---
+  apiVersion: apiextensions.k8s.io/v1beta1
+  kind: CustomResourceDefinition
+  metadata:
+    name: fluxhelmresources.integrations.flux
+  spec:
+    group: integrations.flux
```
squaremo: The advice I have seen for naming this suggests including the FQDN of one's organisation, as well as something representing the group itself. So, I suggest `integrations.flux.weave.works`, or if you want to be really specific, `helm.integrations.flux.weave.works`.

deploy/helm-operator-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+    names:
+      kind: FluxHelmResource
+      listKind: FluxHelmResourceList
+      plural: fluxhelmresources
+    scope: Namespaced
+    version: v1
```
squaremo: `v1alpha` at this point, I think.

deploy/helm-operator-deployment.yaml
```diff
@@ -0,0 +1,54 @@
+---
+  apiVersion: apiextensions.k8s.io/v1beta1
+  kind: CustomResourceDefinition
+  metadata:
+    name: fluxhelmresources.integrations.flux
+  spec:
+    group: integrations.flux
+    names:
+      kind: FluxHelmResource
```
squaremo: Everything is a resource, so that doesn't need to be in the name. "Flux" is in the API group, but commands and code don't always (or even) refer to that, so probably worth keeping to avoid clashes. 
`Chart` is more specific than `Helm`, unless you reckon any later code dealing with packages will use this definition too. Lastly, probably worth having `Release` in the name, since that what the resources will represent. So, either `FluxChartRelease` or `FluxHelmRelease`, I reckon.

integrations/artifacts/fluxhelmresource-crd.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
squaremo: Oh right, this is the canonical version of this file. It'd be good to have it only in one place. See my comments in deploy/helm-operator.yaml above.

It'd be good to have this only in one place. Is there a downside to the operator applying it on startup?

squaremo:
@stefanprodan, @errordeveloper If you have time for a skim of this PR it would be appreciated. Just looking at the surface detail (i.e., names of things and how it's run), for the minute.

deploy/helm-operator-deployment.yaml
squaremo: To be fair, it was the only advice I found (https://docs.openshift.org/latest/admin_guide/custom_resource_definitions.html)

hack/update_codegen.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
tamarakaufler: "hack" because every  custom operator implementation I looked at, that uses these regeneration scripts used this naming convention. I found it a  bit strange but went with the common naming. No problem to rename.

We spoke about this at the beginning. I initially tried not to have the generated code in the repo and generate it during CI, but that  did not work. So I committed the generated code and left improvements for later.

deploy/helm-operator-deployment.yaml
tamarakaufler: Will change. _helm.integrations.flux.weave.works_ seems good to me (in case we integrate more), even if quite long

deploy/helm-operator-deployment.yaml
tamarakaufler: :+1:

deploy/helm-operator-deployment.yaml
tamarakaufler: I think FluxHelmRelease is a good name

hack/update_codegen.sh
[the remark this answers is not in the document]
tamarakaufler: I dont seem to be able to comment on your comment regarding the flags. Yes, as I mentioned to you a couple of days ago, that part needs tidying up. Too early for a PR ....

integrations/artifacts/fluxhelmresource-crd.yaml
tamarakaufler: yes, apiextensions.k8s.io/v1beta1 is the k8s API version for creating CRDs. I shall make the changes you suggested for deploy/helm-operator.yaml.

I don't think there is a downside to that..

squaremo: The comment's there because that's the first line of the file :)

hack/update_codegen.sh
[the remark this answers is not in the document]
squaremo: > I initially tried not to have the generated code in the repo and generate it during CI, but that did not work.

How did it fail to work?

integrations/artifacts/fluxhelmresource-crd.yaml
tamarakaufler: Actually there is a preferred sequence as to how the three parts of the integration functionality are released:
1) CRD ... both the fluxd and helm-operator depend on it
2) fluxd ... help-operator depends on it for the secret
3) helm-operator

Saying that,  helm-operator can apply the CRD at the startup and wait for fluxd to create the keys.

deploy/helm-operator-deployment.yaml
tamarakaufler: Yes, I saw that one too. I considered other options, very similar to yours, but could not decide. I thought about the FQDN, but then the naming seemed so long. So I just started with integrations.flux with the prospect of a change (which came :) )

Renamed.

deploy/helm-operator-deployment.yaml
tamarakaufler: done

deploy/helm-operator-deployment.yaml
tamarakaufler: done

deploy/helm-operator-deployment.yaml
stefanprodan: It should be `v1alpha1` if we want to use Kubernetes version naming.

integrations/artifacts/fluxhelmresource-crd.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: fluxhelmresources.integrations.flux
+spec:
+  group: integrations.flux
+  names:
+    kind: FluxHelmResource
+    listKind: FluxHelmResourceList
+    plural: fluxhelmresources
+  scope: Namespaced
+  version: v1
```
stefanprodan: This should match the package name

squaremo:
This needs a rebase, and a CircleCI stanza to publish the image.

stefanprodan:
Proposition: Instead of the verbose `name/value` pair inside the CR values we could have the chart `values.yaml` content.

Example:

Chart: charts/redis/values.yaml:

```yaml
image: bitnami/redis:4.0.8-r0
metrics:
  enabled: true
  image: oliver006/redis_exporter
  imageTag: v0.11
```

Current CR:

```yaml
kind: FluxHelmRelease
metadata:
  name: dev-redis
  namespace: default
  labels:
    chart: redis
spec:
  chartGitPath: redis
  releaseName: dev-redis
  values:
    - name: image
      value: bitnami/redis:4.0.8-r0
    - name: metrics.enabled
      value: true
    - name: metrics.image
      value: oliver006/redis_exporter
    - name: metrics.imageTag
      value: v0.11
```

Proposed CR:

```yaml
kind: FluxHelmRelease
metadata:
  name: dev-redis
  namespace: default
  labels:
    chart: redis
spec:
  chartGitPath: redis
  releaseName: dev-redis
  values:
    image: bitnami/redis:4.0.8-r0
    metrics:
      enabled: true
      image: oliver006/redis_exporter
      imageTag: v0.11
```

rade:
@stefanprodan good idea, but surely a separate issue.

tamarakaufler:
@stefanprodan 
Response regarding your proposition: Instead of the verbose name/value pair inside the CR values we could have the chart values.yaml content.

Initially I considered having a customisation setup in the CR compatible with the values.yaml file. This proved difficult as there needs to be a data structure description of each CR property so that kubernetes can interpret it. Helm template properties are varied; can be single values or nested to different depths; ie not known beforehand. I tried to use the generic interface which, unsurprisingly, did not work.

The current approach, where the customisation, ie the values field, is described as a map[string] string is:
a) easily processed on our side into the form that helm API needs
b) can be, i think, easily understood and applied by users

I welcome any better way to achieve the same with similar benefits.

rade:
Even just flattening the structure to
```
  values:
    - image: bitnami/redis:4.0.8-r0
    - metrics.enabled: true
    - metrics.image: oliver006/redis_exporter
    - metrics.imageTag: v0.11
```
would be an improvement.

But can we please, please discuss this in a separate issue rather than cluttering this PR.

tamarakaufler:
Yes, this is possible. Cannot recall from the top of my head why I chose my approach (oh now I do, but wont explain here, just will say the reason for this disappeared, but I left this data structure. So your suggestion will work fine).

An improvement to make when the time comes.

rade:
> An improvement to make when the time comes.

File an issue, please.

tamarakaufler:
Created: https://github.com/weaveworks/flux/issues/951

PR #962, opened 2018-02-26
Let fluxd answer queries without a git repo supplied

squaremo:
Instead of waiting for the git repo to be ready before responding to API calls (aside from status), just start the repo syncing when starting up, and respond according to its current state.

The API (ListServices) will now mark controllers that aren't in the git repo -- or all controllers if there's no git repo -- as being read-only, supplying a reason in the field `ReadOnly`. An empty field indicates that the controller is (probably) writable. Old daemons won't populate this field, but since they don't answer ListServices until the repo is ready anyway, the assumption that the controller is writable will still usually be correct (and not disastrous when it's wrong).

An example API response:

```json
[
  {
    "ID": "hello:cronjob/hellocron",
    "Containers": [
      {
        "Name": "weekly-hello-cron",
        "Current": {
          "ID": "alpine:3.4",
          "Digest": "",
          "ImageID": ""
        },
        "Available": null
      }
    ],
    "ReadOnly": "NotInRepo",
    "Status": "ready",
    "Automated": false,
    "Locked": false,
    "Ignore": false,
    "Policies": {}
  },
  {
    "ID": "hello:deployment/helloworld",
    "Containers": [
      {
        "Name": "helloworld",
        "Current": {
          "ID": "quay.io/weaveworks/helloworld:master-07a1b6b",
          "Digest": "",
          "ImageID": ""
        },
        "Available": null
      },
      {
        "Name": "sidecar",
        "Current": {
          "ID": "quay.io/weaveworks/sidecar:master-a000002",
          "Digest": "",
          "ImageID": ""
        },
        "Available": null
      }
    ],
    "Status": "ready",
    "Automated": false,
    "Locked": false,
    "Ignore": false,
    "Policies": {
      "tag.helloworld": "glob:master-*"
    }
  },
  {
    "ID": "hello:deployment/localhello",
    "Containers": [
      {
        "Name": "helloworld",
        "Current": {
          "ID": "kube-registry.kube-system.svc.cluster.local:5000/mikeb/hello",
          "Digest": "",
          "ImageID": ""
        },
        "Available": null
      },
      {
        "Name": "sidecar",
        "Current": {
          "ID": "quay.io/weaveworks/sidecar:master-a000002",
          "Digest": "",
          "ImageID": ""
        },
        "Available": null
      }
    ],
    "ReadOnly": "NotInRepo",
    "Status": "ready",
    "Automated": false,
    "Locked": false,
    "Ignore": false,
    "Policies": {}
  }
]
```


api/v6/api.go
```diff
@@ -16,9 +16,23 @@ type ImageStatus struct {
...
+	ReadOnlyOK      ReadOnlyReason = ""
+	ReadOnlyMissing ReadOnlyReason = "NotInRepo"
+	ReadOnlySystem  ReadOnlyReason = "System"
+	ReadOnlyNoRepo  ReadOnlyReason = "NoRepo"
+)
+
 type ControllerStatus struct {
 	ID         flux.ResourceID
 	Containers []Container
+	ReadOnly   ReadOnlyReason `json:",omitempty"`
```
sambooo: What's the motivation behind making this `omitempty` but none of the other properties?

It's plausible that a `Controller` could have no `Policies`, or maybe even no `Containers`, just as it may not be `ReadOnly`.

daemon/daemon.go
```diff
@@ -76,21 +76,35 @@ func (d *Daemon) ListServices(ctx context.Context, namespace string) ([]v6.Contr
...
 
 	var services policy.ResourceMap
+	var noRepo bool
 	err = d.WithClone(ctx, func(checkout *git.Checkout) error {
 		var err error
 		services, err = d.Manifests.ServicesWithPolicies(checkout.ManifestDir())
 		return err
 	})
-	if err != nil {
+	switch {
```
sambooo: I like the use of the bare switch statements here and below - keeps all the references to each new bit of state easily within a screenful. Should be easier to read and follow than some of our bigger methods.

daemon/daemon.go
```diff
@@ -76,21 +76,35 @@ func (d *Daemon) ListServices(ctx context.Context, namespace string) ([]v6.Contr
...
+	case err == git.ErrNotReady:
+		noRepo = true
+	case err != nil:
 		return nil, errors.Wrap(err, "getting service policies")
 	}
 
 	var res []v6.ControllerStatus
 	for _, service := range clusterServices {
-		policies := services[service.ID]
+		var readonly v6.ReadOnlyReason
```
sambooo: Should be `readOnly`

git/repo.go
```diff
@@ -21,7 +21,7 @@ const (
 var (
 	ErrNoChanges = errors.New("no changes made in repo")
 	ErrNotReady  = errors.New("git repo not ready")
-	ErrNoConfig  = errors.New("git repo has not valid config")
+	ErrNoConfig  = errors.New("git repo does not have valid config")
```
sambooo: Should this be used somewhere? Since no code references this value, this PR doesn't
> Let fluxd answer queries without a git repo supplied,

so much as it does
>Let fluxd answer queries before the supplied git repo is ready,

right?

sambooo:
Should we wait until we're ready to do a `1.3` release before we merge this? Or, conversely, should we more quickly make sure we're ready to do a `1.3` so that we can merge this?

cluster/cluster.go
```diff
@@ -29,6 +29,10 @@ type Cluster interface {
 type Controller struct {
 	ID     flux.ResourceID
 	Status string // A status summary for display
+	// Is the controller considered read-only because it's under the
+	// control of the platform. In the case of Kibernetes, we simply
+	// omit these controllers; but this may not always be the case.
+	IsSystem bool
```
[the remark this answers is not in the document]
squaremo: Yeah; it _would_ be filled in by the cluster/kubernetes code, but that currently _omits_ the only things that would have it set (which are add-ons).

cmd/fluxd/main.go
```diff
@@ -368,48 +322,23 @@ func main() {
...
+		}()
 	}
 
+	logger.Log(
+		"url", *gitURL,
+		"user", *gitUser,
+		"email", *gitEmail,
+		"sync-tag", *gitSyncTag,
+		"notes-ref", *gitNotesRef,
+		"set-author", *gitSetAuthor)
```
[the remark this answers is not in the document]
squaremo: Under Pike's Eye.

cmd/fluxd/main.go
```diff
@@ -425,14 +354,38 @@ func main() {
 		GitConfig:      gitConfig,
 		Jobs:           jobs,
 		JobStatusCache: &job.StatusCache{Size: 100},
-
-		EventWriter: eventWriter,
-		Logger:      log.With(logger, "component", "daemon"), LoopVars: &daemon.LoopVars{
+		Logger:         log.With(logger, "component", "daemon"),
+		LoopVars: &daemon.LoopVars{
```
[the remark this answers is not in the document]
squaremo: Yup. Odd isn't it -- if I move the fields to the same line, it aligns it with the others, but when they're on their own lines, it puts it there.

api/v6/api.go
squaremo: There's no especial reason, and in fact I think this is the wrong way around for this particular field -- I _want_ a missing value (e.g., from an old daemon) to look like an empty string, for API consumers.

squaremo:
> Should we wait until we're ready to do a 1.3 release before we merge this?

I reckon we can do a release in the meantime.

api/v6/api.go
sambooo: With or without `omitempty`, unmarshalling with a missing value will result in an empty string in the consumer. Only way to distinguish missing versus empty is to make this field a pointer.

I suppose if we care about saving a few bytes we can pepper our structs with `omitempty` and hope any other languages consuming our api Do The Right Thing.

git/repo.go
squaremo: > Since no code references this value

https://github.com/weaveworks/flux/blob/master/git/repo.go#L156

but .. I take your general point, which is that this case isn't handled (or isn't handled well). Possibly there should be another ReadOnlyReason for "No git repo supplied", to distinguish it from "git repo isn't ready".

api/v6/api.go
squaremo: > unmarshalling with a missing value will result in an empty string in the consumer.

From Go, yes. But not from say, JavaScript.

git/repo.go
squaremo: Oh I see, that state is not reachable, because the git.Repo is initialised in the state RepoNew and there are no transitions into RepoNoConfig.

OK, I have fixed that (most recent commit). If the git.Repo is given an empty origin URL, it is initialised in RepoNoConfig. This percolates out to the API, where it's represented as `const ReadOnlyNoRepo ReadOnlyReason = "NoRepo"`.

sambooo: > there are no transitions into RepoNoConfig

I worded my comment poorly enough that it wasn't even obvious to me this is what I actually meant... Yep, there was no code which _assigned_ that value to anything.

If you're going to link code you should use [the permalink](https://github.com/weaveworks/flux/blob/1a860b16de3eb86f4a17c088a805da61c7bc8e4c/git/repo.go#L156). You never know how much time you might save the next code archeologist 😄.

rndstr: > If you're going to link code you should use the permalink

<kbd>y</kbd> is a nice shortcut for this when looking at a branch HEAD

PR #969, opened 2018-02-28
Fix failure to complete syncing due to --git-path

### LINKED ISSUE #967: Flux master isn't moving the sync tag
Some change in #940 broke moving the tag in the upstream repo. Track findings here, or hopefully just close with a simple fix.

squaremo:
The repo methods CommitsBetween and CommitsBefore use `git log
--oneline`, but hitherto did not separate the path argument (used to
restrict the commits of interest to those affecting that path) from
the arguments with `--`.

While we were using a local git clone with a working directory, this
didn't matter, because the path argument was recognisable as such. But
since we're now using a bare git clone, `git` will complain about any
path that looks a bit like a ref, and thereby fail to report syncs
after it's applied them to the cluster, as well as failing to respond
to SyncStatus.

We were also throwing away the error returned from doSync, so it was
more difficult to see what was the matter.

So:
 - use `--` to distinguish the path argument to `git log`
 - log errors returned from `doSync`
 - I found it helpful to trace the `git` invocations; I've left that
 in as a compile-time possibility.

Fixes #967.

### DIFF
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -84,7 +84,9 @@ func (d *Daemon) Loop(stop chan struct{}, wg *sync.WaitGroup, logger log.Logger)
 				default:
 				}
 			}
-			d.doSync(logger)
+			if err := d.doSync(logger); err != nil {
+				logger.Log("err", err)
+			}
 			syncTimer.Reset(d.SyncInterval)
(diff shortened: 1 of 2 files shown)


squaremo:
Thanks Sam!

PR #970, opened 2018-03-01
Include individual resource sync errors in Sync events

### LINKED ISSUE #756: Include errors in sync events
When we run the sync, we log but otherwise ignore error results from `kubectl apply`. The excuse is that in many cases it will be to do with the ordering -- creating a resource before its namespace has been created, for example. But this misses some important cases that we do want to know about, like having an invalid manifest.

I suggest we include these "genuine" errors in the sync event sent upstream, so they can be included in notifications. (If we can be clever about not including ordering problems, either by filtering them out, or #602, so much the better).

squaremo:
A sync can be a partial success, in that we skip lightly over individual resources failing to be applied. This means that bogus manifests don't get noticed, and the symptoms can be difficult to pin down if you are used to thinking `cluster == git`.

This PR surfaces those errors by including them in the Sync event, as described in #756. (It fixes #756, in fact.)

There are other measures we could take, e.g., annotating offending commits (maybe? we may not be able to pin it down); keeping a gauge of sync failures which can be alerted on. But this is a fair start.

### DIFF
--- a/cluster/manifests.go
+++ b/cluster/manifests.go
@@ -20,8 +20,11 @@ type Manifests interface {
 	// Update the definitions in a manifests bytes according to the
 	// spec given.
 	UpdateDefinition(def []byte, container string, newImageID image.Ref) ([]byte, error)
-	// Load all the resource manifests under the path given
-	LoadManifests(paths ...string) (map[string]resource.Resource, error)
+	// Load all the resource manifests under the path given. `baseDir`
+	// is used to relativise the paths, which are supplied as absolute
+	// paths to directories or files; at least one path must be
+	// supplied.
+	LoadManifests(baseDir, first string, rest ...string) (map[string]resource.Resource, error)
 	// Parse the manifests given in an exported blob
 	ParseManifests([]byte) (map[string]resource.Resource, error)
 	// UpdatePolicies modifies a manifest to apply the policy update specified
--- a/cluster/sync.go
+++ b/cluster/sync.go
@@ -11,8 +11,8 @@ import (
 // SyncAction represents either the deletion or application (create or
 // update) of a resource.
 type SyncAction struct {
-	Delete     resource.Resource // ) one of these
-	Apply      resource.Resource // )
+	Delete resource.Resource // ) one of these
+	Apply  resource.Resource // )
 }
 
 type SyncDef struct {
@@ -20,14 +20,14 @@ type SyncDef struct {
 	Actions []SyncAction
 }
 
-type SyncError struct {
+type ResourceError struct {
 	resource.Resource
 	Error error
 }
 
-type SyncErrors []SyncError
+type SyncError []ResourceError
 
-func (err SyncErrors) Error() string {
+func (err SyncError) Error() string {
 	var errs []string
 	for _, e := range err {
 		errs = append(errs, e.ResourceID().String()+": "+e.Error.Error())
--- a/cluster/sync.go
+++ b/cluster/sync.go
@@ -2,24 +2,17 @@ package cluster
 
 import (
 	"strings"
+
+	"github.com/weaveworks/flux/resource"
 )
 
 // Definitions for use in synchronising a cluster with a git repo.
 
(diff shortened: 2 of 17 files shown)


cluster/sync.go
```diff
@@ -2,37 +2,35 @@ package cluster
...
 	Actions []SyncAction
 }
 
-type SyncError map[string]error
+type SyncError struct {
+	resource.Resource
+	Error error
+}
+
+type SyncErrors []SyncError
```
sambooo: Should this be its own commit? Could do with an explanation for what motivated the change, I'm sure future-us would agree.

event/event.go
```diff
@@ -204,6 +204,8 @@ type SyncEventMetadata struct {
 	// policy changes, and "other" (meaning things we didn't commit
 	// ourselves)
 	Includes map[string]bool `json:"includes,omitempty"`
+	// Per-resource errors
+	Errors map[string]string `json:"errors,omitempty"`
```
sambooo: Can we give this a better type, perhaps? `map[T]error` where `T` is some meaningful struct type with a good `String()` implementation would help make this self-documenting.

I think the `ResourceID` code is a good example of making something that _is_ stringy not be so stringy.

cluster/manifests.go
```diff
@@ -21,7 +21,7 @@ type Manifests interface {
 	// spec given.
 	UpdateDefinition(def []byte, container string, newImageID image.Ref) ([]byte, error)
 	// Load all the resource manifests under the path given
-	LoadManifests(paths ...string) (map[string]resource.Resource, error)
+	LoadManifests(base, first string, rest ...string) (map[string]resource.Resource, error)
```
sambooo: What do `base`, `first`, `rest` mean?

event/event.go
squaremo: Yes, probably just 

```
type ErrorSource struct {
    ResourceID string
    Source string
}
```

(those are the two bits of information that get encoded in here). I don't think we much care about unpacking it at the other side.

squaremo: Or,

```
type ResourceError struct {
  ID string
  Path string
  Error string
}

type SyncEventMetadata struct {
...
  Errors []ResourceError
}
```

cluster/manifests.go
squaremo: Worth a comment, do you think?

sambooo: > Worth a comment, do you think?
I do. That or a change to `baseDir` so it's more obvious that it's a directory. The problem with `string` is you can put anything in it and it's basically just `[]byte` with the implication it might be readable when printed.

event/event.go
sambooo: The intended result here is that consumers of this struct (ui, notifications) don't need to search through the code to find out what kind of string we build and put in here.

The most-general, least-informative types we can JSONify are `map[string]string` and `map[string]interface{}`, both of which are basically `Object`. In this case the `V` in our `map[K]V` can't really be narrowed down to more than just arbitrary error text, but the `K` certainly can be.

If there's a way we can't build a key from `resource.Resource` that'd probably be perfect :)

squaremo: I'd rather have a slice of structs than have to pack things into a string so it can be a JSON-able map.

sambooo: Indeed if we don't actually need a map, let's scrap the map! Now let's figure out what a `struct { string, string, string }` needs... 😄 

(`ResourceError` up there doesn't have the same issue because it has named fields, but maybe we can give `ID` similar treatment to the map keys. `resource.ID`?)

squaremo: > maybe we can give ID similar treatment to the map keys

At some point we have to bottom out to JSON types. This is that point.

squaremo: Technically we could probably use a `ResourceID` there, since it'll get serialised to a string anyway.

cluster/manifests.go
squaremo: > The problem with string is you can put anything in it

I shall use the filesystem path types from Go's standard lib. OK done.

squaremo:
PTAL @sambooo

PR #973, opened 2018-03-02
Make notifications at least once delivery

Eochs:
changed flux loop so that if a sync or noteEvent fails to send, just return and do not move tag forward. This way if a notification fails, we wait for the next sync event to retry sending logic.
fixes weaveworks/service#1550
As stated in:
https://github.com/weaveworks/service/issues/1550
Since fluxd is constantly checking and bringing the cluster's state in line with the specified state in git, if an event does not fire for one sync, it will likely fire on the next sync.

### DIFF
--- a/daemon/loop.go
+++ b/daemon/loop.go
@@ -362,11 +362,15 @@ func (d *Daemon) doSync(logger log.Logger) (retErr error) {
 			},
 		}); err != nil {
 			logger.Log("err", err)
+			// Abort early to ensure at least once delivery of events
+			return err
 		}
 
 		for _, event := range noteEvents {
 			if err = d.LogEvent(event); err != nil {
 				logger.Log("err", err)
+				// Abort early to ensure at least once delivery of events
+				return err
 			}
 		}
 	}


daemon/loop.go
```diff
@@ -362,11 +362,13 @@ func (d *Daemon) doSync(logger log.Logger) (retErr error) {
 			},
 		}); err != nil {
 			logger.Log("err", err)
+			return err
```
rndstr: maybe add a comment to explain why we abort early?

sambooo:
Nitpick: that commit message is going to be hard to read 6 months from now.

I would suggest something like this:
```
Leave sync-tag unchanged if LogEvent returns an error

We want to ensure at-least-once delivery of events to Weave Cloud, so if notifying the upstream fails we retry on the next sync. The task of deduplicating events is left to the consumer.
```

Put the 'what' on the first line and then go into 'why' below, and leave the 'how' to the diff!

sambooo:
Have you tested this change against a selectively-failing `LogEvent` handler?

How does the daemon's behaviour change if `LogEvent` constantly returns an error? I would guess without testing it myself that flux basically stops doing anything useful, but maybe that's not the case here, and maybe `LogEvent` always erroring _should_ break things.

squaremo:
> I would guess without testing it myself that flux basically stops doing anything useful

It will keep syncing new things, but not report all events. It would look like the recent breakage (#967), since it'll fail at about the same stage -- after syncing, before moving the high water mark.

Eochs:
@sambooo since all of the sync logic happens before events are sent, if LogEvent fails, syncs will happen, but events will not be sent, and the sync-tag will not be moved. Then LogEvent will retry after every loop of sync.

PR #974, opened 2018-03-02
Move the different API versions into packages of their own

squaremo:
To get the type system to help us with versioning the API, I've moved v6 and v9 APIs to their own packages (those in which there were changes to the interface, rather than the transport or semantics).

(From the HTTP or RPC point of view, things look exactly the same.)

There more versioning types that could be done, though it gets a bit thorny here on in.


cmd/fluxd/main.go
```diff
@@ -278,7 +279,7 @@ func main() {
 	}
 
 	// Indirect reference to a daemon, initially of the NotReady variety
```
sambooo: fwiw, this doesn't seem like much of an indirect reference

squaremo: aw come on github, show me more than three lines of context!

Issue #958, opened 2018-02-22
Weave-flux pod crashing looping with panic:


foot:
```
ts=2018-02-22T11:58:38.568263415Z caller=release.go:100 component=platform method=Sync cmd="kubectl apply -f -" took=3.656547872s err=null output="deployment \"dashboard\" configured\nrole \"fluentd\" configured\ningress \"event-api\" configured\nservice \"default-http-backend\" configured\ndeployment \"goto-nrw\" configured\nservice \"goto-nrw\" configured\nserviceaccount \"nginx-ingress-serviceaccount\" configured\ningress \"klima-api\" configured\nservice \"klima-api\" configured\nrolebinding \"nginx-ingress-role-nisa-binding\" configured\ndeployment \"klima-api\" configured\nconfigmap \"nginx-configuration\" configured\nserviceaccount \"helm\" configured\ndaemonset \"nginx-ingress-controller\" configured\ningress \"dashboard\" configured\ningress \"traffics\" configured\ncronjob \"unims-scraper\" configured\ningress \"goto-nrw\" configured\nconfigmap \"udp-services\" configured\nrole \"nginx-ingress-role\" configured\nservice \"event-api\" configured\ndeployment \"default-http-backend\" configured\nservice \"traffics\" configured\nrolebinding \"fluentd\" configured\ndaemonset \"weave-net\" configured\njob \"unims-scraper\" configured\nservice \"dashboard\" configured\nconfigmap \"tcp-services\" configured\ndeployment \"event-api\" configured\ndeployment \"traffics\" configured"
ts=2018-02-22T11:58:38.568485016Z caller=loop.go:164 component=sync-loop err="default:list/: running kubectl: error: the namespace from the provided object \"kube-system\" does not match the namespace \"default\". You must pass '--namespace=kube-system' to perform this operation."
ts=2018-02-22T11:58:38.598793855Z caller=daemon.go:461 component=daemon event="Sync: 502b797..7892738, default:clusterissuer/letsencrypt, default:clusterrole/r2-cert-manager, default:clusterrolebinding/r2-cert-manager, default:customresourcedefinition/certificates.certmanager.k8s.io, default:customresourcedefinition/clusterissuers.certmanager.k8s.io, default:customresourcedefinition/issuers.certmanager.k8s.io, default:deployment/r2-cert-manager, default:ingress/kibana, default:ingress/kibana-auth, default:serviceaccount/r2-cert-manager" logupstream=true
[...]

Issue #949, opened 2018-02-19
Flux panics when applying CRDs


stefanprodan:
I have the following CRD:

```yaml
---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: fluxhelmreleases.helm.integrations.flux.weave.works
spec:
  group: helm.integrations.flux.weave.works
  names:
    kind: FluxHelmRelease
    listKind: FluxHelmReleaseList
    plural: fluxhelmreleases
  scope: Namespaced
  version: v1alpha
```

And Flux panics with:

```
ts=2018-02-19T11:40:31.216648153Z caller=daemon.go:461 component=daemon event="Sync: 5c14d48..75539f9, default:customresourcedefinition/fluxhelmreleases.helm.integrations.flux.weave.works, weave:deployment/helm-operator" logupstream=true
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xf03876]
```

rade:
@stefanprodan can you get us more of the stack trace, please?

squaremo:
Should be easy enough to reproduce I reckon.

stefanprodan:
I can't reproduce this with the latest Flux version v1.2.3:

```
ts=2018-03-03T07:10:27.803114389Z caller=release.go:100 component=platform method=Sync cmd="kubectl --namespace default apply -f -" took=11.013475214s err=null output="customresourcedefinition \"fluxhelmreleases.helm.integrations.flux.weave.works\" created"
ts=2018-03-03T07:10:27.8059301Z caller=release.go:61 component=platform method=Sync cmd=apply args= count=1
ts=2018-03-03T07:10:46.848320251Z caller=release.go:100 component=platform method=Sync cmd="kubectl apply -f -" took=19.040658437s err=null output="deployment \"weave-flux-helm-operator\" created"
```

PR #975, opened 2018-03-05
Avoid an NPE when parsing LogEvent responses

squaremo:
We assume that if the upstream we're posting events to responds with
`Content-Type: application/json`, it must be a helpful error
(`*github.com/weaveworks/flux/errors.Error`). However, it's possible
to get any old JSON (even no JSON), and in that case, we'd trip
oursevles up by creating a helpful error with a nil pointer, leading
to a panic later.

### DIFF
--- a/http/client/client.go
+++ b/http/client/client.go
@@ -138,6 +138,7 @@ func (c *Client) methodWithResp(ctx context.Context, method string, dest interfa
 	defer resp.Body.Close()
 
 	respBytes, err := ioutil.ReadAll(resp.Body)
+
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rade:
Have we seen this in the wild? #949 and #958 perhaps?

squaremo:
> Have we seen this in the wild? #949 and #958 perhaps?

Not sure. I noticed it after adding a List resource to my test repo, rather than a CRD. It was provoked by the upstream server sending a `400 Bad Request` in response to the posted sync event, and maybe that happens for CRDs too.

PR #976, opened 2018-03-05
[WIP] Use kubeyaml for YAML updates

squaremo:
`squaremo/kubeyaml` uses a round-tripping YAML parser, so can do a more reliable job at updating YAMLs.


cluster/kubernetes/kubeyaml.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// Annotate calls the kubeyaml subcommand `annotate` with the arguments as given.
+func (k KubeYAML) Annotate(in []byte, ns, kind, name string, policies ...string) ([]byte, error) {
+	args := []string{"annotate", "--namespace", ns, "--kind", kind, "--name", name}
+	args = append(args, policies...)
+	return execKubeyaml(in, args)
+}
+
+func execKubeyaml(in []byte, args []string) ([]byte, error) {
```
rade:
```golang
func (k KubeYAML) exec(...)
```
perhaps?

sambooo: Consume an `io.Reader` here! That way we don't have to read in a yaml's `os.File` (which implements `io.Reader`) into a `[]byte` only to wrap it back up in a `bytes.NewBuffer` later.

We only need to go down to the byte-slice level if we actually want to look at what's in there. Once our flux code stops modifying yaml directly, it'll only be doing two things:
- reading in yaml for the sake of parsing (but not editing), and
- forwarding yaml to kubeyaml,
neither of which should actually require us to know anything more than "we have a source of bytes"

Steve Francia says it better than I might hope to: https://youtu.be/29LLRKIL_TI?t=199

sambooo: Produce a reader too, maybe

squaremo: > Consume an io.Reader here!

Great idea.

> Produce a reader too, maybe

Maybe; we will almost always want to write the result back to the file, and returning a reader makes this a bit harder.

squaremo: The only counter-argument I can summon is that we read the input more than once, at least for annotations, so this could create more faff rather than reducing it.

squaremo:
IRL conversation: we could give kubeyaml a server mode, since it only needs to answer simple questions and doesn't need a sophisticated API or anything. That would let us run it as a sidecar, rather than an initContainer, removing some of the ickiness with `$PATH`.

squaremo:
> we could give kubeyaml a server mode, since it only needs to answer simple questions and doesn't need a sophisticated API or anything

Though this would make testing a bit harder.

cluster/kubernetes/kubeyaml.go
sambooo: > Maybe; we will almost always want to write the result back to the file, and returning a reader makes this a bit harder.

How so? Given a reader, `r`, and an open file, `f`, `io.Copy(f, r)` will write out the file.


> we read the input more than once, at least for annotations

For now! I can't see a reason why in future we would need to read the file more than once specifically for annotations. One call to `YAML.Unmarshal(...)` will extract all the info we need, no?

sambooo:
Should unit tests depend on kubeyaml at all, whether it's via `os.Exec` or `http.Client`?

We could:
1. Create a kubeyaml mock and test that flux tells it the right changes to make,
2. Test that kubeyaml itself makes the right changes,
3. (optional) Integration test it all for additional confidence

cluster/kubernetes/kubeyaml.go
squaremo: > One call to YAML.Unmarshal(...) will extract all the info we need, no?

It does; _then_ we want to pass the input to `kubeyaml`.

sambooo: Sounds like we want to use [`io.TeeReader`](https://godoc.org/io#TeeReader)

squaremo:
Just noticed a problem. When we're updating then container-specific tag filters, we parse the configs. But while `kubeyaml` happily deals with Lists and multidoc files, that specific bit of parsing doesn't. I can think of a couple of fixen:

 - be able to parse Lists and multidocs in that place; or, more likely, adapt the code in `cluster/kubernetes/resources` so we can extract the list of containers using it.
 - add a feature to `kubeyaml` so it will do the container enumeration when updating;
 - change how these annotations work so that the value can always be replaced without knowledge of the specific containers

docker/Dockerfile.flux
```diff
@@ -18,6 +18,10 @@ WORKDIR /home/flux
 ENTRYPOINT [ "/sbin/tini", "--", "fluxd" ]
 RUN apk add --no-cache openssh ca-certificates tini 'git>=2.3.0'
 
+# Get the kubeyaml binary (files) and put them on the path
+COPY --from=quay.io/squaremo/kubeyaml:0.1.0 /usr/lib/kubeyaml /usr/lib/kubeyaml/
```
sambooo: Hope to see 1.0 soon

squaremo: Current roadblock there: kubeyaml assumes ruamel works perfectly; however, ruamel does not work perfectly :-/

But: at least with #1094, and an annotation analogue of that, we'd catch cases where ruamel mangled things.

cluster/kubernetes/kubeyaml.go
squaremo: I like this suggestion overall, and I've started implementing it in a branch (`refactor/bytes-to-readers`). The consequences are quite far-ranging though, so best to do in another PR after the fact, I think.

cluster/kubernetes/update_test.go
```diff
@@ -149,7 +149,6 @@ spec:
     metadata:
       labels:
         name: fluxy
-        version: master-a000001
```
awh: This removal isn't meaning-preserving as advertised in the commit message - incidental cleanup perhaps?

cluster/kubernetes/update_test.go
```diff
@@ -194,7 +193,6 @@ spec:
     metadata:
       labels:
         name: fluxy
-        version: "1234567"
```
awh: Ditto

cluster/kubernetes/kubeyaml.go
awh: I agree; a separate PR seems the best way to go.

cluster/kubernetes/update_test.go
squaremo: Yes, I expect I removed this as part of kneading the tests into shape, and forgot to record the fact in the commit message. Sorry!

kubeyaml doesn't attempt to update this label -- it was a vestigial part of ReplicationController support, iirc.

Issue #971, opened 2018-03-01
Flux helm operator doesn't work with non-master branches


sambooo:
The specific error we're seeing is:

```
ts=2018-02-28T22:45:24.276957262Z caller=main.go:225 component=helm-operator error="Failed to clone git repo [git@github.com:username/repo.git, charts, flux]: reference not found
```

Looks like it may be an error in the way we're calling the git library we're using, and it looks like master works.

rossedman:
I can confirm this. This is happening to me as well.

sambooo:
@rossedman if you can find a way to use a master branch, that'll work for the time being. I'll update this issue as I progress.

rossedman:
@sambooo yeah can do, tried to fork and change some code but wasn't sure how to test and deploy with it.

rossedman:
@sambooo You can do checkouts from branch with something like this:

```
	r, err := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
		URL:           repo,
		ReferenceName: plumbing.ReferenceName(fmt.Sprintf("refs/heads/%s", branch)),
		SingleBranch:  true,
		Depth:         1,
	})
```

Seems like the checkout is happening and then branch changing. This example uses in storage memory as well.

sambooo:
@rossedman I just extracted a minimal amount of code to reproduce the failure outside of the helm-operator itself. See: #977. Contributions welcome if you spot a fix before I do!

rossedman:
@sambooo Awesome. Thanks so much. When I run locally with kubeconfig I cannot connect to Tiller, how do you solve this problem when testing locally? Was going to try to debug as well.

sambooo:
@rossedman I didn't author this part of the code, so I'm not sure how that's done. @tamarakaufler might be able to answer your question.

sambooo:
@rossedman I've just built and pushed `quay.io/samb/helm-operator:v1` from the changes in #977. Would you mind putting that in your yaml and letting me know if it fixes the problem?

rossedman:
@sambooo This appeared to work for me. Fresh cluster in Docker Edge and reapplied everything. Received successful clones.

sambooo:
Awesome! I'll rework that PR and get it merged into master at some point. You should be fine to use that image for the time being.

Cheers for the input 👍

rossedman:
@sambooo Thanks so much!

tamarakaufler:
Hello @rossedman :wave: 

[...]

PR #977, opened 2018-03-05
WIP #971 debugging branch

sambooo:
This shouldn't be merged but since I'm timeboxing I'll push up what I've written so far in my efforts, if only to pick it up later.

### DIFF
--- a/cmd/git-test/main.go
+++ b/cmd/git-test/main.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"golang.org/x/crypto/ssh"
+
+	gitssh "gopkg.in/src-d/go-git.v4/plumbing/transport/ssh"
+
+	"github.com/go-kit/kit/log"
+	"github.com/weaveworks/flux/integrations/helm/git"
+)
+
+var (
+	url            string
+	branch         string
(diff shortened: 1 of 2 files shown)


sambooo:
Build with `go build -o build/git-test ./cmd/git-test`, run with `./build/git-test -help`

cmd/git-test/main.go
```diff
@@ -0,0 +1,75 @@
+package main
```
tamarakaufler: Do we need this test file in the master? I would suggest to remove it now that the problem is fixed.

integrations/helm/git/git.go
```diff
@@ -98,9 +100,10 @@ func (ch *Checkout) Clone(ctx context.Context, cloneSubdir string) error {
 	}
 	ch.Dir = repoDir
 
-	repo, err := gogit.PlainClone(repoDir, false, &gogit.CloneOptions{
-		URL:  ch.Config.URL,
-		Auth: ch.auth,
+	repo, err := gogit.CloneContext(ctx, memory.NewStorage(), osfs.New(repoDir), &gogit.CloneOptions{
```
tamarakaufler: Another possibility would be to use gogit.PlainCloneContext and have all info on the filesystem, freeing the memory for other uses. I don't have strong feelings about it though.

As we are cloning the correct branch, do we need the wt.Checkout (original lines 114-120)?

PR #988, opened 2018-03-07
Fix for Helm operator non-master branch clone issue

tamarakaufler:


deploy/helm-operator-deployment.yaml
```diff
@@ -15,23 +15,28 @@
...
         labels:
-          name: helm-operator
+          name: flux-helm-operator
       spec:
-        serviceAccount: flux
         volumes:
         - name: git-key
           secret:
             secretName: flux-git-deploy
+        serviceAccountName: weave-flux
```
squaremo: The service account in https://github.com/weaveworks/flux/blob/master/deploy/flux-account.yaml is named `flux` and is in the default namespace. Better to be consistent with the other resources in this dir.

squaremo:
Why are there two Deployments now?

deploy/helm-operator-deployment.yaml
tamarakaufler: I agree about the consistency. I am aware of the differences regardinf the account and the namespace.

I should have been clear about the purpose of this deployment manifest and should have named it differently. The reason the helm-operator deployment yaml is set up the way it is, is to allow the user to use it as an addon given the weave agent does not set up the helm-operator pod yet. At the moment the purpose of the helm-operator deployment file is not to run it in a standalone mode but rather to set it up manually before the helm operator gets properly integrated.

I should have used another name for the file, which will be there before the helm operator gets  integrated, so as not to confuse it with its standalone cousin. As I used the current name in the Flux-Helm integration setup doc, I think it's perhaps best to leave it, for the time being, as is, not to confuse current Helm users that are giving this already a try. When the helm-operator is properly integrated, then the helm-operator-deployment.yaml will be used for the standalone mode. (I have added helm-operator-deployment.standalone.yaml, which will be in due time renamed to helm-operator-deployment.yaml.

tamarakaufler:
Please see my explanation above

deploy/helm-operator-deployment.yaml
tamarakaufler: Deployment verified.

squaremo: Actually, stepping back from the particulars, I don't think those commits belong here anyway: what have they to do with using a branch other than master? I recommend transplanting them to another PR.

squaremo:
> Please see my explanation above

Right I see; there's a need for one that works standalone, and one that works with weave cloud (because weave cloud won't supply it, yet). Let's move the latter to an `examples/` directory, to maintain consistency within the `deploy/` dir.

deploy/helm-operator-deployment.yaml
tamarakaufler: As per our conversation, I shall move the deployment yaml changes to another PR.

PR #990, opened 2018-03-08
Ensure RPC clients implement `io.Closer`

sambooo:
Since #974 the `Close` method isn't implemented on our RPC clients. This was able to happen because removing it wasn't a compile error. Now it is.

Also contains some cleanup in the general area.

### DIFF
--- a/remote/rpc/clientV6.go
+++ b/remote/rpc/clientV6.go
@@ -17,7 +17,7 @@ import (
 // talking to remote daemons.
 type RPCClientV6 struct {
 	client *rpc.Client
-	baseClient
+	fallback
 }
 
 type clientV6 interface {
--- a/remote/rpc/clientV6.go
+++ b/remote/rpc/clientV6.go
@@ -16,8 +16,8 @@ import (
 // RPCClient is the rpc-backed implementation of a server, for
 // talking to remote daemons.
 type RPCClientV6 struct {
-	*baseClient
 	client *rpc.Client
+	baseClient
 }
 
 type clientV6 interface {
@@ -48,7 +48,7 @@ var supportedKindsV6 = []string{"service"}
 
 // NewClient creates a new rpc-backed implementation of the server.
 func NewClientV6(conn io.ReadWriteCloser) *RPCClientV6 {
-	return &RPCClientV6{&baseClient{}, jsonrpc.NewClient(conn)}
+	return &RPCClientV6{client: jsonrpc.NewClient(conn)}
 }
 
 // Ping is used to check if the remote server is available.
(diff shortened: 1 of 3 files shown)


remote/rpc/clientV6.go
```diff
@@ -16,8 +16,8 @@ import (
 // RPCClient is the rpc-backed implementation of a server, for
 // talking to remote daemons.
 type RPCClientV6 struct {
-	*baseClient
 	client *rpc.Client
+	fallback
```
squaremo: Nice use of embedding

squaremo:
> Since #974 the Close method isn't implemented on our RPC clients. This was able to happen because removing it wasn't a compile error. Now it is.

But it a compile error (or worse, a cast error) when we use it elsewhere?

remote/rpc/fallback.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/weaveworks/flux/job"
+	"github.com/weaveworks/flux/remote"
+	"github.com/weaveworks/flux/update"
+)
+
+// fallback is, as the name suggests, a fallback api.UpstreamServer implementation.
+// Any method not implemented in a given RPC client will terminate here.
+type fallback struct{}
+
+type completeServer interface {
```
squaremo: This will make failure to implement a "meaningful" `Close` method into a runtime error. Wouldn't it be better if it were a compile time error? I'm not sure why we have `fallback` (formerly `baseError`), any more.

sambooo:
> But it is a compile error (or worse, a cast error) when we use it elsewhere?

Yep, https://github.com/weaveworks/service/pull/1850 won't compile if we vendor the current master branch.

remote/rpc/fallback.go
sambooo: > This will make failure to implement a "meaningful" Close method into a runtime error.

Yes it will, good point. I shall rethink this.

> I'm not sure why we have fallback (formerly baseError), any more.

I'll look into the flux-api code implications for removing the fallback at some point. My suspicion is some code duplication will arise, but hopefully not.

sambooo:
Hang on a minute! I think this PR is actually unneeded 😅

We have a reference to the connection in [our consumer code](https://github.com/weaveworks/service/blob/0f309e1b911d2e7e8a88725c53a5402067965314/flux-api/http/server.go#L356), so we can just call `Close()` on that.

Leaving this open (and approved, thanks @squaremo) for now, but I suspect it'll be closed without merge.

sambooo:
As suspected, there's not actually any reason this needs to be done. The consumer of the client must pass in an `io.ReadWriteCloser`, so they may close that instead of the RPC client itself.

PR #993, opened 2018-03-09
Ignore helm charts when scanning manifests

squaremo:
At the minute, if you have any Helm charts in the path you give to flux, it'll probably bork. The templated manifests in `<chart>/templates/` will either not parse as yamels, or be applied with weird field values.

So: when walking the repo looking for manifests, skip any directory that looks like a chart. What looks like a chart? The telltale things are:
 - a `Chart.yaml` file; and,
 - a `values.yaml` file.

Both of these, and only these, are mandatory. Looking for anything else would give false negatives (-> failing to skip a chart); _not_ looking for those would give false positives (-> failing to include a non-chart).

The first commit prepares the ground a little by making the test fixture more varied (i.e., including subdirectories and non-yamels), so chart files can be added to it in the latter commit to verify that everything still works by virtue of the chart being ignored.

### DIFF
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -22,6 +22,11 @@ func Load(roots ...string) (map[string]resource.Resource, error) {
 			if err != nil {
 				return errors.Wrapf(err, "walking %q for yamels", path)
 			}
+
+			if info.IsDir() && looksLikeChart(path) {
+				return filepath.SkipDir
+			}
+
 			if !info.IsDir() && filepath.Ext(path) == ".yaml" || filepath.Ext(path) == ".yml" {
 				bytes, err := ioutil.ReadFile(path)
 				if err != nil {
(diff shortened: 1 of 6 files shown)


stefanprodan:
@squaremo I discovered that not all files inside the `templates` dir are ignored. For example a `PodDisruptionBudget` gets applied even if it's part of a chart.   

```yaml
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: {{ template "memcached.fullname" . }}
spec:
  selector:
    matchLabels:
      app: {{ template "memcached.fullname" . }}
      chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
      release: "{{ .Release.Name }}"
      heritage: "{{ .Release.Service }}"
  minAvailable: {{ .Values.pdbMinAvailable }}
```

With the above file, Flux will run into this error:

```
ts=2018-05-07T08:44:46.764045009Z caller=loop.go:89 component=sync-loop err="loading resources from repo: parsing file at \"/tmp/flux-working640881799/repo/charts/memcached/templates/pdb.yaml\": parsing YAML doc from \"charts/memcached/templates/pdb.yaml\": yaml: unmarshal errors:\n  line 4: cannot unmarshal !!map into string"
```

A chart example can be found here: https://github.com/stefanprodan/weave-flux-helm-demo/tree/master/charts/memcached

squaremo:
> @squaremo I discovered that not all files inside the templates dir are ignored. For example a PodDisruptionBudget gets applied even if it's part of a chart.

It's not getting applied. It _is_ being considered as a resource though, which is still a problem. There is more than one way to load manifests from files, and (evidently) not all of them skip over chart directories. I'm going to log this as another issue.

Issue #951, opened 2018-02-20
Change Custom Resource values data structure


tamarakaufler:
**Problem**

The current implementation of FluxHelmRelease Custom Resources uses the following format for providing customizations:

```
values:
  - name: imageTag
    value: a00000a
  - name: memory
    value: 4Gi
``` 

This implementation can be simplified.

**A/C**

The user can add his customizations in the Custom Resource for a chart release in the form:

```
values:
  imageTag: a00000a
  memory: 4Gi
```

errordeveloper:
I agree that the 2nd option is more user-friendly.
However, it's important to understand that first option is commonly used in Kubernetes, it's often referred to as "keyed list", and I believe it's mostly for patching functionality there is a common method called `patchMergeKey` and user-facing aspects are documented at the following pages:
- https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/
- https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarative-config/

Maybe we don't need to support patching, but I'm not sure.

tamarakaufler:
The data structure change mimicking the helm values.yaml file has been proposed by Stefan, with Matthias

tamarakaufler:
A good point, patching may be needed.

sambooo:
In this case patching the object isn't something that makes sense to support. The expectation is that updates to our CustomResource will be made by Flux, where the old object is simply overwritten.

I don't think there's any compelling reason to use a keyed-list here.

tamarakaufler:
Discussion regarding the pros and cons of different approaches is in the **Discussion about the format of the overriding spec.values** comment in #982

squaremo:
Addressed by #1035.

PR #1006, opened 2018-03-14
982 overriding variables issue

### LINKED ISSUE #982: Helm Operator Doesn't Accept Variables
It appears the Helm Operator does not accept variables properly, it also does not accept boolean values as shown in the documentation. Here is an example of my release file, maybe I am misunderstanding something. I have double checked the variable paths multiple times for my chart, any advice on this? Thanks.

```
---
kind: FluxHelmRelease
metadata:
  name: ingress
  namespace: ingress
  labels:
    chart: ingress
spec:
  chartGitPath: ingress
  values:
  - name: nginx-ingress.controller.daemonset.useHostPorts
    value: "true"
  - name: nginx-ingress.controller.nodeSelector.pool_type
    value: lbe
  - name: nginx-ingress.controller.service.type
    value: ClusterIP
  - name: nginx-ingress.controller.stats.enabled
    value: "true"
  - name: nginx-ingress.controller.extraArgs.default-ssl-certificate
    value: nginx-ingress/default-ssl-certificate
```

tamarakaufler:
Fixes #982

### DIFF
--- a/integrations/helm/release/utils.go
+++ b/integrations/helm/release/utils.go
@@ -39,7 +39,7 @@ func collectValues(logger log.Logger, params []ifv1.HelmChartParam) (map[string]
 			}
 		}
 		pMap := mappifyValueOverride(k, vu)
-		base = mergeMaps(base, pMap)
+		base = mergeOverrides(base, pMap)
 	}
 
 	logger.Log("debug", fmt.Sprintf("override parameters in a data structure: %#v", base))
@@ -83,42 +83,42 @@ func mappifyValueOverride(k string, v interface{}) map[string]interface{} {
 	return inner
 }
 
-// mergeMaps ... merges two, possibly nested, maps
-// (copied from kubernetes/helm/cmd/helm/install.go (mergeValues function))
-func mergeMaps(dest map[string]interface{}, src map[string]interface{}) map[string]interface{} {
+// mergeOverrides ... merges two, possibly nested, maps
+// (copied from kubernetes/helm/cmd/helm/install.go (mergeValues function)
+// with redundant code removed)
+// mergeOverrides merges map related to one parameter override
+//
+//		- if a key k in the dest map exists:
+//				- if the value src[k] is not a map => the destMap value is overridden by the srcMap value
+//				- if the value src[k] is a map and value dest[k] is not a map => prefer the src[k] map
+//				- if both src[k] and value dest[k] are maps => merge the two maps
+func mergeOverrides(dest map[string]interface{}, src map[string]interface{}) map[string]interface{} {
 	for k, v := range src {
-		// If the key doesn't exist already, then just set the key to that value
 		if _, exists := dest[k]; !exists {
 			dest[k] = v
 			continue
 		}
 		nextMap, ok := v.(map[string]interface{})
-		// If it isn't another map, overwrite the value
+		// If the new value is not another map, overwrite the current value with it
 		if !ok {
 			dest[k] = v
 			continue
 		}
-		// If the key doesn't exist already, then just set the key to that value
-		if _, exists := dest[k]; !exists {
-			dest[k] = nextMap
-			continue
-		}
-		// Edge case: If the key exists in the destination, but isn't a map
+		// Edge case: If the key exists in the destination, but the corresponding value
+		// isn't  a map => overwrite with the new value
 		destMap, isMap := dest[k].(map[string]interface{})
-		// If the source map has a map for this key, prefer it
 		if !isMap {
 			dest[k] = v
 			continue
 		}
-		// If we got to this point, it is a map in both, so merge them
-		dest[k] = mergeMaps(destMap, nextMap)
+		// maps present in both the source and the destination => merge them
+		dest[k] = mergeOverrides(destMap, nextMap)
 	}
 	return dest
 }
 
 // unwrap ... unmarshals a string that is a serialised list
 func unwrap(v string) (interface{}, error) {
-	//out := []interface{}{""}
 	var out []interface{}
 	err := json.Unmarshal([]byte(v), &out)
 	if err != nil {
--- a/integrations/helm/release/utils.go
+++ b/integrations/helm/release/utils.go
@@ -1,0 +1,105 @@
+package release
+
+import (
+	"strings"
+
+	ifv1 "github.com/weaveworks/flux/apis/helm.integrations.flux.weave.works/v1alpha"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// collectValues ... assembles overriding parameters and outputs a
+// serialised map suitable for passing to helm client API.
+// Parameters with names containing "." will result in nested maps.
+// Maps from all parameters are merged together, with a possible overwriting
+// if a parameter is provided twice with different values.
+func collectValues(params []ifv1.HelmChartParam) ([]byte, error) {
+	base := map[string]interface{}{}
+	if params == nil || len(params) == 0 {
+		return yaml.Marshal(base)
+	}
+
+	for _, p := range params {
+		k, v := cleanup(p.Name, p.Value)
+		if k == "" {
+			continue
+		}
+
+		pMap := mappifyValueOverride(k, v)
+		base = mergeValues(base, pMap)
+	}
+
+	return yaml.Marshal(base)
+}
+
+func cleanup(k, v string) (string, string) {
+	k = strings.TrimSpace(k)
+	k = strings.Trim(k, "\n")
+
+	v = strings.TrimSpace(v)
+	v = strings.Trim(v, "\n")
+
+	return k, v
+}
+
+func reverse(s []string) []string {
+	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
+		s[i], s[j] = s[j], s[i]
+	}
+	return s
+}
+
+// mappifyValueOverride ... takes a parameter and its value, and creates
+// a corresponding map suitable for passing to helm client API
+// to override default values
+func mappifyValueOverride(k, v string) map[string]interface{} {
+	nests := reverse(strings.Split(k, "."))
+
+	inner := map[string]interface{}{}
+	outer := map[string]interface{}{}
+	for i, n := range nests {
+		switch i {
+		case 0:
+			inner[n] = v
+		default:
+			outer = map[string]interface{}{
+				nests[i]: inner,
+			}
+			inner = outer
+		}
+
+	}
+	return inner
+}
+
+// mergeValues ... merges two, possibly nested, maps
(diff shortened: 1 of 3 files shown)


tamarakaufler:
Fixes #982

tamarakaufler:
Fixes #982

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+			continue
+		}
+		nextMap, ok := v.(map[string]interface{})
+		// If it isn't another map, overwrite the value
+		if !ok {
+			dest[k] = v
+			continue
+		}
+		// If the key doesn't exist already, then just set the key to that value
+		if _, exists := dest[k]; !exists {
```
aaron7: This check is the same as the one on line 95 and will never be true.
The lines 99 to 104 don't modify `dest` or `k` so anything caught here would be caught by line 95.
Is this correct?

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+			inner = outer
+		}
+
+	}
+	return inner
+}
+
+// mergeValues ... merges two, possibly nested, maps
+// (copied from kubernetes/helm/cmd/helm/install.go)
+func mergeValues(dest map[string]interface{}, src map[string]interface{}) map[string]interface{} {
```
aaron7: nit: `mergeMaps` is more understandable

integrations/helm/release/utils_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	ifv1 "github.com/weaveworks/flux/apis/helm.integrations.flux.weave.works/v1alpha"
+)
+
+var noopLogger = log.NewNopLogger()
+
+type mappifyTestCase struct {
```
aaron7: would it be easier to read tests if these structs are defined right before the test, or in the test case themselves (they don't need to be global)?

integrations/helm/release/utils_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+type unwrapTestCase struct {
+	desc     string
+	value    string
+	expected []interface{}
+}
+
+// fixtures for mappoifyValueOverride tests
+var mappifyExpectedSimple = map[string]interface{}{
```
aaron7: these fixtures can all be `const`s

integrations/helm/release/utils_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestUnwrap(t *testing.T) {
+	for _, c := range []unwrapTestCase{
+		{"list with simple values", `["AAA","BBB","CCC"]`, []interface{}{"AAA", "BBB", "CCC"}},
+		{"list with maps", `[{"key":"DDD","value":"ddd"},{"key":"EEE","value":"eee"}]`, []interface{}{map[string]interface{}{"key": "DDD", "value": "ddd"}, map[string]interface{}{"key": "EEE", "value": "eee"}}},
+	} {
+		testUnwrap(t, c)
+	}
+}
+
+func testMappify(t *testing.T, c mappifyTestCase) {
```
aaron7: would it be easier to read tests if this code is in the test function? this code is not used elsewhere.

integrations/helm/release/utils_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"image": map[string]interface{}{
+		"repository": map[string]interface{}{
+			"name": "bitnami/kibana",
+		},
+	},
+	"resources": map[string]interface{}{"memory": []interface{}{map[string]interface{}{"key": "aaa", "value": "val"}, map[string]interface{}{"key": "bbb", "value": "val2"}}}}
+
+// fixtures for unwrap tests
+func TestMappifyValueOverride(t *testing.T) {
+	for _, c := range []mappifyTestCase{
```
aaron7: would it be easier to read if the fixtures were in the test function? they don't seem to be shared and currently have to scroll up and down

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+	v = strings.Trim(v, "\n")
+
+	return k, v
+}
+
+func reverse(s []string) []string {
+	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
+		s[i], s[j] = s[j], s[i]
+	}
+	return s
```
squaremo: This reverses in place; either construct a fresh slice and return it, or have a void return.

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+// collectValues ... assembles overriding parameters and outputs a
+// map suitable for passing to helm client API.
+// Parameters with names containing "." will result in nested maps.
+// Maps from all parameters are merged together, with a possible overwriting
+// if a parameter is provided twice with different values.
+func collectValues(logger log.Logger, params []ifv1.HelmChartParam) (map[string]interface{}, error) {
+	base := map[string]interface{}{}
+	if params == nil || len(params) == 0 {
```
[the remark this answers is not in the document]
tamarakaufler: done

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	logger.Log("debug", fmt.Sprintf("override parameters in a data structure: %#v", base))
+
+	return base, nil
+}
+
+func cleanup(k, v string) (string, string) {
+	k = strings.TrimSpace(k)
+	k = strings.Trim(k, "\n")
```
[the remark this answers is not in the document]
tamarakaufler: done

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+// if a parameter is provided twice with different values.
+func collectValues(logger log.Logger, params []ifv1.HelmChartParam) (map[string]interface{}, error) {
+	base := map[string]interface{}{}
+	if params == nil || len(params) == 0 {
+		return base, nil
+	}
+
+	var vu interface{}
+	var err error
+	regx := regexp.MustCompile(`^\[.*\]$`)
```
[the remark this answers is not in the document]
tamarakaufler: done

integrations/helm/release/utils.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		// If we got to this point, it is a map in both, so merge them
+		dest[k] = mergeValues(destMap, nextMap)
+	}
+	return dest
+}
+
+// unwrap ... unmarshals a string that is a serialised list
+func unwrap(v string) (interface{}, error) {
+	out := []interface{}{""}
```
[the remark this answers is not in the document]
tamarakaufler: why is this considered superior?

integrations/helm/release/utils.go
tamarakaufler: done

integrations/helm/release/utils.go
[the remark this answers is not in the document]
squaremo: What you have there constructs a value which is subsequently thrown away. Declaring it without initialising it avoids that, and is more idiomatic therefore less jarring to the reader. (I read it and thought "why is this situation special?" before figuring out that it wasn't)

integrations/helm/release/utils_test.go
tamarakaufler: It's quite common to have all the definitions at the top of the test file. I shall move them just before the tests where they are used. With more complex test setup I think it's better to have the structs defined separately and refer to them in the test cases. I consider it more readable but perhaps it's a personal thing.

integrations/helm/release/utils_test.go
tamarakaufler: Constants can only be numbers, characters (runes), strings or booleans.

aaron7: yep, sorry this was wrong

integrations/helm/release/utils_test.go
tamarakaufler: I moved the fixtures towards the tests, so hopefully more readable.  If it still bothers you greatly I shall insert the fixtures inside (this was done from start for the unwrap tests as the test input is small). Though I would prefer not to have to do it :)

integrations/helm/release/utils.go
tamarakaufler: A good point. Will change.

integrations/helm/release/utils_test.go
aaron7: yep, if the structs are being used across multiple test cases then makes sense to define them at the top and use across multiple test cases.

otherwise it seems common to include the struct and test cases inside the test function itself. some examples:
- https://github.com/weaveworks/flux/blob/master/release/releaser_test.go#L137
- https://github.com/weaveworks/flux/blob/master/sync/sync_test.go#L69
- https://github.com/weaveworks/service/blob/master/users/tokens/tokens_test.go#L13
- https://github.com/weaveworks/cortex/blob/master/pkg/chunk/chunk_store_test.go#L105

integrations/helm/release/utils.go
tamarakaufler: done

integrations/helm/release/utils.go
tamarakaufler: This method was copied from kubernetes/helm/cmd/helm/install.go, so, to tell the truth, I did not spend much time looking at it.
You are right about your observation from what I can see. I shall write a test for this method to confirm it behaves.

integrations/helm/release/utils.go
tamarakaufler: If I made no changes to the function I think it would have been better to leave it named as it was originally (in kubernetes/helm) but as I am changing it (based on your feedback) it's fine to rename it.

integrations/helm/release/utils.go
tamarakaufler: Actually, need to think about it more. It's good to think ....

tamarakaufler: Removed the duplicate, renamed the function and added tests. The tests confirm you were right, that the functionality truly does not change whether or not those lines are there.

integrations/helm/release/utils_test.go
tamarakaufler: Is it ok to leave the tests (with the struct definition moved towards the relevant tests) this time as is?

integrations/helm/release/utils_test.go
tamarakaufler: no worries :+1:

integrations/helm/release/utils_test.go
aaron7: sure

tamarakaufler: Thank you.

squaremo:
NB this is waiting for it be determined whether #951 would be a better way forward (at addressing #982, and otherwise).

PR #1007, opened 2018-03-14
Generate keys in a separate tmpfs volume

### LINKED ISSUE #1002: fluxd dies with "read-only file system" error
_From @crw on March 6, 2018 23:41_

Part of my job is to watch every FullStory and send every trialist an email. One scenario I occasionally see is a user getting through setup to the point that the agents are "Installing," but the session ends without the agents having completed setup. This is frequently combined with the user bailing on the installation screen.

I do not know if our agents are in fact not installing, or if FullStory is not capturing data, or what. But, I am going to start collecting examples of this "problem" in this ticket. 

_Copied from original issue: weaveworks/service-ui#2064_

squaremo:
In Kubernetes >= 1.10, secrets (and config-maps) will be mounted read-only. This means we cannot use the tmpfs volume used for the deploy secret as a workspace for generating new keys (and that we have to mount the secret with the right mode, since we won't be able to `chmod` it).

Instead, allow _another_ tmpfs to be mounted, and use that when directed to. The new flag `--ssh-keygen-dir` is for providing the path; otherwise, it will use the path for the mounted secret, as before.

So we can still use stable paths, both possible stable locations of the private key (in the mounted secret or in the mounted tmpfs workspace) are mentioned in ~/.ssh/config (in the Docker image).

The example is also updated to reflect the new way of doing things, compatible with Kubernetes 1.10; existing configs will continue to work.

Fixes #1002.

### DIFF
--- a/ssh/keygen.go
+++ b/ssh/keygen.go
@@ -2,6 +2,7 @@ package ssh
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


squaremo:
@awh MBOI

ssh/keygen.go
```diff
@@ -156,9 +157,9 @@ type PublicKey struct {
 // ExtractPublicKey extracts and returns the public key from the specified
 // private key, along with its fingerprint hashes.
 func ExtractPublicKey(privateKeyPath string) (PublicKey, error) {
-	keyBytes, err := exec.Command("ssh-keygen", "-y", "-f", privateKeyPath).Output()
+	keyBytes, err := exec.Command("ssh-keygen", "-y", "-f", privateKeyPath).CombinedOutput()
```
squaremo: This is merely to provide a better error when it fails (e.g., when the file is missing).

PR #1011, opened 2018-03-16
Add ability to append [ci skip] to commit messages

### LINKED ISSUE #856: Excluding Flux commits from CI
Someone asked in weaveworks public slack:

> is there any way to control the commit message that flux uses when it contributes to a repo?
> the reason is: we use the repo that flux is automating deployments from to deploy additional resources 
> that aren’t tracked by flux — we have a process that publishes these resources on every commit (PR 
> merge) — but we don’t want to fire this when flux makes a change… we can filter `***NO_CI***`

squaremo:
Appending `[ci skip]` to a commit message will instruct CI to skip the
build for that commit, at least in these systems:

 - https://circleci.com/docs/1.0/skip-a-build/
 - https://docs.travis-ci.com/user/customizing-the-build/#Skipping-a-build
 - https://docs.gitlab.com/ee/ci/yaml/#skipping-jobs

So: I've added a flag which switches that will do exactly that, to all
commits fluxd creates. To account for other CI systems that have a
similar mechanism but don't use that convention, the flag
`--git-ci-skip-message` lets you supply the text to append.

Resolves #856.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -48,8 +48,9 @@ const (
 
 	// There are running systems that assume these defaults (by not
 	// supplying a value for one or both). Don't change them.
-	defaultGitSyncTag  = "flux-sync"
-	defaultGitNotesRef = "flux"
+	defaultGitSyncTag     = "flux-sync"
+	defaultGitNotesRef    = "flux"
+	defaultGitSkipMessage = "\n\n[ci skip]"
 )
 
 func optionalVar(fs *pflag.FlagSet, value ssh.OptionalValue, name, usage string) ssh.OptionalValue {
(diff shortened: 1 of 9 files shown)


cmd/fluxd/main.go
```diff
@@ -145,6 +148,12 @@ func main() {
 		}
 	}
 
+	switch {
```
sambooo: Took a (brief) moment to grok this with context, so for contextless future sam I think the intent would be more clearly expressed here with a plain if statement.

sambooo: Oh no github, we couldn't possibly want to see the lines _after_ this one, don't be silly...

squaremo: I don't know it seemed simpler to write it as a `switch`.

PR #1013, opened 2018-03-18
Add force flag for work on forced deployments

Eochs:
This will ensure the change will also be vendored into flux-api so it 
can handle the force-flag as well.

fixes weaveworks/service#1851


sambooo:
> fixes weaveworks/service#1851

Resolves, perhaps 😄

PR #1015, opened 2018-03-20
Make upstream.Close() wait for shutdown

squaremo:
We let main() exit, and block on graceful shutdown in a `defer`ed
procedure. We _also_ `defer` a call to `upstream.Close()`, which stops
the upstream reconnecting and closes it.

The two deferred procedure calls used to be in such an order that by
coincidence, `upstream.Close()` was _also_ blocked on graceful
shutdown. But this order was shuffled in PR #962.

This commit makes the `upstream.Close()` explicitly wait for shutdown.

Fixes weaveworks/service-conf#2053.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -398,7 +398,10 @@ func main() {
 				os.Exit(1)
 			}
 			daemon.EventWriter = upstream
-			defer upstream.Close()
+			go func() {
+				<-shutdown
+				upstream.Close()
+			}()
 		} else {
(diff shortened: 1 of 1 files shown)


squaremo:
> I did briefly wonder what'd happen if main exited before upstream.Close() were called. Seems fine to assume the runtime will clean up for us in that case.

Yeah; this is just a last-chance cleanup. (Better would be to use the same shutdown protocol for the upstream; let's save that for a bigger refactor)

PR #1017, opened 2018-03-22
Log the daemon version on startup

sambooo:
If a new flux release isn't adding to our checkpoint-api, there's some potentially misleading log output on daemon startup.

Specifically, we were seeing this output when running flux 1.2.5:
```
ts=2018-03-22T08:57:19.895914653Z caller=checkpoint.go:24 component=checkpoint msg="up to date" version=1.2.4
```


squaremo:
It'd be nice to change the uptodate log message too -- I always see it and think it's reporting the current version.

sambooo:
@squaremo I've changed the `version` label to `latest` instead, because I reckon that's more clear. go-kit wouldn't let me use `latest-version`.

Issue #1019, opened 2018-03-22
Delete all yaml-modifying code


sambooo:
We have the requirement that modifying yaml belonging to the user must preserve their comments, since the output may still be read by humans. There is no existing library for Go which supports this behaviour, and as a result we manipulate yaml with a collection of regular expressions in [`cluster/kubernetes/update.go`](https://github.com/weaveworks/flux/blob/f220230cb0b6c420babbb347a32899b27e7bb893/cluster/kubernetes/update.go#L129) and [`cluster/kubernetes/policies.go`](https://github.com/weaveworks/flux/blob/f220230cb0b6c420babbb347a32899b27e7bb893/cluster/kubernetes/policies.go#L80).

This code is difficult to maintain, changes to it are difficult to review, and it's difficult to be sure it does the right thing for all possible inputs.

We should delete it all and offload the yaml-modification to `kubeyaml`, as is being done in #976.

squaremo:
As of #976 being merged, we don't modify YAMLs from the go code any more.

sambooo:
Awesome!

squaremo:
For the record: https://github.com/weaveworks/flux/issues/1120

sambooo:
No `help wanted` labels in sight

squaremo:
> No help wanted labels in sight

Magic labeling machine waved!

PR #1020, opened 2018-03-22
Add warning comments to all YAML-modifying regular expressions

sambooo:
This should be sufficient to ensure that anyone considering modifying these bits
of code understands it's probably not a good idea.

See #1019

### DIFF
--- a/cluster/kubernetes/update.go
+++ b/cluster/kubernetes/update.go
@@ -119,6 +119,8 @@ func tryUpdate(def []byte, container string, newImage image.Ref, out io.Writer) 
 	}
 
 	// Detect how indented the "containers" block is.
+	// TODO: delete all regular expressions which are used to modify YAML.
+	// See #1019. Modifying this is not recommended.
 	newDef := string(def)
 	matches := regexp.MustCompile(`( +)containers:.*`).FindStringSubmatch(newDef)
 	if len(matches) != 2 {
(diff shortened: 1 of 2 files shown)


cluster/kubernetes/update.go
```diff
@@ -169,14 +175,16 @@ func tryUpdate(def []byte, container string, newImage image.Ref, out io.Writer)
...
+	// See #1019. Modifying this is not recommended.
 	// Replacing labels: these are in two places, the container template and the selector
 	// TODO: This doesn't handle # comments
 	// TODO: This encodes an expectation of map keys being ordered (i.e. name *then* version)
 	// TODO: This assumes that these are indented by exactly 2 spaces (which may not be true)
 	replaceLabelsRE := multilineRE(
 		`((?:  selector|      labels):.*)`,
 		`((?:  ){2,4}name:.*)`,
-		`((?:  ){2,4}version:\s*) (?:` + optq + `[-\w]+` + optq + `)(\s.*)`,
+		`((?:  ){2,4}version:\s*) (?:`+optq+`[-\w]+`+optq+`)(\s.*)`,
```
squaremo: grr who wasn't running gofmt (possibly me)

sambooo: Checking `git blame` might surprise you... or not since you're aware of why we don't change this code.

PR #1027, opened 2018-03-27
Break dependencies among git, job, event packages

squaremo:
These all use each others types, and that makes it difficult to
refactor other packages without introducing cycles. To break the
dependencies, I needed to:

 - make git accept any (JSON serialisable) value as a note
 - stop referring to an event type in the job package, and thereby the
   API and client

### DIFF
--- a/daemon/note.go
+++ b/daemon/note.go
@@ -1,11 +1,11 @@
-package git
+package daemon
 
 import (
 	"github.com/weaveworks/flux/job"
 	"github.com/weaveworks/flux/update"
 )
 
-type Note struct {
+type note struct {
 	JobID  job.ID        `json:"jobID"`
 	Spec   update.Spec   `json:"spec"`
 	Result update.Result `json:"result"`
--- a/git/operations.go
+++ b/git/operations.go
@@ -128,7 +128,7 @@ func getNotesRef(ctx context.Context, workingDir, ref string) (string, error) {
 	return strings.TrimSpace(out.String()), nil
 }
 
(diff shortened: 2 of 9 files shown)


daemon/note.go
```diff
@@ -1,4 +1,4 @@
-package git
+package daemon
```
sambooo: `Note` can be unexported.

Github won't let me comment on an unchanged but moved file, because nothing notable could possibly happen when moving a file...

git/operations.go
```diff
@@ -128,28 +128,27 @@ func getNotesRef(ctx context.Context, workingDir, ref string) (string, error) {
...
 	b, err := json.Marshal(note)
 	if err != nil {
 		return err
 	}
 	return execGitCmd(ctx, workingDir, nil, "notes", "--ref", notesRef, "add", "-m", string(b), rev)
 }
 
-// NB return values (*Note, nil), (nil, error), (nil, nil)
-func getNote(ctx context.Context, workingDir, notesRef, rev string) (*Note, error) {
+// NB first return value indicates whether there was a note; second was an error
```
sambooo: Can this comment be replaced by named return values instead? `(ok bool, err error)` seems like a good choice.

daemon/note.go
squaremo: Yes, and so (elsehwere) can `daemonJobFunc`.

squaremo:
Thanks Sam. I've adjusted as requested.

PR #1032, opened 2018-03-29
Added some docs around tag image filtering

mikecann:
As mentioned in issue https://github.com/weaveworks/flux/issues/912 there are no docs around tag filtering and so as I mentioned in https://github.com/weaveworks/flux/issues/1025 I decided to go ahead and add some.


rade:
It would be nice to include examples of tag filters for versioned images since that is quite a common use case, i.e. "I want the the latest _released_ version (or maybe the latest released 1.x version), and not any random development image that happens to have been published to the image repo".

squaremo:
> It would be nice to include examples of tag filters for versioned images

Yeah maybe, but it would be better to have semver support, then write about _that_ :)

PR #1033, opened 2018-03-29
Refactor update and release packages to break some dependencies

squaremo:
~Just one commit -- the other will be rebassed away once #1027 is merged.~ (EDIT: the rebass is accomplished)

The main change here is to move policy update specs into update/ and their application into release/, thus putting them in the same place as image updates (i.e, releases) and treating them more or less uniformly.

Part of this was breaking the dependence of the cluster package on the update package, which itself was a whisker away from depending on cluster. This'll make it easier to move more things around.

### DIFF
--- a/cluster/kubernetes/policies.go
+++ b/cluster/kubernetes/policies.go
@@ -13,16 +13,16 @@ import (
 	"github.com/weaveworks/flux/policy"
 )
 
-func (m *Manifests) UpdatePolicies(in []byte, update policy.Update) ([]byte, error) {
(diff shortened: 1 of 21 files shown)


2opremio:
@squaremo are you planning to revive this at some point or should we close it?

squaremo:
> are you planning to revive this at some point or should we close it?

Good question -- aside from a bit of renaming faff, it does have some useful tidies. I'll give reviving a go, and close if that's at all tricky.

squaremo:
Oh yeah, this comes off rather badly in a collision with #2201. Never mind, maybe another time ..

PR #1034, opened 2018-04-01
Make mode of ssh identity files RO

mt-inside:
This causes flux-helm-operator not to try to chmod them to this mode,
which in kubernetes 1.9.4+ and 1.10+, breaks, as secrets volumes are now
read-only filesystems, to mitigate CVE-2017-1002102. This effectively
reapplied the same fix to the main fluxd deployment.

### DIFF
--- a/deploy-helm/helm-operator-deployment.yaml
+++ b/deploy-helm/helm-operator-deployment.yaml
@@ -17,6 +17,7 @@ spec:
       - name: git-key
         secret:
           secretName: flux-git-deploy
+          defaultMode: 0400 # when mounted read-only, we won't be able to chmod
       containers:
       - name: flux-helm-operator
         # There are no ":latest" images for helm-operator. Find the most recent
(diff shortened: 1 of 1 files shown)


squaremo:
> does [the Helm operator's git-handling code] work OK if you make the secret volume read-only?

@tamarakaufler May be worth a look ^

mt-inside:
@squaremo I assumed the code was different, because the error was different :) And yeah, it was broken on 1.9.6 (which has the CVE fix making secrets volumes read-only if you want it or not), and this fixed it. I didn't find and read the code, but empirically it must be checking whether the mode is acceptable and not try to change it.

squaremo:
> it must be checking whether the mode is acceptable and not try to change it.

If it has a common ancestor with the fluxd code, then yes most likely this is what it does.

squaremo:
NB for future visitors, including myself: it is fine for this volume to be mounted read-only, because the helm operator does not attempt the key generation in the same way fluxd does.

mt-inside:
@squaremo I did wonder about that - fluxd now mounts an extra, emptyDir volume RW in order to be able to keygen. I wasn't sure if the helm operator would need that too. If that code ever gets ported across, I guess it will.

PR #1035, opened 2018-04-04
helm: Use chartutil.Values in FluxHelmReleaseSpec

### LINKED ISSUE #982: Helm Operator Doesn't Accept Variables

aaron7:
- Move to the nested map structure helm values uses
- Implement DeepCopyInto for chartutils.Values because the generation tool does not like `map[string]interface`
- The latest versions of dep create a new line for the package lists, hence the large change in Gopkg.lock

Fixes #982


aaron7:
Will update docs now...

aaron7:
will add some tests

site/helm/helm-integration.md
```diff
@@ -44,8 +44,7 @@ Flux-Helm Integration implementation consists of two parts:
     chartGitPath: mongodb
     releaseName: mongo-database
     values:
-      - name: image
-        value: bitnami/mongodb:3.7.1-r1
+      image: bitnami/mongodb:3.7.1-r1
```
tamarakaufler: It might be useful to add an example of a nested parameter and an example of a parameter with a values in a list format.

tamarakaufler: @aaron7 
Thanks for the added examples.
Can you please also add a case where the value of a parameter, whether nested or not is a list? Thanks.

apis/helm.integrations.flux.weave.works/v1alpha/types_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}
+
+	for i, tc := range testCases {
+		output := &FluxHelmValues{}
+		tc.original.DeepCopyInto(output)
+		assert.Exactly(t, tc.expectedCopy, tc.transformer(output), "copy was not mutated. test case: %d", i)
+		assert.Exactly(t, tc.expectedOriginal, tc.original, "original was mutated. test case: %d", i)
+	}
+}
```
tamarakaufler: :+1:

aaron7:
Tested with https://github.com/weaveworks/flux-helm-test/

apis/helm.integrations.flux.weave.works/v1alpha/types.go
```diff
@@ -21,15 +21,15 @@ type FluxHelmRelease struct {
 // FluxHelmReleaseSpec is the spec for a FluxHelmRelease resource
 // FluxHelmReleaseSpec
 type FluxHelmReleaseSpec struct {
-	ChartGitPath string         `json:"chartGitPath"`
-	ReleaseName  string         `json:"releaseName,omitempty"`
-	Values       FluxHelmValues `json:"values,omitempty"`
+	ChartGitPath   string `json:"chartGitPath"`
+	ReleaseName    string `json:"releaseName,omitempty"`
+	FluxHelmValues `json:",inline"`
```
tamarakaufler: This indicates that FluxHelmValues will slurp whatever fields are not chartGitPath or releaseName within the FluxHelmReleaseSpec part. However as FluxHelmValues struct has a field defined as `json:"values,omitempty"`, it will only pick "values". Is that correct? Has it been empirically confirmed that an additional field, different to chartGitPath, releaseName and values will be discarded? (I think it should be given the definition.)

aaron7: Here's an example showing multiple `json:",inline"`s and additional values: https://play.golang.org/p/T1D3YikyKQ2

I believe the schema that is used to unmarshal the string takes into account the inlines at that point. Additional values would be discarded if not defined in the struct as usual.

tamarakaufler: Thank you :+1:

stefanprodan:
This is a breaking change, maybe we should increment the version to 

helm.integrations.flux.weave.works/v1alpha2

tamarakaufler:
A very good point.

tamarakaufler:
On the other hand we are still quite at the beginning

stefanprodan:
We have a couple of users on Slack that are using the Helm operator, I think we should increment the version each time we are introducing a breaking change. Also the current version `v1alpha` should be `v1alpha1` if we want to follow the Kubernetes convention.

tamarakaufler:
👍 I shall create a ticket

tamarakaufler:
https://github.com/weaveworks/flux/issues/1041 created

PR #1046, opened 2018-04-17
Allow namespace tag policies

### LINKED ISSUE #1045: Add higher-level defaults for tag filters
Suppose you build all your images with a tag pattern like `branch-commithash`, then you might want to set a cluster-wide default filter of `master-*`.

Possibly you might want a per-namespace default.

aaron7:
Fixes #1045.

- If no tag policy is set on the service, we check if one exists on the namespace.
- This PR does not implement updating namespace tags via flux.
- Decided to create new functions to fetch Namespace objects - we could instead include namespace objects as a service resource kind and remove some of the duplicated code I've added here? Otherwise the separation works well for now.

e.g.
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: cortex
  annotations:
    flux.weave.works/tag.cortex: glob:master-*
```

### DIFF
--- a/cluster/kubernetes/files.go
+++ b/cluster/kubernetes/files.go
@@ -29,3 +29,23 @@ func (c *Manifests) FindDefinedServices(path string) (map[flux.ResourceID][]stri
 	}
 	return result, nil
 }
+
+// FindNamespaces finds all the namespaces defined under the directory
+// given, and returns a map of namespace IDs (default:namespace:<namespace>)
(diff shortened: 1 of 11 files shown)


rade:
> This PR does not implement updating namespace tags via flux.

Please file an issue.

rade:
Also, surely we should document this.

aaron7:
Closing for now - can reopen/refer to in the future.

Issue #1048, opened 2018-04-23
Helm Operator removes all releases from a cluster 


stefanprodan:
When changing a field in the values section inside a FluxHelmRelease file the Helm Operator:
a) deletes all Helm releases including itself if it's installed with Helm
b) deletes a release whose related FHR was updated, instead of upgrading

When a FluxHelmRelease is updated the Operator deletes all Helm releases from a cluster. I would expect for the Operator to only instrument the releases managed with FluxHelmReleases not all releases. Also if a FluxHelmRelease values has been modified the Operator should perform a release upgrade not a delete. 

Affected version: `master-6f427cb`

How to reproduce:

```
helm repo add sp https://stefanprodan.github.io/k8s-podinfo

# Fork https://github.com/stefanprodan/weave-flux-helm-demo

helm install --name cd \
--set git.url=git@github.com:MY-USENAME/weave-flux-helm-demo \
--set helmOperator.create=true \
--namespace flux \
sp/weave-flux

# find the SSH public key with:
export FLUX_POD=$(kubectl get pods --namespace flux -l "app=weave-flux,release=cd" -o jsonpath="{.items[0].metadata.name}")
kubectl -n flux logs $FLUX_POD | grep identity.pub | cut -d '"' -f2 | sed 's/.\{2\}$//'

# copy the public key and create a deploy key with write access on the GitHub repository

# wait for the first sync
# increment the replica count in MY-USENAME/weave-flux-helm-demo/config/frontend.yaml
# wait for the 2nd sync

helm list

# all helm releases have been deleted without purging including weave-flux
```

tamarakaufler:
helm-integration image based on the latest flux master has manual release sync implemented. This has the following implications:
```
a) 
helm-operator checks at regular intervals whether there are:

      aa) any new manually installed Chart releases => deletes them
            The assumption is that all cluster releases are described/configured through FHRs. 
            Not ideal and will be removed.
      ab) deleted Chart releases which have its counterpart in the provided git repo => 
            reinstalled if these had been purged
      ac) manually upgraded releases are reverted to correspond to their corresponding 
            FHR declaration
            
b)
helm-operator and flux agent should not be currently Chart releases:
      This happens due to aa). When aa) is fixed, then there can be any additional 
      Chart releases that do not have their FHR counterparts. 
```

PR #1049, opened 2018-04-23
Removes delete action from helm-operator manual release sync

tamarakaufler:
The original assumption was that user's cluster only contains
Chart releases defined by FluxHelmReleases stored in git repo.
As a result the provided implementation deleted any release
not having a FHR counterpart. This resulted in unwanted deletions
such as helm-operator deleting itself when deployed as a Chart etc


stefanprodan:
This will fix #1048

PR #1052, opened 2018-04-24
Push helm operator release versions

aaron7:


.circleci/config.yml
```diff
@@ -33,12 +33,16 @@ jobs:
               docker login -u "$DOCKER_REGISTRY_USER" -p "$DOCKER_REGISTRY_PASSWORD" quay.io
               docker push "quay.io/weaveworks/flux:$(docker/image-tag)"
             fi
+            if echo "${CIRCLE_TAG}" | grep -Eq "helm-^[0-9]+(\.[0-9]+)*(-[a-z]+)?$"; then
```
tamarakaufler: Shouldn't this be:

"^helm-[0-9]+(\.[0-9]+)*(-[a-z]+)?$"

?

tamarakaufler: When we merge into the master, the helm-operator gets the same tag as flux. Is that what we want?

squaremo: Yes, that's fine. Well, it's fine that they are both named for the commit and branch. It's a bit annoying that changes to the operator code result in a new (but actually unchanged) daemon image, and vice versa.

.circleci/config.yml
```diff
@@ -33,12 +33,16 @@ jobs:
...
 
 workflows:
   version: 2
   build-and-push:
     jobs:
       - build:
           filters:
-            tags:  
+            tags:
               only: /[0-9]+(\.[0-9]+)*(-[a-z]+)?/
```
squaremo: You may have to change this

.circleci/config.yml
tamarakaufler: :+1:   I remember we spoke about this problem a while ago.

aaron7: oops, thanks

squaremo: Still needs changing

.circleci/config.yml
squaremo: (.. and probably still needs changing)

.circleci/config.yml
```diff
@@ -33,12 +33,16 @@ jobs:
               docker login -u "$DOCKER_REGISTRY_USER" -p "$DOCKER_REGISTRY_PASSWORD" quay.io
               docker push "quay.io/weaveworks/flux:$(docker/image-tag)"
             fi
+            if echo "${CIRCLE_TAG}" | grep -Eq "helm-[0-9]+(\.[0-9]+)*(-[a-z]+)?$"; then
+              docker login -u "$DOCKER_REGISTRY_USER" -p "$DOCKER_REGISTRY_PASSWORD" quay.io
+              docker push "quay.io/weaveworks/helm-operator:$(docker/image-tag | cut -c 5-)"
```
squaremo:
```
$ echo helm-0.1.0-alpha | cut -c 5-
-0.1.0-alpha
```

(also, you may as well use ${CIRCLE_TAG} here)

PR #1066, opened 2018-05-04
Fix formatting in standalone install docs

millar:


squaremo:
Thank you @millar (and thanks @evnsio for checking it) :star:

Issue #985, opened 2018-03-07
Release images to FluxChartRelease resources


squaremo:
`FluxChartRelease` resources are effectively workload controllers, or will usually be, since they will typically

 - refer to a chart that includes at least one core workload controller
 - specify particular images and tags (versions) as `values`

To support release automation, the flux machinery will need to understand how to interpret these resources, and how to update them.

tamarakaufler:
## Investigation

Some possible formats of image info in Helm Chart's values.yaml:

```
image: memcached:1.5.6-alpine
```
```
image: itzg/minecraft-server
imageTag: latest
```
```
image:
  repository: quay.io/monax/db
  tag: 0.17.1
```
```
image:
  registry: docker.io
  repository: bitnami/mariadb
  tag: 10.1.32
```
```
artifactory:
  image:
    repository: "docker.bintray.io/jfrog/artifactory-pro"
    version: 5.9.1
```
```
jupyter:
  image:
    repository: "daskdev/dask-notebook"
    tag: "0.17.1"
```

tamarakaufler:
There is a lot of variation in how the image information can be provided in a Chart. We can provide solution that will cope with most of the variations. Alternatively we can select one format meaning users would need change their Charts to use this particular format of the image info.

A) The former will result in:

> higher complexity
>        vs
> allowing flexible image info format with generally no need to alter the Chart
> 

B) The latter option will result in:

> simpler implementation
> vs
> possible need to alter a Chart to cope with our chosen format(s)

I think the A) option is preferable, especially as it is possible to adjust the level of complexity based on which of the image info formats should be supported.  

#### Other than FluxHelmRelease custom resources

To extend image automation to other CRDs (eg https://github.com/openfaas-incubator/faas-o6s/blob/master/artifacts/figlet-armhf.yaml), we should also consider having the image info directly under the Spec, ie not under Spec.Values.

tamarakaufler:
### Implementation thoughts:

https://docs.google.com/document/d/1AoO8nMY_Ju9gh2LTgKgUbHFTEmV3ighDO4QKG2m9eTk/edit#heading=h.k9ts9ic80ezl

tamarakaufler:
The selected approach is to register a new resource kind and provide the implementation specifics for a kind during the registration.
github branch: https://github.com/weaveworks/flux/tree/985-release-images-to-FHRs-v2

 
[...]

PR #1074, opened 2018-05-10
Dedup parsing in UpdatePolicy and UpdateImage

squaremo:
This rewrites Kubernetes UpdatePolicies/UpdateImages to use the
generic resource parsing (in cluster/kubernetes/resource), rather than
doing its own. It's not possible to avoid entirely, because the way
the updating is written, it needs the annotations rather than the
policy.

As a side effect, this introduces a ResourceID into the arguments for both methods; this will be useful for supporting multidocs and lists, later.

The internals of UpdateImage can also be rewritten, to a lesser
extent. I also removed some of the code that dealt with
ReplicationControllers, which we have not supported for a while.

### DIFF
--- a/cluster/kubernetes/manifests.go
+++ b/cluster/kubernetes/manifests.go
@@ -1,6 +1,7 @@
 package kubernetes
 
 import (
+	"github.com/weaveworks/flux"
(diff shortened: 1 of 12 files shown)


tamarakaufler:
A nice improvement! Helps with the #985 ticket, makes the code clearer :sparkles:

PR #1076, opened 2018-05-13
Fix #1067 by filtering out chart dirs from manifest loading of changed files

### LINKED ISSUE #1067: Not all ways of loading manifests skip over chart files
From @stefanprodan in #993:

I discovered that not all files inside the templates dir are ignored. For example a PodDisruptionBudget gets applied even if it's part of a chart.

```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: {{ template "memcached.fullname" . }}
spec:
  selector:
    matchLabels:
      app: {{ template "memcached.fullname" . }}
      chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
      release: "{{ .Release.Name }}"
      heritage: "{{ .Release.Service }}"
  minAvailable: {{ .Values.pdbMinAvailable }}
```

With the above file, Flux will run into this error:

ts=2018-05-07T08:44:46.764045009Z caller=loop.go:89 component=sync-loop err="loading resources from repo: parsing file at \"/tmp/flux-working640881799/repo/charts/memcached/templates/pdb.yaml\": parsing YAML doc from \"charts/memcached/templates/pdb.yaml\": yaml: unmarshal errors:\n  line 4: cannot unmarshal !!map into string"

A chart example can be found here: https://github.com/stefanprodan/weave-flux-helm-demo/tree/master/charts/memcached

ncabatoff:
I haven't written any tests yet because I first wanted to validate with the project maintainers whether this was the right approach.  Another way of tackling this might be to filter out everything under the --git-charts-path option.  Though in that case maybe we should also change the other manifest loading code path?

### DIFF
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -28,11 +28,11 @@ func Load(base, atLeastOne string, more ...string) (map[string]resource.Resource
 				return errors.Wrapf(err, "walking %q for yamels", path)
 			}
 
-			if charts.isChart(path) {
+			if charts.isDirChart(path) {
 				return filepath.SkipDir
 			}
 
-			if charts.inChart(path) {
+			if charts.isPathInChart(path) {
 				return nil
 			}
 
@@ -88,11 +88,11 @@ func newChartTracker(root string) (chartTracker, error) {
 	return chartTracker(chartdirs), nil
 }
 
-func (c chartTracker) isChart(path string) bool {
+func (c chartTracker) isDirChart(path string) bool {
 	return c[path]
 }
 
-func (c chartTracker) inChart(path string) bool {
+func (c chartTracker) isPathInChart(path string) bool {
 	p := path
 	root := fmt.Sprintf("%c", filepath.Separator)
 	for p != root {
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -24,7 +24,7 @@ func Load(base, atLeastOne string, more ...string) (map[string]resource.Resource
 				return errors.Wrapf(err, "walking %q for yamels", path)
 			}
 
-			if info.IsDir() && looksLikeChart(path) {
+			if info.IsDir() && LooksLikeChart(path) {
 				return filepath.SkipDir
 			}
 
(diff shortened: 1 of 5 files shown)


squaremo:
Hi @ncabatoff , thanks for the PR :star:

With respect to the approach: the `Manifests` interface is intended to be platform neutral, i.e., not particular to Kubernetes (notwithstanding, there are no implementations other than for Kubernetes). Charts are Kubernetes-specific concern. So, I would prefer if they were not mentioned in that interface.

There are a couple of ways this can be achieved, I think.

1. Just make sure that all methods in `cluster/kubernetes` that obtain `Resource`s from files, skip charts. In other words, this bakes in the fact that files within charts shouldn't be interpreted as manifests. To my mind this is the right way to go -- the current code in master is really an incomplete implementation of this idea.

2. A fallback might be to decide on a more neutral name for `LooksLikeChart` -- say, `ShouldBeIgnored` -- and its derivatives, so that other (hypothetical) implementations of `Manifests` could also implement it. But then you'd have to ask "If this always gets called after loading manifests, why don't you just do it all the time in the implementation of loading manifests?" (i..e, why don't you just do 1.) above).

ncabatoff:
Fair points.  I like this approach better too.

cluster/kubernetes/resource/load.go
```diff
@@ -18,16 +18,24 @@ import (
...
+		return nil, errors.Wrapf(err, "walking %q for chartdirs", base)
+	}
 	for _, root := range roots {
 		err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
 			if err != nil {
 				return errors.Wrapf(err, "walking %q for yamels", path)
 			}
 
-			if info.IsDir() && looksLikeChart(path) {
+			if charts.isChart(path) {
```
squaremo: `isChart` and `inChart` are a little too similar -- I can imagine typo-flips between them. How about `isDirChart` and `isPathInChart`, or similar?

ncabatoff:
Thanks, I'm glad you're happy with the revised version.  

Re the ambiguous method names: agreed, I had that same thought when I was writing them and then forgot to fix before pushing.

Re going up as far as the filesystem root, rather than just the chart root: I think for safety we need to check for when we reach the real filesystem root anyway.  Consider what happens if the method is called with a path argument that doesn't match the root the chartTracker was constructed from; that would be a bug, I admit, but it would be a bad bug (infinite loop) if we don't have the loop terminate upon reaching the fs root.  So given that we need that test anyway for safety, and we'd only be saving a few extra I/Os by also testing to see if we've reached the root the chartTracker was constructed from, and performance isn't critical here, I prefer the current simpler approach.

squaremo:
> Consider what happens if the method is called with a path argument that doesn't match the root the chartTracker was constructed from; that would be a bug, I admit, but it would be a bad bug (infinite loop) if we don't have the loop terminate upon reaching the fs root.

I suppose I am a bit nervous about `/` being particular to Unix-like filesystems -- not that anything in flux runs on Windows anyway!

You are right that individual path arguments aren't necessarily in the root path. They should either be supplied as relative to the root, or it should be an error to supply a path that isn't contained. I was probably a bit confused about the requirements while I wrote this, leading to neither of these being enforced. It can be the subject of a later PR, anyway.

squaremo:
Thanks @ncabatoff :star:

PR #1080, opened 2018-05-15
Add `fluxctl sync` command

### LINKED ISSUE #1079: fluxctl command to sync while-u-wait
As mentioned in https://www.youtube.com/watch?v=aQz3H9bIH8Y, it's occasionally useful -- e.g., in a demo -- to be able to tell flux to pull from upstream and sync _right now_.

The very least that could be done is to expose the git hook in the daemon API, and add a command that pokes it. However, as a `fluxctl` command you would expect it to do what you tell it synchronously; in other words, to report the HEAD it pulls, and when it gets synced.

To do that first bit, the fetch really has to be a job, which `fluxctl` can then poll for the result. Provided it gets a commit ID back, it can then call the API SyncStatus to check when it's been applied.

squaremo:
This adds the ability to manually trigger a sync (pull commits from
the git origin, and apply to cluster), and a command `fluxctl sync`
which triggers a manual sync and reports its progress.

Since a sync is potentially a long-running task, it's run using the
job machinery: the command posts an `update.Spec`, which is submitted
as a job and the ID returned. The command then proceeds more or less
as it would for a release or policy change.

Because the different kinds of update require different handling, I've
factored the creation of jobs out into some combinators (or near
combinators):

 - run a job that needs a git clone
 - run a job that results in a commit, and report the result back to
   the upstream server

Fixes #1079.

### DIFF
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -67,6 +67,7 @@ func (opts *rootOpts) Command() *cobra.Command {
 		newControllerPolicy(opts).Command(),
 		newSave(opts).Command(),
 		newIdentity(opts).Command(),
+		newSync(opts).Command(),
 	)
 
 	return cmd
--- a/cmd/fluxctl/sync_cmd.go
+++ b/cmd/fluxctl/sync_cmd.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/weaveworks/flux/git"
+	"github.com/weaveworks/flux/update"
+)
+
+type syncOpts struct {
+	*rootOpts
+}
+
+func newSync(parent *rootOpts) *syncOpts {
+	return &syncOpts{rootOpts: parent}
+}
+
+func (opts *syncOpts) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sync",
+		Short: "synchronize the cluster with the git repository, now",
+		RunE:  opts.RunE,
+	}
+	return cmd
+}
+
+func (opts *syncOpts) RunE(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return errorWantedNoArgs
+	}
+
+	ctx := context.Background()
+
+	gitConfig, err := opts.API.GitRepoConfig(ctx, false)
+	if err != nil {
+		return err
+	}
+
+	switch gitConfig.Status {
+	case git.RepoNoConfig:
+		return fmt.Errorf("no git repository is configured")
+	case git.RepoReady:
+		break
+	default:
+		return fmt.Errorf("git repository %s is not ready to sync (status: %s)", gitConfig.Remote.URL, string(gitConfig.Status))
+	}
+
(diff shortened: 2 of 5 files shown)


cmd/fluxctl/sync.go
```diff
@@ -0,0 +1,74 @@
...
+		return fmt.Errorf("git repository %s is not ready to sync (status: %s)", gitConfig.Remote.URL, string(gitConfig.Status))
+	}
+
+	fmt.Fprintf(cmd.OutOrStderr(), "Synchronizing with %s\n", gitConfig.Remote.URL)
+
+	updateSpec := update.Spec{
+		Type: update.Sync,
+		Spec: update.ManualSync{},
+	}
+	jobID, err := opts.API.UpdateManifests(ctx, updateSpec)
```
sambooo: Should this be something other than an `UpdateManifests` call? No manifests are updated in the repo so this feels like existing-API abuse.

squaremo: I considered adding another API method just for the sync call, but since it had almost everything in common with updates (goes in the job queue, result is a revision, which you then wait for), I decided to use the same mechanism. There are some small mismatches, aside from the name -- there's no cause, or `update.Result`, associated with the sync.

And, to be honest, I felt dread at the prospect of adding an API method.

Arguably, this method should really be called `SubmitJob`. Maybe in the next API version we can finesse it a little.

squaremo:
`frontend.dev.weave.works` won't understand the new bit of API until we merge this and revendor it. To try it without going through a service, use a port-forward:

```
kubectl port-forward -n weave deployment/flux 3030:3030 &
fluxctl --url "http://localhost:3030/api/flux" sync
```

foot:
> you may need to alter the deployment/flux to suit how you're running flux

I had a fiddle, adding a port:3030 section on deployment.containers as in the fluxsrc/deploy folder but the port forward connection was refused w/ curl/fluxctl via localhost:3030. How else might you alter `deployment/flux`?

foot:
Oh right I think I see what you mean now sorry. The command rather than the deployment itself.

It might be the katacoda k8s env. I will try again w/ minikube in a bit.

foot:
Current:

```
Synchronizing with git@github.com:foot/flux-example
Job ID e49b2c70-23ec-906b-2cfc-21b7552ec5e2
HEAD of master is cd3b7425ea968dea0396dba04c0b542777c313d7
Applied cd3b7425ea968dea0396dba04c0b542777c313d7
Done.
```

With more ellipsis's. 
```
Synchronizing with git@github.com:foot/flux-example
HEAD of master is cd3b7425ea968dea0396dba04c0b542777c313d7
Applying cd3b7425ea968dea0396dba04c0b542777c313d7 to cluster...
Done.
```

Or on error:
```
Job ID: e49b2c70-23ec-906b-2cfc-21b7552ec5e2 failed: "cannot ff"
```

foot:
Oh yeah `--help` has `[flags]` in it. Add docs about flags or remove that reference perhaps.

squaremo:
> Oh yeah --help has [flags] in it. Add docs about flags or remove that reference perhaps.

Ah yeah, a bit odd when there are no flags (aside from help itself -- meta!). But consistent with the other usage messages, at least.

Issue #913, opened 2018-01-24
/api/flux/v6/images?service=<all> returns context deadline exceeded


foot:
The query is probably too big to lookup/encode/transmit in 10s.

Use cases: 
 - "Deploy Status" Dashboard
 - building web-UI equivalent `fluxctl release --all --update-image=library/hello:v2`

We need to choose a subset of the tags.

Ideal response:
 - `/images?service=<all>&tagCreatedAt=gtCurrent`: Include all the tags where tag.CreatedAt > Current.CreatedAt
     - This would allow us to display in the UI:
        - "You are 4 tags behind latest"
        - "You are 3 weeks behind latest tag"

Small caveat: the ideal response might be too big sometimes? The next idealist response would be:
 - `/images?service=<all>&tagLimit=1&onlyFilteredTags=true`: include 1 tag for each image that matches whatever tagFilter is set.
    - Which would allow us to say in the ui:
        - "You are 3 weeks behind latest tag"

foot:
If we could declare a "tagsBefore" and "tagsAfter" additionally to "tagsLimit" (mirroring the /history?params) we would have a lot flexibility in the UI to overlay image information in a timeline like manner.

foot:
Migration strategy would be to:
  1. Try the current api w/ the extra params.
     - expect an un`limit`ed and un`before/after`ed response from the BE even if the values are provided and snip on the FE side.
     - expect a context-deadline exceeded most of the time. sock-shop/dev respond with one. - maybe takes 10s, then show the old ui, which a "maybe upgrade fluxd?" message? Can't say w/ confidence this was a freak context-deadline-exceeded.
  2. Bump the api:
     - add a `v7/images?` endpoint that supports more options, if we immediately get a 404 show old ui and show a "please upgrade" dialog with confidence.

I like 2.

In the future we could respond w/ a 400 if the client provides extra unsupported query string params, then the client would know its an old fluxd even if the api hasn't technically "broken". Some fluxd's will support more options that the client can make use of.

Does this work when deprecating options? E.g. no more before/after, please use something else? Old fluxd will respond w/ "i don't have support that prop".... oft, maybe just bumping api versions is nicer.

squaremo:
A bit of evidence for what's happening ..

If I curl the API at our dev environment, I get context exceeded (i.e., a timeout). If I port-forward to the fluxd pod and curl the same API endpoint, I get a full result -- it's about 3MB.

This makes me suspect one of a couple of causes:
[...]

PR #1084, opened 2018-05-17
Add ListImages summary fields

aaron7:
Part of #913 to allow a HTTP client to receive image summary information instead of fetching all of the images.

- Creates new fields in `containers`:
  - `LatestFiltered` - image.Info for the latest image
  - `AvailableImagesCount` - count of all available images
  - `NewAvailableImagesCount` - count of all new available images
  - `FilteredImagesCount` - count of all images after tag filter
  - `NewFilteredImagesCount` - count of all new images after tag filter

- Allows specifiying which containerFields to return in the response via a `?containerFields=...` query parameter
  - e.g. `images?containerFields=Name,Current,AvailableImagesCount`

- Refactors some of the variable names around image fetching to make it clearer what is happening (image repo stuff)
- Changes `ListImages` interface and so we will have to update this in WC service
- Don't require queryParams like `?service=` - defaults to `<all>`

### DIFF
--- a/api/v6/api.go
+++ b/api/v6/api.go
@@ -42,19 +42,19 @@ type ControllerStatus struct {
 }
 
 type Container struct {
-	Name           string
-	Current        image.Info
-	LatestFiltered image.Info
+	Name           string     `json:",omitempty"`
+	Current        image.Info `json:",omitempty"`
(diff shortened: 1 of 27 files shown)


squaremo:
Nice work Aaron :star: I'm rerun the failing CI, which failed for a silly reason.

PR #1094, opened 2018-05-22
Verify container changes

### LINKED ISSUE #1081: Verify changes when updating manifests
The manifest update code (that changes the images for a deployment, e.g.) is quite fiddly and delicate. It occasionally fails, and occasionally -- worse -- makes unintended changes.

To make sure that latter doesn't happen, we could reload the manifests and check that only the intended changes were made.

squaremo:
This PR adds a verification step to the release process. The verification step checks that exactly the changes that were intended were made to the files in the git repo clone. This guards against the file update code doing something silly, like [changing every container in a deployment](https://github.com/weaveworks/flux/issues/1082), and this being committed.

Fixes #1081.

### DIFF
--- a/release/errors.go
+++ b/release/errors.go
@@ -19,7 +19,7 @@ deal with. See
 for those limitations.
 
 If your files appear to meet the requirements, it may simply be a bug
-in Flux. PLease report it at
+in Flux. Please report it at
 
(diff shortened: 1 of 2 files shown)


squaremo:
@aaron7 PTAL, and I will autosquash if you're happy.

PR #1101, opened 2018-05-27
Add integration tests to test helm+flux

### LINKED ISSUE #919: Integration test in test/bin/test-flux has rotted
Repro steps:
1. Run `make integration-test`
2. Wait for it to time out

Expected:
It finishes

Actual:
It times out. Pod logs show `git clone: fatal: Could not read from remote repository.`

Root cause:
After #892, ssh config is written loaded into the container image from disk. In the integration test, we mount the `known_hosts` file to the `/root/.ssh` directory:
https://github.com/weaveworks/flux/blob/master/test/flux-deploy-all.yaml#L86

This overwrites the directory and nukes the ssh_config, which contains the location of the identity file mounted at /etc/fluxd/ssh/identity: https://github.com/weaveworks/flux/blob/master/docker/ssh_config

Setting `StrictHostKeyChecking no` and removing the `volumeMount` works around the issue, but that won't work long term.

Kubernetes issue: https://github.com/kubernetes/kubernetes/issues/51165
Possible working configuration (although it is a slightly different use case): https://github.com/dshulyak/kubernetes.github.io/commit/d58ba7b075bb4848349a2c920caaa08ff3773d70

ncabatoff:
See the README for more details.  This fixes #919, and adds tests for flux+helm.  Besides the tests copied from test-flux (test that the sync tag gets updated, test that automation works) it adds tests to ensure that charts in the git repo get deployed, that changes made to charts in the git repo result in a helm release upgrade, and that helm changes made outside of git get reverted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,8 @@ testdata/sidecar/sidecar-linux-amd64
 testdata/sidecar/.sidecar
 docker/fluxy-dumbconf.priv
 test/profiles
+test/deps
+test/bin/yq
+test/bin/kubectl
+test/bin/helm
+test/bin/minikube
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,44 @@
+# flux integration tests
+Integration tester for [flux](weaveworks/flux).
+
+This is a translation to Go of the bash script 
+[test-flux](https://github.com/weaveworks/flux/blob/master/test/bin/test-flux).
+
+# Usage
+
+```
+cd test && ./download-prereqs.sh && go test -tags integration_test -start-minikube=true
+```
+
+WARNING: This will blow away your existing minikube "minikube" profile.
+See below for why.
+
+## Current status
+
+The main differences with test-flux:
+- Not broken (see [#919](https://github.com/weaveworks/flux/issues/919))
+- By default doesn't start/delete minikube, assumes one is already running
+- Requires specific minikube and k8s versions
+- Deploys flux via a helm chart
+- Adds support for testing flux's helm-operator.
+
+## Background
+
+There's a PR ([#921](https://github.com/weaveworks/flux/pull/921)) which 
+attempts to fix some of the brokenness in test-flux.  I had many issues
+trying to get it to work for me, mostly relating to minikube and k8s.
+- the fix relies on subpath, which is broken in k8s 1.9 prior to 1.9.5
+  (see [#61076](https://github.com/kubernetes/kubernetes/issues/61076))
+- minikube with the default options has various problems in recent 
+  releases (0.26.0+)
+- minikube has broken the -profile option in recent releases
+  (see [#2717](https://github.com/kubernetes/minikube/issues/2717))
+- minikube with localkube doesn't support a 1.9 release >1.9.4
+- if I'm going to run minikube without localkube, I'd like to use the
+  latest and greatest
+
+What it boiled down to was that any fix I tried to make to test-flux would've
+been quiet disruptive already. Moreover, I was tired of the slow feedback
+loop from doing a minikube delete/start for each invocation, and I wanted to
+introduce more tests (e.g. for helm), and I've written enough shell to know
+when it's a good idea to stop and rewrite in a better language.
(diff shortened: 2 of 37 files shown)


squaremo:
> since I rely on the ssh-known-hosts configmap being mounted

Ahh yes I see, that does make it a bit tricky. Maybe @stefanprodan will be kind enough to comment here on how it could be adapted to serve both purposes (I know he is a fan of mounting `known_hosts` from a configmap, from #1113).

Aside from converging on the Helm chart, it would be really nice if we can get this to run in CI -- that would help avoid the situation that prompted it (the prior tests rotting). Can it be done with the `minikube --vm-driver=none` do you reckon? We've used this with some success elsewhere: https://github.com/weaveworks/launcher/blob/master/.circleci/config.yml#L87

squaremo:
>> since I rely on the ssh-known-hosts configmap being mounted
> 
> Ahh yes I see, that does make it a bit tricky. Maybe @stefanprodan will be kind enough to comment here on how it could be adapted to serve both purposes (I know he is a fan of mounting known_hosts from a configmap, from #1113).

The helm chart now has a parameter for supplying a host key, which gets mounted into the fluxd container. Is that enough to rework this PR around the chart herein? (I will attempt so, eventually, but feel free to beat me to it :-)

squaremo:
The `-profile` bug in minikube appears to have been fixed in 0.28.1 -- I've had some success reworking this branch a bit to use that version.

ncabatoff:
I'm finally coming back to this patch.  Thanks for pointing out the fix in 0.28.1.

I will look into working with the upstream charts soon.  For now my focus has been on getting this working with TravisCI, as I agree with you that should be done to prevent rot.  Yes, I'm almost certain it can be done with minikube --vm-driver=none.  Main change required is that since we don't have a host node to use as our git repo host any longer, I'm deploying a pod to be the git server instead.  I've got it working that way locally.  In travis I've got it to the point where kubernetes comes up and we can interact with it, but the tests are still timing out while waiting to see the flux-sync tag get updated.  I expect I should have that fixed and an updated PR issued within the week.

squaremo:
@ncabatoff Great! For testing against self-hosted repos, I mashed together https://github.com/squaremo/git-serv at one point -- I can adapt it to what's needed here, if that would be helpful.

ncabatoff:
I've been continuing my experiments in my flux-tester repo: I feel a bit more free to do whatever since it's more isolated from your repo and I still don't really know what I'm doing git-wise.

Sadly I fear circleci is not going to be viable for running k8s integration tests, at least with minikube.  I really wanted it to work, because the config format seems nicer and better documented and the system is much more responsive and reliable, based on the little I've used it so far.  

However:
- CircleCI 1.0 will not be supported as of August 31
- It's unclear what support they plan to provide for building in a VM environment instead of a docker container, especially at the free tier (they hint it may start costing money at some point)
- They've [responded](https://discuss.circleci.com/t/ubuntu-16-04-xenial-xerus-build-image-support/3585/12) to requests for Ubuntu 16 with "use a docker container in CircleCI 2.0"; it sounds like they'd prefer to drop the VM option
- Minikube with vm-driver none seems to depend on systemd and won't run in Ubuntu 14

An alternative might be https://github.com/kubernetes-sigs/kubeadm-dind-cluster, but it doesn't look anywhere near as stable or well supported as Minikube.

My experience today with travis, incidentally (it ran builds every time I tried a tweak to the circleci config.yml) was pretty positive today: out of 9 builds, 7 passed.  Unfortunately I think it's a bit hit or miss, but it may still be our best option for now.

squaremo:
Via @hiddeco: looks like it may still be possible to run `minikube --vm-driver=none` in CircleCI 2.0, but it needs a particular container image (not the Go image we use for building the binaries). I wonder if we can use jobs or workflows or whatever to do it. https://github.com/gavinzhou/ci-minikube

squaremo:
Rebased and modestly worked-on branch at `git checkout pr/1101`. I realised that working integration tests are a lot of value for #1382, and don't have to work in CircleCI straight away.

dholbach:
We have the Flux dev meeting coming up next week. One of the action item from last time was to take a look at this effort and see if the remaining work could be broken up into smaller pieces.

In broad strokes: what's still missing and where could people help?

squaremo:
We now have an end-to-end smoke test, which does not quite exercise as much of flux as here, but is likely to be the base on which more end-to-end tests are added.

PR #1106, opened 2018-05-29
Purge release on FHR delete

stefanprodan:
Currently, when a FluxHelmResource gets deleted, the corresponding Chart release is only marked as deleted in Tiller. This prevents our users to redeploy the same release at a later time. This PR adds the purge option to the delete operation.

### DIFF
--- a/integrations/helm/release/release.go
+++ b/integrations/helm/release/release.go
@@ -246,7 +246,7 @@ func (r *Release) Install(checkout *helmgit.Checkout, releaseName string, fhr if
 	}
 }
 
-// Delete deletes Chart release
+// Delete purges a Chart release
 func (r *Release) Delete(name string) error {
 	ok, err := r.canDelete(name)
 	if !ok {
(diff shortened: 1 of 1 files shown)


squaremo:
Why does Tiller only mark releases as "deleted" (i.e., what is the purpose of distinguishing between that and purging deletions)

stefanprodan:
The default behaviour is to prevent a release name to be reuse. I think the main reason would be to track what apps you have discontinued from your cluster.

stefanprodan:
The main problem with not purging releases is that you would expect from a GitOps Operator to undo any manual changes on a cluster. But if you delete a FHR by mistake, the Helm Operator will error out forever since Tiller will prevent it from applying the state from Git.

squaremo:
> you would expect from a GitOps Operator to undo any manual changes on a cluster.

What happens if someone deletes a release manually (without purging)? A different problem, I grant you ..

PR #1111, opened 2018-05-31
Support FluxHelmRelease resources in fluxd

### LINKED ISSUE #985: Release images to FluxChartRelease resources

squaremo:
This adds support for treating FluxHelmRelease releases as workloads (i.e., things that refer to images and result in pods being created). They will appear in ListImages and ListServices API results, and can be updated and automated like other workloads.

Fixes #985.

### DIFF
--- a/cluster/kubernetes/resource/fluxhelmrelease.go
+++ b/cluster/kubernetes/resource/fluxhelmrelease.go
@@ -7,6 +7,18 @@ import (
 	"github.com/weaveworks/flux/resource"
 )
 
+// ReleaseContainerName is the name used when flux interprets a
+// FluxHelmRelease as having a container with an image, by virtue of
+// having a `values` stanza with an image field:
+//
+// spec:
+//   ...
+//   values:
+//     image: some/image:version
+//
+// The name refers to the source of the image value.
+const ReleaseContainerName = "chart-image"
+
 // FluxHelmRelease echoes the generated type for the custom resource
 // definition. It's here so we can 1. get `baseObject` in there, and
 // 3. control the YAML serialisation of fields, which we can't do
@@ -14,8 +26,7 @@ import (
 type FluxHelmRelease struct {
 	baseObject
 	Spec struct {
-		ChartGitPath string `yaml:"chartGitPath"`
-		Values       map[string]interface{}
+		Values map[string]interface{}
 	}
 }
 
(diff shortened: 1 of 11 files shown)


squaremo:
Hmm that scheme of naming the container for the path is not great -- for one thing, it could have slashes, and container names are `dns_label`s I think. Either giving it a hard-wired name (suggestive of the format), or at least using the basename of the path, would be better.

cluster/kubernetes/resource/fluxhelmrelease.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Containers returns the containers that are defined in the
+// FluxHelmRelease. At present, this assumes only one image in the
+// Spec.Values, which is then named for the chart. If there is no such
+// field, or it is not parseable as an image ref, no containers are
+// returned.
+func (fhr FluxHelmRelease) Containers() []resource.Container {
+	values := fhr.Spec.Values
+	if imgInfo, ok := values["image"]; ok {
+		imgInfoStr := imgInfo.(string)
```
awh: Change this from a type assertion to a test?

squaremo:
> giving it a hard-wired name (suggestive of the format)

I came up with `chart-values-image`. That is suggestive of the format, but only if you know what the format is ... Just `image` would be less mysterious (but still mysterious).

I don't think there's any such name that's going to be entirely self-explanatory. It would be nice to synthesise a name that's meaningful to the chart, but we don't have the chart to hand, and using the path seems unreliable (what if someone supplies `"."` as the chart path?).

squaremo:
> giving it a hard-wired name (suggestive of the format)
> 
> I came up with chart-values-image. That is suggestive of the format, but only if you know what the format is ... Just image would be less mysterious (but still mysterious).

Settled on `chart-image`. Less weirdly specific; indicates how the image value is used. I need to build it into kubeyaml before this will work though.

squaremo:
> I need to build it into kubeyaml before this will work though.

(built into kubeyaml 0.3.1, which is used in the most recent commit)

awh:
:+1:

PR #1117, opened 2018-06-04
Sort resources using dependency order before applying

squaremo:
Kubernetes resource kinds have a partial order relation of (loosely) "may refer to": a Deployment may mount a ConfigMap as a volume; most resources are scoped to a namespace; a RoleBinding refers to a Role or ClusterRole; and so on.
    
Usually, you want the referenced resources to be created or changed before the referring resources. In general, Kubernetes is designed so that it will sort itself out eventually, but things go more smoothly if things are present before they are needed.

Here I've boiled the ordering down to a small number of ranks, with each rank containing kinds that refer only to kinds in the ranks before. When applying resources, they are sorted by rank, so that resources that depend on other resources will get updated or created after those others.

We _used_ to separate resources into those that don't mention a namespace and those that do, before applying each set in turn. This was for two reasons:
 - namespaces don't have a namespace, so these would come first
 - we wanted to supply a `--namespace` argument to kubectl, because otherwise un-namespaced resources would be created in the same namespace as flux; but, if you supply that argument, applying anything that _does_ have a namespace can be an error.

If we give kubectl a context by putting one in `/root/.kube/config` though, we don't have to use the `--namespace` argument, and can apply everything together.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ build/.%.done: docker/Dockerfile.%
 		-f build/docker/$*/Dockerfile.$* ./build/docker/$*
 	touch $@
 
-build/.flux.done: build/fluxd build/kubectl docker/ssh_config
+build/.flux.done: build/fluxd build/kubectl docker/ssh_config docker/kubeconfig
 build/.helm-operator.done: build/helm-operator build/kubectl docker/ssh_config
 
 build/fluxd: $(FLUXD_DEPS)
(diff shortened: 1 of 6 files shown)


sambooo:
> If we give kubectl a context by putting one in `/root/.kube/config` though, we don't have to use the `--namespace` argument, and can apply everything together.

Oh nice

PR #1129, opened 2018-06-08
Create new ListImagesWithOptions method in v10

aaron7:
We originally added options to `ListImages` in https://github.com/weaveworks/flux/pull/1084 however we found out we couldn't make this change because of the RPC method signature.

In this PR we revert the `ListImages` change and instead create `ListImagesWithOptions` in v10. For API versions before v10 `ListImagesWithOptions` will continue to work, with the WC service polyfilling the container fields filter option.


remote/logging.go
```diff
@@ -43,13 +44,22 @@ func (p *ErrorLoggingServer) ListServices(ctx context.Context, maybeNamespace st
...
 		}
 	}()
-	return p.server.ListImages(ctx, spec, opts)
+	return p.server.ListImages(ctx, spec)
+}
+
+func (p *ErrorLoggingServer) ListImagesWithOptions(ctx context.Context, opts v10.ListImagesOptions) (_ []v6.ImageStatus, err error) {
+	defer func() {
+		if err != nil {
+			p.logger.Log("method", "ListImages", "error", err)
```
squaremo: `"ListImagesWithOptions"`

remote/rpc/clientV10.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"io"
+	"net/rpc"
+
+	"github.com/weaveworks/flux/api/v10"
+	"github.com/weaveworks/flux/api/v6"
+	"github.com/weaveworks/flux/remote"
+)
+
+// RPCClientV10 is the rpc-backed implementation of a server, for
```
squaremo: Bit of copypasta in this comment? (referring to v8; if anything, it ought to refer to v9)

remote/rpc/clientV6.go
```diff
@@ -6,7 +6,11 @@ import (
 	"net/rpc"
 	"net/rpc/jsonrpc"
 
+	"github.com/weaveworks/flux/api/v10"
+	"github.com/weaveworks/flux/policy"
+
 	"github.com/weaveworks/flux/api/v6"
+	"github.com/weaveworks/flux/daemon"
```
squaremo: It is ... a bit strange to be importing the `daemon` package here, since it is near the top of the import graph (almost like a main package). Importing it here makes it more likely that a cycle can arise accidentally.

I'd put the adapter functions (`BuildContainer`, `GetTagPattern`) in one of the versioned API packages, and use them from both places.

api/v10/api.go
```diff
@@ -0,0 +1,25 @@
+// This package defines the types for Flux API version 9.
```
squaremo: Which version?

aaron7: thanks, fixed

Issue #1110, opened 2018-05-31
Show release status for FluxHelmRelease


squaremo:
When we list workload resources ("controllers"), we give a short status string. For e.g., deployments, this encodes the number of replicas that are ready. For `FluxHelmRelease`s, we could show the helm release status, which is available through the ReleaseStatus method of the helm API. *Where* we want to do that is a different story -- perhaps the operator could monitor the releases and update `fhr.Status`.

squaremo:
Fixed in #1134.

PR #1134, opened 2018-06-11
Link helm-created resources to FluxHelmReleases

squaremo:
This does two things:

 - it records the "release status" of a Helm release that corresponds to a FluxHelmRelease, in the FluxHelmRelease

 - it marks each resource that's created as a consequence of a FluxHelmRelease as linked to that FHR. This is then included in the ListControllers response in the API, as the field `Antecedent`, in the form of a stringified resource ID.

For example,

```json
[
  {
    "ID": "default:deployment/default-mariadb-mariadb",
    "Containers": [
      {
        "Name": "mariadb",
        "Current": {
          "ID": "bitnami/mariadb:_"
        },
        "LatestFiltered": {
          "ID": ""
        }
      }
    ],
    "ReadOnly": "NotInRepo",
    "Status": "ready",
    "Antecedent": "default:fluxhelmrelease/mariadb",
    "Labels": {
      "app": "default-mariadb-mariadb",
      "chart": "mariadb-2.1.3",
      "heritage": "Tiller",
      "release": "default-mariadb"
    },
    "Automated": false,
    "Locked": false,
    "Ignore": false,
    "Policies": {}
  },
  {
    "ID": "default:fluxhelmrelease/mariadb",
    "Containers": [
      {
        "Name": "chart-image",
        "Current": {
          "ID": "bitnami/mariadb:_"
        },
        "LatestFiltered": {
          "ID": ""
        }
      }
    ],
    "ReadOnly": "",
    "Status": "DEPLOYED",
    "Antecedent": "",
    "Labels": {
      "chart": "mariadb"
    },
    "Automated": true,
    "Locked": false,
    "Ignore": false,
    "Policies": {
      "automated": "true"
    }
  }
]
```


squaremo:
Fixes #1134 and fixes #1110.

integrations/helm/status/status.go
```diff
@@ -0,0 +1,116 @@
...
+						logger.Log("namespace", ns.Name, "resource", fhr.Name, "err", err)
+						continue
+					}
+				}
+			}
+		}
+
+		// Get the corresponding HelmRelease status
+		// Annotate/Status things
+		// (will probably need to keep a worklist)
```
aaron7: are these left over comments?

integrations/helm/status/status.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (a *Updater) Loop(stop <-chan struct{}, logger log.Logger) {
+	ctx := context.Background()
+	var logErr error
+
+bail:
+	for {
+		select {
+		case <-stop:
+			break bail
```
aaron7: this break is jumping before the for loop. When shutdown/stop occurs, should it be re-entering the loop?

I could be misunderstanding the use of the stop channel in this case.

integrations/helm/status/status.go
```diff
@@ -0,0 +1,116 @@
...
+		default:
+		}
+
+		if err := a.rateLimiter.Wait(ctx); err != nil {
```
aaron7: should we use a ticker here in this select loop so we can immediately stop when we get something on the stop channel?

integrations/helm/status/status.go
squaremo: `break label` exits the block so labelled; i.e., it jumps _after_ the loop.

integrations/helm/status/status.go
squaremo: Yes, probably _/me scurries away to throw the comments down the git memory hole_

integrations/helm/status/status.go
squaremo: A ticker would work as well, since we're using it in just this one site. I'll look at rewriting it.

integrations/helm/status/status.go
aaron7: ah yes, sorry

squaremo:
@aaron7 Wanna check the fixups? Then I'll squash them and meeerge.

Issue #1113, opened 2018-06-01
Include known_hosts in secret to make it a bit easier to mount into container


squaremo:
Either as the standard way of deploying it, or as an alternative means of overriding known_hosts (as in #724).

squaremo:
Maybe a docker-config too? (e.g., as generated from sidecars, like the workaround posted in #539)

NB one thing to keep in mind: if you mount a secret or configmap using subPath, the files won't get updated when the resources do.

hiddeco:
> NB one thing to keep in mind: if you mount a secret or configmap using subPath, the files won't get updated when the resources do.

You can get around this by specifying [`.configMap.items` in combination with a `mountPath`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#add-configmap-data-to-a-specific-path-in-the-volume) as [mounted ConfigMaps are updated automatically](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically).

The downside of this approach is that it overwrites all the original data in the `mountPath` (if any).

You can however get around this (again) by creating a symlink to your `mountPath` (e.g. `/home/flux/.ssh/known_hosts` -> `/home/flux/config/known_hosts`).

supra08:
@2opremio I would like to work on this issue.

PR #1142, opened 2018-06-13
Make it easier to mount known_hosts from a ConfigMap in Helm chart

squaremo:
Addresses #1113, and helps with #1101 by making the chart usable from the integration test (I think).

WIP until I've actually tried the Helm chart install myself.


squaremo:
Note to self, this may need to alter the howto bit of setup.md, and possibly FAQ.

stefanprodan:
@squaremo you could also bump Flux version to 1.4.0

squaremo:
PR for the first bit is #1154.

PR #1143, opened 2018-06-14
Make sure the secret creation example names file

squaremo:
I think at least a couple of people have run into this -- it's not obvious that the secret has to name the private key `identity`.


squaremo:
Thanks @rade :stars:

PR #1146, opened 2018-06-15
Fix flux ListImagesWithOptions v10

aaron7:
From https://github.com/weaveworks/flux/pull/1129

- Use an interface for dispatching ListImages and ListServices to the rpc clients
- Add RegisterDaemonV10 constant


remote/rpc/clientV6.go
```diff
@@ -119,7 +119,16 @@ func (p *RPCClientV6) ListImages(ctx context.Context, spec update.ResourceSpec)
 }
 
 func (p *RPCClientV6) ListImagesWithOptions(ctx context.Context, opts v10.ListImagesOptions) ([]v6.ImageStatus, error) {
-	images, err := p.ListImages(ctx, opts.Spec)
+	return listImagesWithOptions(ctx, p, opts)
+}
+
+type listImagesWithOptionsClient interface {
```
squaremo: Do you reckon it's worth a comment here?

PR #1156, opened 2018-06-20
Enable running out of cluster for faster development

andersjanmyr:


squaremo:
Nice idea! There's a few other things I think we need to handle, to be able to run outside the cluster (maybe you've already thought these through):

 - fluxd will try to create a secret to hold its SSH key. We'd need disable this behaviour (and just use the user/global SSH config)
 - flux expects a memcached to be available; I suppose it could just be a requirement that you have one running. But it currently looks for a SRV record in DNS, and it would still need to be told to use an exact host and port.

andersjanmyr:
@squaremo Yes, that sounds right. I'll update the PR to make this work.

2opremio:
@andersjanmyr any chance of having some time to finish up this PR?

andersjanmyr:
@2opremio I have been distracted by other things and I currently don't have time to work on this. Sorry. :(

hiddeco:
@squaremo can you think of any side-effects my last commit may have due to assumptions in other parts of Flux that there is a key ring present?

squaremo:
@hiddeco Am I reading it correctly, that `*kubeConfig` being empty distinguishes between using a keyring and not using a keyring? That's a bit dangerous, since it's possible to mount a kubeconfig into a pod, and use the argument to point at that. (Yes, the argument is a new one, but no doubt it would make its way into the chart, etc.)

It might be better to explicitly signal not to use a keyring with the `k8s-secret-*` arguments. These default to non-empty values, but perhaps an _empty_ value could be used to switch off the keyring. Or: an override switch, like `--k8s-secret-keyring=false` (flags that default to true can be tricky, though).

hiddeco:
> It might be better to explicitly signal not to use a keyring [...]

I agree.

> Or: an override switch, like `--k8s-secret-keyring=false` (flags that default to true can be tricky, though).

In this case I think it is the best option, as I can only imagine people setting it to `false` if they really know what they are doing (or want it to work this way), and I prefer having an explicit flag over logic tied to empty strings.

2opremio:
@hiddeco are you planning to revive this work or should I close this PR?

hiddeco:
This should work as is (except for the required rebase), SSH keyring backed by Kubernetes can be disabled with the flag, the `SSH_AUTH_SOCK` is inherited from the system, and gpg is configurable by setting the `GNUPGHOME` variable.

bzon:
Any update for this? I tried the same code change locally and **fluxd** works just fine. However, I do get **panic** errors though when **fluxctl** commands are invoked. For example; `fluxctl -u http://localhost:3030/api/flux identity`

hiddeco:
Closing this as #2140 has been merged.

PR #1158, opened 2018-06-21
More FAQs

squaremo:
Batching these ... (add to the list as they occur)

 - [x] how do I use my own git host?
 - [x] does flux delete resources when I remove them from git?
 - [x] Why don't my images show up?
   - ECR funny auth
   - Using digests instead of tags
   - "fat manifests" e.g., NATS
   - (also worth mentioning these in requirements.md and maybe troubleshooting.md)
 - [x] Do I have to have my app code and my config in the same git repo?
 - [x] Can I make flux sync with git faster or more often?
 - [x] Why does my CI pipeline keep getting triggered?
 - [x] Can I restrict the namespaces that flux can see / operate on?

### DIFF
--- a/site/faq.md
+++ b/site/faq.md
@@ -186,3 +186,45 @@ How to do this is documented in
 Not at present. It's tricky to come up with a safe and unsurprising
 way for this to work. There's discussion of some possibilities in
 [weaveworks/flux#738](https://github.com/weaveworks/flux/issues/738).
+
+### Why does my CI pipeline keep getting triggered?
+
+There's a couple of reasons this can happen.
+
+The first is that Flux pushes commits to your git repo, and if that
+repo is configured to go through CI, usually those commits will
+trigger a build. You can avoid this by supplying the flag `--ci-skip`
+so that Flux's commit will append `[ci skip]` to its commit
+messages. Many CI system will treat that as meaning they should not
+run a build for that commit. You can use `--ci-skip-message`, if you
+need a different piece of text appened to commit messages.
+
+The other thing that can trigger CI is that Flux pushes a tag to the
+upstream git repo whenever it has applied new commits. This acts as a
+"high water mark" for Flux to know which commits have already been
+seen. The default name for this tag is `flux-sync`, but it can be
+changed with the flags `--git-sync-tag` and `--git-label`. The
+simplest way to avoid triggering builds is to exclude this tag from
+builds -- how to do that will depend on how your CI system is
+configured.
+
+Here's the relevant docs for some common CI systems:
+
+ - [CircleCI](https://circleci.com/docs/2.0/workflows/#git-tag-job-execution)
+ - [TravisCI](https://docs.travis-ci.com/user/customizing-the-build#Building-Specific-Branches)
+ - [GitLab](https://docs.gitlab.com/ee/ci/yaml/#only-and-except-simplified)
+
+### Can I restrict the namespaces that Flux can see or operate on?
+
+Yes, though support for this is experimental at the minute.
+
+Flux will only operate on the namespaces that its service account has
+access to; so the most effective way to restrict it to certain
+namespaces is to use Kubernetes' role-based access control (RBAC) to
+make a service account that has restricted access itself. You may need
+to experiment to find the most restrictive permissions that work for
+your case.
+
+You will need to use the (experimental) command-line flag
+`--k8s-namespace-whitelist` to enumerate the namespaces that Flux
+attempts to scan for workloads.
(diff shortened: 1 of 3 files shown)


squaremo:
- [ ] Why don't my images show up?
   - ECR funny auth
   - Using digests instead of tags
   - "fat manifests" e.g., NATS

dholbach:
This looks good to me!

site/faq.md
```diff
@@ -64,6 +64,13 @@ This
 [flux (daemon) operator](https://github.com/justinbarrick/flux-operator)
 project may be of use for managing multiple daemons.
 
+### Do I have to have my application code and config in the same git repo?
```
rade: I feel that the preceding question should perhaps reference this one and/or re-iterate that when we are talking about "one git repo" we mean "one git repo for all your config", not "one git repo for all your applications".

squaremo: Good thought :+1:

stephenmoloney:
Great - Very useful information !

Some ones that I grappled with:

- How does flux poll a private docker image registry ?
- Do I need to set my credentails for the private docker image  registry ?
- How does flux know what images and image tag versions to pull from the private docker image registry?

justinbarrick:
If you’re still working on this, a good thing to add would be detailing the purpose of the sync tags. I still haven’t figured out what they’re for!

site/faq.md
```diff
@@ -112,3 +161,54 @@ setting. Flux will read this value and parse the Kubernetes secret.
...
+
+Not at present. It's tricky to come up with a safe and unsurprising
+way for this to work. There's discussion of some possibilities in
+[weaveworks/flux#738](https://github.com/weaveworks/flux/issues/738).
+
+### Why does my CI pipeline keep getting triggered?
+
+There's a couple of reasons this can happen.
+
+The first is that Flux pushes commits to your git repo, and if you
```
luisbarrueco: There's an extra `you` word that might have been mistakenly left after rephrasing.
This should read: `..., and if that repo is configured to go...`

squaremo:
I think I have covered everything that I set out to, plus extra suggestions made here, and adjustments pointed out. Gonna stop there in un-WIP it.

site/faq.md
```diff
@@ -104,11 +170,127 @@ Now restart fluxd to re-read the k8s secret (if it is running):
...
+[OpenContainers pre-defined annotations](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys).
+
+### How do I use a private git host (or one that's not github.com, gitlab.com, or bitbucket.com)?
+
+As part of using git+ssh securely from the Flux daemon, we make sure
+`StrictHostKeyChecking` is on in the
+[SSH config](http://man7.org/linux/man-pages/man5/ssh_config.5.html). This
+mitigates against man-in-the-middle attacks.
+
+We bake host keys for `github.com`, `gitlab.com`, and `bitbucket.com`
```
dholbach: `bitbucket.org`, right?

site/faq.md
```diff
@@ -104,11 +170,127 @@ Now restart fluxd to re-read the k8s secret (if it is running):
...
+`--git-sync-tag` and `--git-label`) to the git repository. This gives
+it a persistent high water mark, so even if it is restarted from
+scratch, it will be able to tell where it got to.
+
+Technically, it only needs this to be able to determine which image
+releases (including automated upgrades) it has applied, and that only
+matters if it has been asked to report those with the `--connect`
+flag. Future versions of Flux may be more sparing in use of the sync
+tag.
+
```
justinbarrick: You would need `—k8s-namespace-whitelist` set here or flux will fail to start when it can’t read some namespaces.

site/faq.md
squaremo: Owps! Thank you

site/faq.md
squaremo: Does this relate to the following section? I had not grasped that it fails utterly if you restrict the permissions, but don't also tell it to restrict its view.

squaremo: OK I tried this, by giving the flux service account these things:

 - a rolebinding to a do-anything role limited to a namespace
 - a clusterrolebinding to "get", "list" namespaces only

and flux responds to `list-controllers` with a permissions error. I'll add a note in the FAQ about using `--k8s-namespace-whitelist` (and maybe file a bug about doing better when we can't read things in a namespace).

squaremo: @justinbarrick If you get a chance, let me know if the fixup reads correctly -- ta!

justinbarrick: Looks good!

PR #1166, opened 2018-06-23
Mount the sshdir into the helm-operator too

stephenmoloney:
What is this MR for ?

- Mounting the sshdir into the helm operator container

- The known_hosts is only mounted in the flux container thus far.

I'm not 100% sure that the `known_hosts` are meant to be available in the `helm-operator` but 
my guess is that they are. Perhaps someone can confirm?

### DIFF
--- a/deploy-helm/helm-operator-deployment.yaml
+++ b/deploy-helm/helm-operator-deployment.yaml
@@ -14,6 +14,14 @@ spec:
     spec:
       serviceAccount: flux
       volumes:
+      # The following volume is for using a customised known_hosts
+      # file file, which you will need to do if you host your own git
+      # repo rather than using github or the like. You'll also need to
+      # mount it into the container, below.
+      # - name: sshdir
+      #   configMap:
+      #     name: flux-ssh-config
+      #     defaultMode: 0600
       - name: git-key
         secret:
           secretName: flux-git-deploy
@@ -26,6 +34,12 @@ spec:
         image: quay.io/weaveworks/helm-operator:0.1.0-alpha
         imagePullPolicy: IfNotPresent
         volumeMounts:
+        # Include this if you need to mount a customised known_hosts
+        # file; you'll also need the volume declared above.
+        # - name: sshdir
+        # mountPath: /root/.ssh/known_hosts
+        # subPath: known_hosts
+        # readOnly: true
         - name: git-key
           mountPath: /etc/fluxd/ssh
           readOnly: true # this will be the case perforce in K8s >=1.10
--- a/chart/flux/templates/helm-operator-deployment.yaml
+++ b/chart/flux/templates/helm-operator-deployment.yaml
@@ -24,6 +24,10 @@ spec:
       serviceAccountName: {{ template "flux.serviceAccountName" . }}
       {{- end }}
       volumes:
+      - name: sshdir
+        configMap:
+          name: {{ template "flux.fullname" . }}-ssh-config
+          defaultMode: 0600
       - name: git-key
         secret:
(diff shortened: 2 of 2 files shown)


stephenmoloney:
note to self: I should add this change to the helm-deploy folder too. Lets see if in principle this change will be accepted first, then I'll add it to the helm-deploy too.

squaremo:
Yes, this looks good and right.

I think we'll also need to change the helm-operator in line with https://github.com/weaveworks/flux/pull/1154, so that mounting into `/root/.ssh` doesn't wipe out the config from the image.

stephenmoloney:
I had an alternative thought on how to handle the known hosts.
Rather than bake the github.com, gitlab.com bitbucket.com hosts known_keys into the docker image, maybe a different approach might be to specify the hosts as a list in the yaml values in helm .
Then an init container will run and generate the known hosts at startup.
I kinda like this idea as from a user's perspective, entering multiline keys in the yaml under `known_hosts` is actually quite cumbersome.

squaremo:
> Rather than bake the github.com, gitlab.com bitbucket.com hosts known_keys into the docker image, maybe a different approach might be to specify the hosts as a list in the yaml values in helm .
> Then an init container will run and generate the known hosts at startup and save them into the /root/.ssh/known_keys file.

This would work. But the reason we switch on `StrictHostKeyChecking` though is that it prevents man-in-the-middle attacks, by making someone supply the key(s) ahead of time[1]. Scanning for the keys on startup undermines this, because you're introducing an opportunity for someone to give you a duff host key.

If you provided the fingerprint alongside the host name in the initial config, that would be enough to avoid that problem, while still being more convenient than assembling the configmap.

[1] We don't do this properly, mind you -- the docs ought to advise that you check the fingerprints; and, we really ought to check in a known_hosts file for github etc., rather than scanning for keys in the build.

stephenmoloney:
Ok, the `Dockerfile` gave me the impression that the practice of scanning for the hostkey was ok. 

So if it is not ok, the Dockerfile is missing a step to check the host keys - we should open an issue for that so that it is checked in the Dockerfile.

Also, in this case, the idea of an init_container doesn't work.

So using `known_hosts` is fine. 

Also you are right about the known hosts are added to the volume, they should really be appended to the existing file rather than outright overriding the existing file. I suppose that operation could be put into an init_container.

squaremo:
> So if it is not ok, the Dockerfile is missing a step to check the host keys - we should open an issue for that so that it is checked in the Dockerfile.

https://github.com/weaveworks/flux/issues/1169

stephenmoloney:
> I think we'll also need to change the helm-operator in line with #1154, so that mounting into /root/.ssh doesn't wipe out the config from the image.

I've changed it to use a subPath now and a specific file so the `/root/.ssh/config` crystallized inside the docker image is unaffected.

stephenmoloney:
Also, would you prefer I squash this or leave as is?

deploy-helm/helm-operator-deployment.yaml
```diff
@@ -14,6 +14,10 @@ spec:
     spec:
       serviceAccount: flux
       volumes:
+      - name: sshdir
+        configMap:
+          name: {{ template "flux.fullname" . }}-ssh-config
```
squaremo: This one isn't in a template. In the flux deployment YAML, I've commented it and the mount out, with an explanation, since they require the configmap to exist.

stephenmoloney:
@squaremo I think the git history is a bit cleaner now if you want to have another look.

squaremo:
Brill, thanks for your perseverance @stephenmoloney.

PR #1171, opened 2018-06-26
Include more information re "git repo not ready"

### LINKED ISSUE #1170: Be more specific with the feedback from git
When git has not quite been set up, because it needs a custom `known_hosts` or whatnot, the flux daemon constantly logs

    caller=images.go:21 component=sync-loop error="getting unlocked automated services: git repo not ready"

and this is something of a misdirection. In part because it's not the underlying problem, just the visible one; but mainly because it doesn't say anything about why.

Logging the git mirroring attempts, and what is not happening, would be a good start. Suppressing the above line, or at least only reporting on state changes, would also be useful.

squaremo:
Setting up git can be quite fiddly. One thing that would make it easier to troubleshoot is if fluxd did a better job of logging why it hadn't cloned the git repo yet.

This commit introduces a struct for not-ready-errors, so they can include the underlying problem. Since we now _expect_ there to be an underlying reason, create errors for "just haven't tried yet", and enforce the invariant of `not ready -> has error`.

Fixes #1170.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -81,8 +81,9 @@ func (d *Daemon) getPolicyResourceMap(ctx context.Context) (policy.ResourceMap, 
 
 	// The reason something is missing from the map differs depending
 	// on the state of the git repo.
+	_, notReady := err.(git.NotReadyError)
 	switch {
-	case err == git.ErrNotReady:
+	case notReady:
 		globalReadOnly = v6.ReadOnlyNotReady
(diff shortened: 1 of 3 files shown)


squaremo:
> I don't suppose it's easy to write some automated tests for these flows?

Fair question! Off the top of my head, it would be possible to set up various scenarios then make sure the git state machine arrived at the correct outcome. Slightly harder would be to check each transition, by stepping through a loop at a time.

PR #1184, opened 2018-07-02
Add --k8s-namespace-whitelist setting that specifies namespaces to watch.

### LINKED ISSUE #1181: Flux does not work properly for cluster with large number of namespaces
# The problem

Works great on all the clusters I manage until I get to a snowflake that has 30k+ namespaces. When that happens the daemon won't output any logs and will max out on it's limit of 50 requests per second to the API server.

I assume it is because it is trying to list any deployments happening in those namespaces.

# The solution

As clusters with this many namespaces are quite rare I would like to propose the following solution.

I think that being able to whitelist some namespaces that you care about for Flux management would be a good way to solve this. If no namespaces are whitelisted then it would default to its current behaviour. If any namesapces are provided then it will only look for objects within the specified namespaces. This means that current clusters are unaffected while large clusters can pick and choose.

# Symptoms and debugging

Example logs from cluster that has 30k+ namespaces:

```
us-east-1_sync --docker-config=/docker-creds/config.json --registry-trace=true
ts=2018-06-29T05:28:24.9841875Z caller=main.go:138 version=1.4.1
ts=2018-06-29T05:28:25.006700363Z caller=main.go:227 component=cluster identity=/etc/fluxd/ssh/identity
ts=2018-06-29T05:28:25.006744553Z caller=main.go:228 component=cluster identity.pub="ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCZ4gypLgAlw/Ja5SWTpXP26HoAvaswz7S+0550I8AdZx5bwGw23StBgFZ4KrfqNlpbtNPx9dwRLu4SK9uFaP5ocJ/9ghNfv6VoBtvdvOGmVZxljMrFAQza3adZef67WkOaV23FwPJ0EYfEKRK5EHR2Ju1IJWwaqsLqjmEoRYQGVUaNgcA76UQfFmNE8yQmvXcyVoBZc2Go3ILMjnk4RD3CK63I+EUhOiYTuyAuWY05Dc9XFctalNcpHzli/Eas/79qRyAeg1zyaAI/a7trULE9SPXlKXwfch7hZQvYRtfLqPKu5B8XjvfRga3sFHDETgLxtjmyALaHJFYe7AKLTiQP"
ts=2018-06-29T05:28:25.006774386Z caller=main.go:229 component=cluster host=https://10.155.0.1:443 version=kubernetes-v1.10.3
ts=2018-06-29T05:28:25.006843526Z caller=main.go:241 component=cluster kubectl=/usr/local/bin/kubectl
ts=2018-06-29T05:28:25.007820146Z caller=main.go:249 component=cluster ping=true
ts=2018-06-29T05:28:25.00939803Z caller=main.go:368 url=ssh://git@stash.atlassian.com:7997/kube/goliath-flux user="Weave Flux" email=support@weave.works sync-tag=bbci/prod/us-east-1_sync notes-ref=flux set-author=false
ts=2018-06-29T05:28:25.009441423Z caller=main.go:423 upstream="no upstream URL given"
ts=2018-06-29T05:28:25.009678217Z caller=loop.go:89 component=sync-loop err="git repo not ready"
ts=2018-06-29T05:28:25.009843024Z caller=images.go:15 component=sync-loop msg="polling images"
ts=2018-06-29T05:28:25.010617777Z caller=images.go:21 component=sync-loop error="getting unlocked automated services: git repo not ready"
ts=2018-06-29T05:28:25.010434008Z caller=main.go:440 addr=:3030
ts=2018-06-29T05:28:25.307131203Z caller=checkpoint.go:24 component=checkpoint msg="up to date" latest=1.4.1
ts=2018-06-29T05:28:30.360228703Z caller=loop.go:102 component=sync-loop event=refreshed url=ssh://git@stash.atlassian.com:7997/kube/goliath-flux branch=bbci/prod/us-east-1 HEAD=7bb72f2c1dff479e90fcec2e8a96f09aba65d6a3
```

Graphs showing Flux smashing the API server with requests.

screen shot 2018-06-29 at 4 27 23 pm

justinbarrick:
Add --k8s-namespace-whitelist setting that specifies namespaces to watch.
    
Fixes #1181
    
Currently, Flux expects to have access to all namespaces, even if no manifests
in the repository reference another namespace, it will check all namespaces
for controllers to update.
    
This change adds a --k8s-namespace-whitelist setting which, if set, will restrict
Flux to only watch the specified namespaces and ignore all others.
    
Intended for clusters with large amounts of namespaces or restrictive RBAC
policies. If provided Flux will only monitor workloads in the given namespaces.
This significantly cuts the number of API calls made.
    
An empty list (i.e. not provided) yields the usual behaviour.

### DIFF
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -175,7 +175,7 @@ func (c *Cluster) SomeControllers(ids []flux.ResourceID) (res []cluster.Controll
 // AllControllers returns all controllers matching the criteria; that is, in
 // the namespace (or any namespace if that argument is empty)
 func (c *Cluster) AllControllers(namespace string) (res []cluster.Controller, err error) {
-	namespaces, err := c.getNamespaces()
+	namespaces, err := c.getAllowedNamespaces()
 	if err != nil {
 		return nil, errors.Wrap(err, "getting namespaces")
 	}
@@ -261,7 +261,7 @@ func (c *Cluster) Ping() error {
 func (c *Cluster) Export() ([]byte, error) {
 	var config bytes.Buffer
 
-	namespaces, err := c.getNamespaces()
+	namespaces, err := c.getAllowedNamespaces()
 	if err != nil {
 		return nil, errors.Wrap(err, "getting namespaces")
 	}
@@ -375,7 +375,7 @@ func mergeCredentials(c *Cluster, namespace string, podTemplate apiv1.PodTemplat
 func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 	allImageCreds := make(registry.ImageCreds)
 
-	namespaces, err := c.getNamespaces()
+	namespaces, err := c.getAllowedNamespaces()
 	if err != nil {
 		c.logger.Log("err", errors.Wrap(err, "getting namespaces"))
 		return allImageCreds
@@ -413,12 +413,12 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 	return allImageCreds
 }
 
-// getNamespaces returns a list of namespaces that the Flux instance is expected
+// getAllowedNamespaces returns a list of namespaces that the Flux instance is expected
 // to have access to and can look for resources inside of.
 // It returns a list of all namespaces unless a namespace whitelist has been set on the Cluster
 // instance, in which case it returns a list containing the namespaces from the whitelist
 // that exist in the cluster.
-func (c *Cluster) getNamespaces() ([]apiv1.Namespace, error) {
+func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
 	nsList := []apiv1.Namespace{}
 
 	namespaces, err := c.client.Namespaces().List(meta_v1.ListOptions{})
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -170,13 +170,13 @@ func (c *Cluster) SomeControllers(ids []flux.ResourceID) (res []cluster.Controll
 // AllControllers returns all controllers matching the criteria; that is, in
 // the namespace (or any namespace if that argument is empty)
 func (c *Cluster) AllControllers(namespace string) (res []cluster.Controller, err error) {
-	namespaces, err := c.client.Namespaces().List(meta_v1.ListOptions{})
+	namespaces, err := c.getNamespaces()
 	if err != nil {
 		return nil, errors.Wrap(err, "getting namespaces")
 	}
 
 	var allControllers []cluster.Controller
-	for _, ns := range namespaces.Items {
+	for _, ns := range namespaces {
 		if namespace != "" && ns.Name != namespace {
 			continue
 		}
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -104,11 +104,12 @@ func isAddon(obj k8sObject) bool {
 // Cluster is a handle to a Kubernetes API server.
 // (Typically, this code is deployed into the same cluster.)
 type Cluster struct {
-	client     extendedClient
-	applier    Applier
-	version    string // string response for the version command.
-	logger     log.Logger
-	sshKeyRing ssh.KeyRing
+	client      extendedClient
+	applier     Applier
+	version     string // string response for the version command.
+	logger      log.Logger
+	sshKeyRing  ssh.KeyRing
+	nsWhitelist []string
 
 	mu sync.Mutex
 }
@@ -118,7 +119,8 @@ func NewCluster(clientset k8sclient.Interface,
 	ifclientset ifclient.Interface,
 	applier Applier,
 	sshKeyRing ssh.KeyRing,
-	logger log.Logger) *Cluster {
+	logger log.Logger,
+	nsWhitelist []string) *Cluster {
 
 	c := &Cluster{
 		client: extendedClient{
@@ -132,6 +134,7 @@ func NewCluster(clientset k8sclient.Interface,
 		applier:    applier,
 		logger:     logger,
 		sshKeyRing: sshKeyRing,
+		nsWhitelist: nsWhitelist,
 	}
 
 	return c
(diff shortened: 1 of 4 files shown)


mwhittington21:
It seems we've basically implemented the same thing except my version allows for multiple namespaces. See #1186. I do like your test code better though, I'm new to Kube API testing.

squaremo:
Great minds think alike! Having looked at both PRs, may I suggest that this PR is adapted

 - to admit multiple namespaces
 - to use the flag name `--k8s-namespace-whitelist`
 - to include the bit of documentation from the other PR for that flag

I _think_ this will be the easier route -- but what you decide between you matters more, since you did the work in the first place.

I have one concern that applies to both PRs: should this also limit which resources are synced, and which resources can be automated or otherwise have new images released?

mwhittington21:
I think that limiting those things you mentioned according to the whitelisted namespaces would be ideal as allowing the user to shoot themselves in the foot with this is not good. It would also provide some protection and isolation were you to want to run multiple Flux daemons in the same cluster.

justinbarrick:
Very cool! I’ll fix up the PR tonight (though, if @mwhittington21 wants to do it in the meantime I have no issue with that).

justinbarrick:
> I have one concern that applies to both PRs: should this also limit which resources are synced, and which resources can be automated or otherwise have new images released?

This seems wise, but I'm not sure the best place to implement it. Can you give me some pointers?

cluster/kubernetes/kubernetes.go
```diff
@@ -402,3 +407,32 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
...
+func (c *Cluster) getNamespaces() ([]apiv1.Namespace, error) {
+	nsList := []apiv1.Namespace{}
+	validNamespaces := map[string]bool{}
+
+	namespaces, err := c.client.Namespaces().List(meta_v1.ListOptions{})
+	if err != nil {
+		return nsList, err
+	}
+
+	for _, namespace := range c.nsWhitelist {
```
mwhittington21: Very minor optimization / nit: might be better to just do this once on struct creation and store a map instead of a list as a struct member.

justinbarrick: Good idea, I think that is a sane way to store it.

mwhittington21:
After a quick look around I think you the best place to stop Flux syncing resources that are not in specific namespaces would be to place some code in sync/sync.go in [prepareSyncApply](https://github.com/weaveworks/flux/blob/9d5eb91961b89a9098fb729efbe68bd677836728/sync/sync.go#L63). It should be possible to inspect namespaces here and only add resources to the apply action if they match the namespace.

Need to be careful about un-namespaced manifests counting as "default" in case that's an edge case. Would this be sufficient @squaremo?

mwhittington21:
I've given this a go over at https://github.com/mwhittington21/flux/tree/1181-continued-add-sync-restriction-on-ns. Feel free to take the work and update this PR with it.

squaremo:
> After a quick look around I think the best place to stop Flux syncing resources that are not in specific namespaces would be to place some code in sync/sync.go in prepareSyncApply. It should be possible to inspect namespaces here and only add resources to the apply action if they match the namespace.

That would work (demonstrably -- you've already done it!). My inclination is to keep the sync package  independent of the cluster implementation, though granted there's only one cluster implementation.

This would mean `sync` remains ignorant of the structure of resource IDs and namespaces, and the filtering happens in `kubernetes.*Cluster.Sync` -- which already has the whitelist. It could report non-whitelisted resources as SyncErrors (meaning they can get surfaced in notifications; and in future versions, by the CLI).

It could also make defaulted namespace handling a bit cleaner, since that's pretty particular to Kubernetes. It will entail looking up the default namespace from the config, though, which is a bit more work.

mwhittington21:
Would it be possible to split the Sync restrictions into a future PR and just merge this one in now? Reducing API calls is the most important thing to my current use case.

squaremo:
> Would it be possible to split the Sync restrictions into a future PR and just merge this one in now? Reducing API calls is the most important thing to my current use case.

Yes. Let's label this as experimental (in the help text), and I'll do a quick review.

cmd/fluxd/main.go
```diff
@@ -114,6 +114,7 @@ func main() {
 		token       = fs.String("token", "", "Authentication token for upstream service")
 
 		dockerConfig = fs.String("docker-config", "", "path to a docker config to use for image registry credentials")
+		k8sNamespaceWhitelist = fs.StringSlice("k8s-namespace-whitelist", []string{}, "Optional, comma separated list of namespaces to monitor for workloads")
```
squaremo: Let's mark this as experimental, for the minute. How about, `"Experimental, optional: restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set."`

cluster/kubernetes/kubernetes.go
```diff
@@ -402,3 +412,27 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 
 	return allImageCreds
 }
+
+// getNamespaces returns a list of namespaces that the Flux instance is expected
+// to have access to and can look for resources inside of.
+// It returns a list of all namespaces unless a namespace whitelist has been set on the Cluster
+// instance, in which case it returns a list containing the namespaces from the whitelist
+// that exist in the cluster.
+func (c *Cluster) getNamespaces() ([]apiv1.Namespace, error) {
```
squaremo: I think I'd prefer `getAllowedNamespaces(...)` or similar, just to make it clear that the point is to filter the namespaces.

cluster/kubernetes/kubernetes.go
```diff
@@ -402,3 +412,27 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
...
+	nsList := []apiv1.Namespace{}
+
+	namespaces, err := c.client.Namespaces().List(meta_v1.ListOptions{})
```
squaremo: @mwhittington21 Is this going to have the API call limiting effect you're after? It's better than asking for all the namespaces _then_ asking for all the things in them. Doing better again might be a tricky balance though, wdyt.

mwhittington21: I've already tested a couple of the commits on my 30k namespace cluster and the result is vastly improved. For starters, Flux isn't frozen due to the amount of API requests it has to make (~10 per namespace). The list operation will take a bit longer, but once it's done there are almost no follow up calls by comparison.

Big thumbs up from me for this current version improving things.

cmd/fluxd/main.go
mwhittington21: Remember to also update the site/daemon.md table with this too.

justinbarrick:
Thanks for the help! I've marked this experimental so we can get this out. I'll likely submit a follow up for the other issues discussed in the next week or so.

squaremo:
Good stuff, thanks @justinbarrick and @mwhittington21. Assuming it looks OK in our dev cluster (which won't use the flag, but that's fine, it's experimental), this should appear in a release pretty soon.

PR #1187, opened 2018-07-02
Helm chart release v0.1.1 and release 0.2.1

stefanprodan:
~- includes Flux v1.4.1 and the Git known hosts customisation option~

EDIT(squaremo): v0.1.1 is already published; this PR now updates the chart to use more recent versions of fluxd and helm-op, and removes the spurious index.yaml (which is maintained in gh-pages). I have kept the original commits, so that there's a record in git of what was published as the 0.1.1 package.


squaremo:
I've rebased this on master and bumped the helm-op and fluxd versions, but I note that the index file includes the old versions and I'm not sure what to do about that.

squaremo:
OK I figured it out: Stefan has been generating the package and updating the index from master branch (or a topic branch like this one). For that reason it's convenient to have the index here, since you need the chart files to make the package tarball anyway.

An alternative is to start in gh-pages, grab the chart files from master with `git clone -b ...` and make the package tarball and update the index, ready to commit. I've added a script to gh-pages to do that, and removed the index from here.

This PR is therefore just about updating the Helm chart to the more recent versions of fluxd and helm-op (the package version 0.1.1 is already published, i.e., committed to the gh-pages branch).

squaremo:
Wonderful, thanks @dholbach :star:

PR #1188, opened 2018-07-02
helm-op: Add git hosts to known hosts in /etc/ssh/ssh_known_hosts

stefanprodan:
Port the know hosts changes from Flux to Flux Helm Operator ref #1154


stephenmoloney:
This change will surely impact the helm chart which uses `/root/.ssh/known_hosts` as the dir ... I'll make a PR for that.

squaremo:
> This change will surely impact the helm chart which uses /root/.ssh/known_hosts as the dir ... I'll make a PR for that.

The git code will look in both places; so it is intended that the image has the config and known_hosts in `/etc/ssh/`, and the chart mounts them into `/root/.ssh/`.

PR #1189, opened 2018-07-02
change sshdir mount path to /etc/ssh/ssh_known_hosts

stephenmoloney:
What does this PR do?

- Changes the mount path for known_hosts

Why is this PR needed?

- To bring the helm templates inline with the changes to the Dockerfile
in PR #1188.

### DIFF
--- a/chart/flux/templates/helm-operator-deployment.yaml
+++ b/chart/flux/templates/helm-operator-deployment.yaml
@@ -38,8 +38,8 @@ spec:
         imagePullPolicy: {{ .Values.helmOperator.pullPolicy }}
         volumeMounts:
         - name: sshdir
-          mountPath: /root/.ssh/known_hosts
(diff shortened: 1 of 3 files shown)


stephenmoloney:
I haven't tested this so please scrutinize it well but I think these changes are required in light of the changes in #1188, thanks!

stefanprodan:
This will break the know hosts logic.

stefanprodan:
To clarify, the `/etc/ssh/ssh_known_hosts` is providing the defaults but you can override it using the root mount via a ConfigMap. The change in this PR will break this feature.

stephenmoloney:
Ah, I see, I must be misunderstanding how the defaults work.  

so we can close this PR. 

Thanks for clarifying.

squaremo:
Sorry for the mixup @stephenmoloney

PR #1192, opened 2018-07-03
Optimize Docker cache usage for flux and helm-operator docker images.

justinbarrick:
Moves the common parts of the Dockerfiles to be in the same order so that the first seven steps are cached.

This is to be just slightly kinder to the networking on my CI system which pulls both images.

### DIFF
--- a/docker/Dockerfile.flux
+++ b/docker/Dockerfile.flux
@@ -1,5 +1,17 @@
 FROM alpine:3.6
 
+WORKDIR /home/flux
+
+RUN apk add --no-cache openssh ca-certificates tini 'git>=2.3.0'
+
+# Add git hosts to known hosts file so we can use
+# StrickHostKeyChecking with git+ssh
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.flux
```diff
@@ -1,5 +1,20 @@
 FROM alpine:3.6
 
+WORKDIR /home/flux
+
+RUN apk add --no-cache openssh ca-certificates tini 'git>=2.3.0'
+
+# Add git hosts to known hosts file so when git ssh's using the deploy
+# key we don't get an unknown host warning.
+RUN mkdir ~/.ssh && touch ~/.ssh/known_hosts && \
+    ssh-keyscan github.com gitlab.com bitbucket.org >> ~/.ssh/known_hosts && \
```
stefanprodan: We are not using /.ssh/known_hosts anymore, please use the latest master dockerfiles.

PR #1193, opened 2018-07-03
helm-op:  Update Kubernetes deps to 1.11 and fix unwanted upgrades

stefanprodan:
* update Kubernetes deps to 1.11 and client-go to 8.0
* generate Helm CRD clientset, informers  and listeners for K8s 1.11
* fix unwanted upgrades on K8s 1.11 by triggering a Helm upgrade only when FHR values changed

Tested on K8s 1.11 and 1.10 with `stefanprodan/helm-operator:k8s11-9d5eb919-wip7`


integrations/helm/operator/operator.go
```diff
@@ -343,3 +341,31 @@ func (c *Controller) deleteRelease(fhr ifv1.FluxHelmRelease) {
 	}
 	return
 }
+
+// needsUpdate compares two FluxHelmRelease and determines if any changes occurred
+func needsUpdate(old, new ifv1.FluxHelmRelease) bool {
```
squaremo: This is a bit brittle, but it'll do for now, unless you can think of a way of making sure it gets updated when the definition is updated.

integrations/client/clientset/versioned/fake/register.go
```diff
@@ -37,7 +37,7 @@ func init() {
 //
 //   import (
 //     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kuberentes/scheme"
+//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
```
squaremo: Was this generated or not? I am lost

integrations/client/clientset/versioned/scheme/register.go
```diff
@@ -37,7 +37,7 @@ func init() {
 //
 //   import (
 //     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kuberentes/scheme"
+//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
```
squaremo: And here it is again

Gopkg.lock
```diff
@@ -649,15 +647,17 @@
...
     "storage/v1alpha1",
     "storage/v1beta1"
   ]
-  revision = "b378c47b2dcba7f5c3ef97d8a5a0b3821ec6a001"
+  revision = "072894a440bdee3a891dea811fe42902311cd2a3"
+  version = "kubernetes-1.11.0"
 
 [[projects]]
-  branch = "master"
+  branch = "release-1.11"
```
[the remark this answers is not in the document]
stefanprodan: We should keep this pinned on the latest stable release, master is not stable.

integrations/client/clientset/versioned/fake/register.go
stefanprodan: All files inside client are generated.

Gopkg.toml
```diff
@@ -18,16 +18,25 @@ required = ["k8s.io/code-generator/cmd/client-gen"]
...
-  branch = "release-1.9"
+  branch = "release-1.11"
 
-# goland/dep does not follow k8s.io/code-generator's Godeps.json and gengo's master
-# version does not work with release-1.8 k8s.io/code-generator. So we have to
-# override it.
 [[override]]
-  name = "k8s.io/gengo"
-  revision = "b58fc7edb82e0c6ffc9b8aef61813c7261b785d4"
+  name = "github.com/json-iterator/go"
```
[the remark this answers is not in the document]
stefanprodan: All kubernetes projects are using godep and dep can't parse that, this pinning is from the godep file as the json-iterator master doesn't work with 8.0

Issue #1175, opened 2018-06-28
Support FluxHelmRelease values format with more than one image


squaremo:
At present fluxd knows how to interpret and update FluxHelmRelease manifests that use the format

```
values:
  image: repo/image:tag
```

which admits only pretty simple charts, since it only lets you change one image. An improvement would be

```
values:
  foo:
    image: repo/foo:tag
  bar:
    image: repo/bar:tag
```

where each image is under a label, which can be treated as the "container" for that image (most likely that's what the templates will use the values for).

PR #1196, opened 2018-07-03
Fix problem with multiple image support in FluxHelmReleases

squaremo:
#1175 didn't quite do the job.

When FluxHelmRelease resources are obtained via the Kubernetes API,
maps in the free-form `.spec.values` field are typed
`map[string]interface{}` since it is deserialised from JSON.

So, we have to account for this case as well as the
`map[interface{}]interface{}` that comes from YAML (i.e., manifest
files), if we want to factor the interpretation of `.spec.values` into
a single procedure.


squaremo:
Thanks Stefan :+1:

PR #1199, opened 2018-07-04
Fix ResourceIDRegexp to allow dots in resource names

justinbarrick:
Some resources (particularly CustomResourceDefinitions) have dots in their names, these
are currently considered invalid by ResourceIDRegexp.

I do not know if this causes an issue inside of Flux, but I import it for fluxcloud and
it breaks unmarshalling events:

```
error decoding: parsing default:customresourcedefinition/clustersecrets.clustersecret.codesink.net: invalid service ID
```


squaremo:
Beat you to it, _just_, in #1197.

PR #1204, opened 2018-07-05
Update CHANGELOG and example deployment for v1.4.2

squaremo:


dholbach:
Good work! Let's bring 1.4.2 on the road!

PR #1212, opened 2018-07-08
Support automatically port forwarding to flux instances in a Kubernetes cluster.

justinbarrick:
Simplify accessing flux instances in a Kubernetes cluster by automatically creating
port forwards, just like Helm does.

This is based on the port forward implementations from kubectl and Helm:

* https://github.com/kubernetes/helm/blob/master/pkg/kube/tunnel.go
* https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/portforward.go

The flow is:

* If neither --flux-namespace nor FLUX_NAMESPACE are set, then this code will do nothing
  and everything will work as before.
* If it is, it will instantiate a Kubernetes client and search for pods in FLUX_NAMESPACE
  that have the `name=flux` labels set.
* If the number of pods with `name=flux` in the namespace are not exactly 1, it will
  return an error.
* It will find an empty port (by first binding to port 0 and then closing the port) and
  start a port forward on that port that forwards to the flux pod on port 3030.
* It will set the flux url to `http://127.0.0.1:$port/api/flux`.
* The port forward goroutine will get cleaned up automatically when fluxctl terminates.

Use-cases:

* Easier use of standalone flux instances.
* Discourage exposing flux apis to the internet (hopefully nobody does this).
* More easily manage large numbers of flux instances.


cmd/fluxctl/port_forward.go
```diff
@@ -0,0 +1,171 @@
...
+
+	ports := []string{fmt.Sprintf("%d:3030", freePort)}
+
+	discard := ioutil.Discard
+	pf, err := portforward.New(dialer, ports, p.stopChan, p.readyChan, discard, discard)
+	if err != nil {
+		return err
+	}
+
+	go pf.ForwardPorts()
```
justinbarrick: `ForwardPorts()` should be handled somehow.

justinbarrick:
Fission also has a port forward implementation: https://github.com/fission/fission/blob/master/fission/portforward/portforward.go

This seems like something useful to a lot of projects, so I think it would be better for me to factor this out into a separate library and pull that in here.

squaremo:
Another thing that occurred to me: it'd be wonderful if this were the default mode of operation, so people have no reason to be tempted to expose the API. This is a bit tricky though, since we'd have to account for people that have already made shell scripts, etc. Though I am not adverse to a small amount of breakage in the interest of encouraging people to use port-forwarding instead of e.g., a NodePort, when either would work.

justinbarrick:
I also saw `name=fluxd` on some old documentation on your website just now, too. So I’ll have it search for a couple labels if you don’t specify one.

I’m not sure which command line I prefer. The first one is nice because if we are searching by label hopefully all you need to change is the namespace so it’s less verbose. But the second one keeps in line with the rest of flux and is more flexible since it lets you pass in a resource name (especially if you have more than one flux in a namespace).

I wasn’t sure about making it the default since it looks like weavcloud connect through the Weave Cloud API. Is that not the case?.

squaremo:
> I’m not sure which command line I prefer. The first one is nice because if we are searching by label hopefully all you need to change is the namespace so it’s less verbose. But the second one keeps in line with the rest of flux and is more flexible since it lets you pass in a resource name (especially if you have more than one flux in a namespace).

On balance I think I like the first one, since it's progressive -- you can just flip the switch, or you can supply a namespace, or you can supply the whole lot if you need to.

> I wasn’t sure about making it the default since it looks like weavcloud connect through the Weave Cloud API. Is that not the case?.

That is the case, yup. You have to supply an auth token as well (either with `--token,-t` or `FLUX_SERVICE_TOKEN` or `WEAVE_CLOUD_TOKEN`), so we can in principle detect when people have it set up that way. We would get a false positive in cases that people have set up an ingress or NodePort or the like; those ones I think I'd be prepared to break, temporarily, for the greater good. We can deal with changing the default after this PR, though -- I don't want to pile more requirements on.

squaremo:
> I also saw name=fluxd on some old documentation on your website just now, too. So I’ll have it search for a couple labels if you don’t specify one.

I think I have conflated `.metadata.name=flux` and selection by label -- does that change anything?

justinbarrick:
> Though I am not adverse to a small amount of breakage in the interest of encouraging people to use port-forwarding instead of e.g., a NodePort, when either would work.

Thinking about this more, I don't actually think we'll break anything at all here! fluxctl with no arguments was never a valid configuration - you always had to either pass a token or a URL in order for fluxctl to work. Now, if you pass neither of those, it will attempt a port forward instead.

justinbarrick:
Okay, I've implemented everything I described above:

* The command line interface logic.
* We search for pods that match the label selector `name in (flux, fluxd, weave-flux-agent)`

This is good to go from my end!

justinbarrick:
Demo:

Default port forward fails, since my flux is in kube-system:

```
➜  flux git:(master) ✗ ./fluxctl list-controllers
Error: Could not find pod for selector: labels "name in (flux,fluxd,weave-flux-agent)"
Run 'fluxctl list-controllers --help' for usage.
```

Search for flux in kube-system and create a port forward:

```
➜  flux git:(master) ✗ ./fluxctl -f kube-system list-controllers
CONTROLLER                                              CONTAINER                      IMAGE                                                                  RELEASE  POLICY
default:daemonset/elasticsearch-operator-sysctl         sysctl-conf                    busybox:1.26.2                                                         ready                                                                                                                    
```

Try to connect to weave cloud with fake token:

```
➜  flux git:(master) ✗ ./fluxctl -t mytoken list-controllers
== Error ==

The request failed authentication

This most likely means you have a missing or incorrect token. Please
make sure you supply a service token, either by setting the
environment variable FLUX_SERVICE_TOKEN, or using the argument --token
with fluxctl.
```

Set a URL and connect directly:

```
➜  flux git:(master) ✗ ./fluxctl -u http://127.0.0.1:3030/ list-controllers
Error: Get http://127.0.0.1:3030/v6/services?namespace=default: dial tcp 127.0.0.1:3030: connect: connection refused
Run 'fluxctl list-controllers --help' for usage.
```

cmd/fluxctl/root_cmd.go
```diff
@@ -74,11 +82,38 @@ func (opts *rootOpts) Command() *cobra.Command {
 }
 
 func (opts *rootOpts) PersistentPreRunE(cmd *cobra.Command, _ []string) error {
+	opts.Namespace = getFromEnvIfNotSet(cmd.Flags(), "k8s-forward-namespace", opts.Namespace, envVariableNamespace)
+	opts.Token = getFromEnvIfNotSet(cmd.Flags(), "token", opts.Token, envVariableToken, envVariableCloudToken)
 	opts.URL = getFromEnvIfNotSet(cmd.Flags(), "url", opts.URL, envVariableURL)
+
+	if opts.Token != "" {
```
squaremo: We use `fluxctl` against development environments of Weave Cloud, and for this we need to be able to set both the URL and the auth token. I think it should be (in pseudo-code): if either URL or token are set, use the URL and token arguments (with defaults as they are now); otherwise, use port forwarding.

| use | set | outcome |
|-------|-----|--------------|
| OSS install | - | uses port forwarding to default namespace |
| OSS install | k8s-forward-namespace=foo | port forward to namespace "foo" |
| weave cloud | token=abc | uses default URL (https://cloud.weave.works/api/flux) and token |
| weave cloud dev | token=abc, url=http://$(minikube ip):$port/api/flux | uses URL and token |

cmd/fluxctl/root_cmd.go
```diff
@@ -74,11 +82,38 @@ func (opts *rootOpts) Command() *cobra.Command {
 }
 
 func (opts *rootOpts) PersistentPreRunE(cmd *cobra.Command, _ []string) error {
+	opts.Namespace = getFromEnvIfNotSet(cmd.Flags(), "k8s-forward-namespace", opts.Namespace, envVariableNamespace)
```
squaremo: There's a difference between the default namespace (which is determined by your kubeconfig), and the namespace `"default"`. Some people use contexts in kubeconfig -- or different kubeconfigs entirely -- when operating in different namespaces. Since the port forwarding library is implicitly getting things from a kubeconfig anyway, to be able to connect at all, could it also pick up the default namespace?

cmd/fluxctl/root_cmd.go
```diff
@@ -49,7 +54,10 @@ func (opts *rootOpts) Command() *cobra.Command {
 		SilenceErrors:     true,
 		PersistentPreRunE: opts.PersistentPreRunE,
 	}
-	cmd.PersistentFlags().StringVarP(&opts.URL, "url", "u", "https://cloud.weave.works/api/flux",
+
+	cmd.PersistentFlags().StringVarP(&opts.Namespace, "k8s-forward-namespace", "f", "default",
```
squaremo: Not sure how I feel about having a shorthand for this -- and `-f` is usually used for `file` (albeit we don't have any commands that accept a file; and, we are stuck with `-u` and `-t` for connection flags, which I wouldn't have chosen ..).

We can shorten the long name, it is a _little_ long: `--k8s-fwd-namespace` or even `--k8s-fwd-ns`. wdyt?

site/using.md
```diff
@@ -20,46 +20,28 @@ You need to tell `fluxctl` where to find the Flux API and to make
...
 ```
 
-### Local endpoint
+This setting can also be set in an environment variable (`FLUX_FORWARD_NAMESPACE`).
 
-**Beware**: this exposes the Flux API, unauthenticated, over an
-insecure channel. Do not do this _unless_ you are operating Flux
-entirely locally; and arguably, only to try it out.
+If you are not able to use the port forward to connect, you can connect by URL with
+`--url` or `FLUX_URL`:
```
squaremo: This is misleading, since there are no preceding instructions for setting up a NodePort or equivalent. I'm happy to remove those, but we need to be clear that the example given won't just work.

cmd/fluxctl/root_cmd.go
justinbarrick: Good point! I'm still working on trying to figure out how to fetch the default namespace from the kube config (it's conspiciously missing from https://godoc.org/k8s.io/client-go/rest#Config).

cmd/fluxctl/root_cmd.go
justinbarrick: Makes sense to me, changed.

site/using.md
justinbarrick: Updated it to indicate that you need to setup some way of connecting to it.

cmd/fluxctl/root_cmd.go
justinbarrick: Good catch, fixed!

cmd/fluxctl/root_cmd.go
squaremo: I think https://godoc.org/k8s.io/client-go/tools/clientcmd has the bits needed. `k8s.io/client-go` really is labyrinthine isn't it :-/

Issue #1211, opened 2018-07-06
Cannot undo ignore annotation


tamirms:
I added an `flux.weave.works/ignore: "true"` annotation to some of my kubernetes deployments. In a subsequent commit to my kubernetes repo, I removed the annotations. However, flux still seems to ignore my deployments. I've tried killing the flux and flux memcached pods but that still hasn't fixed the problem. I've also tried renaming my manifest file.

some more info from slack:

```
in the weave flux agent logs I see
```ts=2018-07-06T17:41:26.352552842Z caller=sync.go:70 component=sync-loop resource=default:deployment/pb-production-accounts-deployment ignore=apply```
so for some reason the ignore tag is stuck to the deployment
I have triple checked our manifest file
and there are no annotations at all

stefan [7:47 PM]
what version are you running?

tamir [7:47 PM]
how can I find that info?

stefan [7:47 PM]
`kubectl get deployment/flux -oyaml`
check the image tag in the output (edited)

tamir [7:54 PM]
I ran `kubectl get deployment/weave-flux-agent -o yaml --namespace=weave`
@stefan the image tag is `quay.io/weaveworks/flux:1.4.2`

stefan [7:56 PM]
that's the latest version, it should work... maybe @mbridgen has any idea how to diagnose this

tamir [8:06 PM]
I’ve tried reinstalling as well
doesn’t seem to do anything

stefan [8:07 PM]
try renaming the deployment file

tamir [8:07 PM]
ok

tamir [8:14 PM]
it didn’t work
`ts=2018-07-06T18:11:55.108793942Z caller=sync.go:70 component=sync-loop resource=default:deployment/pb-production-accounts-deployment ignore=apply`
still shows up in the logs (edited)
I guess I’ll have to do things manually
@mbridgen @stefan let me know if you require any more info from me to diagnose this issue

stefan [8:16 PM]
one more try, can you please delete the flux pod and the memcache one

tamir [8:17 PM]
ok

tamir [8:29 PM]
still didn’t work :disappointed:
is there some volume memcached is using to persist its state?
maybe I can delete that as well?

stefan [8:37 PM]
Nop, my guess is that the git pull that Flux does gets somehow an older version. Can you check in Weave Cloud that you have the right branch

tamir [8:41 PM]
i see this
[...]

Issue #749, opened 2017-09-22
Annotations are ignored for some kinds of resources when syncing


squaremo:
In `sync/sync.go`, the manifests (files) are compared with the resources in the cluster, by `cluster.Export`ing and parsing the latter. But export only includes the kinds of resource we "care" about -- hitherto, namespaces, services and deployments. So an annotation on another kind of resource will not be in the results, and not be taken into account.

This is a problem when using the `flux.weave.works/ignore` annotation on say, a daemonset resource, because the file will be `kubectl apply`ed regardless.

#747 will expand the set of resources we take notice of (to include daemonsets, among others), but ultimately, we want to know about anything that might appear as a manifest.

squaremo:
The rewrite of the syncing code in #1442 removed this limitation. Sorry for the confusing leftover issue.

PR #1216, opened 2018-07-09
docs: document  ignore  in FAQ

### LINKED ISSUE #955: Document "ignore" and its limitations
.. and possibly change it to "ignored", or allow that as a synonym.

dholbach:
Document `ignore` annotation in the FAQ, take into account #1211 and #749.

Closes #955.

### DIFF
--- a/site/faq.md
+++ b/site/faq.md
@@ -300,3 +300,43 @@ your case.
 You will need to use the (experimental) command-line flag
 `--k8s-namespace-whitelist` to enumerate the namespaces that Flux
 attempts to scan for workloads.
+
+### Can I temporarily make flux ignore a deployment?
+
+Yes. The easiest way to do that is to use the following annotation
+*in the manifest files*:
+
+```yaml
+    flux.weave.works/ignore: true
+```
+
+To stop ignoring these annotated resources, you simply remove the
+annotation from the manifests in git. A live example can be seen
+[here](https://github.com/stefanprodan/openfaas-flux/blob/master/secrets/openfaas-token.yaml).
+This will work for any type of resource.
+
+Sometimes it might be easier to annotate a *running resource in
+the cluster* as opposed to committing a change to git. Please note
+that this will only work with resources of the type `namespace`
+and the set of controllers in
+[resourcekinds.go](https://github.com/weaveworks/flux/blob/master/cluster/kubernetes/resourcekinds.go),
+namely `deployment`, `daemonset`, `cronjob`, `statefulset` and
+`fluxhelmrelease`).
+
+If the annotation is just carried in the cluster, the easiest way
+to remove it is to run:
+
+```sh
+kubectl annotate <resource> "flux.weave.works/ignore"-
+```
+
+Mixing both kinds of annotations (in-git and in-cluster), can make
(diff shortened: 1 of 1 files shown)


site/faq.md
```diff
@@ -112,3 +112,30 @@ setting. Flux will read this value and parse the Kubernetes secret.
...
+```yaml
+    flux.weave.works/ignore: true
+```
+
+A live example of this can be seen
+[here](https://github.com/stefanprodan/openfaas-flux/blob/master/secrets/openfaas-token.yaml).
+
+Please note that [flux#749](https://github.com/weaveworks/flux/issues/749)
+limits this currently to resources of the type `namespace`, `service`
+and `deployment`.
```
rade: This isn't true; see #749, which is referenced in the above.

site/faq.md
```diff
@@ -112,3 +112,30 @@ setting. Flux will read this value and parse the Kubernetes secret.
...
+
+To stop ignoring deployments, you simply remove the annotation from
+the manifests in git.
```
rade: not just 'deployments'.

site/faq.md
dholbach: Yes, I saw that. I was waiting for confirmation of @squaremo on https://github.com/weaveworks/flux/pull/1210#pullrequestreview-135073547

site/faq.md
```diff
@@ -300,3 +300,32 @@ your case.
 You will need to use the (experimental) command-line flag
 `--k8s-namespace-whitelist` to enumerate the namespaces that Flux
 attempts to scan for workloads.
+
+### Can I temporarily make flux ignore a deployment?
```
squaremo: If you put it in the manifest file, it'll work for any resource.

site/faq.md
```diff
@@ -300,3 +300,32 @@ your case.
...
+
+A live example of this can be seen
+[here](https://github.com/stefanprodan/openfaas-flux/blob/master/secrets/openfaas-token.yaml).
+
+Please note that not all resources are tracked by flux at the moment,
+we are currently limited to resources of the type `namespace` and the
+set of controllers in
+[resourcekinds.go](https://github.com/weaveworks/flux/blob/master/cluster/kubernetes/resourcekinds.go),
+namely `deployment`, `daemonset`, `cronjob`, `statefulset` and `deployment`).
+
```
squaremo: That's not accurate as stated. Well, depending on what's meant by "tracked".

The full story is this: flux looks at the files and the running resources when deciding whether what to apply. But it gets the running resources by exporting them from the cluster, and that only returns the kinds of resource mentioned above. So, annotating a running resource only works if it's one of those kinds; putting the annotation in the file always works.

site/faq.md
```diff
@@ -300,3 +300,32 @@ your case.
...
+To stop ignoring these annotated resources, you simply remove the
+annotation from the manifests in git.
+
+If the annotations are carried in the cluster as well (this should
+not happen by default), you need to remove them from the cluster as well,
```
squaremo: Annotating the resource in the cluster is an intended mode of use (it can be a bit easier than committing a change to git).

PR #1218, opened 2018-07-09
Add release spec format to selectively update containers in resources

### LINKED ISSUE #1140: Allow for updating multiple workloads to arbitrary images in a single release
Currently, there isn't a way to set the image for multiple workloads within a single release (commit). The `/update-images` endpoint only allows for multiple workloads which reference the same container image to be updated simultaneously.

We need a way to update the images for different workloads at the same time. Example state-change (pseudo JSON):

```
{
  "default:deployment/ui-server": {
    "Containers": [
      {
        "Name": "ui-server",
        "ID": "quay.io/myorg/ui-server:xyz789" => "quay.io/myorg/ui-server:tuv456"
      }
    ]
  },
  "default:deployment/auth": {
    "Containers": [
      {
        "Name": "auth",
        "ID": "quay.io/myorg/auth:abc123" => "quay.io/myorg/auth:def456"
      }
    ]
  }
}

```

^^ This "transaction" should be completed within the same release.

rndstr:
This adds a new release spec format providing update requests on container
level.

Sample POST request body:
```
{
  "type": "containers",
  "cause": {
    "Message": "sample request body",
    "User": "alice"
  },
  "spec": {
    "ContainerSpecs": {
      "default:deployment/nginx": [
        {
          "Container": "nginx",
          "Current": "nginx:1.14.0",
          "Target": "nginx:1.15.0"
        }
      ],
      "default:deployment/helloworld": [
        {
          "Container": "helloworld",
          "Current": "quay.io/weaveworks/helloworld:master-07a1b6b",
          "Target": "quay.io/weaveworks/helloworld:master-a000004"
        },
        {
          "Container": "sidecar",
          "Current": "quay.io/weaveworks/sidecar:master-a000001",
          "Target": "quay.io/weaveworks/sidecar:master-a000002"
        }
      ]
    },
    "Kind": "plan",
    "IgnoreMismatches": false
  }
}
```

The request will fail if any of the `Current` container image requirements
are not met or containers cannot be found. To have partial updates go
through and ignore any failures, one can pass `true` for `IgnoreMismatches`.

Also, `Current` can be left empty to disable any precondition check.

Closes #1140

### DIFF
--- a/release/releaser_test.go
+++ b/release/releaser_test.go
@@ -2,12 +2,14 @@ package release
 
 import (
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"testing"
 	"time"
 
 	"github.com/go-kit/kit/log"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/cluster"
 	"github.com/weaveworks/flux/cluster/kubernetes"
@@ -23,6 +25,8 @@ var (
 	// This must match the value in cluster/kubernetes/testfiles/data.go
 	helloContainer   = "greeter"
 	sidecarContainer = "sidecar"
+	lockedContainer  = "locked-service"
+	testContainer    = "test-service"
 
 	oldImage      = "quay.io/weaveworks/helloworld:master-a000001"
 	oldRef, _     = image.ParseRef(oldImage)
@@ -52,7 +56,7 @@ var (
 	oldLockedRef, _ = image.ParseRef(oldLockedImg)
 
 	newLockedImg     = "quay.io/weaveworks/locked-service:2"
-	newLockedID, _   = image.ParseRef(newLockedImg)
+	newLockedRef, _  = image.ParseRef(newLockedImg)
 	lockedSvcID, _   = flux.ParseResourceID("default:deployment/locked-service")
 	lockedSvcSpec, _ = update.ParseResourceSpec(lockedSvcID.String())
 	lockedSvc        = cluster.Controller{
@@ -60,19 +64,20 @@ var (
 		Containers: cluster.ContainersOrExcuse{
 			Containers: []resource.Container{
 				{
-					Name:  "locked-service",
+					Name:  lockedContainer,
 					Image: oldLockedRef,
 				},
 			},
 		},
 	}
 
-	testSvc = cluster.Controller{
-		ID: flux.MustParseResourceID("default:deployment/test-service"),
+	testSvcID = flux.MustParseResourceID("default:deployment/test-service")
+	testSvc   = cluster.Controller{
+		ID: testSvcID,
 		Containers: cluster.ContainersOrExcuse{
 			Containers: []resource.Container{
 				{
-					Name:  "test-service",
+					Name:  testContainer,
 					Image: testServiceRef,
 				},
 			},
@@ -104,7 +109,7 @@ var (
 				CreatedAt: timeNow,
 			},
 			{
-				ID:        newLockedID,
+				ID:        newLockedRef,
 				CreatedAt: timeNow,
 			},
 		},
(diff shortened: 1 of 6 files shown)


squaremo:
I've reread #1140 and I understand a bit better what the purpose of `IgnoreFailedControllers` is -- it's to address this bit, I think:

> You might find that, for example, releases can hardly ever succeed, because something always changes before you hit the button. In that case it would be useful to be able to relax the all-or-nothingness

In that case, I reckon it should switch between the mode of "succeed only if all preconditions are met" and "succeed for all containers where the preconditions are met, leave others as they are".

Unconditionally updating containers (being able to not supply a precondition) is a different matter -- sorry for the distraction -- though may in itself be useful, I don't know.

release/releaser_test.go
```diff
@@ -527,6 +532,154 @@ func Test_UpdateList(t *testing.T) {
...
+					Container: sidecarContainer,
+					Current:   sidecarRef,
+					Target:    newSidecarRef,
+				},
+			},
+			Expected: update.ControllerResult{
+				Status: update.ReleaseStatusFailed,
+				Error:  fmt.Sprintf(update.ContainerTagMismatch, helloContainer),
+			},
+			Commit: "Release containers\n\ndefault:deployment/helloworld failed: container tag mismatch: greeter",
```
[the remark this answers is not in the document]
rndstr: The success was initially not mentioned because any failure for a container also meant the workload fails. I changed this now to allow partial workload updates. In this case, we don't have access to the errors since the workload can only have one result (success or failed).

rndstr:
@squaremo yes, I adjusted the logic a bit to make things clearer with regard to failed/skipped states.

It's either
- only update if all preconditions match (default)
- update all containers whose preconditions match (`IgnoreMismatches`–open to renaming)
- skip preconditions (don't provide them)

update/containers.go
```diff
@@ -0,0 +1,172 @@
...
+		var containerUpdates []ContainerUpdate
+		for _, spec := range s.ContainerSpecs[u.ResourceID] {
+			container, ok := containers[spec.Container]
+			if !ok {
+				notfound = append(notfound, spec.Container)
+				continue
+			}
+
+			// An empty spec for the current image skips the precondition
+			empty := image.Ref{}
```
squaremo: Worth lifting this out of the loops?

update/containers.go
```diff
@@ -0,0 +1,171 @@
...
+			// to Skipped instead of Failed. We still want to provide an error
+			// message that conveys why we skipped it.
+			if len(notfound) > 0 {
+				skippedError = notFoundError
+			} else if len(mismatch) > 0 {
+				skippedError = mismatchError
+			}
+		}
+
+		if !s.IgnoreMismatches && len(notfound) > 0 {
```
squaremo: This might be more compact, without losing readability, as a `switch`

update/containers.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+
+			containerUpdates = append(containerUpdates, spec)
+		}
+
+		skippedError := ImageUpToDate
+		notFoundError := fmt.Sprintf(ContainerNotFound, strings.Join(notfound, ", "))
+		mismatchError := fmt.Sprintf(ContainerTagMismatch, strings.Join(mismatch, ", "))
+
+		if s.IgnoreMismatches {
```
squaremo: The only place this matters if in the case `len(containerUpdates) == 0` below, so you could move it there.

update/containers.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return updates
+}
+
+func (s ContainerSpecs) ReleaseKind() ReleaseKind {
+	return s.Kind
+}
+
+func (s ContainerSpecs) ReleaseType() ReleaseType {
```
squaremo: Do we still use this and `ReleaseKind()`? Maybe we should remove them.

squaremo: (doesn't have to be done here, I just noticed here)

update/containers.go
```diff
@@ -0,0 +1,171 @@
...
+		for _, spec := range cs {
+			containers[spec.Name] = spec
+		}
+
+		var mismatch, notfound []string
+		var containerUpdates []ContainerUpdate
+		for _, spec := range s.ContainerSpecs[u.ResourceID] {
+			container, ok := containers[spec.Container]
+			if !ok {
+				notfound = append(notfound, spec.Container)
```
squaremo: I still think not finding the named container should be an outright error, i.e., not excused with `IgnoreMismatches`. That's on the basis that `IgnoreMismatches` is closer to "I don't have to do all these in one go", than to "I want to ignore problems".

update/containers.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/weaveworks/flux"
+	"github.com/weaveworks/flux/image"
+	"github.com/weaveworks/flux/resource"
+)
+
+// ContainerSpecs defines the spec for a `containers` manifest update.
+type ContainerSpecs struct {
+	Kind             ReleaseKind
+	ContainerSpecs   map[flux.ResourceID][]ContainerUpdate
+	IgnoreMismatches bool
```
squaremo: `SkipMismatches` is closer to the effect it has, I think.

squaremo:
Looks good, let's give it a go :smile: 

`git rebase -i master` to get rid of the feedback commits, depending on your personal attitude towards git neat-and-tidiness (and present level of busyness).

jpellizzari:
@rndstr To which endpoint does one do the POST and with what URL params, if any?

rndstr:
@jpellizzari `/v9/update-manifests`; no URL params, all the options are in the JSON body

PR #1230, opened 2018-07-13
Read default image creds for each request

squaremo:
Part of the point of being able to provide a docker config is to be
able to mount it from a volume, and thereby be able to update it on
the fly (e.g., with a sidecar).

But this won't work if it's read only at startup. So, just check it's
OK when we start, and otherwise read it every time we are asked for
credentials.


mwhittington21:
This should solve most of the ECR problem provided the user can refresh the creds themselves. Nice!

squaremo:
thanks @aaron7 :space_invader:

PR #1231, opened 2018-07-13
 Add `flux release --interactive`

### LINKED ISSUE #1227: Interactive `fluxctl release` command
At present you can

    fluxctl release --dry-run ...

to show you what would have been released at that moment, or

    fluxctl release ...

to just release what will be released at that moment. But you can't either release exactly what the first thing tells you (and do it all in one go), or pick and choose what to do (and do it all in one go).

With #1218 it's possible to release an exact specification of images; so it's possible to get the plan and turn that into a spec (if it looks good), and either succeed or fail in one go.

rndstr:
This adds an `--interactive` option to the `release` command that lets
you pick and choose which containers to update. It uses the new
`containers` specification in the /v9/update-manifest call.

Once the containers are selected, the release is sent off with
the constraint that the state of the containers need to be the exact
same as when the results were presented to the user. If anything changed
in the meanwhile, the whole release will not be processed.

Closes #1227

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -883,6 +883,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "06f3e7d274e6c3037b637df418c1cf8772a5e25bdc5e858c9cf9fe44bd80ceb4"
+  inputs-digest = "753bd801312bfb0aacc2d33e343346c5a73bbdab2cde843d0b31aabefe14895f"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -329,6 +329,15 @@
   version = "v0.8.0"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/pkg/term"
+  packages = [
+    ".",
+    "termios"
+  ]
+  revision = "cda20d4ac917ad418d86e151eff439648b06185b"
+
+[[projects]]
   name = "github.com/pmezard/go-difflib"
   packages = ["difflib"]
   revision = "792786c7400a136282c1664665ae0a8db921c6c2"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -49,3 +49,7 @@ required = ["k8s.io/code-generator/cmd/client-gen"]
 [[constraint]]
 	name = "github.com/justinbarrick/go-k8s-portforward"
 	version = "v1.0.0"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/pkg/term"
--- a/cmd/fluxctl/release_cmd.go
+++ b/cmd/fluxctl/release_cmd.go
@@ -3,10 +3,12 @@ package main
 import (
 	"context"
 	"fmt"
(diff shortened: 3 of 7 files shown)


rndstr:
@squaremo I knew there was something left to do.. fixed.

squaremo:
We don't currently build windows binaries for `fluxctl`, but it has been requested and does apparently build. Judging by the imports, I would expect the interactive stuff here to only work for linux; can you add a shim specific to Windows that at least says something like "Sorry, this doesn't work in Windows yet".

squaremo:
This doesn't build for windows (try adding `windows` to the list of OSes in the `release-bins` target in the Makefile), but it looks like it could in the future: https://github.com/pkg/term/issues/8

I think you'll have to funnel the interactive bit through an OS-specific file, and make the windows one return an apologetic error.

squaremo:
On the UI itself, it is goodly simple and I like it.

It was not obvious to me what was going on to start off with, because my first attempt had only one change. But the instructions are clear enough (maybe you could mention arrow keys too). Is there a nice right-pointing arrow glyph, since we're glyphing?

There's some glitching:

```
$ fluxctl release -n hello --controller=deployment/helloworld --update-all-images --interactive                                                                                 
Submitting dry-run release...                                                                                                                                                   
Handling connection for 3030                                                                                                                                                    
   CONTROLLER              STATUS   UPDATESDATES                                                                                                                                
 ◉ hello:deployment/helloworld  success  helloworld: quay.io/weaveworks/helloworld:master-a000002 -> master-07a1b6b                                                             
>◯                                       sidecar: quay.io/weaveworks/sidecar:master-a000001 -> master-a000002                                                                   
                                                                                                                                                                                
Use [Space] to deselect containers and hit [Enter] to release selected. 
```

(the columns don't line up, and `UPDATESDATES`)

The "STATUS" column is a bit out of place -- what does "success" mean here? (I realise this is inherited from the dry run output, where it is also a bit odd).

rndstr:
I tried making it run in Windows but it needs a bit more work to read a single char cross-platform. existing libraries attempt to capture the whole terminal space by clearing it which i would rather not have.

> maybe you could mention arrow keys too

doing so now :+1: 

> Is there a nice right-pointing arrow glyph, since we're glyphing?

⇒

> the columns don't line up, and UPDATESDATES

tabwriter didn't like the escape sequences.

> The "STATUS" column is a bit out of place -- what does "success" mean here?

`success` means that it is updateable. if you provide `--verbose` you will also see the other statuses of updates that have *not* been selected for various reasons. i agree it looks odd while there is no other status.

squaremo:
> success means that it is updateable. if you provide --verbose you will also see the other statuses of updates that have not been selected for various reasons. i agree it looks odd while there is no other status.

OK, ignore this one, it's not critical. (We can revisit how the results are shown, some other time)

squaremo:
image

 - I reckon put a space between the arrow and the toggle (possibly it's just my font scaling messing with things).
 - instead of "... to deselect containers" I suggest "... to toggle updates"

Sorry to drag this out :-/

rndstr:
@squaremo made the changes, PTAL

PR #1234, opened 2018-07-16
Update helm-op changelog for release 0.1.1-alpha

squaremo:
Update changelog for the Helm operator v0.1.1-alpha release. After this is merged, we can tag it so a release image is built.


squaremo:
> Can we for future releases include who contributed stuff?

I would like to do that, but it's a ton more work than reporting the changes.

PR #1238, opened 2018-07-19
git repo improvements for helm operator

squaremo:
Here are some improvements to the git mirroring, so that it's usable for the Helm operator.

 - allow a git repo be read-only, so that you can give the helm-op a read-only SSH key (as we do in local env)
 - add a "files only" clone, `git.Export`, so the helm operator can get at a working tree without all the config baggage that fluxd uses in its working directories

### DIFF
--- a/daemon/daemon_test.go
+++ b/daemon/daemon_test.go
@@ -597,9 +597,9 @@ func mockDaemon(t *testing.T) (*Daemon, func(), func(), *cluster.Mock, *mockEven
 	}
 
 	start := func() {
-		wg.Add(1)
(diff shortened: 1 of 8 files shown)


squaremo:
The substance is in https://github.com/weaveworks/flux/pull/1238/files#diff-4e28ed4007c9ceb9cb3efd309643ca04, in which the state machine loop is factored out so it can be used synchronously (in `repo.Ready(...)` ) but also set running in `repo.Start(...)`.

squaremo:
Thanks Aaron :spaghetti:

PR #1245, opened 2018-07-24
Add note on git repo layout to FAQ (and correct example URL, while we're here)

squaremo:


squaremo:
Thanks Daniel :smile:

PR #1246, opened 2018-07-24
chart: fix known_hosts

malvex:
If you use one host key for known_hosts file, it will dump value right after pipe and fails on syntax error.

```Error: YAML parse error on flux/templates/ssh.yaml: error converting YAML to JSON: yaml: line 5: did not find expected comment or line break```


stefanprodan:
Thanks @malvex

Issue #1127, opened 2018-06-07
automation toggled between two images


rade:
For some unknown reason flux decided to deploy the most-recent-but-one image of itself, and then immediately re-deployed the most-recent image:
![screenshot from 2018-06-07 15-00-58](https://user-images.githubusercontent.com/109109/41104694-ddadcd76-6a63-11e8-951c-9112a2c9806c.png)

There is no apparent trigger - the two images were built yesterday, 6 hours apart.

squaremo:
Please posts logs or other clues, this isn't enough to go on.

rade:
Scanning the logs around the time for actions relating to the two image ids...
```
ts=2018-06-07T13:22:56.658856019Z caller=images.go:67 component=sync-loop service=weave:deployment/weave-flux-agent container=flux currentimage=quay.io/weaveworks/flux:master-180d2e3 msg=\"added image to changes\" newimage=quay.io/weaveworks/flux:master-4fd139b\n
ts=2018-06-07T13:23:07.537069912Z caller=daemon.go:549 component=daemon event=\"Automated release of quay.io/weaveworks/flux:master-4fd139b\" logupstream=true\n
ts=2018-06-07T13:24:13.021477431Z caller=images.go:67 component=sync-loop service=weave:deployment/weave-flux-agent container=flux currentimage=quay.io/weaveworks/flux:master-4fd139b msg=\"added image to changes\" newimage=quay.io/weaveworks/flux:master-180d2e3\n
ts=2018-06-07T13:24:23.8406753Z   caller=daemon.go:548 component=daemon event=\"Automated release of quay.io/weaveworks/flux:master-180d2e3\" logupstream=true\n
```

rade:
We had a slew of these over the weekend, all for the same image, this time not flux.

Interesting observations:
- the flip is always from the most-recent to the 2nd-most-recent (taking into account the tag filter), and then back
- the flip back happens within less than 2 minutes
- the flips were approx an hour apart. Never less, though sometimes a few hours were skipped. E.g. we had flips at 1:32, 2:35, 4:35, 5:35, 6:35, 11:35

The one hour is a smoking gun: @squaremo says that the memcached expiry is set to that.

rade:
We've had some more flipping. Slack said
```
9:16:44pm
CONTROLLER                STATUS   UPDATES
default:deployment/users  success  users: quay.io/weaveworks/users:master-20cda2fd -> master-7a30379f

9:17:56pm
[...]

PR #1247, opened 2018-07-24
make image sort stable, and correct

rade:
The ByCreatedDesc sort was broken in two ways:

1. If there were two images with a zero CreatedAt time, the sort order
would depend on the order in which they appeared in the
orginal (unsorted) list, thus making the order unstable.

2. Images with a zero CreatedAt time were considere more recent than
any images with a non-zero CreatedAt time, which is the exact opposite
of what we want.

These bugs _may_ be the cause of the 'flipping' between different
images we are seeing in #1127.


image/image_test.go
```diff
@@ -183,13 +183,28 @@ func TestImageInfoCreatedAtZero(t *testing.T) {
...
+	// more stability checks
+	for i := len(imgs)/2 - 1; i >= 0; i-- {
+		opp := len(imgs) - 1 - i
+		imgs[i], imgs[opp] = imgs[opp], imgs[i]
+	}
+	sort.Sort(ByCreatedDesc(imgs))
+	checkSorted(t, imgs)
+}
+
+func checkSorted(t *testing.T, imgs []Info) {
```
squaremo: Instead of this we could check the _property_ of being sorted, that is, `a_n < a_n+1`.

PR #1249, opened 2018-07-24
Be more explicit and concise during automation run

squaremo:
- say why an image was added for updating
 - add another safeguard against dubious updates -- check the image
   creation date is not zero
 - don't bother logging when nothing happens

Not intended to cl0se #1127, but may give us more information about what it think it's doing, if it does it again.

### DIFF
--- a/daemon/images.go
+++ b/daemon/images.go
@@ -41,6 +41,7 @@ func (d *Daemon) pollForNewImages(logger log.Logger) {
 
 	changes := &update.Automated{}
 	for _, service := range services {
+	containers:
 		for _, container := range service.ContainersOrNil() {
 			currentImageID := container.Image
 			pattern := policy.GetTagPattern(candidateServicesPolicyMap, service.ID, container.Name)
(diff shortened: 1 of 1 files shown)


daemon/images.go
```diff
@@ -41,28 +42,35 @@ func (d *Daemon) pollForNewImages(logger log.Logger) {
 	changes := &update.Automated{}
 	for _, service := range services {
 		for _, container := range service.ContainersOrNil() {
-			logger := log.With(logger, "service", service.ID, "container", container.Name, "currentimage", container.Image)
-
 			currentImageID := container.Image
-			if err != nil {
-				logger.Log("error", err)
-				continue
-			}
```
rade: that `err != nil branch` was dead code, right?

daemon/images.go
```diff
@@ -41,28 +42,35 @@ func (d *Daemon) pollForNewImages(logger log.Logger) {
...
 
 			filteredImages := imageRepos.GetRepoImages(repo).Filter(pattern)
 
 			if latest, ok := filteredImages.Latest(); ok && latest.ID != currentImageID {
 				if latest.ID.Tag == "" {
-					logger.Log("msg", "untagged image in available images", "action", "skip", "available", repo)
+					logger.Log("warning", "untagged image in available images", "action", "skip")
+					continue
+				}
+				if latest.CreatedAt.IsZero() {
```
rade: Because of the sort order, if there is any image with a non-zero `CreatedAt` then it will be considered newer than a zero-timestamp image. So if there is some weirdness going on where the `CreatedAt` timestamp of an image comes and goes, then this check does not prevent 'flipping'.

The ultra-cautious approach would be to skip updating if _any_ of the `filterImages` have a zero `CreatedAt`.

squaremo:
@rade points out that this won't avoid the situation where the image that would otherwise be considered the latest gets shuffled to the end of the list through having a zero timestamp.

daemon/images.go
squaremo: Well, we do look through the whole back catalogue, so may as well check this.

daemon/images.go
squaremo: Right; it used to parse the image string, but that gets done earlier now.

squaremo:
@rade Now checks all the images for a zero timestamp, and skips if it finds one. Care to take a quick look again?

PR #1252, opened 2018-07-25
refactor: make manifest fetch loop slighly more intelligible

rade:
I really should have tagged this onto #1251.

### DIFF
--- a/registry/client.go
+++ b/registry/client.go
@@ -75,9 +75,10 @@ func (a *Remote) Manifest(ctx context.Context, ref string) (image.Info, error) {
 	}
 	var manifestDigest digest.Digest
 	digestOpt := client.ReturnContentDigest(&manifestDigest)
-	manifest, fetchErr := manifests.Get(ctx, digest.Digest(ref), digestOpt, distribution.WithTagOption{ref})
+	dgst := digest.Digest(ref)
 
-interpret:
+fetch:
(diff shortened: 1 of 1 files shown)


registry/client.go
```diff
@@ -75,9 +75,10 @@ func (a *Remote) Manifest(ctx context.Context, ref string) (image.Info, error) {
 	}
 	var manifestDigest digest.Digest
 	digestOpt := client.ReturnContentDigest(&manifestDigest)
-	manifest, fetchErr := manifests.Get(ctx, digest.Digest(ref), digestOpt, distribution.WithTagOption{ref})
+	dgst := digest.Digest(ref)
 
-interpret:
+fetch:
+	manifest, fetchErr := manifests.Get(ctx, dgst, digestOpt, distribution.WithTagOption{ref})
```
squaremo: This isn't the same as the code it changes: here it will use the manifests.Get(...) option `distribution.WithTagOption{ref})` every single time, while before it only used that when fetching the initial manifest. Are you certain that is OK?

rade: ah, varargs. No, I am not sure.

squaremo: Looked it up again. `WithTagOption{ref}` makes the manifest fetcher look for the tag given, rather than the digest; in the lower down invocation of `manifests.Get(...)`, we actually want to fetch using the digest, rather than the tag. Adapting to that difference would probably cancel out the improved readability :-(

PR #1254, opened 2018-07-25
Rewrite operator to use chartsync

squaremo:


squaremo:
@stefanprodan Thanks for finding that problem -- I reproduced it, and fixed it in subsequent commits.

squaremo:
(merged using my "Get out of alpha free" card)

Issue #1205, opened 2018-07-05
Extract useful info from site/helm/helm-integration*.md


dholbach:
The content of `site/helm/helm-integration*.md` reads like a feature design document. Maybe we should extract all the relevant information for users into other parts of our docs and archive the rest.

squaremo:
Yep I agree, that material needs a bit of repurposing.

dholbach:
Largely fixed via #1357, remaining bits tracked in #1359 and #1360.

PR #1255, opened 2018-07-26
Fix docs for standalone flux pod

stephenmoloney:
What does this MR do?

- Minor doc fix where if the doc is for standalone, then
the helm operator should not be started

Why is this MR needed?

- To avoid confusion

### DIFF
--- a/site/helm/get-started.md
+++ b/site/helm/get-started.md
@@ -59,23 +59,33 @@ Add the Flux repository of Weaveworks:
 helm repo add weaveworks https://weaveworks.github.io/flux
 ```
 
-In this next step you install Weave Flux using `helm`. Simply:
+In this next step you install Weave Flux using `helm`. Simply
 
  1. Fork [flux-helm-test](https://github.com/weaveworks/flux-helm-test)
     on Github and
- 1. install Weave Flux and its Helm Operator by specifying your fork
+ 1. Install Weave Flux and its Helm Operator by specifying your fork
     URL:
 
-    *Just make sure you replace `YOURUSER` with your GitHub username
-    in the command below:*
-    ```sh
-    helm install --name flux \
-    --set helmOperator.create=true \
-    --set git.url=ssh://git@github.com/YOURUSER/flux-helm-test \
-    --set git.chartsPath=charts \
-    --namespace flux \
-    weaveworks/flux
-    ```
+      *Just make sure you replace `YOURUSER` with your GitHub username
+      in the command below:*
+      
+    - Using a public git server from `bitbucket.com`, `github.com` or `gitlab.com`:
+    
+      ```sh
+      helm install --name flux \
+      --set helmOperator.create=true \
+      --set git.url=ssh://git@github.com/YOURUSER/flux-helm-test \
+      --set git.chartsPath=charts \
+      --namespace flux \
+      weaveworks/flux
+      ```
+      
+    - Using a private git server:
+       
+      When deploying from a private repo, the known_hosts of the git server needs 
+      to be configured into a kubernetes configmap so that `StrictHostKeyChecking` is respected.
+      See [chart/flux/README.md](https://github.com/weaveworks/flux/blob/master/chart/flux/README.md#to-install-flux-with-the-helm-operator-and-a-private-git-repository)
+      for further installation instructions in this case.
 
 Allow some time for all containers to get up and running. If you're
 impatient, run the following command and see the pod creation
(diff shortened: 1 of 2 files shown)


squaremo:
That _is_ a little confusing. If you're using helm to install flux, then we might presume you want the helm operator to run, too. But, this is in instructions for running flux, as you say ...

stephenmoloney:
Agreed - which is why I am wondering if we should move this out of `site/standalone` completely and then move it into `site/helm` instead?

dholbach:
Yes. I agreed, it'd make sense to move it under `site/helm`. Maybe into its `README.md`?

stephenmoloney:
This is the current structure of helm readme:

```
./site/helm
|-- get-started.md
|-- helm-integration-requirements.md
`-- helm-integration.md
```

It doesn't fit in the latter 2, only `get-started.md` or a new file makes sense.

I could break up `get-started` into

1. For a public git repo do ... 

2. For a private git repo do...

sound ok ?

dholbach:
The helm-integration*.md files are not very helpful right now. I filed #1205 about them.

I meant `chart/flux/README.md` earlier.

@squaremo What do you think about @stephenmoloney's suggestion?

stephenmoloney:
I've made the suggested change in `site/helm/get_started.md` and removed from `site/standalone` so you can review and see what you think... 

https://github.com/stephenmoloney/flux/blob/b39a028e369f0485a40658939388b68a77153408/site/helm/get-started.md

dholbach:
My idea of `get-started.md` to have a self-contained guide which takes you along well-defined steps to having flux up and running in front of you, kind of like a hello-world example.

Maybe document in the chart's README?

squaremo:
> @squaremo What do you think about @stephenmoloney's suggestion?

It's reasonable to mention how to use known_hosts in a prominent place. The `get-started.md` file assumes you're using github, though, so it doesn't apply there (though it would be OK to have a warning, perhaps).

What we really need is more detailed and general instructions for installing it all with helm, but that would be rather unfairly expanding the scope of this PR :-) So let's just clear up the confusion, and separate Helm instructions from vanilla flux instructions.

site/helm/get-started.md
```diff
@@ -59,23 +59,42 @@ Add the Flux repository of Weaveworks:
...
+      --set git.url=ssh://git@github.com/YOURUSER/flux-helm-test \
+      --set git.chartsPath=charts \
+      --namespace flux \
+      weaveworks/flux
+      ```
+      
+    - Using a private git repo:
+
+      ```sh
+      KNOWN_HOSTS='domain ssh-rsa line1
```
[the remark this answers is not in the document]
stephenmoloney: I think these are valid points that should be addressed. 

### 1. How to get the value for known_hosts?

^ I'd like some advice for this part

### 2. How to pass the value for known_hosts into the `helm install` command

My own opinion is that we should show both ways

 1. Based on a `helm install --set-string known_hosts=${known_hosts}` method
 2. Based  on a `helm install --set-file known_hosts=path/to/file` method

The question then becomes, with this amount of information, where to put it without taking away from the flow of `getting_started.md` ...

stephenmoloney:
```
My idea of get-started.md to have a self-contained guide which takes you along well-defined steps to having flux up and running in front of you, kind of like a hello-world example.
```
I can appreciate that - so an alternative option is to simply point a link to another `flux/chart/flux/README.md` with a little note `if using a private repo, take a look at flux/chart/flux/README.md`

And another reference again at the bottom of `getting_started.md` to `flux/chart/flux/README.md`.

If we agree this is a reasonable approach, then I can put that into a separate PR later and as @squaremo  said, just remove that documentation for now from `standalone/...`

dholbach:
Sounds good to me.

Thanks a lot for working on this!

stephenmoloney:
I've changed it again. So I'd like to hold off until I make another PR for `chart/flux/README.md` changes with details about the helm install. Will be a few days.

site/helm/get-started.md
```diff
@@ -59,23 +59,34 @@ Add the Flux repository of Weaveworks:
...
+      --set git.url=ssh://git@github.com/YOURUSER/flux-helm-test \
+      --set git.chartsPath=charts \
+      --namespace flux \
+      weaveworks/flux
+      ```
+      
+    - Using a private git repo:
+       
+      When deploying from a private repo, the known_hosts of the git repo needs 
+      to be configured into a kubernetes secret so that `StrictHostKeyChecking` is respected.
```
squaremo: It's a ConfigMap in the chart -- nitpick!

stephenmoloney: yea, I'm writing the docs now for the README.md and I noticed that -- those are not exactly secret!

PR #1257, opened 2018-07-26
Document how to install flux/flux helm operator on private git host

stephenmoloney:
What does this MR/Commit do?

- Documents ways to install flux/flux helm operator from the command
line using ssh.known_hosts

Why is this MR needed?

- To enhance documentation examples

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -31,7 +31,7 @@ Add the weaveworks repo:
 helm repo add weaveworks https://weaveworks.github.io/flux
 ```
 
-To install the chart with the release name `flux`:
+#### To install the chart with the release name `flux`:
 
 ```console
 $ helm install --name flux \
@@ -40,7 +40,7 @@ $ helm install --name flux \
 weaveworks/flux
 ```
 
-To connect Flux to a Weave Cloud instance:
+#### To connect Flux to a Weave Cloud instance:
 
 ```console
 helm install --name flux \
@@ -49,7 +49,7 @@ helm install --name flux \
 weaveworks/flux
 ```
 
-To install Flux with the Helm operator (alpha version):
+#### To install Flux with the Helm operator (alpha version):
 
 ```console
 $ helm install --name flux \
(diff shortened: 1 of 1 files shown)


chart/flux/README.md
```diff
@@ -59,7 +59,63 @@ $ helm install --name flux \
...
+* Get the `ssh.known_hosts` keys by running the following command:
+
+```bash
+ssh-keyscan <your_git_server_domain>
+```
+
+To prevent a potential man-in-the-middle attack, it is desirable to
+run the ssh-keyscan from within the internal network of the git server
+to verify the ssh keys match expectations. Eg `ssh-keyscan 127.0.0.1` from
+within the git server and the returned keys should be identical to those
```
squaremo: Can we give people an ready-made example of verifying the host key from within the network? (Scanning this, I thought it was advising me to construct the known_hosts from the host itself; giving an explicit example would also guard against  that misapprehension).

stephenmoloney: Part of me would like to remove any instructions on how to get `known_hosts` because 

1. This is not my area of expertise - help wanted here

and 

2. Seems like something that goes beyond the remit of weaveworks flux to do

chart/flux/README.md
```diff
@@ -59,7 +59,63 @@ $ helm install --name flux \
 weaveworks/flux
 ```
 
-The [configuration](#configuration) section lists the parameters that can be configured during installation.
+#### To install Flux with the Helm operator and a private git repository:
```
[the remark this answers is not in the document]
stephenmoloney: `private git host` sounds right

chart/flux/README.md
```diff
@@ -59,7 +59,63 @@ $ helm install --name flux \
...
+
+When using a private git server, there is an additional requirement for
```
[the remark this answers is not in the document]
stephenmoloney: Yea, I'll rework it.

chart/flux/README.md
squaremo: Yep, I can understand those sentiments. I tried to find a reasonably definitive guide to initializing `known_hosts`, but couldn't. The advice is generally the same -- use ssh-keyscan, then check the fingerprints -- but never authoritative (and generally in the midst of instructions for doing something else ..).

stephenmoloney: The missing thing here is that by doing a simple `ssh-keyscan git.example.com` for example, one is not protected against a man in the middle attack. 
If doing it that way, one has to use an out-of-band method to verify those ssh keys returned.
So, in my case, this would be to login to the OVH panel through a web browser. Get the VPS I am using my gitlab server on. Then do an ssh-keyscan on the localhost network, `ssh-keyscan 127.0.0.1`. And you are right about generating the fingerprints - those fingerprints should match the ones generated using the console and without the console.

I would write this up as it pertains to OVH and leave it to the user to figure it out in their own case.

It would probably require some screenshots. 

This is very much a manual process, it's not just a couple of command line statements and done.

squaremo: It's a minefield isn't it. OK, the only thing I was really worried about here was that the second `ssh-keyscan` mention could be confusing, and people would end up with the server key from their own laptop or something along those lines. So perhaps it'd be better to just say that the means of verifying the host key(s) will differ and leave it at that.

stephenmoloney: Yea, I'll simplify it - and inform the user that they should `verify the ssh keys for known_hosts using an out of band method`

I'll update this PR in next day or two.

squaremo:
@stefanprodan or @dholbach Would you mind doing a quick check of these known_hosts+chart instructions? Merge if all good (or let me know). Thanks!

PR #1258, opened 2018-07-26
Support initContainers and {image, tag} and {image: {repository, tag}} Helm values

### LINKED ISSUE #1226: Support typical chart image format in FluxHelmRelease automation
The current FluxHelmRelease automation looks for `image: name:tag` in the values.

Most Helm charts use a different format:

```
image:
  repository: name
  tag: tag
```

It looks like it would be pretty trivial to support both formats, just have different handling for maps than strings: https://github.com/weaveworks/flux/blob/2d58c32ae195fda77450177e7d16bbc554c26975/cluster/kubernetes/resource/fluxhelmrelease.go#L63

Example:

```
$ helm create hello
$ cd hello
$ head values.yaml
# Default values for hello.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.

replicaCount: 1

image:
  repository: nginx
  tag: stable
  pullPolicy: IfNotPresent
$
```

Supporting this format would make helm-operator work with most charts out of the box.

squaremo:
1. Support initContainers in deployments, daemonsets etc., via

 - read and write to initContainers as well as containers in
   manifests, so we can update them;
 - report on the initContainers of Kubernetes resources

2. Many charts use an encoding for images of

```
    image: image/name
    tag: tag
```

rather than putting the whole image ref in one field. Support that
(after kubeyaml) by interpreting the values fields, or its values,
depending on whether a "tag" entry is present.

3. Still more charts use

```
    image:
      repository: org/name
      tag: v1
```

including those generated by `helm create` (and not substantially altered afterwards). Support that too.

Fixes #1226.

### DIFF
--- a/cluster/kubernetes/resource/spec.go
+++ b/cluster/kubernetes/resource/spec.go
@@ -22,13 +22,12 @@ type PodTemplate struct {
 
 func (t PodTemplate) Containers() []resource.Container {
 	var result []resource.Container
+	// FIXME(https://github.com/weaveworks/flux/issues/1269): account for possible errors (x2)
 	for _, c := range t.Spec.Containers {
-		// FIXME(michael): account for possible errors here
 		im, _ := image.ParseRef(c.Image)
 		result = append(result, resource.Container{Name: c.Name, Image: im})
 	}
 	for _, c := range t.Spec.InitContainers {
-		// FIXME(michael): account for possible errors here
 		im, _ := image.ParseRef(c.Image)
 		result = append(result, resource.Container{Name: c.Name, Image: im})
 	}
--- a/cluster/kubernetes/resource/spec.go
+++ b/cluster/kubernetes/resource/spec.go
@@ -27,11 +27,22 @@ func (t PodTemplate) Containers() []resource.Container {
 		im, _ := image.ParseRef(c.Image)
 		result = append(result, resource.Container{Name: c.Name, Image: im})
 	}
+	for _, c := range t.Spec.InitContainers {
+		// FIXME(michael): account for possible errors here
+		im, _ := image.ParseRef(c.Image)
+		result = append(result, resource.Container{Name: c.Name, Image: im})
+	}
 	return result
 }
 
 func (t PodTemplate) SetContainerImage(container string, ref image.Ref) error {
 	for i, c := range t.Spec.Containers {
+		if c.Name == container {
+			t.Spec.Containers[i].Image = ref.String()
+			return nil
+		}
+	}
+	for i, c := range t.Spec.InitContainers {
 		if c.Name == container {
 			t.Spec.Containers[i].Image = ref.String()
 			return nil
@@ -44,6 +55,7 @@ type PodSpec struct {
 	ImagePullSecrets []struct{ Name string }
 	Volumes          []Volume
 	Containers       []ContainerSpec
+	InitContainers   []ContainerSpec `yaml:"initContainers"`
 }
 
 type Volume struct {
--- a/bin/kubeyaml
+++ b/bin/kubeyaml
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker run --rm -i quay.io/squaremo/kubeyaml:0.4.0 "$@"
+docker run --rm -i quay.io/squaremo/kubeyaml:0.4.1 "$@"
--- a/bin/kubeyaml
+++ b/bin/kubeyaml
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker run --rm -i quay.io/squaremo/kubeyaml:0.3.3 "$@"
+docker run --rm -i quay.io/squaremo/kubeyaml:0.4.0 "$@"
--- a/bin/kubeyaml
+++ b/bin/kubeyaml
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker run --rm -i quay.io/squaremo/kubeyaml:0.4.1 "$@"
+docker run --rm -i quay.io/squaremo/kubeyaml:0.4.2 "$@"
--- a/cluster/kubernetes/resource/fluxhelmrelease.go
+++ b/cluster/kubernetes/resource/fluxhelmrelease.go
@@ -59,56 +59,94 @@ func FindFluxHelmReleaseContainers(values map[string]interface{}, visit func(str
 	// values:
 	//   image: 'repo/image:tag'
 	// ```
-	if imgInfo, ok := values["image"]; ok {
-		if imgInfoStr, ok := imgInfo.(string); ok {
-			imageRef, err := image.ParseRef(imgInfoStr)
-			if err == nil {
-				return visit(ReleaseContainerName, imageRef, func(ref image.Ref) {
-					values["image"] = ref.String()
-				})
-			}
-		}
+	if image, setter, ok := interpret_stringmap(values); ok {
+		visit(ReleaseContainerName, image, setter)
+		return nil
 	}
+
 	// Second most simple format:
 	// ```
 	// values:
 	//   foo:
-	//     image: repo/foo:tag
+	//     image: repo/foo:v1
 	//   bar:
-	//     image: repo/bar:tag
+	//     image: repo/bar:v2
 	// ```
+	// with the variation that there may also be a `tag` field:
+	// ```
+	// values:
+	//   foo:
+	//     image: repo/foo
+	//     tag: v1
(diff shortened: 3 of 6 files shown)


justinbarrick:
Helm charts typically use:

```
image:
  repository: nginx
  tag: latest
```

Are you intentionally using a different format than is commonly used? See #1226 for an example

squaremo:
> Are you intentionally using a different format than is commonly used?

Argh, no, I think I just made it up in my head :-(

squaremo:
Hey @justinbarrick, if you get a chance, can you have a glance at this (code and or trying out)? I do not trust my head to get it right ...

justinbarrick:
Your code and tests look good! You have a test that covers:

```
values:
  container:
    image:
      repository: blah
      tag: blah
```

Maybe you could cover the really basic case as well:

```
values:
  image:
    repository: blah
    tag: blah
```

And I’ll give this a try on my cluster tonight!

justinbarrick:
So I tried this out in my personal Kubernetes cluster, which is primarily managed using charts and FluxHelmReleases. Some notes:

* Tags that are parsed as numbers in YAML are ignored by flux (e.g., initially [this image](https://hub.docker.com/r/linuxserver/qbittorrent/tags/) confused it until I quoted the tag).
* On all of my FluxHelmReleases, I'd actually only included `image.tag` and left off repository, since I used the default from the chart's `values.yaml`. Since helm-operator has you vendor the charts, typically I vendor them and modify the `values.yaml` to set the defaults for my cluster, then in FluxHelmReleases I only put the values that change between application instances. I'm not sure if other people do it this way and I recognize this is probably unsolvable since Flux doesn't necessarily have access to the chart repository, but figured I'd mention it.
* A couple of third party charts do not follow any of these patterns:
  * [brigade](https://github.com/Azure/brigade/blob/master/charts/brigade/values.yaml#L40-L45) uses `controller.registry` and `controller.name` to form the image name.
  * [stable/influxdb](https://github.com/helm/charts/blob/master/stable/influxdb/values.yaml#L3-L5) uses `image.repo` instead of `image.repository`.
  * [stable/external-dns](https://github.com/helm/charts/blob/master/stable/external-dns/values.yaml#L2-L4) uses `image.name`.
* All of my other charts (a mix of stable / incubator / random githubs / my own) appear to be compatible.
* Upon adding the images to my FluxHelmReleases, flux upgraded the FluxHelmReleases that already had the automation annotation on them.
* I was able to use fluxctl to automate all of the charts I've been wanting to automate.
  * However, it doesn't appear to be upgrading these images, even though it detects they are out of date (maybe I am impatient):

```
ts=2018-07-28T19:47:06.772203252Z caller=images.go:74 component=sync-loop service=media:fluxhelmrelease/jackett container=chart-image repo=linuxserver/jackett pattern=* current=linuxserver/jackett:208 info="added update to automation run" new=linuxserver/jackett:221 reason="latest 221 (2018-07-27 22:19:04.471110059 +0000 UTC) > current 208 (2018-07-07 06:59:05.611229667 +0000 UTC)"
ts=2018-07-28T19:47:06.772338995Z caller=images.go:74 component=sync-loop service=media:fluxhelmrelease/plex container=chart-image repo=plexinc/pms-docker pattern=* current=plexinc/pms-docker:1.12.3.4973-215c28d86 info="added update to automation run" new=plexinc/pms-docker:public reason="latest public (2018-07-19 18:49:15.490859368 +0000 UTC) > current 1.12.3.4973-215c28d86 (2018-04-18 20:45:18.328855429 +0000 UTC)"
ts=2018-07-28T19:47:06.772457747Z caller=images.go:74 component=sync-loop service=media:fluxhelmrelease/qbittorrent container=chart-image repo=linuxserver/qbittorrent pattern=* current=linuxserver/qbittorrent:73 info="added update to automation run" new=linuxserver/qbittorrent:87 reason="latest 87 (2018-07-27 22:32:53.488895884 +0000 UTC) > current 73 (2018-05-11 22:30:52.167200023 +0000 UTC)"
```

justinbarrick:
This is a crappy patch that fixes the issue I described above: https://github.com/justinbarrick/flux/commit/acdc513144dda99063f1c3877a505a77a1cb19bf

It makes it work correctly in my cluster, but I'm not sure if that's how you want to fix it.

justinbarrick:
Another case to test might be:

```
values:
  image:
    repository: myimage
    tag: mytag
  anothercontainer:
    image:
      repository: otherimage
      tag: othertag
```

squaremo:
> This is a crappy patch that fixes the issue I described above: justinbarrick@acdc513

Ahh yeah I see the mistake I've made, thanks for puzzling that out. I wish I could think of a better way of dealing with the different types .. I'm tempted to just recursively transform the `map[sinterface{}]interface{}` values to `map[string]interface{}`. (But of course, you'd need a reference to the original value for the setter, well fiddly).

squaremo:
Thank you very much for checking it over @justinbarrick, that's great feedback. I'll do another round to fix up the things you pointed out.

> Another case to test might be:

At present I expect one or other of the formats, whichever is seen first -- I don't think there's any reason not to accept all entries that match _any_ format, though.

squaremo:
- Added test for image object at the top level

```
values:
  image:
    repository: foo
    tag: v1
```

and rewrote the FluxHelmRelease values "interpretation" code to cover the missing case shown in justinbarrick/flux@acdc513.

justinbarrick:
Just tried it in my cluster and it's working well!

squaremo:
> Tags that are parsed as numbers in YAML are ignored by flux (e.g., initially this image confused it until I quoted the tag).

 - [ ] Can probably deal with this just by checking for all the possible ground types YAML can put in a field.

> On all of my FluxHelmReleases, I'd actually only included image.tag and left off repository, since I used the default from the chart's values.yaml. Since helm-operator has you vendor the charts, typically I vendor them and modify the values.yaml to set the defaults for my cluster, then in FluxHelmReleases I only put the values that change between application instances. I'm not sure if other people do it this way and I recognize this is probably unsolvable since Flux doesn't necessarily have access to the chart repository, but figured I'd mention it.

Right; as it's implemented now, fluxd doesn't have access to the charts, so it won't know that the chart has a `repository` field that doesn't appear in the FluxHelmRelease. That's a limitation that stems from keeping the flux daemon and the helm operator separate.

You can put the repository value in the `FluxHelmRelease` -- a bit of a pain, and duplicates information, but a workaround.

> A couple of third party charts do not follow any of these patterns:
> brigade uses controller.registry and controller.name to form the image name.
> stable/influxdb uses image.repo instead of image.repository.
> stable/external-dns uses image.name.
> All of my other charts (a mix of stable / incubator / random githubs / my own) appear to be compatible.

We could extend the interpretation of FluxHelmRelease to cover these -- I'm sure there will always be exceptions, whatever we do. I've open an issue for these varieties anyway (#1262).

> Upon adding the images to my FluxHelmReleases, flux upgraded the FluxHelmReleases that already had the automation annotation on them.
> I was able to use fluxctl to automate all of the charts I've been wanting to automate.
> However, it doesn't appear to be upgrading these images, even though it detects they are out of date (if you write the test I described above, this bug appears):

@justinbarrick Did this problem still appear the more recent time you tried it?

squaremo:
>> Tags that are parsed as numbers in YAML are ignored by flux (e.g., initially this image confused it until I quoted the tag).
> 
>  Can probably deal with this just by checking for all the possible ground types YAML can put in a field.

Arp, nope. You just have to quote values, at least those that look like floats or ints. The reason is this: YAML will parse those number-like values as ints or floats, and the actual string cannot be reliably recovered. (OK, it can if it's an int, but not if it's a float and I'd rather have no special cases).

squaremo:
@aaron7 With Justin's thumbs up, I reckon this is ready for code review -- sorry for the premature review request

justinbarrick:
Oh good point in regards to floats! Octal and hex integers might have been an issue if you tried that too.

cluster/kubernetes/resource/spec.go
```diff
@@ -27,6 +27,11 @@ func (t PodTemplate) Containers() []resource.Container {
 		im, _ := image.ParseRef(c.Image)
 		result = append(result, resource.Container{Name: c.Name, Image: im})
 	}
+	for _, c := range t.Spec.InitContainers {
+		// FIXME(michael): account for possible errors here
```
aaron7: perhaps we can create an issue and write the issue number here?

squaremo: Good idea!

PR #1270, opened 2018-08-04
Manual releases comply with container image tag filters; and --force

### LINKED ISSUE #941: Allow to force deployment of a locked controller
In some occasions it would be nice to be able to force release a locked controller w/o unlocking and locking again. Has the benefit of retaining the lock message.

Something like
```
fluxctl release --force --update-all-images --controller=foo:deployment/bar
```
maybe with a sentry in case `--force` is being unintentionally used with `--all`?

### LINKED ISSUE #1054: tag filters should apply to manual updates
At the moment, tag filters are only taken into account for automatic updates, not manual updates. This is somewhat surprising behaviour and can easily lead to users accidentally releasing the wrong images when, say, running `--update-all` and expecting the tag filters to be taken into account.

The only time the filter should be ignored is when manually releasing a specific version of an image with `--update-image=...`.

rndstr:
This PR makes manual releases to comply with container image tag filters (previously only automated releases did so), and adds the option to `force` a release outside of these constraints. `force` also ignores locked controllers when targeted at specific controllers.

Please have a lock at individual commit messages for further details. 

Closes #941 
Closes #1054

### DIFF
--- a/update/release.go
+++ b/update/release.go
@@ -231,7 +231,12 @@ func (s ReleaseSpec) calculateImageUpdates(rc ReleaseContext, candidates []*Cont
 		for _, container := range containers {
 			currentImageID := container.Image
 
-			filteredImages := imageRepos.GetRepoImages(currentImageID.Name).FilterAndSort(policy.PatternAll)
+			tagPattern := policy.PatternAll
+			if pattern, ok := u.Resource.Policy().Get(policy.TagPrefix(container.Name)); ok {
+				tagPattern = policy.NewPattern(pattern)
+			}
+
+			filteredImages := imageRepos.GetRepoImages(currentImageID.Name).FilterAndSort(tagPattern)
 			latestImage, ok := filteredImages.Latest()
 			if !ok {
 				if currentImageID.CanonicalName() != singleRepo {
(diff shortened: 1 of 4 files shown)


squaremo:
Should this be for merge into master?

update/release.go
```diff
@@ -231,7 +234,15 @@ func (s ReleaseSpec) calculateImageUpdates(rc ReleaseContext, candidates []*Cont
 		for _, container := range containers {
 			currentImageID := container.Image
 
-			filteredImages := imageRepos.GetRepoImages(currentImageID.Name).FilterAndSort(policy.PatternAll)
+			tagPattern := policy.PatternAll
+			// Only consider _all_ images if specific image requested and forced
+			if s.ImageSpec == ImageSpecLatest || !s.Force {
```
squaremo: Similar dissonance with the comment vs the condition. Here I think it'd be better to have an `if..else`, since the "all" case is the special one:

```
var tagPattern policy.Pattern
// Use the container's filter unless given a specific image and forced
if s.Force && s.ImageSpec != ImageSpecLatest {
  tagPattern = policyPatternAll
} else {
  .. get it from the resource policy
}
```

rndstr: the `else` won't cover the case where `Get`ting the tag fails, though.

rade:
Shouldn't there have been some docs updates in this? E.g. in https://github.com/weaveworks/flux/blob/master/site/using.md#image-tag-filtering, where it says "When automation is turned on...". And surely the `--force` flag deserves a mention somewhere.

PR #1282, opened 2018-08-09
Allow colons in the name component of resource IDs

### LINKED ISSUE #1277: Names with : breaks things
I get the following errors

```
ts=2018-08-09T06:07:40.234783734Z caller=loop.go:376 component=sync-loop err="executing HTTP request: 400 Bad Request parsing default:clusterrolebinding/metrics-server:system:auth-delegator: invalid service ID\n"
ts=2018-08-09T06:07:40.237234687Z caller=loop.go:89 component=sync-loop err="executing HTTP request: 400 Bad Request parsing default:clusterrolebinding/metrics-server:system:auth-delegator: invalid service ID\n"
```

When trying to apply https://github.com/kubernetes-incubator/metrics-server/blob/master/deploy/1.8%2B/resource-reader.yaml. I realize that technically names in kubernetes can't have a `:` in most cases, however this is a very common practice for cluster roles. As is seen by this example coming directly from a core kubernetes component. There are quite a few other services out there that also attempt to do this, and they all cause this problem in flux.

squaremo:
There are published manifests that people want to be use with flux,
but use colons in resource names. Kubernetes admits these, so we may
as well too.

Fixes #1277. Will need to be vendored into weaveworks/service before being included in a release (ideally well before, say immediately after merging this).

### DIFF
--- a/resourceid_test.go
+++ b/resourceid_test.go
@@ -1,0 +1,38 @@
+package flux
+
+import (
+	"testing"
+)
+
+func TestResourceIDParsing(t *testing.T) {
+	type test struct {
+		name, id string
+	}
+	valid := []test{
+		{"full", "namespace:kind/name"},
+		{"legacy", "namespace/service"},
+		{"dots", "namespace:kind/name.with.dots"},
+		{"colons", "namespace:kind/name:with:colons"},
+		{"punctuation in general", "name-space:ki_nd/punc_tu:a.tion-rules"},
+	}
+	invalid := []test{
+		{"unqualified", "justname"},
+		{"dots in namespace", "name.space:kind/name"},
+		{"too many colons", "namespace:kind:name"},
+	}
+
+	for _, tc := range valid {
+		t.Run(tc.name, func(t *testing.T) {
+			if _, err := ParseResourceID(tc.id); err != nil {
+				t.Error(err)
+			}
+		})
+	}
+	for _, tc := range invalid {
+		t.Run(tc.name, func(t *testing.T) {
+			if _, err := ParseResourceID(tc.id); err == nil {
+				t.Errorf("expected %q to be considered invalid", tc.id)
+			}
+		})
+	}
+}
--- a/flux.go
+++ b/flux.go
@@ -13,9 +13,14 @@ import (
 var (
 	ErrInvalidServiceID = errors.New("invalid service ID")
 
-	LegacyServiceIDRegexp       = regexp.MustCompile("^([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)$")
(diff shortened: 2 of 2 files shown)


resourceid_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestResourceIDParsing(t *testing.T) {
+	type test struct {
+		name, id string
+	}
+	valid := []test{
+		{"full", "namespace:kind/name"},
+		{"legacy", "namespace/service"},
+		{"dots", "namespace:kind/name.with.dots"},
+		{"colons", "namespace:kind/name:with:colons"},
+		{"punctuation in general", "name-space:ki_nd/punc_tu:a.tion-rules"},
```
nicolerenee: You can actually have more than one `:`. In this case it's not specifying the namespace and name on opposite sides of it. This is a pattern that seems to only happen with clusterroles (from what I've seen). 

```
ubectl get clusterroles | grep ':'
system:aggregate-to-admin                                              1d
system:aggregate-to-edit                                               1d
system:aggregate-to-view                                               1d
system:auth-delegator                                                  1d
system:aws-cloud-provider                                              1d
system:basic-user                                                      1d
system:certificates.k8s.io:certificatesigningrequests:nodeclient       1d
system:certificates.k8s.io:certificatesigningrequests:selfnodeclient   1d
system:controller:attachdetach-controller                              1d
system:controller:certificate-controller                               1d
system:controller:clusterrole-aggregation-controller                   1d
system:controller:cronjob-controller                                   1d
system:controller:daemon-set-controller                                1d
<trimed>
```

nicolerenee: For example it leads to invalid names for deployments and other objects
```
k apply -f nginx.yaml
The Deployment "flux:nginx" is invalid: metadata.name: Invalid value: "flux:nginx": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.'
```

squaremo: Not sure what you're saying here. What's the content of nginx.yaml and where did it come from?

squaremo: Also: how did those cluster roles get those names, if kubernetes (well, kubectl) rejects names like that?

nicolerenee: The nginx.yaml just had a name set that was `flux:nginx` but was a very basic nginx deployment with nothing else in it. I'm saying the `:` is only allowed for roles and role bindings. Both namespaces and at the cluster level. Not for all resources. But also that it can have numerous `:` in it. The `system` ones above are standard on a new K8s cluster spun up with kubeadm.

nicolerenee: That works for me. Just need to update to allow multiple colons.

squaremo: The second to last test case above has two colons in the name.

PR #1283, opened 2018-08-11
Verify known_host ssh keys during image build

stephenmoloney:
What does this commit/MR do?

- Verifies the known_hosts match expectations after running a
ssh-keyscan on the git hosts

Why is this commit/MR needed?

- To mitigate (albeit unlikely) man-in-middle attacks


squaremo:
Cool, thank you. If we're going to check against hard-wired fingerprints, could we just do that locally, and commit the `known_hosts` file?

squaremo:
> If so, I thought about that and I think it is no harm to leave it there - because it verifies that the keys have not changed - the build should abort if one of the keys has changed which will flag this fact to the user.

Yes, that's a good point, we'd want to know if the keys have been changed, either way.

> Are keys ever changed ? I don't know about that really.

I cannot find any indication that github changes its keys, but it's considered good practice to do so I believe. :shrug:

squaremo:
(Yes I did go and find the fingerprints myself and check them :-)

PR #1284, opened 2018-08-12
Update go-k8s-portforward to latest version to provide kubectl auth plugins for GKE, AWS, to support managed clusters.

justinbarrick:


ByteFlinger:
It seems this issue is fixed for 1.6.0 however when I run 1.6.0 it towards my GKE cluster I keep getting the error `Error: No Auth Provider found for name "gcp"`

Is there anything specific one needs to do to get it to work?

squaremo:
@ByteFlinger I think that missing piece was added in #1334, which will be included in 1.6.1, released shortly (probably this week).

ByteFlinger:
Yeah someone on the slack channel said the same. I managed to get it
working with master.

Thank you for the update

PR #1287, opened 2018-08-14
Simplify verify_known_hosts and make deterministic

squaremo:
We can obtain the fingerprints for all entries in known_hosts at once;
so simplest way to determine if they are correct is to compare the
sorted output with a known good, sorted value.

This also avoids depending on the order of the known_host entries, by
comparing sorted values.


stephenmoloney:
This is cleaner. LGTM

PR #1288, opened 2018-08-14
Add a note on minimal node size for a Weave Flux test.

petervandenabeele:
As this are experiments, I was using 2 "default" nodes from terraform (`n1-standard-1`) and this is not large enough. Combined, the `kube-system` resources and `flux` and `memcahced`and `helloworld` together consumed 700m to 800m and this was enough to just block the Rolling Upgrade strategy with a forever pending pods.  Switching to a cluster with 3 times a `n1-standard-2` made the upgrade pass perfectly.


petervandenabeele:
Thanks for the quick merge :-)

PR #1289, opened 2018-08-14
Add documentation for --force

rndstr:
Follow-up to #1270

### DIFF
--- a/site/using.md
+++ b/site/using.md
@@ -329,6 +329,15 @@ CONTROLLER                     STATUS   UPDATES
 default:deployment/helloworld  success
 ```
 
+# Releasing an image to a locked controller
+
+It may be desirable to release an image to a locked controller while
+maintaining the lock afterwards. In order to not having to modify the
+lock policy (which includes author and reason), one may use `--force`:
+```
+fluxctl release --controller=default:deployment/helloworld --update-all-images --force
+```
+
 # Unlocking a Controller
 
 Unlocking a controller allows it to have manual or automated releases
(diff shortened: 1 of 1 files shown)


site/using.md
```diff
@@ -389,6 +396,22 @@ individually:
 fluxctl policy --controller=default:deployment/helloworld --tag='helloworld=prod-*' --tag='sidecar=prod-*'
 ``` 
 
+Manual releases without explicit mention of the target image will
```
rade: Does 'image' here include the tag? For a controller with multiple containers and images, does flux support updating a specific image to the latest version, e.g. with
```
fluxctl release --controller=default:deployment/helloworld --update-image=helloworld
```
? If so, would the tag filter be applied? It should.

Indeed, does flux support something like
```
fluxctl release --update-image=helloworld
```
for releasing the latest version of an image to all controllers that run that image? If so, the tag filter(s) should apply there too.

squaremo: > Indeed, does flux support something like
> 
> fluxctl release --update-image=helloworld
> for releasing the latest version of an image to all controllers that run that image? If so, the tag filter(s) should apply there too.

It doesn't -- firstly, taken literally that would replace the image field with an untagged image, which would simply run whichever image is present on the host; secondly, we check that the image exists in the image DB, and an untagged image won't pass that check.

site/using.md
```diff
@@ -329,6 +329,13 @@ CONTROLLER                     STATUS   UPDATES
 default:deployment/helloworld  success
 ```
 
+# Releasing a locked controller
```
squaremo: It's not really the controller that's being released, but the image, so this should read "Releasing an image to a locked controller". I think it's also worth describing the circumstances in which you might want to do so, and pointing out that the controller remains locked afterward.

site/using.md
```diff
@@ -389,6 +396,22 @@ individually:
...
+
+To release an image outside of tag filters, either specify the image:
+```
+fluxctl release --controller=default:deployment/helloworld --update-image=helloworld:dev-abc123
+```
+or use `--force`:
+```
+fluxctl release --controller=default:deployment/helloworld --update-all-images --force
+```
+
```
squaremo: Worth mentioning that automation may immediately undo such a release.

site/using.md
```diff
@@ -329,6 +329,15 @@ CONTROLLER                     STATUS   UPDATES
 default:deployment/helloworld  success
 ```
 
+# Releasing an image to a locked controller
+
+It may be desirable to release an image to a locked controller while
+maintaining the lock afterwards. In order to not having to modify the
+lock policy (which includes author and reason), one may use `--force`:
+```
```
squaremo: :100:

PR #1291, opened 2018-08-16
Get image pull secrets via serviceAccounts

### LINKED ISSUE #1043: imagePullSecrets must be duplicated on every controller manifest
The [recommended approach to using `imagePullSecrets`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account) is to attach them to a service account (usually `default`). This eliminates the need to duplicate the `imagePullSecret` on every controller that requires it.

However, this doesn't work with Flux today. Unless the controller explicitly specifies an `imagePullSecret`, Flux won't be able to poll its images.

I feel it would be much easier to add an imagePullSecret to the `flux` ServiceAccount like this:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    name: flux
  name: flux
imagePullSecrets:
  - name: my-private-registry
```

squaremo:
It is possible, and often more economical, to give ImagePullSecrets to
a service account (perhaps the default for the namespace), rather than
each workload individually.

So that flux will still find these secrets, look up the service
account, and its imagePullSecrets, if any. This involves one more RPC
to the API server (to get the service account), per workload.

Fixes #1043.

### DIFF
--- a/cluster/kubernetes/images.go
+++ b/cluster/kubernetes/images.go
@@ -19,7 +19,8 @@ func mergeCredentials(log func(...interface{}) error, client extendedClient, nam
 	if saName == "" {
 		saName = "default"
 	}
-	sa, err := client.ServiceAccounts(namespace).Get(saName, meta_v1.GetOptions{})
+
+	sa, err := client.CoreV1().ServiceAccounts(namespace).Get(saName, meta_v1.GetOptions{})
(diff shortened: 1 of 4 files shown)


squaremo:
Cheers :sparkler:

PR #1292, opened 2018-08-17
Image filtering with regular expressions

### LINKED ISSUE #1239: Support regex image tag filtering
At the moment Flux only offers support for wildcard filtering (e.g. `prod-*`, `v1.8.*`, etc).

This works perfectly fine if you are in control of how the images are getting tagged, however, if you are using images from an external vendor you will run in to problems pretty fast.

It is a common pattern to build additional images on top of a base image and tag them with a suffix describing the addition (e.g. `1.0` and `1.0-jmx`, see for an example https://hub.docker.com/r/datadog/agent/tags/). These images will get rolled out last so Flux will detect them as the eligible new image (as it looks for the last published images that matches the filter requirements).

I briefly discussed this issue some weeks ago with @squaremo on Slack who proposed a `semver:^1.8` solution (as an addition to `glob:1.8.*`). I do however think this solution would be insufficient as semver tagging is as common as using the latest commit hash.

Instead I would like to propose a regex filter option that looks for pattern matches, this is highly configurable and can get as complex (and nasty) as required. Eliminating issues with whatever weird tag the user wants to keep track of.

hiddeco:
Closes #1239

### DIFF
--- a/site/using.md
+++ b/site/using.md
@@ -416,6 +416,20 @@ fluxctl release --controller=default:deployment/helloworld --update-all-images -
 
 Please note that automation might immediately undo this.
 
+## Filter pattern types
+
+Flux currently offers support for `glob`, `semver` and `regexp` based filtering.
+
+### Glob
+
+The glob (`*`) filter is the simplest filter Flux supports, a filter can contain
+multiple globs:
+```
+fluxctl policy --controller=default:deployment/helloworld --tag-all='glob:master-v1.*.*'
+```
+
+### Semver
+
 If your images use [semantic versioning](https://semver.org) you can filter by image tags
 that adhere to certain constraints:
 ```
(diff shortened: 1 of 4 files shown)


site/using.md
```diff
@@ -430,6 +430,11 @@ fluxctl policy --controller=default:deployment/helloworld --tag-all='semver:*'
 Using a semver filter will also affect how flux sorts images, so
 that the higher versions will be considered newer.
 
+If your images have complex tags you can filter by regular expression:
+```
+fluxctl policy --controller=default:deployment/helloworld --tag-all='regexp:^([a-zA-Z]+)$'
+```
+
```
squaremo: I reckon it is worth noting that if you want to match the whole tag, you must bookend your pattern with `^` and `$`.

squaremo:
Nicely done, thank you!

PR #1296, opened 2018-08-20
Remove spurious ServicesWithPolicies

squaremo:
The method `ServicesWithPolicies` and its associated type,
`policy.ResourceMap` aren't needed. At best, the required value is
easily calculated at the call site, and at worst, the invocation
repeats a lot of work for a value (e.g., a resource's policy set) that
is already available in scope.


squaremo:
Thanks Aaron! :star:

PR #1297, opened 2018-08-20
Support multiple `-git-path` arguments

### LINKED ISSUE #1267: find yamls in more than one dir
I know there exists an option for ignoring specific manifests through an annotation in the manifest, but it would be nice to be able to configure flux to ignore entire directories. The use case I see for this is two-fold:
  * We'd like to be able to store both our production and non-production manifests in a single repository. We have a large number of [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets), where they are partitioned out by environment. We don't wish to deploy credentials for a staging environment onto the production cluster and vice versa.
  * Flux seems to get choked up over helm chart template files that are suffixed as `.yaml`. Ideally, we'd like to store all of our custom helm charts and config in the same repo, but don't wish to use the FluxHelmOperator just yet. Flux is still walking our helm chart directory tree and erring out when it finds a template file.

How are people getting around this limitation currently, and do you see this as being hard to implement?

squaremo:
The idea is to let people overlay bits of configuration by supplying more than one `--git-path` argument.

Much of the machinery for dealing with manifest files (and directories) already expected more than one path; but, it wasn't necessarily convenient to supply all of them in one go, so I've updated the Manifests interface and followed that through.

The git path appears in API results -- it's just an opaque string, though, so joining the now multiple paths and returning that should be fine.

Fixes #1267.

### DIFF
--- a/daemon/images.go
+++ b/daemon/images.go
@@ -96,6 +96,7 @@ func (r resources) IDs() (ids []flux.ResourceID) {
 }
 
 // getUnlockedAutomatedServices returns all the resources that are
+// both automated, and not locked.
 func (d *Daemon) getUnlockedAutomatedResources(ctx context.Context) (resources, error) {
 	resources, _, err := d.getResources(ctx)
 	if err != nil {
--- a/cluster/kubernetes/manifests.go
+++ b/cluster/kubernetes/manifests.go
@@ -10,8 +10,8 @@ import (
 type Manifests struct {
 }
 
-func (c *Manifests) LoadManifests(base, first string, rest ...string) (map[string]resource.Resource, error) {
-	return kresource.Load(base, first, rest...)
+func (c *Manifests) LoadManifests(base string, paths []string) (map[string]resource.Resource, error) {
+	return kresource.Load(base, paths)
 }
 
 func (c *Manifests) ParseManifests(allDefs []byte) (map[string]resource.Resource, error) {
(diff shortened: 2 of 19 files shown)


squaremo:
> it's just an opaque string, though, so joining the now multiple paths and returning that should be fine.

This is not quite true; if you round-trip this, and supply comma-delimited paths as the value for `--git-path`, it won't work. You need to break it into multiple `--git-path` arguments.

daemon/images.go
```diff
@@ -81,18 +86,28 @@ func (d *Daemon) pollForNewImages(logger log.Logger) {
...
+type resources map[flux.ResourceID]resource.Resource
+
+func (r resources) IDs() (ids []flux.ResourceID) {
+	for k, _ := range r {
+		ids = append(ids, k)
+	}
+	return ids
+}
+
+// getUnlockedAutomatedServices returns all the resources that are
```
rndstr: the suspension is real!

squaremo: :joy: owps, will try harder to complete my sentences.

squaremo:
Rebassed and cliffhanger comment resolved.

stephenmoloney:
Just wondering:

1. if the docs show how this works ?

2. If there is an open issue to support something similar for the helmOperator ?

squaremo:
> 1. if the docs show how this works ?

Possibly only the usage message :-S

PR #1298, opened 2018-08-21
Get individual namespaces when given whitelist

squaremo:
Asking for the full list of namespaces requires a cluster-scoped permission of listing namespaces; however, a common scenario for using the whitelist is that you want to restrict permissions.

If we simply Get the whitelisted namespaces, ignoring those we're forbidden to see (or that don't exist, as before), we don't need the cluster-scoped permission and can just be given permissions per namespace.

The trade is that we do an API request per whitelisted namespace. I expect there to be relatively few, though, so I don't think this is a huge deal.

### DIFF
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -313,9 +313,13 @@ func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
 	if len(c.nsWhitelist) > 0 {
 		nsList := []apiv1.Namespace{}
 		for _, name := range c.nsWhitelist {
-			if ns, err := c.client.CoreV1().Namespaces().Get(name, meta_v1.GetOptions{}); err == nil {
+			ns, err := c.client.CoreV1().Namespaces().Get(name, meta_v1.GetOptions{})
+			switch {
+			case err == nil:
 				nsList = append(nsList, *ns)
-			} else if !(apierrors.IsNotFound(err) || apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err)) {
+			case apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err) || apierrors.IsNotFound(err):
+				c.logger.Log("warning", "whitelisted namespace unauthorized, forbidden, or not found", "namespace", name)
+			default:
 				return nil, err
 			}
 		}
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -101,7 +101,7 @@ type Cluster struct {
 	version     string // string response for the version command.
 	logger      log.Logger
 	sshKeyRing  ssh.KeyRing
-	nsWhitelist map[string]bool
+	nsWhitelist []string
 
 	mu sync.Mutex
 }
@@ -114,11 +114,6 @@ func NewCluster(clientset k8sclient.Interface,
 	logger log.Logger,
 	nsWhitelist []string) *Cluster {
 
-	nsWhitelistMap := map[string]bool{}
-	for _, namespace := range nsWhitelist {
-		nsWhitelistMap[namespace] = true
-	}
-
 	c := &Cluster{
 		client: extendedClient{
 			clientset,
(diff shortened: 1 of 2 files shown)


petervandenabeele:
I presume this would be part of a fix for the question I raised recently:
https://weave-community.slack.com/messages/C2ND76PAA/convo/C2ND76PAA-1534779886.000100/
(the other part using an explicit whitelist of namespaces).

squaremo:
> I presume this would be part of a fix for the question I raised recently:
> https://weave-community.slack.com/messages/C2ND76PAA/convo/C2ND76PAA-1534779886.000100/
> (the other part being to use an explicit whitelist of namespaces).

This changes how whitelisting is implemented, so you need to give flux slightly fewer permissions. I'll do a follow-up that documents exactly what you do need, but with this PR, it should be a little as:

 - A role that can patch the SSH key secret in flux's namespace
 - A role that can list, get, create and update resources you want managed, for each namespace in the whitelist

justinbarrick:
Awesome!

cluster/kubernetes/kubernetes.go
```diff
@@ -315,20 +310,21 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
 // instance, in which case it returns a list containing the namespaces from the whitelist
 // that exist in the cluster.
 func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
-	nsList := []apiv1.Namespace{}
+	if len(c.nsWhitelist) > 0 {
+		nsList := []apiv1.Namespace{}
+		for _, name := range c.nsWhitelist {
+			if ns, err := c.client.CoreV1().Namespaces().Get(name, meta_v1.GetOptions{}); err == nil {
+				nsList = append(nsList, *ns)
+			} else if !(apierrors.IsNotFound(err) || apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err)) {
```
rade: Do we really want to be silent here? Won't that make it rather hard for users to diagnose why a namespace they specified in their whitelist is getting ignored by flux?

squaremo: It's a tricky one. If the namespace doesn't exist (yet), we'll get either NotFound (if the permissions allow listing namespaces) or Forbidden (if they don't). Failing the whole operation, which underlies much of the API, seems a bit punitive.

We could log it as a warning, I suppose -- that way at least it's flagged up to anyone investigating problems.

squaremo: Now logs it -- could be pretty verbose, but no straight-forward way to avoid that.

justinbarrick:
If you wanted to avoid spamming the logs you could cache in memory which namespaces you have logged errors for and rate limit it.

cluster/kubernetes/kubernetes.go
```diff
@@ -315,20 +313,29 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
...
+		nsList := []apiv1.Namespace{}
+		for _, name := range c.nsWhitelist {
+			ns, err := c.client.CoreV1().Namespaces().Get(name, meta_v1.GetOptions{})
+			switch {
+			case err == nil:
+				c.nsWhitelistLogged[name] = false // reset, so if the namespace goes away we'll log it again
+				nsList = append(nsList, *ns)
+			case apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err) || apierrors.IsNotFound(err):
+				if !c.nsWhitelistLogged[name] {
+					c.logger.Log("warning", "whitelisted namespace unauthorized, forbidden, or not found", "namespace", name)
```
rade: It would be rather nice to know _which_ of the three errors it is. Why not simply log `err`? AFAICT (from looking at https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go) that should produce something reasonably intelligible.

rade: I suppose one issue here is that the error may change. If we care enough, we could stash the last error message (well, ideally code, but getting hold of that requires some casting) in the map.

squaremo: I changed it to log the error, so that will give people an indication of what's wrong. Logging transitions between the different errors is overkill I reckon.

PR #1299, opened 2018-08-22
Helm chart: Allow helm-op to sync with a different git repo than Flux

stefanprodan:


chart/flux/templates/helm-operator-deployment.yaml
```diff
@@ -67,9 +70,9 @@ spec:
         {{- end }}
         {{- end }}
         args:
-        - --git-url={{ .Values.git.url }}
-        - --git-branch={{ .Values.git.branch }}
-        - --git-charts-path={{ .Values.git.chartsPath }}
+        - --git-url={{ $gitURL }}
```
squaremo: If you're using a different repo, you almost certainly need to mount a different SSH key. Unlike fluxd, we don't try to generate one (or give people a way to ask for it). The best that can be done for now is for people to create a secret with a private key, install the public key at github (or wherever), and supply the secret name as a parameter.

hiddeco: You can simply get around this by just reusing the same key as Flux uses with read only rights?

squaremo: github won't let you use the same SSH key for more than one repo.

hiddeco: Oh, you're completely right. I totally forgot about this restriction.

stefanprodan: @squaremo I've added the option to specify a SSH key secret for helm-op

chart/flux/values.yaml
```diff
@@ -33,6 +33,11 @@ helmOperator:
     url: ""
     branch: ""
     chartsPath: ""
+    # generate a SSH key named identity: ssh-keygen -q -N "" -f /tmp/identity
```
squaremo: I wouldn't put a private key in `/tmp`, anyone could find it!

stefanprodan: Show I change it to `$HOME/identity` ?

chart/flux/values.yaml
```diff
@@ -22,13 +22,22 @@ helmOperator:
...
-
+    keyFile: "tls.key"
+    certFile: "tls.crt"
+    caContent: ""
+  # override Flux git settings
+  git:
+    url: ""
+    branch: ""
+    chartsPath: ""
+    # generate a SSH key named identity: ssh-keygen -q -N "" -f $HOME/.ssh/identity
```
squaremo: Sorry to quibble again .. this could overwrite another key that's in use. I would just advise people to create it wherever they are, since 1. that's likely to be within their home directory and 2. ssh-keygen will make the file mode 0400 anyway.
2.) counts for `/tmp` too, but I'd still be a bit wary of putting keys there.)

You could also recommend people delete the file after creating the secret, too. They can always create another key if the secret disappears.

PR #1300, opened 2018-08-22
Add Flux Helm Operator FAQ

stefanprodan:


site/faq.md
```diff
@@ -352,3 +352,85 @@ running resources by exporting them from the cluster, and that
 only returns the kinds of resource mentioned above. So,
 annotating a running resource only works if it's one of those
 kinds; putting the annotation in the file always works.
+
+## Flux Helm Operator questions
+
+### My Helm charts have more than one container image. How can I automate the image tag update for all my containers?
+
+A container image list must have the following format:
```
squaremo: This raises the question "what is a container image list?". I suggest a little more preamble, to explain that the values.yaml in the chart, and consequently the values field in the FluxHelmRelease, must have one of the formats that follow.

squaremo: (I guess that really belongs in usage docs, rather than here)

stefanprodan:
@squaremo I've removed the multi-image example, better to have a dedicated docs section for this.

PR #1265, opened 2018-08-01
Record unusable image manifests, rather than fail

squaremo:
At present the registry client considers a non-Linux/AMD64 image an
error, which provokes the image "cache" warmer to fail for that entire
image repo. As more and more image repositories have Windows images,
this mean flux won't be able to automate or even report on many
workloads.

What we actually want to do is just exclude unusable images from the
result; and for bonus points, record the fact that it wasn't a usable
image. To this end, I've wrapped the cache entries in a struct that
also includes an "excluded reason". If that field is present, the
fetch is considered a success, but the image is not included in the
result for the repo.

The cache entry version doesn't need to be bumped, because old entries
can be parsed as new entries. Happily, no-one has to refresh their
image DB from scratch.

Fixes #741.

PR #1304, opened 2018-08-23
Correctly store Excluded manifests

squaremo:
I noticed that fluxd was continually fetching image manifests that
were "excluded", that is, had the wrong arch or otherwise weren't
suitable.

The design in #1265 was to mark those manifests as excluded, but count
them as successful fetches; and, to save the fact in memcached. The
implementation failed on three counts:

 - it would unconditionally use a cached result, even if excluded;

 - it was using the ID from excluded results (which have no ID value)
   as a key, so they got effectively thrown away anyway;

 - the encoding of results to JSON omitted the ExcludedReason field,
   because they embed the image.Info struct and thus inherit its JSON
   encoding methods.

This commit corrects those problems.


squaremo:
@rndstr Would you have a look at this when you have some review-time, please?

squaremo:
Thanks for the review, Roli :green_apple:

PR #1307, opened 2018-08-23
Doc/customized known hosts

petervandenabeele:
* a simple typo (double `file`)
* also adding a pointer to the doc (it was not immediately clear in this file, how to exactly do this; with the site documentation, it is OK).
* explain that `Maintainer` permissions are needed in the Gitlab context.
* explain that to create the configmap for the `known_hosts` on a container, the namespace needs to be explicitly given
* explain that the configmap can also be generated from a proper template.

### DIFF
--- a/site/standalone/setup.md
+++ b/site/standalone/setup.md
@@ -121,21 +121,24 @@ has been installed properly first, then come back. `ssh -vv $GITHOST`
 from within the container may help debug it.
 
 If it _did_ work, you will need to make it a more permanent
-arrangement. Back in that shell:
+arrangement. Back in that shell, create a configmap for the cluster. To make
+sure the configmap is created in the namespace of the flux or weave deployment,
+the namespace is set explicitly:
 
 ```sh
-container$ kubectl create configmap flux-ssh-config --from-file=$HOME/.ssh/known_hosts
+container$ kubectl create configmap flux-ssh-config --from-file=$HOME/.ssh/known_hosts -n $(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
 configmap "flux-ssh-config" created
 ```
 
-It will be created in the same namespace as the flux daemon, since
-you're creating it from within the flux daemon pod.
-
 To use the ConfigMap every time the Flux daemon restarts, you'll need
 to mount it into the container. The example deployment manifest
-includes an example of doing this, commented out. Uncomment that (it
-assumes you used the name above for the ConfigMap) and reapply the
+includes an example of doing this, commented out. Uncomment that:
+* an ssh-config volume
+* an ssh-config volumeMount
+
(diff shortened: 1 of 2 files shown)


site/standalone/setup.md
```diff
@@ -41,7 +41,11 @@ can restart `flux` and it will continue to use the same key.
 Flux connects to the repository using an SSH key.
 
 ***The SSH key must be configured to have R/W access to the repository***.
- 
+
+More specifically, in the case of standalone Flux, the ssh key must be able to
+create and update tags. E.g. in Gitlab, that means it requires `Maintainer`
+permissions. The `Developer` permission can create tags, but not update them.
+
```
squaremo: It's slightly more graded than that; if you want to be able to do releases through flux (automated, or otherwise), it has to be able to push commits, too.

site/standalone/setup.md
```diff
@@ -118,21 +125,43 @@ has been installed properly first, then come back. `ssh -vv $GITHOST`
...
-It will be created in the same namespace as the flux daemon, since
-you're creating it from within the flux daemon pod.
-
 To use the ConfigMap every time the Flux daemon restarts, you'll need
 to mount it into the container. The example deployment manifest
-includes an example of doing this, commented out. Uncomment that (it
-assumes you used the name above for the ConfigMap) and reapply the
+includes an example of doing this, commented out. Uncomment that:
+* an ssh-config volume
+* an ssh-config volumeMount
```
squaremo: This parses a bit awkwardly; maybe you could say "Uncomment these two blocks:" and just show the two entries in code fences -- they are pretty small.

site/standalone/setup.md
petervandenabeele: Agree :-)

I was writing up something, but I do not have enough background for those scenario's. Can someone else complete this explanation then ?

site/standalone/setup.md
petervandenabeele: Agree :-)

I will add a small new commit to fix this.

petervandenabeele: I tried making it clear. Not entirely sure if this now really a good way to describe it (the code snippets without context seem not very instructive ...).

Anyone has better ideas on this ? Thx :-)

site/standalone/setup.md
petervandenabeele: Hi @squaremo to get this PR not get stale, I could also simply remove the sentence above, so that at least the other aspects of this PR are not lost.

squaremo: Good idea; let's defer that bit.

squaremo: Maybe change it to list the things it must be able to do? Which is:

 - push commits to your chosen branch
 - force push the sync tag

petervandenabeele:
Thanks :-)

PR #1314, opened 2018-08-28
Add docker-config to chart and example deployment

Alien2150:


chart/flux/templates/deployment.yaml
```diff
@@ -51,6 +55,9 @@ spec:
             readOnly: true
           - name: git-keygen
             mountPath: /var/fluxd/keygen
+          - name: docker-creds
+            mountPath: /docker-creds
```
squaremo: Can you make this `/etc/fluxd/docker`, just for consistency's sake? (OK, there is not a lot of consistency on display otherwise, but we have to start somewhere ..)

deploy/flux-config-map.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
squaremo: If this gets applied again, it will overwrite a value that's been updated by something else (e.g., a sidecar, or a controller, creating a docker config for ECR). Perhaps I'm misunderstanding how you're intending this to be used, though?

chart/flux/templates/deployment.yaml
```diff
@@ -34,6 +34,10 @@ spec:
       - name: git-keygen
         emptyDir:
           medium: Memory
+      - name: docker-creds   
+        secret:
+          secretName: {{ template "flux.fullname" . }}-docker-creds
```
squaremo: It doesn't look like this secret will be created when the chart is installed; will that mean the deployment doesn't run?

deploy/flux-deployment.yaml
```diff
@@ -25,6 +25,11 @@ spec:
       - name: git-keygen
         emptyDir:
           medium: Memory
+       
+      # This is holding the reference to docker-credentials if provided from the outside 
+      - name: docker-creds
+        configMap:
```
squaremo: I think this is better as a secret, as for the chart.

chart/flux/templates/deployment.yaml
Alien2150: Yeah my fail. Not sure how it went in there. Probably i messed around with one version. Fixing

squaremo:
Would you mind rebasing on master, rather than merging it into the branch -- it makes the history easier to follow. Thanks! (I think it'll rebase cleanly, but if not, I can give it a go.)

deploy/flux-config-map.yaml
Alien2150: Hey good catch. So the problem with ECR atm is that the token is running out after 12hrs. So I would need to recreate it. But maybe i can use a variable here to make it a bit more flexible. Or do you have a better idea of how to make the config.json file available but overwriteable from outside?

Alien2150:
Can try rebasing :)

chart/flux/templates/deployment.yaml
Alien2150: Sure 👍

deploy/flux-config-map.yaml
Alien2150: A btw @squaremo. Do i need to make those charts adjustment in deploy, charts or both? Was confusing me a lot that there actually 2 places

squaremo: The hard requirement is that the secret exists at the time pods are created, so it can be mounted. I think it's OK if another process goes and updates the secret _after_ it's been created.

If the secret doesn't contain the actual file, that is also OK. I will need to make sure that `fluxd` doesn't break if the docker-config file isn't present; but otherwise, if you update the secret from another pod or by hand, the file will be created or updated in the volume (after some minutes), and `fluxd` will be able to use it.

Crucially, if the secret doesn't contain a `data` field, it won't overwrite the existing contents when you reapply the manifest (e.g., if you upgrade the chart). We rely on this for the SSH key generation (see `deploy/flux-secret.yaml`).

squaremo: > Do i need to make those charts adjustment in deploy, charts or both?

You can just add it to the chart if you wish. The files in `deploy/` are just examples, for people who aren't using helm -- if you do add things there, consider comments explaining what the various bits are.

deploy/flux-deployment.yaml
Alien2150: indeed.

stephenmoloney:
Just wondering, what are the docker-creds for ?

Alien2150:
@stephenmoloney u can provide docker credentials when pulling images. This is needed when pulling images from a private registry (Like ECR)

stephenmoloney:
But I found that I needed to add docker credentials for `pullSecrets` on a per namespace basis so how does this work across all namespaces ?

squaremo:
> So I still need to add the secret on a per namespace basis though, is that right ?

If you're giving deployments `imagePullSecrets`, yes. (The next release of fluxd will also look for secrets linked to the service account; but, those are still namespaced)

Alien2150:
So tried rebasing. Hopfully I did it right.

Alien2150:
@squaremo So the idea is to overwrite the secret (So setting it via --set. E.g helm .... --set docker.registry.username=AWS --docker.registry.password would actually be the AWS ECR token (Just copied that one from helm: https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md)). Would helm not need something like a default value to being able to render the values in the secret.yaml?

deploy/flux-config-map.yaml
Alien2150: Will fix this one here. Actually the example should also use a secret

stefanprodan:
I would not create the docker secret via Helm, but just pass in the secret name like we are doing here https://github.com/weaveworks/flux/blob/master/chart/flux/values.yaml#L49

Alien2150:
So just paste it as an example into deploy and remove it from charts? @stefanprodan

stefanprodan:
After looking at https://github.com/weaveworks/flux/issues/1324 and @bzon ECR sidecar https://github.com/bzon/ecr-k8s-secret-creator I think it's best to create an empty secret from the chart like we are doing for the SSH key here https://github.com/weaveworks/flux/blob/master/chart/flux/templates/secret.yaml

stefanprodan:
@Alien2150 I would remove the user/pass from the chart values and keep `useCredentials` flag only. In the future we could add @bzon sidecar as an optional component.

Alien2150:
@johnraz let me recap what has happended / what was released. Got the impression that mentioned sidecar is a bit more advanced and is pointing to the right direction.

johnraz:
Sure - just FTR I need to use GCR not ECR, so I guess having a "forever there" secrets would be enough in my case.

Alien2150:
@johnraz do you mean google? Cause that would be supported natively iirc

johnraz:
@Alien2150 Yep, Google Container Registry! I've been looking around quite a bit and ended up here, do you mind pointing me to the doc maybe?

johnraz:
https://github.com/weaveworks/flux/blob/master/site/faq.md#how-do-i-give-flux-access-to-an-image-registry

Maybe I've been confused by this and should try imageCreds ...

Alien2150:
Iirc the issue was (why this PR the was made) is that ECR requires a token every 12 hours while gcr will work with fixed credentials.

johnraz:
The problem I'm facing is to be able to provide a docker config via the Helm chart, relying to the regular deployment and using `--docker-config` would fix my problem, I just can't use docker config in the current state of the Chart (without your changes that is).

> Iirc the issue was (why this PR the was made) is that ECR requires a token every 12 hours while gcr will work with fixed credentials.

Yes but I'm still not sure if Google still requires the use of a docker config, that doesn't change every 12 hours...

The doc says:
> To work around the exceptional cases, you can mount a docker config into the Flux container. See the argument --docker-config in the daemon arguments reference.

I guess I should first give it a try instead of reading the doc and get confused 😝 I'll let you know if it works by using imagePullSecrets.

stefanprodan:
Closing this in favour of #1762

PR #1321, opened 2018-08-29
 Exclude locked workloads for container type release

### LINKED ISSUE #1313: Exclude locked workloads for container type release
Locked workloads should be excluded from a `container` type release the same way as they are excluded for automated and other manual releases. Adding support for `--force` will allow to override this.

rndstr:
For `containers` type releases, locked workloads are now excluded. In
case one wants to include them nevertheless, a `Force` option can be set
to ignore the lock-flag. For `fluxctl`, the `--force` flag will allow
updating containers of locked workloads interactively.

Fixes #1313

### DIFF
--- a/cmd/fluxctl/release_cmd.go
+++ b/cmd/fluxctl/release_cmd.go
@@ -78,7 +78,6 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 		return err
 	}
 
-
 	switch {
 	case len(opts.controllers) <= 0 && !opts.allControllers:
 		return newUsageError("please supply either --all, or at least one --controller=<controller>")
@@ -160,7 +159,7 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 			return err
 		}
 
-		spec, err := promptSpec(cmd.OutOrStdout(), result, opts.verbosity)
+		spec, err := promptSpec(cmd.OutOrStdout(), result, opts.verbosity, opts.force)
 		if err != nil {
 			fmt.Fprintln(cmd.OutOrStderr(), err.Error())
 			return nil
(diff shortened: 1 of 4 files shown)


cmd/fluxctl/release_cmd.go
```diff
@@ -160,7 +159,7 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 			return err
 		}
 
-		spec, err := promptSpec(cmd.OutOrStdout(), result, opts.verbosity)
+		spec, err := promptSpec(cmd.OutOrStdout(), result, opts.verbosity, opts.force)
```
squaremo: Given that the `force` argument just gets written into the spec, you could do that here after calling an unmodified `promptSpec`.

But I wonder if this is the most useful interpretation of "force" in the interactive mode. To use it, you have to decide ahead of time whether you need it -- and you don't necessarily know what would be left out if not used, so you have to figure that out first. Interactively, you might expect to be shown updates that would occur _except that_ they are locked, and select them individually. Not a requirement for this PR.

release/releaser.go
```diff
@@ -34,7 +34,7 @@ func Release(rc *ReleaseContext, changes Changes, logger log.Logger) (results up
 	before, err := rc.LoadManifests()
 	updates, results, err := changes.CalculateRelease(rc, logger)
 	if err != nil {
-		return nil, err
+		return results, err
```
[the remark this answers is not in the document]
rndstr: I wonder why the other unhappy paths return it as well. Looking at all the places it is called (in this repo), `Result` is never inspected if it errors. I switched all of them to return `nil`.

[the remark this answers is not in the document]
squaremo: Probably my fault -- no return type discipline, me :-/

release/releaser.go
```diff
@@ -39,19 +39,19 @@ func Release(rc *ReleaseContext, changes Changes, logger log.Logger) (results up
 
 	err = ApplyChanges(rc, updates, logger)
 	if err != nil {
-		return results, MakeReleaseError(errors.Wrap(err, "applying changes"))
+		return nil, MakeReleaseError(errors.Wrap(err, "applying changes"))
```
squaremo: Thanks for making this consistent

release/releaser_test.go
```diff
@@ -891,44 +901,104 @@ func Test_UpdateContainers(t *testing.T) {
...
 			}
-			specs.SkipMismatches = ignoreMismatches
-			results, err := Release(ctx, specs, log.NewNopLogger())
-			if tst.Expected[ignoreMismatches].Status != "" {
-				assert.Equal(t, tst.Expected[ignoreMismatches], results[hwSvcID], name)
-				assert.Equal(t, tst.Commit[ignoreMismatches], specs.CommitMessage(results), name)
-			} else {
-				assert.Error(t, err, name)
-			}
+			t.Run(name, func(t *testing.T) {
```
squaremo: Neat and tidy.

PR #1315, opened 2018-08-28
docs: add CONTRIBUTING.md and DCO docs (from scope, modified for flux)

dholbach:
Let me know if there is anything special about the way Flux and its development works and I'll make sure to cover it.

PR #1323, opened 2018-08-30
Update contributor docs

dholbach:
This is a follow-up to #1315, adds link to CoC, is clearer in a few places.

@squaremo Can you take a brief look at https://github.com/weaveworks/flux/blame/master/CONTRIBUTING.md#L66 - it talks about an ACK from two maintainers. That's not actually what we want, right? :-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,6 +62,14 @@ These things will make a PR more likely to be accepted:
  * tests for old code!
  * new code and tests follow the conventions in old code and tests
  * a good commit message (see below)
+ * All code must abide [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments)
+ * Names should abide [What's in a name](https://talks.golang.org/2014/names.slide#1)
+ * Code must build on both Linux and Darwin, via plain `go build`
+ * Code should have appropriate test coverage, invoked via plain `go test`
+
+In addition, several mechanical checks are enforced.
+See [the lint script](/lint) for details.
+
 
 In general, we will merge a PR once two maintainers have endorsed it.
 Trivial changes (e.g., corrections to spelling) may get waved through.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,7 +66,8 @@ These things will make a PR more likely to be accepted:
 - All code must abide [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments)
 - Names should abide [What's in a name](https://talks.golang.org/2014/names.slide#1)
 - Code must build on both Linux and Darwin, via plain `go build`
-- Code should have appropriate test coverage, invoked via plain `go test`
+- Code should have appropriate test coverage and tests should be written
+  to work with `go test`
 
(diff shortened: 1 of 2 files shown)


squaremo:
> it talks about an ACK from two maintainers. That's not actually what we want, right? :-)

It is not -- we are happy to merge with one approval from a maintainer.

CONTRIBUTING.md
```diff
@@ -37,31 +38,38 @@ This is a rough outline of how to prepare a contribution:
...
- * a good commit message (see below)
+- a well-described requirement
+- tests for new code
+- tests for old code!
+- new code and tests follow the conventions in old code and tests
+- a good commit message (see below)
+- All code must abide [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments)
+- Names should abide [What's in a name](https://talks.golang.org/2014/names.slide#1)
+- Code must build on both Linux and Darwin, via plain `go build`
+- Code should have appropriate test coverage, invoked via plain `go test`
```
squaremo: We're a bit more flexible than this suggests -- we have some integration tests that need external dependencies for example. But this'll do for now.

CONTRIBUTING.md
```diff
@@ -37,31 +38,38 @@ This is a rough outline of how to prepare a contribution:
...
+
+In addition, several mechanical checks are enforced.
+See [the lint script](/lint) for details.
```
squaremo: That doesn't get run, it takes way too long and I hate lint. Best to remove this bit :-)

dholbach:
Thanks, updated.

CONTRIBUTING.md
```diff
@@ -37,34 +38,40 @@ This is a rough outline of how to prepare a contribution:
...
+- a good commit message (see below)
+- All code must abide [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments)
+- Names should abide [What's in a name](https://talks.golang.org/2014/names.slide#1)
+- Code must build on both Linux and Darwin, via plain `go build`
+- Code should have appropriate test coverage, invoked via plain `go test`
 
-In general, we will merge a PR once two maintainers have endorsed it.
-Trivial changes (e.g., corrections to spelling) may get waved through.
+In addition, several mechanical checks are enforced.
+See [the lint script](/lint) for details.
```
squaremo: That doesn't get run, it takes way too long and I hate lint. Best to remove this bit :-)

dholbach: Should the script in general be removed/fixed/updated? (Happy to remove the line in any case.)

CONTRIBUTING.md
squaremo: `go test` on its own will just run tests in the current directory, and there are few tests in the top directory.

I think the advice is intended to mean "you should be able to run your tests using just go test with appropriate arguments", i.e., you don't need to use any other tooling. But it's ambiguous. Perhaps it would be clearer to say

> Tests should be written to work with `go test`

Do you think that resolves the ambiguity?

dholbach: Yes, sounds good.

dholbach:
I think I covered all your comments.

PR #1332, opened 2018-09-01
fluxctl: Skip port forwarding for version command

### LINKED ISSUE #1318: "fluxctl version" does not report back, if fluxd can't be reached
```sh
$ fluxctl version
Error: Get https://192.168.39.52:8443/api/v1/namespaces/default/pods?labelSelector=name+in+%28flux%2Cfluxd%2Cweave-flux-agent%29: dial tcp 192.168.39.52:8443: i/o timeout
Run 'fluxctl version --help' for usage.
$
```

It'd be good if `fluxctl` would print out its own version first, then error out if it can't determine the `fluxd` version. That'd be mimicking the behaviour of `kubectl`.

stefanprodan:
Fix #1318

### DIFF
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -92,6 +92,12 @@ func (opts *rootOpts) Command() *cobra.Command {
 }
 
 func (opts *rootOpts) PersistentPreRunE(cmd *cobra.Command, _ []string) error {
+	// skip port forward for version command
+	switch cmd.Use {
+	case "version":
+		return nil
+	}
+
 	opts.Namespace = getFromEnvIfNotSet(cmd.Flags(), "k8s-fwd-ns", opts.Namespace, envVariableNamespace)
 	opts.Token = getFromEnvIfNotSet(cmd.Flags(), "token", opts.Token, envVariableToken, envVariableCloudToken)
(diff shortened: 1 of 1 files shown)


stefanprodan:
I've rerun the CircleCI build, they had an outage 3h ago.

cmd/fluxctl/root_cmd.go
```diff
@@ -92,6 +92,12 @@ func (opts *rootOpts) Command() *cobra.Command {
 }
 
 func (opts *rootOpts) PersistentPreRunE(cmd *cobra.Command, _ []string) error {
+	// skip port forward for version command
+	switch cmd.Use {
+	case "version":
+		return nil
+	}
+
```
rade: I am not familiar with this code; could you please explain how this works? In particular, how does this result in `fluxctl` printing out it's own version _and then_ the `fluxd` version? Which is what #1318 proposes (with the latter step failing if it cannot connect). By my naive reading of the code this change will stop `fluxctl` from connecting to `fluxd` altogether.

stefanprodan: @dholbach assumed that the `version` command prints the client and server versions (like kubectl does) but in fact fluxctl only prints the client version.

stefanprodan: If we want the same behaviour like kubectl we could make a new issue for this feature. My PR prevents the `version` command from crashing if fluxd is not accessible.

PR #1342, opened 2018-09-05
helm charts: override the ./kube/config file

brandon-bethke-neudesic:
Override the ./kube/config for in the flux pod when using helm to deploy flux

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -23,6 +23,9 @@ spec:
       serviceAccountName: {{ template "flux.serviceAccountName" . }}
       {{- end }}
       volumes:
+      - name: kubedir
+        configMap:
+          name: {{ template "flux.fullname" . }}-kube-config
       - name: sshdir
         configMap:
           name: {{ template "flux.fullname" . }}-ssh-config
@@ -43,6 +46,8 @@ spec:
             containerPort: 3030
             protocol: TCP
           volumeMounts:
+          - name: kubedir
+            mountPath: /root/.kube
           - name: sshdir
             mountPath: /root/.ssh
             readOnly: true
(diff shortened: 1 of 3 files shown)


stefanprodan:
@brandon-bethke-neudesic how did you test this? If you install the chart with the defaults is Flux still working?

brandon-bethke-neudesic:
We tested by deploying flux.  We verrified, that the flux container had the correct kube/config file mounted as specified in the values.yaml. We haven't encountered any issues thus far and flux is deploying the FHR and the artifacts in our charts without issue. Are you encountering issues?

The updated flux chart is mounting the ./kube/config file to the same location as specified in Dockerfile.flux `COPY ./kubeconfig /root/.kube/config`
And the default value in values.yaml is the same value as found in file: flux/docker/kubeconfig

**The Reason for the change**
We needed to override the default namespace because we need to transition/promote releases between namespaces. Therefore putting the namespace in the FHR is not an option or portable and there was no other way, we could find, to force flux to install artifacts to a specific namespace.

Or to put it more generally. We have flux installed in a namespace and we want all kube artifacts to be installed in that namespace without having to specify the namespace in the FHR file.

brandon-bethke-neudesic:
Yes. Our plan is to run a flux instance in each of our namespaces which represent the environments: test, stage, prod, etc
We have a single 'release repository' which contains only helm charts and FHR files. Each flux instance is scanning that single repo, but looking at different branches. To promote a release, we simply merge a branch.

stefanprodan:
Besides FHRs and charts do you also have secrets or ingress definitions? I image you can't merge an ingress definition from one namespace to another without changing the public domain name.

brandon-bethke-neudesic:
We have those types of artifacts. I am not familiar with the details, but we have an instance of traefik that is configured to route requests to another instance of traefik in the desired namespace based on a cookie.

PR #1346, opened 2018-09-07
Excluding Goland idea based IDE meta files

casibbald:
Incomplete .gitignore.
Please may you accept the additional ignore line.


stefanprodan:
This should be in your `~/.gitignore`

Issue #1352, opened 2018-09-10
Support id filter for listing services


rndstr:
`ListServices` should also support a `services` filter to only return requested ids.
In case the caller only cares about a subset we should not waste time requesting the whole bunch from Kubernetes.

PR #1355, opened 2018-09-13
Provide detailed pod status when listing services

lelenanam:
For deployments, daemonSet, statefulSet provide structure with:
 - status
 - numbers of desired, updated, ready, outdated pods
 - pod conditions with message and reason

Fixes https://github.com/weaveworks/flux/issues/1303

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,8 +4,8 @@
 [[projects]]
   name = "cloud.google.com/go"
   packages = ["compute/metadata"]
-  revision = "64a2037ec6be8a4b0c1d1f706ed35b428b989239"
-  version = "v0.26.0"
+  revision = "c728a003b238b26cef9ab6753a5dc424b331c3ad"
+  version = "v0.27.0"
 
 [[projects]]
   name = "github.com/Azure/go-autorest"
@@ -228,7 +228,7 @@
     "openstack/utils",
     "pagination"
   ]
-  revision = "07d15af37699505fb5535c0962f49a10379d12ff"
+  revision = "13af1c2a38cfbc60a7a13062338838301035d91d"
 
 [[projects]]
   name = "github.com/gorilla/context"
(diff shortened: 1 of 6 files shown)


squaremo:
> * pod conditions with message and reason

This appears to increase the API use from `O(namespaces x kinds)` to `O(namespaces x kinds + controllers)`, where `controllers` is likely to be at least as larger as the other addend and probably much larger. Putting some numbers in there, our dev environment has 14 namespaces, and we look for five different kinds; and there about one hundred controllers running. It'll also add to the size of the response -- where it scaled with the number of controllers, now it'll scale with the number of pods.

I don't think flux should be reporting the status of individual pods. If you (the user) need to investigate why a particular rollout seems to be stuck, there should be better tools out there for getting that information.

rndstr:
Agree that we don't need the full pod status for every service listing. Maybe we can only provide it if explicitly requested through an option? Combined with #1352 this would reduce the calls dramatically.

> I don't think flux should be reporting the status of individual pods. If you (the user) need to investigate why a particular rollout seems to be stuck, there should be better tools out there for getting that information.

The pod states primarily serve as an indicator that a rollout is stuck.

squaremo:
> The pod states primarily serve as an indicator that a rollout is stuck.

I'm still not keen on including information for all pods, individually. And diagnosis of problems with pods belongs in another component (doesn't scope look at individual pods?).

But I'm not against flux saying so if it can detect that a rollout is stuck (when asked specifically about it). On that basis, may I suggest:

 - getting info on the pods themselves is behind an option in the ListControllers API
 - it's not fetched if the controller is in a full ready state (and if it's possible with client-go, the query is limited to pods that aren't ready)
 - the information from the pods is interpreted and summarised at little, such that the CLI or UI can say "the rollout is stuck because the new image can't be pulled" or some such.

lelenanam:
Close this and open https://github.com/weaveworks/flux/pull/1380 for controller status without all pods details.

PR #1364, opened 2018-09-14
Don't create git secret if name is overridden

mt-inside:
I want to supply my own private keys for github access (I am creating them, uploading them to github, and to k8s all with terraform. This is completely automatic and avoids the manual process of: install flux; wait; copy out public key; upload to github).

The problem is that because the secret flux-git-deploy already exists, install of the whole chart fails. Automatic creation of this secret (which by dint of being empty causes flux to generate a keypair) should be optional.

I've made it so
"The manifest operator now takes a secretName like the Helm operator
does. If the secretName is specified (i.e. is non-default), the user is
assumed to be bringing their own secret and none is created."

Open to opinions on this: avoiding creation of the key is now inferred from the use of any non-default name. Maybe this is too magic and we want a separate "don't make the secret" flag? Is there a use case for auto-generated keys being stored in custom-named secrets?


stefanprodan:
The Helm Operator will try to mount the default secret, if you specify a different name for Flux than this should also be supplied to helm-op. 

In helm-operator-deployment.yaml there is:

```yaml
      - name: git-key
        secret:
          {{- if .Values.helmOperator.git.secretName }}
          secretName: {{ .Values.helmOperator.git.secretName }}
          {{- else }}
          secretName: {{ template "flux.fullname" . }}-git-deploy
          {{- end }}
```

Should be something like:

```yaml
      - name: git-key
        secret:
          {{- if .Values.helmOperator.git.secretName }}
          secretName: {{ .Values.helmOperator.git.secretName }}
          {{- else }}
          {{- if .Values.git.secretName }}
          secretName: {{ .Values.git.secretName }}
          {{- else }}
          secretName: {{ template "flux.fullname" . }}-git-deploy
          {{- end }}
          {{- end }}
```

stefanprodan:
closing this in favour of #1391

mt-inside:
Sorry I didn't get a chance to keep up with this. Thanks for the more comprehensive fix!

PR #1365, opened 2018-09-17
Avoid timeouts by testing with a stable image repo

squaremo:
One of the integration tests checks that the image metadata fetch
works with a real world image repository. It was using `busybox` (at
DockerHub) as that repo; but, busybox has enough manifests that the
test can time out waiting for the fetch to complete.

To keep the running time bounded, use a repo that has a small (but >
burst) number of images already, but isn't getting new tags added to
it too fast; and, tune the rate limiting down so that it won't make
too much of an impression on the registry.

### DIFF
--- a/registry/cache/memcached/integration_test.go
+++ b/registry/cache/memcached/integration_test.go
@@ -19,7 +19,7 @@ import (
 
 // memcachedIPs flag from memcached_test.go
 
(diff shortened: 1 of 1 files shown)


squaremo:
Merged by playing the admin card. This affects the integration tests, and should make PR and `master` builds more reliable.

PR #1368, opened 2018-09-17
Bump fluxd and helm-op versions, and Helm chart version

stefanprodan:
- includes Flux v1.7.0 and Flux Helm Operator v0.2.1

### DIFF
--- a/chart/flux/Chart.yaml
+++ b/chart/flux/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "1.6.0"
+appVersion: "1.7.0"
 description: Flux is a tool that automatically ensures that the state of a cluster matches what is specified in version control
 name: flux
-version: 0.3.2
+version: 0.3.3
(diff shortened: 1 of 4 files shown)


squaremo:
I'm confused about when we need to update the chart with new versions. What's the policy?

stefanprodan:
We need to update the chart every time we do a Flux/helm-op release or when we modify something in the yamls.

squaremo:
> We need to bump the chart version every time we do a Flux/helm-op release

Why? The version of both those things are parameters to the chart.

stefanprodan:
but that means that all our tutorials should include hardcoded versions. Also when someone will use the Helm chart with the default install command they will get older versions.

stefanprodan:
Here is an example from the official Helm repo https://github.com/helm/charts/commits/master/stable/drupal/Chart.yaml

squaremo:
Why have the versions as parameters then?

PR #1369, opened 2018-09-17
Reduce memcache logging by default

alanjcastonguay:
Unless debugging the image data being cached in memcached,
there is no value in recording all requests and responses between
fluxd and memcached into the logs. Reduce the logging, but
leave behind the comment in case someone needs it.


alanjcastonguay:
I deployed this manifest without thinking, and paid about $248 last month
to store (replicated) memcached logs. That's a lot of beer. I have since
modified my deployment, but I hope to save someone else from having the
same initial experience with Flux.

PR #1372, opened 2018-09-18
Include initContainers in images to fetch

### LINKED ISSUE #1371: init containers are not updated
If the init container image is not present in containers Flux will not update it.

Different images (the busybox never gets updated):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: podinfo
  namespace: default
  annotations:
spec:
  replicas: 1
  selector:
    matchLabels:
      app: podinfo
  template:
    metadata:
      labels:
        app: podinfo
    spec:
      initContainers:
      - name: init
        image: busybox:1.25
        command:
        - sleep
        - "1"
      containers:
      - name: podinfod
        image: quay.io/stefanprodan/podinfo:1.2.0
        command:
        - ./podinfo
        - --port=9898
        ports:
        - name: http
          containerPort: 9898
```

squaremo:
Symptom: images used in initContainers, but not elsewhere, don't get
metadata fetched for them, and don't get automated updates.

Fix: the ImagesToFetch method of the cluster, supplied to the image DB
as the way to get a list of all the images to fetch, didn't look in
initContainers. So, make it do that.

Fixes #1371.

### DIFF
--- a/cluster/kubernetes/images.go
+++ b/cluster/kubernetes/images.go
@@ -89,6 +89,14 @@ func mergeCredentials(log func(...interface{}) error, client extendedClient, nam
 		}
 		imageCreds[r.Name] = creds
 	}
+	for _, container := range podTemplate.Spec.InitContainers {
+		r, err := image.ParseRef(container.Image)
+		if err != nil {
+			log("err", err.Error())
+			continue
+		}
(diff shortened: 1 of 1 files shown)


squaremo:
Brill, thanks very much for the quick turnaround on review Stefan :candy:

PR #1374, opened 2018-09-19
fix wrong link to using.md from helm-get-started.md

tlvu:
They are both in the same folder.


dholbach:
Thanks for the fix @tlvu - my bad.

PR #1375, opened 2018-09-20
 Add ListServicesWithOptions in API v11

### LINKED ISSUE #1352: Support id filter for listing services

rndstr:
Adds the endpoint `ListServicesWithOptions` supporting `namespace` and
`services` as a parameter. `services` filters the result by resource
ids.

It introduces a new API v11 which builds on top of v10 but adds the
`ListServicesWithOptions` call. It is polyfilled back to API v6 by
using `ListServices` and then manually removing the unwanted services.

Closes #1352

### DIFF
--- a/http/transport.go
+++ b/http/transport.go
@@ -30,6 +30,7 @@ func NewAPIRouter() *mux.Router {
 	r := mux.NewRouter()
 
 	r.NewRoute().Name(ListServices).Methods("GET").Path("/v6/services")
+	r.NewRoute().Name(ListServicesWithOptions).Methods("GET").Path("/v11/services")
 	r.NewRoute().Name(ListImages).Methods("GET").Path("/v6/images")
 	r.NewRoute().Name(ListImagesWithOptions).Methods("GET").Path("/v10/images")
 
--- a/remote/rpc/clientV6.go
+++ b/remote/rpc/clientV6.go
@@ -6,12 +6,11 @@ import (
 	"net/rpc"
 	"net/rpc/jsonrpc"
 
-	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/api/v10"
+	"github.com/weaveworks/flux/api/v11"
 	"github.com/weaveworks/flux/api/v6"
 	fluxerr "github.com/weaveworks/flux/errors"
 	"github.com/weaveworks/flux/job"
-	"github.com/weaveworks/flux/policy"
 	"github.com/weaveworks/flux/remote"
 	"github.com/weaveworks/flux/update"
 )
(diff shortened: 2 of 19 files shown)


http/transport.go
```diff
@@ -30,6 +30,7 @@ func NewAPIRouter() *mux.Router {
 	r := mux.NewRouter()
 
 	r.NewRoute().Name(ListServices).Methods("GET").Path("/v6/services")
+	r.NewRoute().Name(ListServicesWithOptions).Methods("GET").Path("/v6/services")
```
squaremo: Should this be `/v11/services`?

remote/rpc/clientV6.go
```diff
@@ -102,6 +103,48 @@ func (p *RPCClientV6) ListServices(ctx context.Context, namespace string) ([]v6.
 	return services, err
 }
 
+func (p *RPCClientV6) ListServicesWithOptions(ctx context.Context, opts v11.ListServicesOptions) ([]v6.ControllerStatus, error) {
+	return listServicesWithOptions(ctx, p, opts, supportedKindsV6)
+}
+
+type listServicesWithOptionsClient interface {
```
squaremo: Pedantry alert! This should be `listServicesWithoutOptionsClient`, surely ..

remote/rpc/clientV6.go
```diff
@@ -102,6 +103,48 @@ func (p *RPCClientV6) ListServices(ctx context.Context, namespace string) ([]v6.
...
+	ListServices(ctx context.Context, namespace string) ([]v6.ControllerStatus, error)
+}
+
+// listServicesWithOptions polyfills the ListServiceWithOptions()
+// introduced in v11 by removing unwanted resources after fetching
+// all the services.
+func listServicesWithOptions(ctx context.Context, p listServicesWithOptionsClient, opts v11.ListServicesOptions, supportedKinds []string) ([]v6.ControllerStatus, error) {
```
squaremo: This should also check that only one of `opts.Namespace` and `opts.Services` has a value, otherwise you could in principle ask it for specific services, as well a namespace that doesn't include (all of) them.

squaremo: (You could put this cross-version stuff in compat.go)

squaremo: (Oh I see, you're following the precedent set by `listImagesWithOptions`. Carry on.)

rndstr: > Oh I see, you're following the precedent set by listImagesWithOptions. Carry on.)

I actually didn't. I mistakenly thought the `daemon.ListImagesWithOptions()` covers it but it obviously doesn't invoke it.. so will add, thanks!

http/transport.go
rndstr: d'oh, good catch!

PR #1380, opened 2018-09-20
Provide detailed pod status when listing services

lelenanam:
For deployments, daemonSet, statefulSet provide:
 - status of rollout
 - numbers of desired, updated, ready, available, outdated pods
 - slice of errors for controller

Fixes https://github.com/weaveworks/flux/issues/1303

### DIFF
--- a/api/v6/api.go
+++ b/api/v6/api.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/weaveworks/flux"
+	"github.com/weaveworks/flux/cluster"
 	"github.com/weaveworks/flux/git"
 	"github.com/weaveworks/flux/job"
 	"github.com/weaveworks/flux/ssh"
@@ -34,6 +35,7 @@ type ControllerStatus struct {
 	Containers []Container
 	ReadOnly   ReadOnlyReason
 	Status     string
+	Rollout    cluster.RolloutStatus
 	Antecedent flux.ResourceID
 	Labels     map[string]string
 	Automated  bool
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -21,6 +21,29 @@ type Cluster interface {
 	PublicSSHKey(regenerate bool) (ssh.PublicKey, error)
 }
 
+// RolloutStatus describes numbers of pods in different states and
+// the messages about unexpected rollout progress
+// a rollout status might be:
+// - in progress: Updated, Ready or Available numbers are not equal to Desired, or Outdated not equal to 0
+// - stuck: Messages contains info if deployment unavailable or exceeded its progress deadline
+// - complete: Updated, Ready and Available numbers are equal to Desired and Outdated equal to 0
+// See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status
+type RolloutStatus struct {
+	// Desired number of pods as defined in spec.
+	Desired int32
+	// Updated number of pods that are on the desired pod spec.
+	Updated int32
+	// Ready number of pods targeted by this deployment.
+	Ready int32
+	// Available number of available pods (ready for at least minReadySeconds) targeted by this deployment.
+	Available int32
+	// Outdated number of pods that are on a different pod spec.
+	Outdated int32
+	// Messages about unexpected rollout progress
+	// if there's a message here, the rollout will not make progress without intervention
+	Messages []string
+}
+
 // Controller describes a cluster resource that declares versioned images.
 type Controller struct {
 	ID     flux.ResourceID
@@ -35,6 +58,7 @@ type Controller struct {
 	// in this field.
 	Antecedent flux.ResourceID
 	Labels     map[string]string
+	Rollout    RolloutStatus
 
 	Containers ContainersOrExcuse
 }
--- a/cluster/kubernetes/resourcekinds.go
+++ b/cluster/kubernetes/resourcekinds.go
@@ -1,8 +1,6 @@
 package kubernetes
 
 import (
-	"fmt"
-
 	apiapps "k8s.io/api/apps/v1"
 	apibatch "k8s.io/api/batch/v1beta1"
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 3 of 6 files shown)


api/v6/api.go
```diff
@@ -34,6 +35,7 @@ type ControllerStatus struct {
 	Containers []Container
 	ReadOnly   ReadOnlyReason
 	Status     string
+	Pods       cluster.Pods
```
squaremo: There doesn't seem much point in compiling a list of problems with the controller if it's not communicated in the API. May I suggest that the `ControllerStatus` has a field for rollout status which includes both the counts of pods and the reason it's stuck or still in progress, if it's not completed.

cluster/kubernetes/resourcekinds.go
```diff
@@ -120,27 +122,50 @@ func (dk *deploymentKind) getPodControllers(c *Cluster, namespace string) ([]pod
 	return podControllers, nil
 }
 
+func deploymentErrors(d *apiapps.Deployment) []string {
+	var errs []string
+	for _, cond := range d.Status.Conditions {
```
squaremo: I didn't know much about conditions so I went to read the internet.

My understanding is that each entry in `.Conditions` represents a transition -- and though the Kubernetes authors seem to experience physical pain at the suggestion of this representing a state machine, at least some of the conditions represent state transitions. https://github.com/kubernetes/kubernetes/issues/7856 makes very interesting reading.

Ultimately we want to know "will this rollout succeed?" with possible answers "eventually", "definitely not", and "it already has". I don't know enough to judge whether the result compiled in this loop answers that question, but I suspect that it might end up including error messages that have been superseded. For example, it could pick up a condition saying the deployment wasn't available because of some transient situation which later went away, allowing progress.

If I'm near the mark, we need to take some more care here to interpret the conditions -- e.g., if the most recent condition says "everything's OK" we can ignore the rest. I could be wrong, and just not understanding what you've already thought through -- in which case, I'd appreciate a comment above the code explaining what it's doing.

cluster/kubernetes/resourcekinds.go
```diff
@@ -176,23 +201,30 @@ func (dk *daemonSetKind) getPodControllers(c *Cluster, namespace string) ([]podC
...
-	} else {
-		status = StatusUpdating
 	}
 
 	return podController{
 		apiVersion:  "apps/v1",
 		kind:        "DaemonSet",
 		name:        daemonSet.ObjectMeta.Name,
 		status:      status,
+		pods:        pods,
```
squaremo: DaemonSets have conditions in the same way as deployments, how come we're not looking at those?

lelenanam:
StatefulSet Status is missing UpdatedReplicas field: https://github.com/kubernetes/kubernetes/issues/52653

cluster/kubernetes/resourcekinds.go
lelenanam: > it might end up including error messages that have been superseded. For example, it could pick up a condition saying the deployment wasn't available because of some transient situation which later went away, allowing progress.

I think if error went away, appropriate condition changed as well.
For example, if deployment wasn't available (`cond.Type == apiapps.DeploymentAvailable && cond.Status == apiv1.ConditionFalse`) but now it is available, then this condition is not `False` anymore:
`cond.Type == apiapps.DeploymentAvailable && cond.Status != apiv1.ConditionFalse`

lelenanam:
>* please either explain, or revise, the code that compiles errors from the conditions (I've expended on this in the comments)

added comment

>* the errors, or their replacement, need to be represented in the API response

moved `Errors` to `RolloutStatus` struct

>* daemonsets etc., should get the same treatment as deployments.

I wasn't able to find a way to get errors from `deamonSet` and `statefulSet` conditions.

lelenanam:
@squaremo PTAL

cluster/kubernetes/resourcekinds.go
squaremo: I was quite surprised to see that conditions are totally ignored (i.e., never populated) for everything except for Deployments!

cluster/kubernetes/resourcekinds.go
```diff
@@ -176,23 +197,30 @@ func (dk *daemonSetKind) getPodControllers(c *Cluster, namespace string) ([]podC
...
+		Available: daemonSetStatus.NumberAvailable,
+		Outdated:  daemonSetStatus.CurrentNumberScheduled - daemonSetStatus.UpdatedNumberScheduled,
+	}
+
 	if daemonSetStatus.ObservedGeneration >= objectMeta.Generation {
 		// the definition has been updated; now let's see about the replicas
-		updated, wanted := daemonSetStatus.UpdatedNumberScheduled, daemonSetStatus.DesiredNumberScheduled
-		if updated == wanted {
+		status = StatusUpdating
+		if rollout.Ready == rollout.Desired && rollout.Available == rollout.Desired && rollout.Outdated == 0 {
```
rndstr: doesn't `available` imply `ready`? if so you could remove `ready == desired`

cluster/cluster.go
```diff
@@ -21,6 +21,29 @@ type Cluster interface {
 	PublicSSHKey(regenerate bool) (ssh.PublicKey, error)
 }
 
+// RolloutStatus describes numbers of pods in a different states and
+// the messages about unexpected rollout progress
+// a rollout status might be:
+// - in progress: Updated, Ready or Available numbers are not equal to Desired, or Outdated not equal to 0
+// - stuck: Messages contains info if deploynemt unavailable or exceeded its progress deadline
+// - complete: Updated, Ready and Available numbers are equal to Desired and Outdateds equal to 0
+// See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status
```
squaremo: :100: thank you

cluster/kubernetes/resourcekinds.go
```diff
@@ -120,27 +120,52 @@ func (dk *deploymentKind) getPodControllers(c *Cluster, namespace string) ([]pod
 	return podControllers, nil
 }
 
+// Deployment may get stuck trying to deploy its newest ReplicaSet without ever completing.
+// One way to detect this condition is to specify a deadline parameter in Deployment spec:
+// .spec.progressDeadlineSeconds
+// See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment
```
squaremo: Brill, thanks for this one too

PR #1382, opened 2018-09-21
Support new API version allowing charts from Helm repos, per-release git repos

squaremo:
- [x] update the CRD and types
 - [x] implement the bit where it releases a chart from a repo spec
 - [x] cache the downloaded charts
 - [x] authentication for chart repos (NB: per-operator, via mounting creds)
 - [x] don't exit or block if there's no git repo configured
 - [x] figure out what to do with `requirements.yml`

EDIT: qualification on authentication, so I can check it off


squaremo:
Some open questions:

##### As it stands, this will remove support from fluxd for the v1alpha2 FluxHelmRelease resources.

In other words, you have to do a big changeover and run both the new helm-operator and new fluxd, otherwise you won't be able to see or update `FluxHelmRelease` workloads via flux; and after doing so, the old resources will be ignored, so you have to carefully convert them at the same time.

NB the API group is different, so there's no automagic translation of the old resources into the new (and Kubernetes < 1.10 doesn't support that anyway).

An alternative would be to support both old and new resources, at least in fluxd (so you can see and update workloads of both versions; but the helm operator only releases things from the new versions). One problem with that is Kubernetes will only respond to `kubectl get fluxhelmrelease` with resources from one of the versions, which would be a bit confusing.

##### Should FHRs refer to charts using a repository URL, or a preconfigured repository name (like Helm does)?

Originally I thought all you would need is a repository URL -- and that's true, to some extent. However I'd also understand if people wanted to use, e.g., `stable` to specify the repo. So, maybe we could allow for referring to names in a repositories.yaml? (Though this introduces some problems with caching, annoyingly).

stefanprodan:
Building with make doesn't work on macOS since there is no `sha256sum` on High Sierra. For anyone trying to test this on macOS you can do the following:

```
brew install coreutils
sudo ln -s /usr/local/bin/gsha256sum /usr/local/bin/sha256sum
```

dholbach:
Is `gsha256sum` something you had to install separately as well? (I'm guessing this because of `/usr/local`.) If it is, it might be good to document that as part of the build instructions.

If it's a tool that's generally around in build environments on MacOS, we should probably special-case this.

hiddeco:
@dholbach `gsha256sum` is included in `coreutils`

---

Just want to make a note that installing `coreutils` is optional.

MacOS has a native `shasum` function, so you can easily mimic the command by adding `function sha256sum() { shasum -a 256 "$@"; } && export -f sha256sum` to your `.bashrc`.

squaremo:
> Building with make doesn't work on macOS since there is no `sha256sum` on High Sierra

Ops! I can probably adapt the Makefile to cope with this.

squaremo:
> Not working:

That's not intended to work -- the supported formats are

```
container1:
  image: foo:v1
container2:
  image: foo
  tag: v1
container3:
  image:
    repository: foo
    tag: v1
```

of which I think the the first and third are in common use, and the middle is my confabulation.

squaremo:
Alternative design for providing Helm repo credentials: refer to a secret with a `repositories.yaml` file in it. This might also let you just give e.g., "stable" as the repository name, if there's an entry for that in the file -- not sure if that's a good idea, though, since that means changing the secret could change the meaning of the FHR.

stefanprodan:
Regarding authentication there are two types of auth that we should support. For chartmuseum there is a basic auth option and for Artifactory there is the access token option.

Basic auth header (chartmuseum):

```
Authorization: Basic YWRtaW46YW....
```

Access token header (Jfrog):

```
Authorization: Bearer eyJ2ZXIikwIj....
```

The chartPullSecret could look like:

```yaml
type: Basic/Bearer
token: 
username:
password:
```

site/helm-integration.md
```diff
@@ -1,119 +1,176 @@
...
+---
+title: Using Flux with Helm
+menu_order: 90
+---
 
-Chart release information is described through Kubernetes Custom Resource (CR) manifests.
+# Using Flux with Helm
 
-Flux-Helm Integration implementation consists of two parts:
+You can release charts to your cluster via "GitOps", by combining Flux
```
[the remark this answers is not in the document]
squaremo: It's a bit circular :-/

site/helm-integration.md
```diff
@@ -1,119 +1,176 @@
...
 
-- name of the resource must be unique across all namespaces
-- namespace is where both the Custom Resource and the Chart, whose deployment state it describes, will live
-- chartgitpath ... this Chart's path within the repo
-- releasename is optional. Must be provided if there is already a Chart release in the cluster that Flux should start looking after. Otherwise a new release is created for the application/service when the Custom Resource is created. Can be provided for a brand new release - if it is not, then Flux will create a release names as $namespace-$CR_name
-- customizations section contains user customizations overriding the Chart values
+```
+spec:
+  chart:
+    gitRepository: git@github.com/weaveworks/flux
```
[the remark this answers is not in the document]
squaremo: DOH! I fell into the trap.

site/helm-integration.md
```diff
@@ -1,119 +1,176 @@
...
+## The `FluxHelmRelease` custom resource
 
-## More detail
+Each release of a chart is declared by a `FluxHelmRelease`
+resource. The schema for these resources is given in [the custom
+resource definition](../deploy-helm/flux-helm-release-crd.yaml). They
+look like this:
 
- - Kubernetes Custom Resource (CR) manifests contain all the information needed to do a Chart release. There is 1-2-1 releationship between a Helm Chart and a Custom Resource.
+```
```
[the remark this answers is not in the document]
squaremo: yaw good idea

site/helm-integration.md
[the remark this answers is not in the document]
dholbach: Granted. :-)

docker/helm.version
```diff
@@ -0,0 +1,4 @@
+# NB Helm clients will refuse to play with Tiller that is older. 2.8.2
+# is the first release that had checksums, so it's used.
+HELM_VERSION=2.8.2
```
mellena1: You might want to look into upgrading the helm client to 2.9.0 at least according to: [these artifactory docs](https://www.jfrog.com/confluence/display/RTF/Helm+Chart+Repositories#HelmChartRepositories-ConfiguringtheHelmClient). From what I've seen from the rest of this PR, this is only for testing locally on the container if you choose to? I think if you are testing, having auth working would still be desired.

I could be way off on this one though, I'm fairly new to this repo! Figured I would point it out though.

squaremo: Good point @mellena1 -- I hadn't clocked that version difference. We can specify a minimum version of Helm/Tiller of 2.9 (if it's been out for a while at least).

squaremo: Note to self: probably need to bump the Helm client package version too.

mellena1: Sounds good 😄

mellena1:
@squaremo Just out of curiosity, for this PR, are you planning on adding/have already added a way for flux to update the version number of the helm chart on GitHub. So if a new chart version is available on the repo, update the FluxHelmRelease on Github, similarly to what happens when a new Docker image is pushed?

squaremo:
> Just out of curiosity, for this PR, are you planning on adding/have already added a way for flux to update the version number of the helm chart on GitHub. So if a new chart version is available on the repo, update the FluxHelmRelease on Github, similarly to what happens when a new Docker image is pushed?

Not in this PR, but it's a desirable feature for sure. The trickiness is this: the helm operator doesn't have access to the config repo -- it just looks at what's happening in the cluster. But perhaps it's something that could be built into fluxd, since that is the Keeper of the Git.

squaremo:
Latest commits add support for mentioning a git repo in an FHR:

```yaml
# ...
spec:
  chart:
    git: git@github.com:squaremo/flux-helm-test
    ref: master
    path: charts/podinfo
  values:
  # ...
```

This assumes it will be able to clone the git repo with the SSH config and keys available in the container -- which means, by default, any key you've mounted at `/etc/fluxd/ssh/identity`. In principle you should be able to go further and mount an SSH config mentioning multiple keys, and the multiple keys. (How this can be made _easier_ is TBD).

squaremo:
Now the helm operator uses Conditions like a grown-up controller.

squaremo:
Image with all of the above: quay.io/squaremo/helm-operator:21e593f

squaremo:
> * figure out what to do with `requirements.yml`

#1450 deals with this for the helm.integrations.flux.weave.works/v1alpha2 operator (that currently in master), by having the helm command line client in the image, and calling `helm repo up; helm dep build` before each release. It's that simple because there's only one chart repo, and all charts are from git.

In the case of flux.weave.works/v1beta1 operator (i.e., that in this PR), there's a wider variety of situations. Namely,

 - charts can come from Helm repos

.. but packaged charts come with their own dependencies (`helm package` won't let you proceed unless `charts/` satisfies your requirements.yaml). So we only have to deal with the charts from git.

 - charts can come from more than one git repo

For the purpose of dependencies, I don't think we care where the chart came from, only what it refers to.

 - releases (at some point) may be supplied with different credentials

It seems like the `repositories.yaml` and index/chart cache belong with the credentials, since those decide what you can access, and all releases sharing credentials will have the same access.
Releases can share a cache, since they will copy what they need from it.

squaremo:
Rebass! This now includes the updating dependencies bit, for charts from git. I pushed a newly (locally) built image: quay.io/squaremo/helm-operator:9de22ad

site/helm-integration.md
```diff
@@ -1,119 +1,195 @@
...
+chart. This is as you would put in a `values.yaml` file, but inlined
+into the structure of the resource.
+
+### Using a chart from a Git repo instead of a Helm repo
+
+You can refer to a chart from a _git_ repo, rather than a chart repo,
+with a `chart:` section like this:
+
+```yaml
+spec:
```
stephenmoloney: I'm a bit confused with the resource context here, does this refer to a `FluxHelmRelease`, in other words:

```yaml
---
apiVersion: helm.integrations.flux.weave.works/v1alpha2
kind: FluxHelmRelease
metadata:
  name: flux
  namespace: flux
  labels:
    chart: flux
spec:
  chartGitPath: flux
  releaseName: flux
```

```yaml
---
apiVersion: helm.integrations.flux.weave.works/v1alpha2
kind: FluxHelmRelease
metadata:
  name: flux
  namespace: flux
  labels:
    chart: flux
spec:
  chart:
    git: git@github.com:weaveworks/flux
    ref: master
    path: charts/flux
  releaseName: flux
```

stephenmoloney: @squaremo 
What is the correct version . `flux.weave.works/v1beta1` or `helm.integrations.flux.weave.works/v1alpha2` ?

squaremo: For the helm operator from this branch, `flux.weave.works/v1beta1`. The difference is mainly the option of supplying a chart repo/name/version or a git URL/branch.

stephenmoloney: thanks. I'm hoping to try new master soon for this functionality `https://github.com/weaveworks/flux/pull/1450` but not entirely sure if i need what inside this open PR.

stefanprodan: @stephenmoloney you don't need anything from this PR to run the deps update

demikl:
Hi, thanks for working on this great feature. I'm wondering whether the CRD may have an inconsistancy:

```
chart:
  oneOf:
  - required: ['git', 'path']
    properties:
      gitURL:
        type: string
        format: git # not defined by OAS
      path:
        type: string
      ref:
        type: string
...
```
The `required` keys are defined as `git` and `path`, but there is no `git` in the `properties` map. Should it be `gitURL` ?

squaremo:
> The `required` keys are defined as `git` and `path`, but there is no `git` in the `properties` map. Should it be `gitURL` ?

I think you are right, that property should be just `git` (looking at the Go types in v1beta1/types.go).

squaremo:
Latest "nightly": quay.io/squaremo/helm-operator:dfb5be1

whereismyjetpack:
One thing I'm having a hard time with is migrating from FluxHelmRelease, to HelmRelease.

I have to delete the fluxhelmrelease, and that deletes the helm release. i'd like to find a way to have the helmrelease take over the fluxhelmrelease, so when i delete it it doesn't destroy the helm stuff. mainly hanging pvcs etc.. 

ideas?

💡

squaremo:
> One thing I'm having a hard time with is migrating from FluxHelmRelease, to HelmRelease.
> 
> I have to delete the fluxhelmrelease, and that deletes the helm release. i'd like to find a way to have the helmrelease take over the fluxhelmrelease, so when i delete it it doesn't destroy the helm stuff. mainly hanging pvcs etc..

Hrm yes this could be a complicated dance. As it stands, the only way to do it non-destructively would be to stop the old operator, convert all resources to the new format, then start the new operator and hope you got the conversion right.

It'd be much better to be able to do that incrementally, though, keeping both operators running until the conversion is completed. And that means there's some transfer of responsibility for a release, from the old resource to the new.

A workaround would be to effectively disable each old resource by changing it to use parameters that make it fail; e.g., a chartGitPath (gitChartPath?) that doesn't exist. I _think_ that would mean it fails before attempting `helm install`, and so will never touch the release. But this would be pretty fragile, even if it works as expected.

One way to make it work _without_ needing to change the old operator would be to make the new operator support both formats (and understand that one is a replacement for the other). Then you'd only need to run the new one, and could convert resources at your leisure. On the other hand: more engineering work! On the gripping hand: probably the right thing to do!

squaremo:
> I wonder if the ignore annotations can be levereaged in some way...

Not that I can think of -- the ignore annotation controls fluxd's syncing, and the helm operators don't care how the resources got there.

> we are moving from an alpha API to a beta one, so breakage isn't unexpected

That's true; arguably it's not worth the effort and ongoing complication to try and maintain full compatibility, so long as you can go forwards conveniently.

With that in mind, I think this is fine as a migration process:

 - stop the old operator
 - start the new operator
 - one by one, create new resources to replace the old resources, and check that the new operator is deploying them correctly
 - when you're happy to proceed with the new resources, delete the old ones.

stefanprodan:
> when you're happy to proceed with the new resources, delete the old ones.

When deleting the v1alpha1 CRD the K8s API will remove any resource of that type.

squaremo:
> > when you're happy to proceed with the new resources, delete the old ones.
> 
> When deleting the v1alpha2 CRD the K8s API will remove any resource of that type.

Ah yeah -- good shortcut!

demikl:
I'm testing with your "nightly" images: quay.io/squaremo/helm-operator:dfb5be1 and accompanying fluxd: quay.io/squaremo/flux:dfb5be1

Installing an Helm release from a HelmRelease object works fine, getting the chart from a private helm repo. That's great ! But that breaks the auto-updating feature of fluxd ; whereas I have a newer image available for one of my deployment (managed by a Deployment manifest, not an Helm release), flux won't update it automatically as soon as I have a HelmRelease object in my git repository. A manuel trigger through `fluxctl release ...` works fine.

In the logs of flux, it seems that the automatic sync-loop is interrupted by this error :

```
ts=2018-11-05T14:55:57.98102016Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2018-11-05T14:55:58.020065631Z caller=images.go:33 component=sync-loop error="checking services for new images: Unsupported kind helmrelease"
```

The HelmRelease object in my git repository has the following header :

```
kind: HelmRelease
metadata:
  annotations:
    flux.weave.works/tag.chart-image: glob:1.0.*.PREPROD
...
```

squaremo:
@demikl Hey thanks for trying the nightly out, much appreciated!

Well spotted on the automation failure. I realised today that the code as of the last nightly build doesn't properly understand the new API version (there's _one_ line missing ... argh). I've fixed that, and rebased, and so on. I'll do another nightly build er, today.

squaremo:
Nightlies: quay.io/squaremo/flux:c3d0052 and quay.io/squaremo/helm-operator:c3d0052

squaremo:
Merged with helm operator "Get out of review free" card. Now retiring that card.

squaremo:
> there's no associated updated Flux so creating HelmRelease objects will break the sync loop.

Most stuff will work with the current release of the flux daemon (including syncing resources). Image updates, through fluxctl or automation, won't work until flux 1.8.1, which will be along shortly.

Issue #1379, opened 2018-09-20
Question: Can flux with helm support refer to a helm chart that is not in the same git repo


tlvu:
In the [flux-helm-test](https://github.com/weaveworks/flux-helm-test) repo, the charts are inside the same flux git repo.

In [helm-integration.md](https://github.com/weaveworks/flux/blob/master/site/helm-integration.md) we see "chartGitPath ... path (from repo root) to a Chart subdirectory".  To me, it seems to imply the charts files have to be physically in the same repo as the flux git repo.

What if we want to use published charts.  How can we refer to a published chart?

squaremo:
> To me, it seems to imply the charts files have to be physically in the same repo as the flux git repo.

You supply the helm operator with a git URL (and branch), and the charts are given by paths within that repo. It does not have to be the same git repo as the one you give to fluxd. If you use a different repo, you'll have to supply a different deploy key secret and so on.

> if we want to use published charts. How can we refer to a published chart?

At present, the helm operator supports only releasing charts from git. We'd like to add support for releasing charts from a Helm repo: https://github.com/weaveworks/flux/issues/1131 since that will make it much more useful.

tlvu:
Replying to my own question after testing, chartGitPath can not point to a remote git url.

squaremo:
@tlvu I think #1131 and the PR #1382 constitute an answer to your question -- do you agree? If so, we can close this issue.

tlvu:
Agree that if issue #1131 is done then that satisfy our requirements.  Thanks for the follow up.

PR #1383, opened 2018-09-21
clarify chartGitPath can also be a git url to a remote git repo (fixe…

tlvu:
…s #1379)

Basically integrate the reply I get in that issue here to prevent other users
to ask the same question again.

### DIFF
--- a/site/helm-integration.md
+++ b/site/helm-integration.md
@@ -49,7 +49,10 @@ Flux-Helm Integration implementation consists of two parts:
 
  - name
  - namespace
- - chartGitPath ... path (from repo root) to a Chart subdirectory
+ - chartGitPath ... path (from repo root) to a Chart subdirectory or url to
+   another git repo containing the chart (ex:
+   https://github.com/weaveworks/flux-helm-test/tree/master/charts/mongodb)
+   Follow issue #1131 for using chart from a Helm repo.
 
 ## Optional fields
 


site/helm-integration.md
```diff
@@ -49,7 +49,10 @@ Flux-Helm Integration implementation consists of two parts:
 
  - name
  - namespace
- - chartGitPath ... path (from repo root) to a Chart subdirectory
+ - chartGitPath ... path (from repo root) to a Chart subdirectory or url to
```
squaremo: If supplying a URL to a git repo elsewhere works, it's a coincidence!

tlvu: Wow! Lighting fast feedback.  Are you saying this "feature" was not by design, it's just an accident/coincidence?

We indent to leverage this feature so we can have each service (which is each in a separate repo) provide their own helm chart in their respective repo so the charts do not have to be in the flux repo.

Should we leverage this feature or not?

squaremo: Yes, it's not intended to work that way. I imagine it works because the path ends up looking like an absolute URL, and Helm downloads it .. or something along those lines.

I would not rely on it. For one thing, it'll likely download everything from the git repo _a lot_.

tlvu: So are you saying currently we are stuck with having the all the different charts physically in the same flux repo?

This is really annoying.  We already have a few microservice that have their helm chart in their own respective repo.

Any other possible work-around to use charts from a remote git repo?

tlvu: In fact, no, pointing chartGitPath to a remote git url do not work.  Sorry, I did not test properly.

ts=2018-09-21T19:27:09.692439044Z caller=release.go:175 component=release error="Chart release failed: mongodb-database: &os.PathError{Op:\"stat\", Path:\"/tmp/flux-working999529096/charts/https:/github.com/tlvu/flux-helm-test/tree/master/charts/mongodb\", Err:0x2}"

tlvu:
Closed unmerged PR because chartGitPath can not point to a remote git url.

PR #1391, opened 2018-09-25
Allow custom ssh keys for flux and helm operators

sebastianvoss:
- Add `git.secretName` to allow custom ssh key for flux operator
- `git.secretName` and `helmOperator.git.secretName` can be set to different values to use different keys for both operators
- If only `git.secretName` is set the secret will be used by both operators
- If only `helmOperator.git.secretName` is set it will be used for helm. Flux uses the auto-generated key (this is the same behaviour as with the previous version of the chart)

Should also fix https://github.com/weaveworks/flux/issues/1386


stephenmoloney:
LGTM

squaremo:
I have lost track of which keys go where via the chart, but I know @stephenmoloney  and @stefanprodan  will know :-S

kozejonaz:
Sweet! Thanks

PR #1394, opened 2018-09-26
Git commit signing

scjudd:
This adds support for flux to sign git commits made to the central config repo. This is important for security-minded users who want to make sure that all commits to the config repo come from an authenticated source, and is the first step towards the larger goal of optionally having flux only apply changes if the latest config has a valid signature from a trusted key.


git/gittest/repo.go
```diff
@@ -3,15 +3,84 @@ package gittest
...
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		cleanup()
+		t.Fatal(err)
+	}
+
+	io.WriteString(stdin, "Key-Type: DSA\n")
+	io.WriteString(stdin, "Key-Length: 1024\n")
+	io.WriteString(stdin, "Key-Usage: sign\n")
+	io.WriteString(stdin, "Name-Real: Weave Flux\n")
```
hiddeco: We have configuration flags for `git-user` and `git-email`, why not use them while generating the GPG key?

scjudd: Excellent point! I'll get that in.

scjudd: So we _can_, but we're only generating this key for testing that commits are get signed with the expected key ID. In actual usage of this feature, the key is provided by the user, so it's expected that it will have been generated with the desired UID. Since we don't do any key generation apart from testing, there's no reason to test that the UID matches the `git-user` and `git-email`, because we're not enforcing that anywhere.

hiddeco: You are 100% right, ignore my comment.

docker/flux-entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+gpg_import_dir="$HOME/gpg-import"
+
+if [ -d "$gpg_import_dir" ]; then
+    IFS="$(printf '\n')"
+    for key in "$gpg_import_dir"/*; do
+        if [ -e "$key" ]; then
+            echo "Importing GPG key: $(basename "$key")"
```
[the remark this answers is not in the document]
scjudd: What about calling `fluxd --version` before the key import loop? It wouldn't be output in the normal log format, though.

We could add a `--log-version` flag that does the same as `--version`, but in the log format. Hacky? Maybe.

If we actually incorporated GPG key importing into fluxd itself we could do away with this entrypoint script altogether. What are your thoughts on that? Originally I felt that this was outside of the scope of fluxd's responsibilities, but I'd be glad to discuss that further.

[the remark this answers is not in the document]
squaremo: > If we actually incorporated GPG key importing into fluxd itself we could do away with this entrypoint script altogether. What are your thoughts on that? Originally I felt that this was outside of the scope of fluxd's responsibilities, but I'd be glad to discuss that further.

I'm inclined to agree with you -- if you were running fluxd in another context, you'd arrange for the gpg keyring some other way.

Another alternative (I think): expect the whole keyring to be mounted, already assembled. This means more work to set it up ahead of time, but it'd be easier to update it while fluxd was running (since it would just be replaced, rather than assembled at startup). Not sure if that's a good trade-off.

If that's no good, a middle ground would be something like `--git-import-gpg-keys=/path/to/dir` (and we incorporate the script into flux code, as you mention).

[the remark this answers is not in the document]
squaremo: NB none of this is a condition for merging this PR; it would be good to figure it out before this feature makes its way into a release, though, since that will make the UI to it harder to change.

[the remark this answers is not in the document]
scjudd: > Another alternative (I think): expect the whole keyring to be mounted, already assembled. This means more work to set it up ahead of time, but it'd be easier to update it while fluxd was running (since it would just be replaced, rather than assembled at startup). Not sure if that's a good trade-off.

I considered this, but I wasn't sure if there was a **good** way to take an entire `$GNUPGHOME` directory and upload it as a Kubernetes secret. It also seems more prone to breakage if an updated version of GPG changes the format of files on disk—but I imagine that they would have to account for people having old keyrings and provide some sort of automatic-upgrade path as well.

If there's a good story for getting a whole `$GNUPGHOME` directory in a secret, and then volume-mounting it into the container, that has a lot of advantages. OTOH, I haven't been working with Kubernetes very long, but it seems like secrets are usually just tiny pieces of secret data, not whole directory trees.

We could also consider a sidecar container responsible for importing keys that writes to a shared volume used by the flux container. The keys don't need to be in place before fluxd starts up, just before fluxd tries to perform a signed commit; and, if they aren't available yet for some reason, the commit should fail and succeed on the next attempt. I'd have to flesh this out to see if it'd work like I'm thinking, but what are your initial thoughts?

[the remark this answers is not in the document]
squaremo: > We could also consider a sidecar container responsible for importing keys that writes to a shared volume used by the flux container. The keys don't need to be in place before fluxd starts up, just before fluxd tries to perform a signed commit; and, if they aren't available yet for some reason, the commit should fail and succeed on the next attempt. I'd have to flesh this out to see if it'd work like I'm thinking, but what are your initial thoughts?

You could put them in an `initContainer`, then it has to have run before fluxd starts up. One trickiness is choosing the container image such that there's not another bulky image to download -- one possibility is to bake the script (and gnupg) into the flux container, since we know they will have that image anyway.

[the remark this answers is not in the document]
scjudd: The `initContainer` solution seems like it'd work perfectly. I will try to get that done as soon as I've got some cycles!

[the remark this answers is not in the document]
squaremo: Aces -- I think we've puzzled it out! Thanks for your patience dealing with my comments here @scjudd, I appreciate it.

hiddeco:
@scjudd, this PR has been stale for a while now and we would like to get it merged!

Are you still interested in finishing it up or do you want me to take over the remaining part? In case of the latter, can you please flip the setting which lets us push to the PR branch? Thanks in advance :medal_sports:

squaremo:
Rebased -- almost all the conflicts were trivial (unrelated changes that happened to touch adjacent lines). The exceptions:

 - in git/operations.go, there's a new argument introduced here for environment entries, while in master, there were changes to how environment entries were prepared. I've made those mesh again, in 
fe8046e.
 - I elided the changelog update, because in the meantime I decided it was better for people to just suggest the entry in their PR and for it to be collated later. The suggested entry is this:

```
- Flux is now capable of signing commits made to the user git repo
  [Git commit signing](https://github.com/weaveworks/flux/blob/master/site/git-commit-signing.md)
```

squaremo:
I'm implementing the "inline the GPG key import into fluxd" option; working in git-signing in this repo for now (i.e., it won't appear here yet)

scjudd:
Please feel free to finish this up, sorry for not being able to contribute recently! Thank you all!

Makefile
```diff
@@ -53,6 +53,7 @@ build/.%.done: docker/Dockerfile.%
 		-f build/docker/$*/Dockerfile.$* ./build/docker/$*
 	touch $@
 
+
```
2opremio:
```suggestion
```

cmd/fluxd/main.go
```diff
@@ -186,6 +192,16 @@ func main() {
 		*sshKeygenDir = *k8sSecretVolumeMountPath
 	}
 
+	// Import GPG keys, if we've been told where to look for them
+	if *gitImportGPG != "" {
+		keyfiles, err := gpg.ImportKeys(*gitImportGPG)
+		if err != nil {
+			logger.Log("error", "failed to import GPG keys", "err", err.Error())
```
2opremio: Shouldn't we refuse to start if the import doesn't succeed?

git/operations.go
```diff
@@ -250,13 +267,13 @@ func changed(ctx context.Context, path, ref string, subPaths []string) ([]string
...
 
-	if err := execGitCmd(ctx, path, out, args...); err != nil {
+	if err := execGitCmd(ctx, path, out, nil, args...); err != nil {
 		return nil, err
 	}
 	return splitList(out.String()), nil
 }
 
-func execGitCmd(ctx context.Context, dir string, out io.Writer, args ...string) error {
+func execGitCmd(ctx context.Context, dir string, out io.Writer, extraEnv []string, args ...string) error {
```
2opremio: Since the `extraEnv` parameter is only used exceptionally, I would  propose to split the function in two: `execGitCmd` (without env variables) and `execGitCmdWithEnvs`

Alternatively, since the amount of arguments is running out of hand, we could use a struct and omit the `extraEnv` field when not needed (since it would default to nil).

gpg/gpg.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		}
+	default:
+		files = []string{src}
+	}
+
+
+	var imported []string
+	for _, path := range files {
+		if err := gpgImport(path); err != nil {
```
2opremio: I don't think we want this to simply fail silently (at least not in all cases)

gpg/gpgtest/gpg.go
```diff
@@ -0,0 +1,77 @@
+package gpgtest
```
2opremio: I've been thinking for a while that we should have a package for testing machinery. Wouldn't it make sense to have a common place were to put all the stuff?

On the other hand having the test machinery closer to the code in which it's used probably makes it easier to understand.

cmd/fluxd/main.go
hiddeco: Only write operations to git will fail if the import does not succeed (but only if `gitSigningKey` is set), all other parts of Flux will continue to work.

Refusing to start because Flux _may_ not be able to do a part of its job is a bit extreme imho.

gpg/gpg.go
hiddeco: In what cases would we want to inform the user, and almost as important; how detailed do we want to present it to the user?

cmd/fluxd/main.go
2opremio: My concern is that the import failure can go undetected unless we fail to start. The user can choose to avoid the import as a response.

No super strong opinion though

gpg/gpg.go
2opremio: Things like badly formated data, for instance

git/operations.go
hiddeco: Did the latter.

cmd/fluxd/main.go
hiddeco: I think due to expectations of the user about what Flux does this will not go undetected as it halts the rollout of new images (and will throw an error in the logs every time Flux attempts to update a manifest).

gpg/gpgtest/gpg.go
squaremo: There are a few surprises, like test files way down in kubernetes/cluster/testfiles being used in several places. (Maybe the solution to that is not to use the same fixtures in several places ..)

git/operations.go
```diff
@@ -22,12 +22,21 @@ const trace = false
 // Env vars that are allowed to be inherited from the os
 var allowedEnvVars = []string{"http_proxy", "https_proxy", "no_proxy", "HOME"}
 
+type gitCmd struct {
```
squaremo: This approach (using a struct to package up arguments) works well in some situations:

 - providing optional arguments, i.e., where some of them may be omitted (and thereby get a zero value)
 - bundling several closely related arguments, all of which must be supplied, to declutter

Here it's not really either case; but, it _could_ be the first if the struct had only the things that were optional. Looking at `execGitCmd`, those are dir, env, and out. The args and context are always required, so I'd expect those to be their own arguments.

daemon/daemon.go
```diff
@@ -13,10 +13,10 @@ import (
 
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/api"
-	"github.com/weaveworks/flux/api/v10"
-	"github.com/weaveworks/flux/api/v11"
-	"github.com/weaveworks/flux/api/v6"
-	"github.com/weaveworks/flux/api/v9"
+	v10 "github.com/weaveworks/flux/api/v10"
```
squaremo: These packages should already have those names; i.e., they don't need aliases.

hiddeco: Introduced by tooling as you assumed (VSC with no Golang related config except for `go.autocompleteUnimportedPackages` and `go.testFlags` set).

I will take a look and see if I can track down the thing that thinks this change should be made and shut it down.

gpg/gpgtest/gpg.go
hiddeco: > On the other hand having the test machinery closer to the code in which it's used probably makes it easier to understand.

I agree.

> (Maybe the solution to that is not to use the same fixtures in several places ..)

This could be a solution although I see downsides to this too (my experience with almost identical fixtures in multiple places is that it results in high maintenance, new issues and/or frustrations).

---

Given the above, shall we:

a) move this to a new issue so we can fix it throughout the whole codebase
b) create our own fixtures in `gpg/gpgtest` so we have it fixed there
c) ...

squaremo:
Looks good -- I am evaluating at a second or third derivative, now -- Fons, mind having a fresh look?

2opremio:
Will do! I am off today though.

git/operations.go
```diff
@@ -51,47 +58,47 @@ func mirror(ctx context.Context, workingDir, repoURL string) (path string, err e
...
-	if err := execGitCmd(ctx,
-		workingDir, nil,
-		"commit",
-		"--no-verify", "-a", "-m", commitAction.Message,
-	); err != nil {
+	if commitAction.SigningKey != "" {
+		args = append(args, fmt.Sprintf("--gpg-sign=%s", commitAction.SigningKey))
+	}
+	if gpgHome != "" {
+		env = []string{fmt.Sprintf("GNUPGHOME=%s", gpgHome)}
```
2opremio: nit: use `+` instead of `Sprintf`

git/working.go
```diff
@@ -20,8 +20,10 @@ type Config struct {
 	NotesRef    string
 	UserName    string
 	UserEmail   string
+	SigningKey  string
 	SetAuthor   bool
 	SkipMessage string
+	GPGHomeDir  string
```
2opremio: This only seems to be used for testing and it is passed on to various commands. 

I am not sure if it can constitute an attack vector, but at least the code would be cleaner by simply defining the `GNUPGHOME` variable in the tests.

gpg/gpgtest/gpg.go
2opremio: I would go for a)

hiddeco: Created an issue for it.

git/working.go
hiddeco: This means adding it to the allowed env var list - which does result in better protection - but is cleaner.

2opremio: How about just temporarily adding it `allowedEnvVars` during the test?

hiddeco: I have been playing around with the idea (after the brief discussion we had in private) and I think `GNUPGHOME` should still be configurable from _outside_ the package.

Overwriting the var in a test in the `git` package is great but prevents packages using the machinery to sign commits as they have no control over `GNUPGHOME`.

This leaves us with two options; we either put it into the `Config` struct or we allow inheritance of the `GNUPGHOME` env variable. The latter being a lot cleaner and even helpful to our users as they now have control over where their imported keys are stored.

2opremio:
Let's go for the envelope variable then!

git/working.go
hiddeco: I just found a 3rd reason: in combination with #1156 this would work outside a cluster.

hiddeco:
@scjudd it only took us 5 months but this has finally landed in `master`. Thanks a lot for your initial work on this, much appreciated :100: :1st_place_medal:

Issue #1303, opened 2018-08-22
Provide pod status when listing services


rndstr:
The current service status when listing services is either `ready`, `updating`, or `X out of N updated`. It would be helpful to have a detailed pod status that tells a broader picture what is currently going on.

rade:
> a broader picture what is currently going on.

Like what?

squaremo:
I think the mode of calling this API will matter too -- e.g., if you want to know the status of your release,

 - you'll probably need to be able to ask for the status of a particular set of services;
 - you also need to know whether your release is actually in the process of being rolled out yet (versus fully rolled out), so you might have to give it a baseline;
 - you will want to know if there was a problem applying the change for any particular controller, since that means you won't see any rollout progress.

rndstr:
Talked this over with @lelenanam and a structure to provide more info about a workload's pod rollout progress could be
```
{
  "default:deployment/work-load": {
    …
    Status: "updating", // deprecated
    PodStatus: {
      Status: "updating",
      Desired: 10,
      Updated: 4,
      Ready: 3,
      Outdated: 7,
      Pods: {
        "work-load-123456-123": {
          …
        },
        …
      }
    }
  },
  …
}
```

lelenanam:
We can use deployment conditions, for example:
```
Type=Progressing
Status=False
Reason=ProgressDeadlineExceeded
```

squaremo:
> * you also need to know whether your release is actually in the process of being rolled out yet (versus fully rolled out), so you might have to give it a baseline;

Is this correct: if you know that your new version of a manifest has been applied, then all the pod status numbers will be in relation to that new revision, so if it says they are all ready, it must have completed rollout.

lelenanam:
I think so, yes.

squaremo:
OK so the related issues are:

> you'll probably need to be able to ask for the status of a particular set of services;

Covered in #1352.

> you will want to know if there was a problem applying the change for any particular controller, since that means you won't see any rollout progress.

Still a concern; covered in #1010.

lelenanam:
Pod rollout progress structure example:
```
type ControllerPodStatus struct {
	// Status of controller pod
	Status string
	// Desired number of pods as defined in spec.
	Desired int32
	// Updated number of pods that are on the desired pod spec.
	Updated int32
	// Ready number of pods that are on the desired pod spec and ready.
	Ready int32
	// Outdated number of pods that are on a different pod spec.
	Outdated int32
[...]

PR #1400, opened 2018-09-28
Polyfill rollout status for older daemons

rndstr:
Polyfills the newly introduced rollout status (#1303) for daemons older than
commit 38d15071.

It parses the `v6.ControllerStatus.Status` field which contains count
of updated pods during rollout and populates `v6.ControllerStatus.Rollout`.

### DIFF
--- a/remote/rpc/compat.go
+++ b/remote/rpc/compat.go
@@ -9,6 +9,7 @@ import (
 	"github.com/weaveworks/flux/api/v10"
 	"github.com/weaveworks/flux/api/v11"
 	"github.com/weaveworks/flux/api/v6"
+	"github.com/weaveworks/flux/cluster/kubernetes"
 	"github.com/weaveworks/flux/policy"
 	"github.com/weaveworks/flux/remote"
 	"github.com/weaveworks/flux/update"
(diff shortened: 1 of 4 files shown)


remote/rpc/compat.go
```diff
@@ -71,6 +72,25 @@ func contains(ss []string, s string) bool {
...
+	for i := range ss {
+		// Polyfill for daemons that list pod information in status ('X out of N updated')
+		if n, _ := fmt.Sscanf(ss[i].Status, "%d out of %d updated", &ss[i].Rollout.Updated, &ss[i].Rollout.Desired); n == 2 {
+			// Daemons on an earlier version determined the workload to be ready if updated == desired.
+			//
+			// Technically, 'updated' does *not* yet mean these pods are ready and accepting traffic. There
+			// can still be outdated pods that serve requests. The new way of determining the end of a rollout
+			// is to make sure the desired count equals to the number of 'available' pods and zero outdated
+			// pods. To make older daemons reach a "rollout is finished" state we set 'available', 'ready',
+			// and 'updated' all to the same count.
```
squaremo: Great explanation, thank you for including this :scroll: 

Does this mean older daemons will give an inaccurate measure of rollout progress, since they conflate updated and available? I am OK with this -- just checking I've understood.

rndstr: that's exactly what it means

PR #1404, opened 2018-09-28
Adhere to one dep version for the whole project

### LINKED ISSUE #1402: Adhere to one dep version for the whole project
We are currently using different versions of `dep` throughout the project and on our own machines.

- CircleCI uses the latest `master` version from `go get github.com/golang/dep/cmd/dep`
- The `Gopkg.*` files in `master` are created with `dep <= 0.4.1`
- [building.md](https://github.com/weaveworks/flux/blob/61812249f17289ab89a747dca7839d07e3147b3c/site/building.md) instructs a new contributor to install the latest official `dep` release

To prevent unforeseen issues during development, test and release builds ([and PRs](https://github.com/weaveworks/flux/pull/1334)) we should adhere to one `dep` version for the whole project.

I discussed this a bit with @squaremo and we think that **adhering to the latest official release would be the best**. Pinpointing to a specific version gives several issues like needing a specific version of `dep` just for the Flux project and keeping track of `dep` releases to see if we are not getting too far behind.

**Taking note of the above the following changes should be made:**

- [x] CircleCI should retrieve the latest official `dep` release instead of using `go get`
- [x] `Gopkg.*` files in master should be regenerated with the latest version of `dep` by running `dep ensure`
- [x] Make a note of this 'policy' in `building.md`, something like:
  > if you see a big diff after running `dep ensure` or other unexpected output, make sure you're using the latest official release of `dep`
- [x] While we are at it we maybe should also mention the current Go version we are building against

hiddeco:
Closes #1402

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
       - checkout
       - setup_remote_docker
 
-      - run: go get github.com/golang/dep/cmd/dep
+      - run: curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
       - run: dep ensure -vendor-only
       - run: make test TEST_FLAGS="-race -tags integration -timeout 60s"
       - run: make all
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,12 +2,15 @@
 
 
 [[projects]]
+  digest = "1:972f38a9c879a4920d1e3a3d3438104b6c06163bfa3e6f4064adb00468d40587"
   name = "cloud.google.com/go"
   packages = ["compute/metadata"]
+  pruneopts = ""
   revision = "c728a003b238b26cef9ab6753a5dc424b331c3ad"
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - checkout
       - setup_remote_docker
 
-      - run: go get github.com/golang/dep/cmd/dep
+      - run: curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
hiddeco: Not ~~sure if this is~~ the best approach -- although it's the simplest.

Alternative would require a lot more lines of code (we need to retrieve the latest release tag by ourselves, download this release from GitHub, set the right permissions and put it in to the right place -- basically the same as what the `install.sh` is doing for us).

squaremo:
Thanks @hiddeco!

PR #1405, opened 2018-09-28
Helm chart release v0.3.4

stefanprodan:
- includes Flux v1.7.1
- allow custom ssh keys for flux and helm-op


chart/flux/README.md
```diff
@@ -173,7 +173,7 @@ The following tables lists the configurable parameters of the Weave Flux chart a
 | Parameter                       | Description                                | Default                                                    |
 | ------------------------------- | ------------------------------------------ | ---------------------------------------------------------- |
 | `image.repository` | Image repository | `quay.io/weaveworks/flux`
-| `image.tag` | Image tag | `1.6.0`
+| `image.tag` | Image tag | `<VERSION>`
```
squaremo: Is this a variable for the reader, just so you don't have to keep having to change it?

stefanprodan: I want to avoid changing this every time, I've seen it used like this in other charts.

PR #1410, opened 2018-09-28
Return sync errors in ListServices

### LINKED ISSUE #1010: Include sync errors in ListControllers result
We now keep track of individual sync errors (failure to `kubectl apply` a resource). We could include this information in the output of `ListControllers`, so it can be shown by `fluxctl list-controllers` (for example).

rndstr:
When applying the Kubernetes YAML fails we now return this error in
`ListServices` as part of the rollout messages.

Closes #1010

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -14,9 +14,11 @@ import (
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/api"
 	"github.com/weaveworks/flux/api/v10"
+	"github.com/weaveworks/flux/api/v11"
 	"github.com/weaveworks/flux/api/v6"
 	"github.com/weaveworks/flux/api/v9"
 	"github.com/weaveworks/flux/cluster"
+	"github.com/weaveworks/flux/cluster/kubernetes"
 	"github.com/weaveworks/flux/event"
 	"github.com/weaveworks/flux/git"
 	"github.com/weaveworks/flux/guid"
@@ -27,7 +29,6 @@ import (
 	"github.com/weaveworks/flux/release"
 	"github.com/weaveworks/flux/resource"
 	"github.com/weaveworks/flux/update"
-	"github.com/weaveworks/flux/api/v11"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cluster/cluster.go
```diff
@@ -17,7 +17,7 @@ type Cluster interface {
 	SomeControllers([]flux.ResourceID) ([]Controller, error)
 	Ping() error
 	Export() ([]byte, error)
-	Sync(SyncDef) error
+	Sync(SyncDef, bool) error
```
squaremo: Alternate suggestion: accept the names of things that are expected to cause problems, i.e., whatever had problems last time.

squaremo:
What's happening with this one? Looks like there's only a couple of things holding it back:

 - the tests don't compile, because there's a fake Applier with an out of date method signature
 - it needs some thought over how sync errors are interpreted with respect to rollout progress

squaremo:
That latest commit looks good 🍊

Regarding the sync error field -- we _could_ think of (or even implement) the sync status as a Condition, in which case we'd probably treat it like we would `Available` or whatever. And then it'd end up in the messages, as it does in the above. I reckon it would still need some interpretation, at least because of the scenario I mentioned above. wdyt @rndstr ?

squaremo:
> Regarding the sync error field -- we _could_ think of (or even implement) the sync status as a Condition, in which case we'd probably treat it like we would `Available` or whatever. And then it'd end up in the messages, as it does in the above. I reckon it would still need some interpretation, at least because of the scenario I mentioned above.

Thinking this through a _little_ more, if the sync status is a condition, then it should be considered as part of the rollout status calculation. That would need control flow to change, since that calculation is all done purely from the cluster resource, at present.
A simple way to do it would be to replace the RolloutStatus field with an interface (or just a procedure) that can be invoked to get the rollout status given the sync status.

squaremo:
Alternative: remove the bit that adds the sync error to the `messages` for now, since that's the controversial bit, and the rest is useful without it.

rndstr:
I finally came to the realization that the sync errors should not be part of the rollout status. The rollout basically provides the current state of a workload. The sync error is from whenever the yaml file last changed (and has been sync'ed past) and has no relation to the current rollout status.
Maybe that's what you already meant @squaremo?

Creating a release does commit & sync, while sync will fail during an user triggered release and provide semi-instant feedback (through polling jobs).

Providing the information in `ListServices` would not necessarily be useful for determining the rollout status but just the general state of the YAML.

I removed the part where it copies it into the rollout status for now and will add this directly as sub field for the `Controller` as `SyncError`.

squaremo:
> I finally came to the realization that the sync errors should not be part of the rollout status. The rollout basically provides the current state of a workload. The sync error is from whenever the yaml file last changed (and has been sync'ed past) and has no relation to the current rollout status.
> Maybe that's what you already meant @squaremo?

Not quite, but this is at least as good a line of reasoning :-)

> I removed the part where it copies it into the rollout status for now and will add this directly as sub field for the `Controller` as `SyncError`.

I like the moving it fully into the cluster implementation -- nice work there.

Issue #1409, opened 2018-09-28
Specify Tag Filter Policies as Code


pcfens:
Is there a way to specify image tag filter policies in the control repository? I've only been able to find how to completely [ignore a deployment ](https://github.com/weaveworks/flux/blob/master/site/faq.md#can-i-temporarily-make-flux-ignore-a-deployment) through an annotation. If it is, then I'm happy to try and put a PR together to update some documentation after I figure out how to use it.

If it's not already a feature, the ability specify the equivalent of `fluxctl policy --controller=default:deployment/helloworld --tag-all='prod-*'` in the git repository that creates `deployment/helloworld` would be a welcome addition to an already great tool.

stefanprodan:
Here is an example with regex, sem ver and glob tag filters:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-dev
  annotations:
    flux.weave.works/automated: "true"
    flux.weave.works/tag.init: regex:^1.*
    flux.weave.works/tag.main: semver:~1.2
    flux.weave.works/tag.sidecar: glob:dev-*
spec:
  template:
    spec:
      initContainers:
      - name: init
        image: busybox:1.25
      containers:
      - name: main
        image: quay.io/stefanprodan/podinfo:1.2.0
      - name: sidecar
        image: quay.io/stefanprodan/podinfo:dev-kb9lm91e
```

pcfens:
Thanks for the help - I've tried to capture this in #1412, though I'm sure it'll need a little editing before it's ready to merge.

PR #1412, opened 2018-10-01
Add automation annotation documentation

pcfens:
The documentation is based on the example provided in #1409.

Because I'm not familiar with everything, I'd appreciate a close look to ensure that everything is accurate (especially the regex vs regexp filter type syntax since I'm not sure about the p in the annotation).


stefanprodan:
Yes it's [regexp](https://github.com/weaveworks/flux/blob/master/policy/pattern.go#L14) I had a typo in my example

site/using.md
```diff
@@ -510,3 +513,14 @@ for b) and c):
...
+
+# Using Annotations
+
+Automation and image tag filtering can also be managed using annotations.
+
+Automation can be enabled with `flux.weave.works/automated: "true"`. Image
+filtering annotations take the form
+`flux.weave.works/tag.container-name: filter-type:filter-value`. Values of
+`filter-type` can be [`glob`](#glob), [`semver`](#semver), and
+[`regex`](#regexp). Filter values use the same syntax as when the filter is
```
squaremo: ^^ `regexp` ^^

rade:
Would it be worth mentioning that, under the covers, the fluxctl commands manipulate these annotations? Rather then operating via an independent, parallel mechanism.

pcfens:
I fixed the typo, and tried to at least reference that fluxctl is using annotations behind the scenes. Thanks for all the help - makes getting started with flux that much easier.

PR #1416, opened 2018-10-01
Add --git-timeout configuration flags

### LINKED ISSUE #1414: Create git timeout config flag
Hello,

We have a quite big private repository, so flux and helm operator fails to clone it due timeout.
As of now I only have helm-operator log.

```
W1001 16:13:51.230107       9 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
ts=2018-10-01T16:13:51.238548609Z caller=helm.go:81 component=helm info="Helm client set up"
ts=2018-10-01T16:13:51.23860388Z caller=main.go:196 component=helm-operator info="Attempting to clone repo ..." url=ssh://git@XXXX.git
ts=2018-10-01T16:14:19.671690154Z caller=main.go:201 component=helm-operator error="git clone --mirror: running git command: git [clone --mirror ssh://git@XXXX.git /tmp/flux-gitclone634843756]: context deadline exceeded"
```
After crash
```
W1001 16:24:54.303852       8 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
ts=2018-10-01T16:24:54.318078555Z caller=helm.go:81 component=helm info="Helm client set up"
ts=2018-10-01T16:24:54.318143302Z caller=main.go:196 component=helm-operator info="Attempting to clone repo ..." url=ssh://git@XXXX.git
ts=2018-10-01T16:25:19.593420485Z caller=main.go:201 component=helm-operator error="git clone --mirror: running git command: git [clone --mirror ssh://git@XXXX.git /tmp/flux-gitclone902263609]: context deadline exceeded"
```
And it stays in that loop.

This is the pod status:
```
flux-helm-operator-84f7849886-b4zwb  0/1    CrashLoopBackOff  10        33m
```

We created a new repository for our flux release, and now it's working, but operator seemed to be ok with the big repository until we deployed it on a different region, so we started to get the error on operator as well.

Either way, would be nice if we could change this timeout!

Great project!

Installed flux using helm.
Flux version: 1.7
Flux helm operator version: 0.2.1

hiddeco:
Closes #1414 

This PR adds a ~`--git-time-out`~ `--git-timeout` flag to both fluxd and the helm operator.

I was unsure what to do with the `gitOpTimeout` constant in [`daemon/loop.go:25`](https://github.com/weaveworks/flux/blob/8f09ae615bc9e12e34567d5bd6e1505998c1f608/daemon/loop.go#L25). If someone could give me guidance in this I am more than happy to make additional changes.

**Resulting builds from changes in this PR can be pulled and tested from:**
[hiddeco/flux:1414-git-timeout-9dc6b2181744](https://hub.docker.com/r/hiddeco/flux/tags/)
[hiddeco/helm-operator:1414-git-timeout-9dc6b2181744](https://hub.docker.com/r/hiddeco/helm-operator/tags/)

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -190,7 +190,7 @@ The following tables lists the configurable parameters of the Weave Flux chart a
 | `git.label` | Label to keep track of sync progress, used to tag the Git branch | `flux-sync`
 | `git.ciSkip` | Append "[ci skip]" to commit messages so that CI will skip builds | `false`
 | `git.pollInterval` | Period at which to poll git repo for new commits | `5m`
-| `git.timeOut` | Duration after which git operations time out | `20s`
+| `git.timeout` | Duration after which git operations time out | `20s`
 | `git.secretName` | Kubernetes secret with the SSH private key | None
 | `ssh.known_hosts`  | The contents of an SSH `known_hosts` file, if you need to supply host key(s) | None
 | `registry.cacheExpiry` | Duration to keep cached image info in memcached | `1h`
@@ -209,7 +209,7 @@ The following tables lists the configurable parameters of the Weave Flux chart a
 | `helmOperator.git.branch` | Branch of git repo to use for Helm charts | `master`
 | `helmOperator.git.chartsPath` | Path within git repo to locate Helm charts (relative path) | `charts`
 | `helmOperator.git.pollInterval` | Period at which to poll git repo for new commits | `git.pollInterval`
-| `helmOperator.git.timeOut` | Duration after which git operations time out | `git.timeOut`
+| `helmOperator.git.timeout` | Duration after which git operations time out | `git.timeout`
 | `helmOperator.git.secretName` | Kubernetes secret with the SSH private key | None
 | `helmOperator.logReleaseDiffs` | Helm operator should log the diff when a chart release diverges (possibly insecure) | `false`
 | `helmOperator.tillerNamespace` | Namespace in which the Tiller server can be found | `kube-system`
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -190,6 +190,7 @@ The following tables lists the configurable parameters of the Weave Flux chart a
 | `git.label` | Label to keep track of sync progress, used to tag the Git branch | `flux-sync`
 | `git.ciSkip` | Append "[ci skip]" to commit messages so that CI will skip builds | `false`
 | `git.pollInterval` | Period at which to poll git repo for new commits | `5m`
+| `git.timeOut` | Duration after which git operations time out | `20s`
 | `git.secretName` | Kubernetes secret with the SSH private key | None
 | `ssh.known_hosts`  | The contents of an SSH `known_hosts` file, if you need to supply host key(s) | None
 | `registry.cacheExpiry` | Duration to keep cached image info in memcached | `1h`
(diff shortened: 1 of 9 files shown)


sfrique:
I don't know who takes care of the helm chart, but it would be nice to add the option there as well.
the current values -> https://github.com/weaveworks/flux/blob/master/chart/flux/values.yaml has quite a few options.

I tested and it fixed the problem I was having, thanks a lot.

cmd/helm-operator/main.go
```diff
@@ -189,12 +191,12 @@ func main() {
...
 
 	gitRemote := git.Remote{URL: *gitURL}
-	repo := git.NewRepo(gitRemote, git.PollInterval(*gitPollInterval), git.ReadOnly)
+	repo := git.NewRepo(gitRemote, git.PollInterval(*gitPollInterval), git.TimeOut(*gitTimeOut), git.ReadOnly)
 
 	// 		Chart releases sync due to Custom Resources changes -------------------------------
 	{
 		mainLogger.Log("info", "Attempting to clone repo ...", "url", gitRemote.URL)
-		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+		ctx, cancel := context.WithCancel(context.Background())
```
justinbarrick: Why did we drop the timeout here?

hiddeco: Because every git operation (step) has its own time-out set. If we leave the time-out here it could cut off early instead of respecting the user’s configuration flag.

cmd/fluxd/main.go
```diff
@@ -93,6 +93,7 @@ func main() {
 		gitSkipMessage = fs.String("git-ci-skip-message", "", "additional text for commit messages, useful for skipping builds in CI. Use this to supply specific text, or set --git-ci-skip")
 
 		gitPollInterval = fs.Duration("git-poll-interval", 5*time.Minute, "period at which to poll git repo for new commits")
+		gitTimeOut      = fs.Duration("git-time-out", 20*time.Second, "duration after which git operations time out")
```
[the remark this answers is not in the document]
hiddeco: > I have reservations over using the same timeout for all git operations -- increasing the timeout for git clone might mean that other operations take much longer to fail.

You are not alone with this, I had/have the same concerns. This can be solved eventually by adding a second flag that defines a budget for all operations together (each operation may take `x` but all together it may not take longer than `y`).

---

As for the spelling - will make adjustments later today.

chart/flux/README.md
```diff
@@ -190,7 +190,7 @@ The following tables lists the configurable parameters of the Weave Flux chart a
 | `git.label` | Label to keep track of sync progress, used to tag the Git branch | `flux-sync`
 | `git.ciSkip` | Append "[ci skip]" to commit messages so that CI will skip builds | `false`
 | `git.pollInterval` | Period at which to poll git repo for new commits | `5m`
-| `git.timeOut` | Duration after which git operations time out | `20s`
+| `git.timeout` | Duration after which git operations timeout | `20s`
```
squaremo: Now I feel like a pedant, but the verb is "to time out"

hiddeco: ~Never drop out of high school to pursue your dreams as an engineer kids, it will bite you later~

Changed like it never happened

latchmihay:
the `gitOpTimeout` constant in [`daemon/loop.go:25`](https://github.com/weaveworks/flux/blob/8f09ae615bc9e12e34567d5bd6e1505998c1f608/daemon/loop.go#L25) which is currently 15s gets hit on slower/remote networks, and it waits for entire loop cycle until its retried.

Is it possible for it to use`--git-timeout` argument for `gitOpTimeout` constant. I understand the concerns that it will slow everything down to timeout, but it should be fine to whoever is changing it. 

Having the `--git-timeout` default of 20s to `gitOpTimeout` constant shouldn't be that big of a deal since its additional 5s

hiddeco:
@latchmihay your reasoning makes sense to me and the additional `5s` probably does not hurt anyone.

I will turn it into a PR tomorrow and see what others think.

Issue #1422, opened 2018-10-03
Flux applies changes to the cluster very slowly


brantb:
Recently we've begun to see a significant increase in the amount of time flux takes to sync the git repository to the cluster:

```
ts=2018-10-03T19:09:42.295546122Z caller=sync.go:121 component=cluster method=Sync cmd=apply args= count=119
ts=2018-10-03T19:14:31.162556177Z caller=sync.go:164 component=cluster method=Sync cmd="kubectl apply -f -" took=4m48.866802454s err=null output="namespace \"cluster-admin\" configured\n <snip>
```

Based on the flux logs, it looks like individual `kubectl apply` calls are taking 6-11 seconds to return:

```
ts=2018-10-02T21:18:17.751129168Z  caller=sync.go:164 component=cluster method=Sync cmd="kubectl apply -f  -" took=4.865805366s err=null output="clusterissuer  \\"redacted\\" configured"
ts=2018-10-02T21:18:24.480141851Z  caller=sync.go:164 component=cluster method=Sync cmd="kubectl apply -f  -" took=6.728899883s err=null output="sealedsecret  \\"redacted\\" configured"
ts=2018-10-02T21:18:28.58635465Z  caller=sync.go:164 component=cluster method=Sync cmd="kubectl apply -f  -" took=4.106128199s err=null output="fluxhelmrelease  \\"redacted\\" configured"
ts=2018-10-02T21:18:33.328764968Z  caller=sync.go:164 component=cluster method=Sync cmd="kubectl apply -f  -" took=4.742312018s err=null output="horizontalpodautoscaler  \\"redacted\\" unchanged"
```

In contrast, syncing the entire git repo from my local machine takes under 5 seconds.

I launched a copy of `k8s.gcr.io/hyperkube:v1.9.0` in the cluster, ran `kubectl apply -Rf` on a clone of our git repository, and it completed quickly there as well. I also tried moving the flux pod to a different node on the cluster to rule out a transient network issue, but there was no effect. It seems only flux is affected.

Let me know if there's any other information I can provide to help troubleshoot this, because I'm stumped.

k8s version: 1.11.2
flux: 1.7.1

([initial discussion](https://weave-community.slack.com/archives/C4U5ATZ9S/p1538583413000100))

justinbarrick:
Do individual applies take 4-6 seconds if you try that locally?

```
time kubectl apply -f file.yaml
```

As I understand it, flux is deploying files or directories one at a time and this might be inefficient.

[...]

PR #1435, opened 2018-10-08
Helm chart: change kubectl config mount path

stefanprodan:
- mount kubectl config to `/root/.kubectl`
- set config path using `KUBECONFIG` env var
- enable kubectl caching by mounting the config to a different dir than `$HOME/.kube/`

Fix: #1422

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -51,7 +51,7 @@ spec:
             protocol: TCP
           volumeMounts:
           - name: kubedir
-            mountPath: /root/.kube
+            mountPath: /root/.kubectl
           - name: sshdir
             mountPath: /root/.ssh
             readOnly: true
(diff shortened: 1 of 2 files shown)


squaremo:
> * mount kubectl config to `/root/.kubectl`

There's a `.kubectl`?? You wouldn't know from the documentation.

squaremo:
> > * mount kubectl config to `/root/.kubectl`
> 
> There's a `.kubectl`?? You wouldn't know from the documentation.

Oh I see, it's an invention for the purpose of moving the config out of the way. As you were.

hiddeco:
Should we also mention this somewhere in the docs(?) (if validated as a fix) for people not using the Helm chart?

stefanprodan:
@squaremo that path is a non-standard one.

I've tested the change by setting the default namespace to `demo` in the values.yaml kube config and I've created a deployment file with no namespace specified. Flux deployed that in the `demo` namespaces instead of `default`, this proves that the kubectl uses the `KUBECONFIG` instead of the `~/.kube/config`

stefanprodan:
Yes I'm aware of the `--git-timeout` change, both Flux and helm-op will need a release before publishing the Helm chart.

PR #1441, opened 2018-10-10
Docs: added note about ignore annotation to fluxctl docs

hiddeco:
User on Slack:

> jonaz [3:33 PM]
Just to confirm, the only way for the Flux Helm Operator to avoid setting up a `FluxHelmRelease` is to use the following annotation
```flux.weave.works/ignore: "true"``` 
It is not enough to avoid adding any annotation?
Never mind, I found https://github.com/weaveworks/flux/blob/master/site/faq.md#can-i-temporarily-make-flux-ignore-a-deployment (edited)
I was looking in https://github.com/weaveworks/flux/blob/master/site/using.md#using-annotations

This adds a note to the annotations section to guide future readers.

### DIFF
--- a/site/using.md
+++ b/site/using.md
@@ -525,3 +525,6 @@ filtering annotations take the form
 `filter-type` can be [`glob`](#glob), [`semver`](#semver), and
 [`regexp`](#regexp). Filter values use the same syntax as when the filter is
 configured using fluxctl.
+
+Annotations can also be used to tell Flux to temporary ignore certain manifests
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: site/using.md [hunk 1]
```diff
@@ -525,3 +525,6 @@ filtering annotations take the form
 `filter-type` can be [`glob`](#glob), [`semver`](#semver), and
 [`regexp`](#regexp). Filter values use the same syntax as when the filter is
 configured using fluxctl.
+
+Annotations can also be used to tell Flux to temporary ignore certain manifests
```
dholbach: temporarily

PR #1442, opened 2018-10-10
Delete resources no longer in git

Timer:
This pull request adds stack<sup>[1]</sup> tracking. Stack tracking attaches metadata to every object created by Flux so resources removed from the "Config Repo" can be pruned.

Every object or updated created by Flux gets assigned a label with the stack's name, e.g. (`flux.weave.works/stack: default`). Additionally, it also gets a checksum which is an aggregate of all resources in that stack: `flux.weave.works/stack-checksum: cf23df2207d99a74fbe169e3eba035e633b65d94`.

After cluster changes have been applied, a full list of cluster resources are retrieved by the stack label. Any resources who's `stack-checksum` do not match the most recently applied checksum are pruned.

### Current Test Image

~~You can test this PR by using the following Docker image: `timer/flux-with-stacks:feature-stack-tracking-1d481d0`.~~
Outdated.

## TODOs

- [ ] Put behind a formal flag, e.g. `--experimental-deletion`
- [ ] `cluster.Cluster`'s `Sync` method should probably receive a more appropriate object than 
2x `map[string]policy.Update` 
- [ ] `cluster.applyMetadata` should be refactored and simplified
- [ ] `cluster.ExportByLabel` should probably be given a few failure/retry cases to handle; what about when we're limited by RBAC?
- [ ] Skipped resources (`v1:ComponentStatus` and `v1:Endpoints`) shouldn't happen at listing level, but probably delete/diff level
- [ ] Should `sync.garbageCollect` mark the object for deletion and only prune it next run? This could help with any race conditions in the K8s API server (I didn't run into any, but you never know).
- [ ] Logger is broken in `sync.Sync` so `fmt.Printf` had to be used. We should figure this out.
- [ ] Tests, tests, tests.
- [ ] `flux.weave.works/immortal`

## Post-merge follow-ups
- [ ] Multi stack support to minimize blast radius
- [ ] Delete K8s resources in `Background` propagation mode

---
<small>
[1]: an arbitrary collection of k8s resources
</small>


hiddeco:
@justinbarrick see https://github.com/weaveworks/flux/issues/738#issuecomment-428590274 and https://github.com/weaveworks/flux/issues/738#issuecomment-428602939

justinbarrick:
What is the purpose of the checksum? It seems to me that if we just added a label to the resources (any label really would work), you could still sweep for resources that have the label but do not match any defined resource in the repository based on type / namespace / name. AFAIK this is how Helm works.

I can see checksums being useful if we just want to know if we need to update a resource, but I would expect just using  kubectl gives something similar to us already.

sync/sync.go
```diff
@@ -42,22 +99,59 @@ func Sync(m cluster.Manifests, repoResources map[string]resource.Resource, clus
...
 
-	return clus.Sync(sync)
+	if err := clus.Sync(sync, resourceLabels, resourcePolicyUpdates); err != nil {
+		return err
+	}
+	if tracks {
+		orphanedResources := make([]string, 0)
+
+		fmt.Printf("[stack-tracking] scanning stack (%s) for orphaned resources\n", stackName)
+		clusterResourceBytes, err := clus.ExportByLabel(fmt.Sprintf("%s%s", kresource.PolicyPrefix, "stack"), stackName)
```
squaremo: Aha, here it is. This is an abstraction leak: you have to know that cluster/kubernetes uses labels for marking stacks, to be able to export the right things. What's really needed here is to export everything* that has been given a stack, rather than things with a specific label.

*Why not things in a _particular_ stack? Because stacks may come and go -- for example, if they were based on top-level directories, and someone deleted a whole top-level directory.

(Side note: the use of Export in here is a misleading convenience -- what we really want is to be able to get all the things in the cluster; it just happens that Export will do that for us. But let's not try to fix that here.)

cluster/cluster.go
```diff
@@ -17,7 +18,8 @@ type Cluster interface {
 	SomeControllers([]flux.ResourceID) ([]Controller, error)
 	Ping() error
 	Export() ([]byte, error)
-	Sync(SyncDef) error
+	ExportByLabel(string, string) ([]byte, error)
+	Sync(SyncDef, map[string]policy.Update, map[string]policy.Update) error
```
squaremo: I would prefer it if the extra metadata were conveyed in the SyncDef's themselves, rather than in parallel in these extra args. This seems a somewhat indirect way of carrying that information.

squaremo:
While working on those commits, I did start wondering why stacks exist. I accept the argument for limiting the "blast radius" of changes (so e.g., every time you change a label somewhere, it doesn't result in every single resource being updated). But why not limit the blast radius to a single resource?

Timer:
A hash per-resource would probably work well.

Timer:
Anything I can help with to push this along? :-)

I think I have some free time coming up.

Timer:
Upgraded to this new version in my cluster and it's working great!

squaremo:
> Anything I can help with to push this along? :-)

I've added some test scaffolding in `sync_test.go`. This exercises pretty much the bare minimum of the sync (and deletion). What would _really_ be useful is to expand the cases covered, especially the corners. For example, I have no tests checking what happens with ignored resources (in fact I don't think they are accounted for in the code, so we'll need to come up with the semantics as well).

squaremo:
Idea: use a sanitised git repo URL as the stack name. That means that when flux supports multiple repos, it will be able to do syncs piece-wise, a repo at a time, and not delete something that simply came from another repo.

Technically, we can partition and name the stacks whatever we like. But it might be nice to have a label saying which repo a resource came from, for other purposes. And, it's useful for the stack to be a less than or equal to the unit of syncing; otherwise, there might be glitches where, say, a resource gets deleted because it has changed stack but that hasn't been applied yet. (Syncing per repo will have this issue, if a resource moves from one repo to another --  but since repos are updated independently, so will syncing all repos at once).

Timer:
Feel free to force push anything you'd like. :-)

squaremo:
I think this is ready for release as an experimental feature (i.e., can be reviewed and merged).

To address the TODOs:

￼ - [x] Put behind a formal flag, e.g. --experimental-deletion
￼ - [x] cluster.Cluster's Sync method should probably receive a more appropriate object than 2x map[string]policy.Update
￼ - [x] cluster.applyMetadata should be refactored and simplified
￼ - [ ] Should sync.garbageCollect mark the object for deletion and only prune it next run? This could help with any race conditions in the K8s API server (I didn't run into any, but you never know).
￼ - [x] Logger is broken in sync.Sync so fmt.Printf had to be used. We should figure this out.
￼ - [x] Tests, tests, tests.
￼ - [x] flux.weave.works/immortal

The annotation "flux.weave.works/ignore" will effectively make something immortal -- flux won't apply it, and (if on a resource in the cluster) won't delete it, even if it's removed from the repo.
These are explained in comments above (or commit comments).

￼ - [ ] cluster.ExportByLabel should probably be given a few failure/retry cases to handle; what about when we're limited by RBAC?

Accounting for the possibilities is a bit mind-boggling. I think it might be easier to release this as an experimental feature, and try it in the real environments (by which I mean disposable environments).

￼ - [ ] Skipped resources (v1:ComponentStatus and v1:Endpoints) shouldn't happen at listing level, but probably delete/diff level

I'm not convinced we have to skip any resources at all, for correctness. We can generalise or embellish this later if necessary, though.

squaremo:
@rndstr Do you mind having a look at this? I won't hold you to a definitive approval, unless you want to give one :-)

rndstr:
@squaremo i likely won't be able to give it a proper look before next Tuesday

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,299 @@ package kubernetes
...
+		mixin["labels"] = mixinLabels
+	}
+
+	if checksum != "" {
+		mixinAnnotations := map[string]string{}
+		mixinAnnotations[checksumAnnotation] = checksum
+		mixin["annotations"] = mixinAnnotations
+	}
+
+	mergo.Merge(&definition, map[interface{}]interface{}{
```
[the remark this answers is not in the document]
squaremo: Yes! Exactly. Well, ideally :-)

marcincuber:
This is absolutely gold PR that is basically the only missing functionality of flux. I would like to test it on one of my clusters asap. So, would it be possible to create a pre-release image and push to quay.io/weaveworks/flux? With this in place functionality I believe we can think of a v2.0.0-alpha-prerelease tag. 
👍

squaremo:
> So, would it be possible to create a pre-release image and push to quay.io/weaveworks/flux

I've built and pushed `quay.io/squaremo/flux:feature-stack-tracking-b59cef8`, which is up-to-date as of the commit "Account for namespace defaulting" above. Modulo code review, I think this works in the right way, or good enough for now.

cluster/kubernetes/fakedynamicclient_test.go
```diff
@@ -0,0 +1,387 @@
...
+/*
+This file was obtained from
+https://github.com/kubernetes/client-go/blob/4b43750b963d2b6e0f7527fe558e71c47bfc5045/dynamic/fake/simple.go
+and modified in the following way(s):
+
+ - the package was changed to `kubernetes`
+
+This file is here because it has a fix for
+https://github.com/kubernetes/client-go/issues/465, which is included
+in client-go v9, which we're not able to vendor at this time.
```
2opremio: What prevents us from vendoring it now (in a separate PR)? (in fact, we should probably bump it to version 10)

2opremio: Also, doesn't this belong to a separate package (e.g. `testutils`) we invoke form the tests? (I have been seeing a need for the same at #1668 )

squaremo: My thought process, no doubt: "ugh this is too much to explain right now, I'm sure I'll be able to figure out again what it means later"

squaremo: Most likely, depending on v9 or v10 either
 1. breaks compatibility with older API servers, or
 2. leads to an unresolvable clash with some other dependency

cluster/kubernetes/resource/load.go
```diff
@@ -14,8 +14,11 @@ import (
 
 // Load takes paths to directories or files, and creates an object set
 // based on the file(s) therein. Resources are named according to the
-// file content, rather than the file name of directory structure.
-func Load(base string, paths []string) (map[string]resource.Resource, error) {
+// file content, rather than the file name of directory structure. The
+// `fallbackNamespace` is assigned to any resource that doesn't
+// specify a namespace; it's only used for identification, and not put
+// in the definition.
+func Load(base, fallbackNamespace string, paths []string) (map[string]resource.Resource, error) {
```
2opremio: In order to consolidate this with #1668 I was thinking we could create something like, 

```
type ParsingContext interface {
  DefaultNamespace() string
  // in #1692 
  // ResourceScope (groupVersion string, string kind) int
}
```

cluster/kubernetes/sync.go
```diff
@@ -2,36 +2,327 @@ package kubernetes
...
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.muSyncErrors.RLock()
+	if applyErrs := c.applier.apply(logger, cs, c.syncErrors); len(applyErrs) > 0 {
+		errs = append(errs, applyErrs...)
+	}
+	c.muSyncErrors.RUnlock()
+
+	if c.GC {
```
2opremio: Sync's size is getting out of hand I would factor this out into a different function (`GarbageCollect()` or so)

cluster/kubernetes/sync.go
```diff
@@ -2,36 +2,327 @@ package kubernetes
...
+				logger.Log("info", "not applying resource; ignore annotation in file", "resource", res.ResourceID(), "source", res.Source())
+				continue
+			}
+			if cres, ok := clusterResources[id]; ok && cres.Policy().Has(policy.Ignore) {
+				logger.Log("info", "not applying resource; ignore annotation in cluster resource", "resource", cres.ResourceID())
+				continue
+			}
+			resBytes, err := applyMetadata(res, stack.Name, checkHex)
+			if err == nil {
+				cs.stage("apply", res.ResourceID(), res.Source(), resBytes)
```
2opremio: strings instead of symbols? :S

2opremio: Unrelated to this PR, but ... is there a good reason to use strings directly instead of constants?

cluster/kubernetes/sync.go
```diff
@@ -2,36 +2,327 @@ package kubernetes
...
-func (c *changeSet) stage(cmd string, o *apiObject) {
-	c.objs[cmd] = append(c.objs[cmd], o)
+func (c *changeSet) stage(cmd string, id flux.ResourceID, source string, bytes []byte) {
+	c.objs[cmd] = append(c.objs[cmd], applyObject{id, source, bytes})
 }
 
 // Applier is something that will apply a changeset to the cluster.
 type Applier interface {
 	apply(log.Logger, changeSet, map[flux.ResourceID]error) cluster.SyncError
+	getDefaultNamespace() string
```
2opremio: As I mentioned above, I think this belongs to the part of the code parsing resources (AFAICT, the fallback namespace is only passed to parsing functions)

cluster/kubernetes/sync.go
```diff
@@ -72,22 +363,45 @@ func (c *Kubectl) connectArgs() []string {
...
+// getDefaultNamespace returns the fallback namespace used by the
+// applied when a namespaced resource doesn't have one specified. This
+// is used when syncing to anticipate the identity of a resource in
+// the cluster given the manifest from a file (which may be missing
+// the namespace).
+func (k *Kubectl) getDefaultNamespace() string {
+	cmd := k.kubectlCommand("config", "get-contexts", "--no-headers")
+	out, err := cmd.Output()
+	if err != nil {
+		return DefaultDefaultNamespace
```
2opremio: I think we should let this function fail since
* `kubectl` is essential to `fluxd`
*  applying changes in `default` if there is a configured default namespace seems like a bad choice

cluster/kubernetes/sync.go
```diff
@@ -72,22 +363,45 @@ func (c *Kubectl) connectArgs() []string {
 	return args
 }
 
+// getDefaultNamespace returns the fallback namespace used by the
+// applied when a namespaced resource doesn't have one specified. This
+// is used when syncing to anticipate the identity of a resource in
+// the cluster given the manifest from a file (which may be missing
+// the namespace).
+func (k *Kubectl) getDefaultNamespace() string {
```
2opremio: I think we should test that this function works with the output of the `kubectl` binary we ship with `fluxd` (in case it changes).

Maybe there are integration to test this already, but, from https://github.com/weaveworks/flux/pull/1101 it seems that's not the case.

cluster/kubernetes/sync.go
```diff
@@ -2,36 +2,327 @@ package kubernetes
...
+	return r.obj.GetLabels()[stackLabel]
+}
+
+func (c *Cluster) getResourcesBySelector(selector, assumedNamespace string) (map[string]*kuberesource, error) {
+	listOptions := meta_v1.ListOptions{}
+	if selector != "" {
+		listOptions.LabelSelector = selector
+	}
+
+	resources, err := c.client.coreClient.Discovery().ServerResources()
```
2opremio: AFAIU this is not cached, which is pretty expensive (~50 requests to the API server)

cluster/kubernetes/sync.go
2opremio: Also, doing this at the end assumes that deletions will always happen after applications. I wonder if that invariant is good enough for all cases.

cluster/kubernetes/sync.go
2opremio: We should try to find an approach which lets use a cached client (or minimize requests) also satisfying the needs of #1668

cluster/kubernetes/resource/load.go
squaremo: That was my thought on combining these as well. If knowing the default is not necessary for #1668, this can be adapted to add it later.

cluster/kubernetes/sync.go
squaremo: It's deliberate -- the GC needs information collected during the application (specifically, which resources were present). There are other designs, but this one is nice and simple: apply everything there, then garbage collect everything you have applied previously but not this time.

cluster/kubernetes/sync.go
squaremo: .. or even different methods -- I think there's a bit of faux generality in here.

cluster/sync.go
```diff
@@ -3,34 +3,41 @@ package cluster
...
-	Delete resource.Resource // ) one of these
-	Apply  resource.Resource // )
+// SyncStack groups a set of resources to be updated. The purpose of
+// the grouping is to limit the "blast radius" of changes. For
+// example, if we calculate a checksum for each stack and annotate the
+// resources within it, any single change will affect only the
+// resources in the same stack, meaning fewer things to annotate. (So
+// why not do these individually? This, too, can be expensive, since
+// it involves examining each resource individually).
+type SyncStack struct {
```
2opremio: `SyncAction` was a really good name. On the other hand `Stack` (which I read about before getting here) was a bit confusing.  It made me wonder if it had to do something with user's application stack (or whether we are explicitly letting the user split their cluster resources in `Stacks`).

I probably don't have a better alternative. Maybe `SyncSet`? or ... `SyncGroup`? Or ... `SyncTargetGroup`?

cluster/kubernetes/sync.go
squaremo: The reason it's here is that this is the interface wrapping `kubectl apply`, and the default namespace is essentially a property of `kubectl apply`.

The only reason we care about the default namespace is that when manifests that don't include one are applied, they will be defaulted. Adding it during parsing is a shortcut -- it's really part of the interpretation of the manifests as things to be applied, not inherent to the manifests themselves.

cluster/sync.go
2opremio: Also, I am wondering whether we can use Merkle Trees (like Cassandra uses for repairs) in order to make this more efficient.

cluster/kubernetes/sync.go
squaremo: Yes, it should probably fail if `kubectl` can't be invoked -- good point.

cluster/kubernetes/sync.go
squaremo: Yeah -- this deserves a regression test at least.

sync/sync.go
```diff
@@ -1,78 +1,36 @@
...
-		for id, res := range clusterResources {
-			prepareSyncDelete(logger, repoResources, id, res, &sync)
-		}
-	}
+// Sync synchronises the cluster to the files under a directory.
+func Sync(repoResources map[string]resource.Resource, clus Syncer) error {
+	// TODO: multiple stack support. This will involve partitioning
+	// the resources into disjoint maps, then passing each to
+	// makeStack.
+	defaultStack := makeStack("default", repoResources)
```
2opremio: Ah! So stacks are not really used yet. Isn't it a bit of over-engineering to add all the infrastructure before they are actually used?

cluster/sync.go
squaremo: `SyncSet` is good. `Stack` comes from the inspiration for the PR, via the author -- I left it in, as homage, but the PR has drifted far enough that it's a bit misleading now.

sync/sync.go
squaremo: A little, yes -- it amounts to a couple of unnecessary loops, and a bit of cognitive burden. If I were to remove them, I'd just strip away the outer struct and the loops.

2opremio: I would vote for removing them for now. My reasoning is:
* This PR doesn't make use of multiple stacks yet.
* Sounds like a premature optimization. Do we really need them? Since we need to load the resources into memory for parsing anyways, I doubt that checksumming of each resource would be too expensive.

squaremo: > * I doubt that checksumming of each resource would be too expensive.

It's done per resource anyway

squaremo:
Precis of discussion with @2opremio about landing this and #1668:

major:
 - ParsingContext / Defaulting the namespace
   - it only matters when applying, so do as close to that
   - does that fit in with 1668 -- whose job is it to filter out the unwanted namespaces?
     - see if it can be done as a post-process step, rather than while parsing; then think about making operations gated on it
 - do GC in separate step?
   - try to break the Sync method down a bit at least
 - needs regression tests
   - yep, especially for getting the context, as the output format could change
 - use caching client / otherwise merge with 1668 approach to namespaces
   - try invalidating on a watch of CRDs (we expect those can change), and otherwise on a conservative schedule (every ten minutes, to account for things we don't expect to change)
 - reduce to just a single SyncSet

minor:
 - try bumping to client v10? the reason not to was to not drag in all the attendant dependency updates, but perhaps worth a try again
   - backup plan: put the duplicated fakedynamicclient in its own package
 - try using ParseGroupVersion -- yep
 - constants for changesets, yes they are a bit funny, but leave for now
 - let it fail when kubectl invocation fails (but not if there are no contexts or no current context)

squaremo:
The WIP commit dd71093 has broken tests for now, but does these things:

> ParsingContext / Defaulting the namespace
> * it only matters when applying, so do as close to that
> * does that fit in with 1668 -- whose job is it to filter out the unwanted namespaces?
>   
>   * see if it can be done as a post-process step, rather than while parsing; then think about making operations gated on it

The namespace adjustment is done in `cluster/kubernetes.Manifests#LoadManifests`, rather than `cluster/kubernetes/resource.load`, which separates it from purely parsing.

> use caching client / otherwise merge with 1668 approach to namespaces

The namespace adjustment is done by a component which combines a means of getting the fallback namespace with a cached Discovery client, so it can see if a given resource is supposed to have a namespace, and if so, apply the default. There's no cache invalidation done at present.

This adds yet more complication to fluxd/main.go, since it has to assemble this `Namespacer` thing to give to the Manifests. But foregoing that would make it more difficult to swap bits out for testing.

cluster/kubernetes/manifests.go
```diff
@@ -7,12 +7,36 @@ import (
 	"github.com/weaveworks/flux/resource"
 )
 
+// namespacer assigns namespaces to manifests that need it (or "" if
+// the manifest should not have a namespace.
+type namespacer interface {
+	correctNamespace(kresource.KubeManifest) error
+}
+
+// Manifests is an implementation Manifests particular to
```
2opremio: _an implementation_ **of cluster.**_Manifests_?

squaremo: I briefly considered blaming golint for forcing me into the awkward grammar but that would be a deflection

cluster/kubernetes/manifests.go
```diff
@@ -7,12 +7,36 @@ import (
...
+// Kubernetes. Aside from loading manifests from files, it does some
+// "post-processsing" to make sure the view of the manifests is what
+// would be applied; in particular, it fills in the namespace of
+// manifests that would be given a default namespace when applied.
 type Manifests struct {
-	FallbackNamespace string
+	Namespacer namespacer
```
2opremio: I find it a bit odd that the field and the constructor are public, but the type isn't.

cluster/kubernetes/manifests.go
```diff
@@ -7,12 +7,36 @@ import (
 	"github.com/weaveworks/flux/resource"
 )
 
+// namespacer assigns namespaces to manifests that need it (or "" if
+// the manifest should not have a namespace.
+type namespacer interface {
```
2opremio: Wouldn't it be more intuitive to declare the interface in `namespacer.go`?

cluster/kubernetes/manifests.go
```diff
@@ -7,12 +7,36 @@ import (
...
+	correctNamespace(kresource.KubeManifest) error
```
2opremio: I am not sure I like the stateful interface. I would probably do something like:

```
type namespacer interface {
  ClusterNamespace(kresource.KubeManifest) (string, error)
}
```

cluster/kubernetes/manifests.go
```diff
@@ -7,12 +7,36 @@ import (
...
-	return kresource.Load(base, c.FallbackNamespace, paths)
+	manifests, err := kresource.Load(base, paths)
+	if err != nil {
+		return nil, err
+	}
+	result := map[string]resource.Resource{}
+	for _, km := range manifests {
+		if c.Namespacer != nil {
+			if err = c.Namespacer.correctNamespace(km); err != nil {
+				return nil, err // TODO(michael) Consider whether it's better to just omit this resource
```
2opremio: I think it's better to report an error, since simply omitting the resource can lead to an unexpected behavior (like not syncing resources _atomically_ (or in one go)).

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	case namespaced && m.GetNamespace() == "":
+		m.SetNamespace(n.fallbackNamespace)
+	case !namespaced:
+		m.SetNamespace("")
+	}
+	return nil
+}
+
+func (n *namespaceViaDiscovery) lookupNamespacedness(m kresource.KubeManifest) (bool, error) {
```
2opremio: s/`lookupNamespacedness`/`isNamespaced`/?

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,56 @@
...
+func NewNamespacer(k *Kubectl, d discovery.DiscoveryInterface) (*namespaceViaDiscovery, error) {
+	fallback, err := k.getDefaultNamespace()
+	if err != nil {
+		return nil, err
+	}
+	cachedDisco := cached.NewMemCacheClient(d)
+	// in client-go v9, the call of Invalidate is necessary to force
+	// it to query for initial data; in subsequent versions, it is a
+	// no-op reset of the cache validity, so safe to call.
+	cachedDisco.Invalidate()
```
2opremio: I am missing some sort of invalidation or watch of CRDTs. Is this coming later?

squaremo: Yeah, coming later, probably in the form of `NewNamespacerViaDiscovery` creating the watch.

Would you elaborate on the consolidation you'd expect to see? Do you mean, should the bit that loads resources from the cluster _also_ adjust namespaces?

cluster/kubernetes/namespacer.go
squaremo: I do so like stretching words out of shape .. but not super appropriate for code

cluster/kubernetes/manifests.go
squaremo: The interface type describes the requirement; for that purpose, it doesn't need to be referred to outside the package. (The compiler will tell you if it turns out what you're putting in that field doesn't meet the requirement).

cluster/kubernetes/manifests.go
squaremo: It's declared next to where it's required. This is a bit unfamiliar from a reference point of nominal interface types (in which you assert that your implementation fits the interface, and the compiler tells you if it doesn't).

But Go's interfaces are structural -- you assert that your _value_ fits the interface of the thing you're assigning to, and the compiler checks that you're right. Since there's no need to name the interface when constructing a value, it doesn't need to be accessible outside the package, and since it's used in one place, the most logical places for its definition is next to that.

Granted, what's strictly necessary is not always what's best; and, the rationale above is not exactly applied consistently through this codebase..

squaremo: ^ said with the conviction of someone who's only recently adopted this position :-)

cluster/kubernetes/namespacer.go
2opremio: I mean using the same cache both for this and getting all the resources for garbage-collection.

squaremo: Oh good question -- I've made that tricky for myself, but it would be nice to be able to do it

squaremo: I don't think you can watch APIResources, which would be the ideal, but I think it's possible to watch custom resource definitions: https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/client/clientset/clientset/typed/apiextensions/v1beta1/customresourcedefinition.go

This is just a more responsive (or accurate) version of "invalidate on a ticker"; there would still be opportunity for getting namespacedness wrong if it were changed. TBD: what happens when it gets the namespacedness wrong...

cluster/kubernetes/kubernetes.go
```diff
@@ -110,6 +92,7 @@ type Cluster struct {
 
 // NewCluster returns a usable cluster.
 func NewCluster(clientset k8sclient.Interface,
+	dynamicClientset k8sclientdynamic.Interface,
```
2opremio: Sorry I didn't catch this before, but ... the number of parameters is out of hand. Wouldn't it make sense to create a structure for this?

squaremo: Yes, or accept an interface, and construct the clientset from outside (the problem there might be that the clientset interfaces have duplicate method definitions).

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,166 @@
...
+	store      toolscache.Store
+	controller toolscache.Controller
+
+	invalidMu sync.Mutex
+	invalid   bool
+}
+
+// The older (v8.0.0) implementation of MemCacheDiscovery refreshes
+// the cached values, synchronously, when Invalidate is called. Since
+// we will invalidate every time something cahnges, but it only
```
2opremio:
```suggestion
// we will invalidate every time something changes, but it only
```

cluster/kubernetes/kubernetes.go
squaremo: > the clientset interfaces have duplicate method definitions

Ah yes, every single clientset has Discovery(), so it can't be done on that level. It might be possible to embed the more specific API-level interfaces ..

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,166 @@
...
+	discovery.CachedDiscoveryInterface
+
+	store      toolscache.Store
+	controller toolscache.Controller
+
+	invalidMu sync.Mutex
+	invalid   bool
+}
+
+// The older (v8.0.0) implementation of MemCacheDiscovery refreshes
```
2opremio: I am a bit confused. We are actually vendoring version v8.0.0 of client-go.

squaremo: Yes; if we were using the newer one (and there was no chance of going back, or it changing its implementation again), this wouldn't be necessary.

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,166 @@
...
+	disco             discovery.DiscoveryInterface
+}
+
+type namespaceDefaulter interface {
+	GetDefaultNamespace() (string, error)
+}
+
+// NewNamespacer creates an implementation of Namespacer
+func NewNamespacer(ns namespaceDefaulter, d discovery.DiscoveryInterface) (*namespaceViaDiscovery, error) {
+	fallback, err := ns.GetDefaultNamespace()
```
2opremio: Since the user can (at least theoretically) mount and modify the kubeconfig file on the fly (including the default namespace), I am not sure it's a good idea to simply read it once and be happy with it.

squaremo: Mounting a new file === creating a new pod, no?

squaremo: Oh, as in, you can modify the contents of the ConfigMap or Secret, and that eventually gets reflected in the pod's filesystem. Yes, that can happen, true. It's not assumed that containers will adapt to those changes -- but, we can fairly cheaply and helpfully do so.

cmd/fluxd/main.go
```diff
@@ -274,6 +309,18 @@ func main() {
 		// There is only one way we currently interpret a repo of
 		// files as manifests, and that's as Kubernetes yamels.
 		k8sManifests = &kubernetes.Manifests{}
+
+		crdClient, err := crd.NewForConfig(restClientConfig)
+		if err == nil {
```
2opremio: I guess this is meant to only check for err once, but I think it makes the code harder to read (since the GO convention is to check for errors first)

cmd/fluxd/main.go
```diff
@@ -274,6 +309,18 @@ func main() {
...
+			disco := kubernetes.MakeCachedDiscovery(clientset.Discovery(), crdClient, shutdown)
+			k8sManifests.Namespacer, err = kubernetes.NewNamespacer(kubectlApplier, disco)
+		}
+		if err == nil {
+		}
```
2opremio: Is this meant to contain code-to-be?

squaremo: No, just copypasta I think.

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,166 @@
...
+	h.first.OnDelete(old)
+	h.next.OnDelete(old)
+}
+
+// MakeCachedDiscovery constructs a CachedDicoveryInterface that will
+// be invalidated whenever the set of CRDs change. The idea is that
+// the only avenue of a change to the API resources in a running
+// system is CRDs being added, updated or deleted. The prehandlers are
+// there to allow us to put extra synchronisation in, for testing.
+func MakeCachedDiscovery(d discovery.DiscoveryInterface, c crd.Interface, shutdown <-chan struct{}, prehandlers ...toolscache.ResourceEventHandler) *cachedDiscovery {
```
2opremio: The `cachedDiscovery` is only used within the `namespaceViaDiscovery`, but I understand you want to reusecachedDiscovery` for the garbage collection during syncing, and thus it makes sense to pass it as a parameter instead of creating in internally (requiring to make the constructor public). 

However, in that case, I think it makes sense to put the `cachedDiscovery` implementation in a separate file.

cmd/fluxd/main.go
squaremo: I really despise that convention :-/, but you are right that it's a convention and going against it is surprising to the reader.

cluster/kubernetes/namespacer.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, resource := range resourceList.APIResources {
+		if resource.Kind == kind {
+			return resource.Namespaced, nil
+		}
+	}
+	return false, fmt.Errorf("resource not found for API %s, kind %s", groupVersion, kind)
+}
+
+// This mainly exists so I can put the controller and store in here,
+// so as to have access to them for testing; and, so that we can do
```
2opremio: Is there a reason not to move `store` and `controller` to the tests (as local variables initialized after calling `MakeCachedDiscovery`? I think that placing them here makes it more difficult to read the code.

cluster/kubernetes/doc.go
```diff
@@ -3,5 +3,4 @@ Package kubernetes provides implementations of `Cluster` and
 `Manifests` that interact with the Kubernetes API (using kubectl or
 the k8s API client).
 */
-
```
2opremio: Was this intentional?

cluster/kubernetes/kubernetes.go
2opremio: Sounds good!

Regardless, I think it probably makes sense to transform the parameters of `NewCluster` into a struct (although maybe not in this PR)

cluster/kubernetes/namespacer.go
squaremo: The fields are assigned during construction, and wouldn't be kept except that it's useful during testing to be able to reach in and check what's been recorded. Could you elaborate on what you're suggesting?

cluster/kubernetes/namespacer.go
2opremio: Ah, older as compared to the latest version released not compared to the one vendored. Gotcha. I think some rephrasing will make it easier to understand though.

cluster/kubernetes/namespacer.go
squaremo: Although -- not sure I want to do an exec every time EffectiveNamespace is called, so this might involve an explicit "Refresh" method on the namespacer, or some indirection

cluster/kubernetes/namespacer.go
2opremio: Uhm, I am also seeing the `chainHandler` thing, which is currently only used in testing. If you are not planning to use it with normal code, I would also move it to testing.

For instance we could create a private `makeCacheDiscovery` which the tests could use:

```golang
func MakeCachedDiscovery(d discovery.DiscoveryInterface, c crd.Interface, shutdown <-chan struct{}) *cachedDiscovery {
  lw, handler, result := makeCachedDiscovery(d, c, shutdown)
  toolscache.NewInformer(lw, &crdv1beta1.CustomResourceDefinition{}, 5*time.Minute, handler)
  return result
}
```

The tests could invoke `makeCachedDiscovery` and then call `toolscache.NewInformer` after chaining the handles and keep the results locally.

It's not as pretty as I would like to, so I will lave it up to you.

squaremo: OK, I get the suggestion now -- thanks Fons

cluster/sync.go
```diff
@@ -3,34 +3,36 @@ package cluster
...
-	// The actions to undertake
-	Actions []SyncAction
+// SyncSet groups the set of resources to be updated. Usually this is
+// the set of resources found in a git repo; in any case, it must
+// represent the complete set of resources, as garbage collection will
+// assume missing resources should be deleted. The name is used to
+// distinguish the resources from a set from other resources -- e.g.,
+// cluster resources not marked as belonging to a set will not be
+// deleted by garbage collection.
+type SyncSet struct {
```
2opremio: After reading the code more carefully I don't see the reason for `SyncSet` at all, particularly since we enforce the whole repository resources to be included for garbage collection and `Name` always being set to `git`

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,308 @@ package kubernetes
...
+	c.muSyncErrors.RLock()
+	if applyErrs := c.applier.apply(logger, cs, c.syncErrors); len(applyErrs) > 0 {
+		errs = append(errs, applyErrs...)
+	}
+	c.muSyncErrors.RUnlock()
+
+	if c.GC {
+		orphanedResources := makeChangeSet()
+
+		clusterResources, err := c.getResourcesInSyncSet(spec.Name)
```
2opremio: I must be missing something but ... by only getting the resources in the syncset (which ultimately are the resources in the Git repo), don't we exclude the ones deleted in the git repo, preventing from garbage collecting them?

2opremio: Scratch that, that's not how `getResourcesInSyncSet` works

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,308 @@ package kubernetes
...
+		if err != nil {
+			return errors.Wrap(err, "collating resources in cluster for calculating garbage collection")
+		}
+
+		for resourceID, res := range clusterResources {
+			actual := res.GetChecksum()
+			expected, ok := checksums[resourceID]
+
+			switch {
+			case !ok: // was not recorded as having been staged for application
```
2opremio: As I understand it, this will cause the deletion of resources whose manifest has been made invalid (potentially by mistake) and thus, not included in the syncset. This means that:

1. corrupting a manifest file by mistake is equivalent to deleting it
2. renaming the extension of a file by mistake is equivalent to deleting it

I am not sure we can/should do anything about 2. But I think we should have a mechanism in place for (1)

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,308 @@ package kubernetes
...
+			}
+		}
+	}
+
+	return result, nil
+}
+
+// exportResourcesInStack collates all the resources that belong to a
+// stack, i.e., were applied by flux.
+func (c *Cluster) getResourcesInSyncSet(name string) (map[string]*kuberesource, error) {
```
2opremio: I would rename this to something more descriptive like (by it's name I thought it got the resources inside a SyncSet data structure).

`getClusterSyncSetLabeledResources` ?
`getSyncSetLabeledResources` ? (which is shorter)

cluster/sync.go
squaremo: Since this ends up being reflected in clusters, I would like to leave at least a bit of room for extending it to more than one set of resources. (If it were entirely internal, I wouldn't care).

cluster/kubernetes/sync.go
squaremo: It's a method on the cluster.

cluster/kubernetes/sync.go
squaremo: There's a few ways a file can be invalid:

 - no longer recognised as being a YAML file: I agree, nothing much we can do about this
 - if it's invalid YAML: we bail on parsing if any files are invalid YAML, so we wouldn't get this far
 - if it's valid YAML, but not accepted by kubectl apply: we still explicitly record these to prevent them being garbage-collected

2opremio: Oh, perfect then!

cluster/sync.go
2opremio: I see

cluster/kubernetes/sync.go
2opremio: Yeah, that's true

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,308 @@ package kubernetes
...
+
+// GetChecksum returns the checksum recorded on the resource from
+// Kubernetes, or an empty string if it's not present.
+func (r *kuberesource) GetChecksum() string {
+	return r.obj.GetAnnotations()[checksumAnnotation]
+}
+
+// GetSyncSet returns the sync set name recorded on the the resource
+// from Kubernetes, or an empty string if it's not present.
+func (r *kuberesource) GetSyncSet() string {
```
2opremio: This is dead code.

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,308 @@ package kubernetes
...
+		// make a record of the checksum, whether we stage it to
+		// be applied or not, so that we don't delete it later.
+		csum := sha1.Sum(res.Bytes())
+		checkHex := hex.EncodeToString(csum[:])
+		checksums[id] = checkHex
+		if res.Policy().Has(policy.Ignore) {
+			logger.Log("info", "not applying resource; ignore annotation in file", "resource", res.ResourceID(), "source", res.Source())
+			continue
+		}
+		if cres, ok := clusterResources[id]; ok && cres.Policy().Has(policy.Ignore) {
```
2opremio: I am not entirely sure this is correct.

1. Git should the source of truth
2. Doesn't this prevent users from ever effectively removing the ignore policy on a git-controlled resource which was synced to the cluster? (i.e. Add resource file in  git -> Sync -> Mark as ignored -> Mark as not ignored -> Make changes in git)

cluster/kubernetes/sync.go
squaremo: True -- because we just select it directly from the API now. Well spotted!

cluster/kubernetes/doc.go
squaremo: Yes, docs ain't docs unless they're immediately before the package declaration (although arguably it doesn't belong in this PR -- it's incidental).

cluster/sync.go
2opremio: After thinking about it a bit more, as it is today you don't even have the possibility of splitting the resources in multiple sets (causing garbage collection to remove resources in the other sets) so I still think we should remove it (since it leads to think you can actually use multiple sets).

daemon/loop.go
```diff
@@ -23,6 +23,7 @@ import (
 const (
 	// Timeout for git operations we're prepared to abandon
 	gitOpTimeout = 15 * time.Second
+	syncSetName  = "git"
```
2opremio: I think that having a constant `syncSetName` can be dangerous since it's used as the label value to gather synced resources:

1. when running multiple `fluxd` instances on non-overlapping resources from different git repos, this will result in a fight to garbage-collect each other's resources.

2. It may also prove problematic once (if we ever) support multiple repos in a single fluxd instance. 

3. It will cause fluxd to remove all the resources in a git repo once we change it's configuration to another git repo with different resources. Intuitively this may be correct, but can be potentially disastrous and easy to miss by users.

As a solution we could append a hash of the git repo details to the label value. When doing this we need to make sure that we overwrite the label value to that of the repository currently used (mergo can be tricky). Alternatively, we can add the hash to the label key, but this can end up leaving garbage behind.

cluster/kubernetes/kubernetes.go
```diff
@@ -91,21 +103,9 @@ type Cluster struct {
...
-func NewCluster(clientset k8sclient.Interface,
-	dynamicClientset k8sclientdynamic.Interface,
-	fluxHelmClientset fhrclient.Interface,
-	applier Applier,
-	sshKeyRing ssh.KeyRing,
-	logger log.Logger,
-	nsWhitelist []string,
-	imageExcludeList []string) *Cluster {
-
+func NewCluster(client ExtendedClient, applier Applier, sshKeyRing ssh.KeyRing, logger log.Logger, nsWhitelist []string, imageExcludeList []string) *Cluster {
```
2opremio: Nice!

daemon/loop.go
squaremo: Yes, it should be something representing the git {repo,branch}. That may leave stragglers if people change the config, but at least they will be easy to find.

cluster/kubernetes/sync.go
squaremo: If it's marked as ignore, it won't be synced to the cluster -- so you can't end up with an unsyncable resource that way.

But you _can_ mark something as ignored directly in the cluster, i.e, with `kubectl annotate ...`. That means it won't be updated when syncing, and won't be garbage collected either. It is an escape hatch from gitops, basically. To return it to control of fluxd, you have to make sure the manifest (if there is one) doesn't have the annotation, and remove the annotation from the cluster resource.

2opremio: Ah, I see, an ignore annotation from git never reaches the cluster. That makes sense. I think that a comment clarifying what are we doing (checking for manually-ignored cluster resources) and why it's ok (flux will never apply ignore annotations from git), would be helpful

2opremio: Ah, I see, an ignore annotation from git never reaches the cluster. That makes sense. I think that a comment clarifying what are we doing (checking for manually-ignored cluster resources) and why it's ok (flux will never apply ignore annotations from git), would be helpful

daemon/loop.go
2opremio: So, are you going to go for a different key per repo?

cluster/kubernetes/cached_disco.go
```diff
@@ -0,0 +1,104 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/discovery"
+	discocache "k8s.io/client-go/discovery/cached"
+	toolscache "k8s.io/client-go/tools/cache"
+)
+
+// This mainly exists so I can put the controller and store in here,
+// so as to have access to them for testing; and, so that we can do
```
2opremio: The first part of the comment doesn't apply anymore

daemon/loop.go
squaremo: hash({URL, branch}) (or even just {URL, branch}, used as the label value, I would think. But then, if you run two fluxd's and point them at different the same repo and branch, but different directories, they will garbage collect each other's resources.

The bounds are, I guess:
 - it's OK to have too much information, so long as it's stable between restarts (otherwise fluxd won't collect things it should collect).
 - having too little information in the label means you can (easily) have two fluxd's that conflict.
Augh..

2opremio: Uhm, I am thinking that, to be safe, we could stop guaranteeing resources being garbage collected between restarts. That is, if a resource is removed in git while fluxd is down, it won't be garbage collected.

That would allow us to use a label value (or key) associated with the current execution/instance of fluxd (e.g. through a random UUID generated at start). In this way, different fluxd instances would only be competing with each other if they are controlling exactly the same resource (which is unavoidable), but would never garbage collect each other's resources.

As an alternative, we could extend the hash to git directories and whatever we think is relevant to avoid mutual garbage collections.

Note that, even when using a git hash of the repo, we would miss garbage collections in restarts if the git repo is moved somewhere else.

cluster/kubernetes/cached_disco.go
squaremo: Oh yes. oops

cluster/kubernetes/kubernetes.go
squaremo: It's a lot tidier, thanks for the suggestion. We could also

 - make NewCluster accept an interface with only the client API methods it uses (though this would mean gathering all the various bits of clientset, verbatim; fiddly, but maybe worth it for being precise about the requirements of the cluster implementation)
 - use an options value for the assorted other arguments passed to NewCluster (`nsWhitelist` and `imageExcludeList` e.g.)

daemon/loop.go
squaremo: > we could stop guaranteeing resources being garbage collected between restarts

I think this would be a bit of a disaster for people operating fluxd -- since fluxd can be restarted arbitrarily, this means you'd never know whether garbage collection had been done, or wasn't (ever) going to be.

> As an alternative, we could extend the hash to git directories and whatever we think is relevant to avoid mutual garbage collections
> [...]
> I don't think we can really guarantee garbage collections between restarts in all cases without risking conflicts between fluxd instances.

Good insight: using the path (either git-path argument, or source path of the resource, or something along those lines) means it's more sensitive to changes in the configuration:

| Labelling | What happens with two fluxds | What happens after reconfiguration |
|-------|--------------------------------------|----------------------------------------|
| constant e.g., "git" | if fluxds are syncing different repos or directories, they will GC each other's resources | everything proceeds as before |
| git repo URL | if fluxds are given the same repo, but _different_ directories, they will GC each other's resources | if the git repo is moved, everything that's still there will get relabelled; anything that's now missing will be orphaned (never collected) |
| git URL and git-path or source path | fluxd's may relabel each other's resources if they are in common, but won't GC each other's resources | if you remove a directory from fluxd's config, it will be orphaned (never collected) |

We could use the `--git-label`, which already needs to be distinct per fluxd (though only if you're using the same repo). Using a unique ID:

| Labelling | What happens with two fluxds | What happens after reconfiguration |
|-------|--------------------------------------|----------------------------------------|
| fluxd ID | fluxds won't collect each other's resources; | if you change who is responsible for a resource, it'll get relabelled |

2opremio: > if you remove a directory from fluxd's config, it will be orphaned (never collected)

I don't think this is a terrible thing (particularly if you consider the possibility of making mistakes in configuration errors)

2opremio: After discussing it offline it was decided to go for a hash of the git URL + git paths

cluster/kubernetes/namespacer.go
2opremio: Happy to create a test for this after merging https://github.com/weaveworks/flux/pull/1765 and rebasing on top of it.

cluster/kubernetes/sync.go
2opremio: Happy to create a test for this after merging #1765 and rebasing on top of it.

cluster/kubernetes/namespacer.go
squaremo: On reflection, I'm not keen on putting in a Refresh method, as the abstractions stand at the minute. I think it'd be better to live with requiring a restart on changes for now, and revisit how the namespacing works afterwards.

cluster/sync.go
squaremo: The name is now calculated based on the git repo config, and supplied im the SyncSet rather than being constant. (In other words, it's necessary to label the SyncSet). Providing more than one, at different times, would be OK since the garbage collection only examines things with the name given -- so it's not misleading.

cluster/kubernetes/sync.go
```diff
@@ -2,31 +2,293 @@ package kubernetes
...
+func (c *Cluster) Sync(spec cluster.SyncSet) error {
+	logger := log.With(c.logger, "method", "Sync")
+
+	// Keep track of the checksum of each resource, so we can compare
+	// them during garbage collection.
+	checksums := map[string]string{}
+
+	// NB we get all resources, since we care about leaving unsynced,
+	// _ignored_ resources alone.
+	clusterResources, err := c.getResourcesBySelector("")
```
squaremo: For follow-up at some point:

Here we want to know about resources corresponding to manifests we have in the git repo, so we can see if they are ignored.

We can do better than asking for all resources. For a start, the manifests in the cluster will only be a subset of the API resources from the cluster, so we need not ask for all of those. Moreover, it's exactly those in the manifests, that aren't already annotated as being ignored, that we need to look up; so we only need to ask for specific namespaces, and in those, specific resources.

Only requesting the API resources needed would come at little cost (a pass through the manifests, or a pull-through cache of some sort). That seems like a good trade.

Restricting the namespaces searched again seems like a good trade, since that reduces the number of requests with little cost.

Asking for the specific resources would probably be a step too far, though -- in general, most resources won't be ignored, and requesting lists of resources is more efficient than asking for individual resources.

2opremio: @squaremo Mind adding a comment/issue covering this?

cluster/kubernetes/namespacer.go
2opremio: Agreed

cluster/kubernetes/namespacer.go
```diff
@@ -4,28 +4,54 @@ import (
...
 	}
 	return &namespaceViaDiscovery{fallbackNamespace: fallback, disco: d}, nil
 }
 
+// getDefaultNamespace returns the fallback namespace used by the
+// when a namespaced resource doesn't have one specified. This is
+// used when syncing to anticipate the identity of a resource in the
+// cluster given the manifest from a file (which may be missing the
+// namespace).
+// A varaible is used for mocking in tests
```
squaremo: "variable"

2opremio:
I think all the comments have been addressed. The only missing thing is documentation.

site/faq.md
```diff
@@ -145,6 +146,18 @@ For more information about Flux commands see [the fluxctl docs](./fluxctl.md).
 
 No. It applies changes to git only when a Flux command or API call makes them.
 
+### Will Flux delete resources when I remove them from git?
+
+Flux has an experimental (for now) garbage collectiion feature,
```
hiddeco: 'collectiion' -> 'collection'

theduke:
Just wanted to give a shoutout and thanks to @Timer , implementation of this feature is much appreciated!

PR #1444, opened 2018-10-12
Explain what is synced to git by flux

### LINKED ISSUE #1443: Document what gets synchronized to the git repo by flux
Reading the documentation it is not at all clear to me what gets automatically synchronized to the git repo by flux. The sentence here https://github.com/weaveworks/flux/blob/master/site/faq.md#why-does-flux-need-a-deploy-key "Flux needs a deploy key to be allowed to push to the version control system in order to read and update the manifests", seems to suggest to me that only the manifests which are in the git repo will be updated, and other manifests will not. I presume that means that if I upload secrets to the cluster using kubectl then they won't end up automatically synced to my git repo by flux, but that's not actually stated and this lack of clairity makes me quite nervous.

timthelion:
Fixes #1443

### DIFF
--- a/site/faq.md
+++ b/site/faq.md
@@ -105,7 +105,7 @@ For example, if you use the following `fluxctl` command:
 
 The image tag will be updated in the git repository upon applying the command.
 
-For more information about flux commands see [the fluxctl docs](./using.md).
+For more information about Flux commands see [the fluxctl docs](./using.md).
 
 ### Does Flux automatically sync changes back to git?
 
--- a/site/faq.md
+++ b/site/faq.md
@@ -72,7 +72,7 @@ daemon and point them at different repos. If you do this, consider
 trimming the RBAC permissions you give each daemon's service account.
 
 This
-[flux (daemon) operator](https://github.com/justinbarrick/flux-operator)
+[Flux (daemon) operator](https://github.com/justinbarrick/flux-operator)
 project may be of use for managing multiple daemons.
 
 ### Do I have to put my application code and config in the same git repo?
@@ -94,10 +94,22 @@ Kubernetes, use `--git-path` to constrain where Flux starts looking.
 See also [requirements.md](./requirements.md) for a little more
 explanation.
 
-### Why does Flux need a deploy key?
+### Why does Flux need a git ssh key with write access?
 
-Flux needs a deploy key to be allowed to push to the version control
-system in order to read and update the manifests.
+There are a number of Flux commands and API calls which will update the git repo in the course of
+applying the command. This is done to ensure that git remains the single source of truth.
+
+For example, if you use the following `fluxctl` command:
+
+    fluxctl release --controller=deployment/foo --update-image=bar:v2
+
+The image tag will be updated in the git repository upon applying the command.
+
+For more information about flux commands see [the fluxctl docs](./using.md).
+
+### Does Flux automatically sync changes back to git?
+
+No. It applies changes to git only when a Flux command or API call makes them.
 
 ### How do I give Flux access to an image registry?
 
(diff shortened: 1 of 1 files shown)


site/faq.md
```diff
@@ -99,6 +99,14 @@ explanation.
 Flux needs a deploy key to be allowed to push to the version control
 system in order to read and update the manifests.
 
+
+### What is auto-matically synced back to git by flux?
```
squaremo: Nothing is synced from the cluster to git -- syncing only goes one way, from git to the cluster.

There are some API calls (and `fluxctl` commands) that make updates to the files in git, and automated updates may also make changes in git, based on what you've told it to do. These go `command or automation -> git -> cluster`

timthelion: So it is potentially possible that in the future flux wouldn't even need write access to the git repo and would be able to operate in a limited mode without it?

squaremo: Yup; in theory (and some designs I've sketched out) it'd be possible to mark a repo as read only, so it will apply things to the cluster, but refuse updates.

timthelion: I've updated the PR. Is that better?

site/faq.md
```diff
@@ -99,6 +99,13 @@ explanation.
...
+
+Nothing is auto-matically sent from the cluster to git! However, some flux commands and API
```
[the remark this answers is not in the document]
timthelion: Fixed

site/faq.md
```diff
@@ -99,6 +99,13 @@ explanation.
 Flux needs a deploy key to be allowed to push to the version control
 system in order to read and update the manifests.
 
+
+### What is auto-matically synced back to git by Flux?
```
squaremo: It may be clearer to answer a more specific, yes/no question. I think the original misdirection was in the sentence above: 

> Flux needs a deploy key to be allowed to push to the version control system in order to read and update the manifests

so the follow-up question could be "Will Flux update git when I change the cluster?", with an unambiguous, unqualified answer.

site/faq.md
```diff
@@ -99,6 +99,13 @@ explanation.
 Flux needs a deploy key to be allowed to push to the version control
 system in order to read and update the manifests.
```
squaremo: This is poorly phrased in itself -- I'll make a note to improve it.

site/faq.md
```diff
@@ -99,6 +99,13 @@ explanation.
...
 
+
+### What is auto-matically synced back to git by Flux?
+
+Nothing is auto-matically sent from the cluster to git! However, some Flux commands and API
+calls will update the git repo so as to reflect the current state of the cluster or the new state
```
squaremo: The `fluxctl` commands don't reflect the current state of the cluster so much as they change the desired state of the cluster, as expressed in the manifests in git. It's not really a qualification -- just a related bit of information. You could mention automated updates here too.

site/faq.md
squaremo: No need for a hyphen here - just "automatically"

timthelion:
What about this version?

site/faq.md
```diff
@@ -96,8 +96,13 @@ explanation.
...
 
-Flux needs a deploy key to be allowed to push to the version control
-system in order to read and update the manifests.
+There are a number of Flux commands and API calls which will update the git repo in order to
+reflect the state of the cluster after applying the command. For more information about flux
+commands see [the fluxctl docs](./using.md).
+
+### Does Flux automatically sync changes back to git?
+
+No. It only syncs with git when a flux command or API call makes an applicable change.
```
squaremo: The second sentence seems to qualify the first, by saying ".. only when..". To make it  less ambiguous, perhaps elaborate by stating what _does_ happen -- that is, changes go from git to the cluster, and commands update git. Commits by people directly to the git repo will _also_ be applied to the cluster.

site/faq.md
```diff
@@ -96,8 +96,13 @@ explanation.
 
 ### Why does Flux need a deploy key?
 
-Flux needs a deploy key to be allowed to push to the version control
-system in order to read and update the manifests.
+There are a number of Flux commands and API calls which will update the git repo in order to
+reflect the state of the cluster after applying the command. For more information about flux
+commands see [the fluxctl docs](./using.md).
+
```
squaremo: Oh yes, updating this bit helps too :-) There's two bits to the answer: one is why we need a deploy key (to get access to the git repo), and the other is why Flux needs both read and write access. You could cover both by explaining the need for read access to sync, and the need for write access to update git when asked.

Secondarily -- "reflect the state of the cluster after applying the command" doesn't follow, because it suggests that commands take what's in the cluster and put that in git. In general, the commands update manifests according to the user's request -- "update the deployment's image to be foo:v1" -- rather than mirroring information from the cluster.

timthelion: If it were a public repo, you wouldn't need the key for read access and I can hardly imagine that people frequently get confused about Flux needing a key to read from private repos. For me, the main source of confusion is why Flux needs write access.

timthelion:
I'm signing off for today, but I've tried to incorporate your suggestions.

site/faq.md
squaremo: That's a good point -- what you've written above explains it well.

PR #1445, opened 2018-10-15
Deal with case of no service request arguments

squaremo:
It's possible to send a request to ListServicesWithOptions that doesn't specify the services of interest.

    fluxctl list-controllers

will do exactly that -- and currently, it gets a 400 Bad Request.

### DIFF
--- a/http/daemon/server.go
+++ b/http/daemon/server.go
@@ -144,13 +144,15 @@ func (s HTTPServer) ListServicesWithOptions(w http.ResponseWriter, r *http.Reque
 	var opts v11.ListServicesOptions
 	opts.Namespace = r.URL.Query().Get("namespace")
 	services := r.URL.Query().Get("services")
-	for _, svc := range strings.Split(services, ",") {
-		id, err := flux.ParseResourceID(svc)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


http/daemon/server.go
```diff
@@ -144,13 +144,15 @@ func (s HTTPServer) ListServicesWithOptions(w http.ResponseWriter, r *http.Reque
 	var opts v11.ListServicesOptions
 	opts.Namespace = r.URL.Query().Get("namespace")
 	services := r.URL.Query().Get("services")
-	for _, svc := range strings.Split(services, ",") {
-		id, err := flux.ParseResourceID(svc)
-		if err != nil {
-			transport.WriteError(w, r, http.StatusBadRequest, errors.Wrapf(err, "parsing service spec %q", svc))
-			return
+	if services != "" {
```
rade: Doesn't the new code behave exactly the same as before, given that `strings.Split("", ",")` is `[]`, and hence the for-loop is a no-op in that case?

squaremo: `strings.Split("", ",")` is not `[]`.

rade: Ah, very true. I was misled by the default formatter.

> 	fmt.Println([]string{""})
> []

Well done, golang, well done.

PR #1446, opened 2018-10-15
Add 'extraEnvs' values to Helm chart

brantb:
If users of the Helm chart need to provide additional environment variables to the pod, they can be added to the 'extraEnvs' or 'helmOperator.extraEnvs' arrays in values.yaml, respectively.

[Relevant discussion in Slack](https://weave-community.slack.com/archives/C4U5ATZ9S/p1539609056000100).


stefanprodan:
@brantb can you please add an issue with the AKS workaround and link it to this PR please?

Thanks

mrrandrade:
Do you folks mind if I throw here an example of this parameter's use? Took me a while to figure out.

```
# Adding CHECKPOINT_DISABLE=true for airgap/offline environment: 
helm upgrade -i flux fluxcd/flux \
   --set 'extraEnvs[0].name=CHECKPOINT_DISABLE' \
   --set 'extraEnvs[0].value="true"' \
   --set git.url=git@yourrepo \
   --namespace flux
```

Issue #1278, opened 2018-08-09
Respect requirements.yaml in helm operator


squaremo:
Charts can come with dependencies: https://docs.helm.sh/developing_charts/#chart-dependencies

I believe the Helm CLI deals with these dependencies -- either downloading the charts from the repository given, or grabbing them from the fileystem before sending to Tiller.

We should come up with a sensible way to let people use the dependency management with the helm operator.

ahmadiq:
+1 We're facing this issue as well. We'd like for the chart dependencies in the requirements.yaml to be downloaded automatically. At the moment we're providing a reference to the public chart repository in the requirements.yaml but that doesn't download, and we seem to need to include the chart within our flux release repo.

requirements.yaml snippet:
```
dependencies:

- name: prometheus
  version: 1.0.19
  repository: https://stakater.github.io/stakater-charts
  alias: prometheus

- name: grafana
  version: 1.0.31
  repository: https://stakater.github.io/stakater-charts  
  alias: grafana
```

stephenmoloney:
Is this issue a duplicate of #1131 ?

ahmadiq:
Hi @stephenmoloney yes my use case seems to be the same as in #1131 
@squaremo can assess from his side as well.

dbluxo:
+1

squaremo:
#1131 is for being able to refer to a chart from a Helm repo, in a FluxHelmRelease, and have that installed.
This bug crosses over releasing charts from git and charts from a Helm repo, both of which may have `requirements.yaml`, though we may do different things with it in each case.

squaremo:
In the case of charts from git, I think it's reasonable to expect the helm operator to (optionally) get the dependencies before applying. I'm not sure whether it makes more sense to have this as a command-line switch, or specify it per FluxHelmRelease.

dbluxo:
Why should this be optional? Specified dependencies should be always resolved.

kozejonaz:
@dbluxo Sometimes the dependency isn't used. A good example is a chart that depends on some sort of DB, but allow you to use your own. I guess it doesn't hurt always resolving the dependency, even in situations where you do not use it, to avoid complexity in Flux.

dbluxo:
I understand, good point. So I would prefer that the dependencies will be resolved as default. Having the possibility to turn it off via FluxHelmRelease.

squaremo:
[...]

PR #1450, opened 2018-10-16
Update chart dependencies before installing

squaremo:
This addresses #1278, for the "alpha" series (i.e., not for #1382). It simply calls `helm dep build` in the chart directory, when installing or upgrading a release.

Note that it bakes in a `repositories.yaml` file (which specifies the repo URLs, and is necessary for the `helm` commands to function). This file specifies _only_ the conventional "stable" repo. We will need to figure out how to allow other repos to be specified; it is not quite good enough to mount your own file, since it lives in a directory helm wants to write to (and the mount would mean it's read-only).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ SUDO := $(shell docker info > /dev/null 2> /dev/null || echo "sudo")
 TEST_FLAGS?=
 
 include docker/kubectl.version
+include docker/helm.version
 
 # NB because this outputs absolute file names, you have to be careful
 # if you're testing out the Makefile with `-W` (pretend a file is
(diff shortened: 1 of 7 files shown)


squaremo:
I need to make sure this can cover a couple of scenarios aside from the bare basics:

### Can you set it up to use a repo other than stable?

This will probably mean mounting a repositories.yaml file; but,

 - those tend to have local paths in them
 - .. and possibly refer to certificates which would also need to be present
 - .. and, Helm really wants this file to be in the same directory as where it's downloading charts, but if you're mounting it from a volume, it'll be read-only

### How do you provide credentials for the repo?

Credentials get put in the repositories.yaml file; so again, if you want to supply credentials, you'd have to mount that file from a secret, with the problems as above.

squaremo:
I've pushed the image quay.io/squaremo/helm-operator:pr-1450 for anyone who wants to try this out.

PR #1453, opened 2018-10-19
Configurable maximum item size for memcache

white-hat:
Setting `max_item_size` to `5m` fixes `SERVER_ERROR object too large for cache`  errors on large deployments

### DIFF
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -125,6 +125,8 @@ registry:
  
 memcached:
   verbose: false
+  max_item_size: 1m
+  max_memory: 64
 
 ssh:
   # Overrides for git over SSH. If you use your own git server, you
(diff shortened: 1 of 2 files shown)


squaremo:
Not that I doubt you ran into this, but wow that is a large amount of metadata for an image repo!

chart/flux/values.yaml
```diff
@@ -125,6 +125,8 @@ registry:
  
 memcached:
   verbose: false
+  max_item_size: 1m
+  max_memory: 64
```
stefanprodan: Please change the vars to maxItemSize and maxMemory

white-hat:
@stefanprodan please let me know if you'd like to have more changes to the PR

stefanprodan:
@white-hat that image that has 1M of metadata is it on a public registry?

PR #1455, opened 2018-10-21
[WIP] Add the ability to support authentication to AWS ECR 

### LINKED ISSUE #539: Support for AWS ECR?
Hi,

is there support for AWS ECR? Couldn't really find documentation or figure out how to get it working...
ECR requires to generate an authorization token which changes every 12h.

> Because the Docker CLI does not support the standard AWS authentication methods, you must authenticate your Docker client another way so that Amazon ECR knows who is requesting to push or pull an image. If you are using the Docker CLI, then use the docker login command to authenticate to an Amazon ECR registry with an authorization token that is provided by Amazon ECR and is valid for 12 hours.

http://docs.aws.amazon.com/AmazonECR/latest/userguide/Registries.html#registry_auth
http://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html

bzon:
This fixes #539. In order for this to work, the **fluxd** container must have the credentials
to access AWS API via API Keys or IAM roles.

**Sidecar integration approach:**

* Run the **ecr-sidecar** container alongside the **fluxd** container (flux Pod).. 
* The **ecr-sidecar**  must use `registry-ids` and `region` flag values which are required when invoking the get token request. 
* The **ecr-sidecar** container fetches the docker login token using [AWS ECR SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/#ECR.GetAuthorizationTokenRequest) with a configurable interval using `fetch-interval` flag and save it in memory.
* The **ecr-sidecar** container exposes a `/awsauth` endpoint which returns the authentication in JSON format.
  ```json
  {
    "auths": {
      "https://12345.dkr.ecr-use-east-1.amazonaws.com": {
        "auth": "QVdTOmVjcnBhc3N3b3JkCg=="
      },
      "https://602401143452.dkr.ecr.us-east-1.amazonaws.com": {
        "auth": "QVdTOnh4eHh4Cg=="
      }
    }
  }
   ```
* **WIP**: Now, when **fluxd** invoke the [credsFor](https://github.com/weaveworks/flux/blob/fd1cddced210acb94a0a005fcf0d75adc3c674ae/registry/credentials.go#L130) function we can add a new condition for AWS related docker repos:

  ```go
  if strings.HasSuffix(host, "amazonaws.com") {
    sidecar := newAWSSidecar()
    if cred, err := sidecar.GetAWSLoginCreds(host); err == nil {
      return cred
    }
  }
  ```
  
  This is not tested yet..

**Considerations:**

Initially, I thought it would be nicer if **GetAWSLoginCreds** function can be executed as a new goroutine during the **fluxd** runtime. But it was taking so much effort for me to understand the end to end flow of this demon. In my opinion, adding a sidecar made it much simpler for me to integrate this feature. We may be able to do the same for Azure authentication in the future.

**References:**

* https://docs.aws.amazon.com/cli/latest/reference/ecr/get-authorization-token.html
* https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/#ECR.GetAuthorizationTokenRequest
* https://github.com/bzon/ecr-k8s-secret-creator

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -41,6 +41,47 @@
   version = "v1.4.0"
 
 [[projects]]
+  digest = "1:eedc89853af7761550e4dbe0627858d8e3c5daf9e702a755cd8edc264fc3b1cf"
+  name = "github.com/aws/aws-sdk-go"
+  packages = [
+    "aws",
+    "aws/awserr",
+    "aws/awsutil",
+    "aws/client",
+    "aws/client/metadata",
+    "aws/corehandlers",
+    "aws/credentials",
+    "aws/credentials/ec2rolecreds",
+    "aws/credentials/endpointcreds",
+    "aws/credentials/stscreds",
+    "aws/csm",
+    "aws/defaults",
+    "aws/ec2metadata",
+    "aws/endpoints",
+    "aws/request",
+    "aws/session",
+    "aws/signer/v4",
+    "internal/ini",
+    "internal/sdkio",
+    "internal/sdkrand",
+    "internal/sdkuri",
+    "internal/shareddefaults",
+    "private/protocol",
+    "private/protocol/json/jsonutil",
+    "private/protocol/jsonrpc",
+    "private/protocol/query",
+    "private/protocol/query/queryutil",
+    "private/protocol/rest",
+    "private/protocol/xml/xmlutil",
+    "service/ecr",
+    "service/ecr/ecriface",
+    "service/sts",
+  ]
+  pruneopts = ""
+  revision = "c0037b52736b07cea4e0cb6521f7995adef833bc"
+  version = "v1.15.59"
+
+[[projects]]
   branch = "master"
   digest = "1:0c5485088ce274fac2e931c1b979f2619345097b39d91af3239977114adf0320"
   name = "github.com/beorn7/perks"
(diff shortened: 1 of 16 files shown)


squaremo:
Thanks for the clear explanation of your thinking in the post. I like the sidecar approach, all in all. It means less special-case stuff in the "core" code.

Can we segregate the AWS-specific bits even more? I am thinking

 - move registry/aws.go to its own package
 - make the opt-in explicit in an argument to fluxd, so you have to provide e.g., the sidecar address, or at least switch it on, and that registers the special case for `".amazonaws.com"`.

One complication this brings up is that it adds another image to be built, etc.. I don't expect it to change all that often, but it does mean more release machinery. We could consider either putting the sidecar in its own repo with releases on their own cadence. (Then I guess we'd need to vendor it to get the client shim. Blargh)

squaremo:
> We could consider either putting the sidecar in its own repo

Actually I am inclined to just version this along with the flux image. That means that occasionally, a change here will have to wait for a new flux release; and, there will be a bunch of releases of the sidecar image that don't actually change anything. But .. :man_shrugging:

bzon:
@squaremo Should we have a **sidecar** package? So we can have aws, azure, etc here in the future. I also like the switch idea via flags maybe? 

Btw, I need help on understanding how to test if flux uses the credentials fetched from the sidecar. I was able to verify that flux actually calls the ecr sidecar endpoint but it wasn't using it.

bzon:
Closing this PR for now. I feel like my approach is not the best way to implement flux authentication to ECR. We should probably first observe how the kubelet deals with ECR authentication here https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/aws/aws_credentials.go and see if we can reuse some of package methods built in to flux instead of creating a sidecar.

PR #1460, opened 2018-10-22
add `globsemver:` pattern

white-hat:
GlobSemver filter is a combination of Glob and Semver filters.
This filter is useful when you want to do glob matching and semantic versioning simultaneously.

For example, one of our development teams pushes several deployments a day, each deployment has to go through multiple CI stages before it finally gets to k8s cluster.
This sometimes triggers race condition -- for example, developer pushes `v1.2.5-rc` tag to CI, then he realizes he had made a logical mistake, fixes it and pushes `v1.2.6-rc` tag.
He expects that soon he will see `v1.2.6-rc` at k8s-staging cluster.
But because the CI node which was testing `v1.2.6-rc` was slower then the one which was testing  `v1.2.5-rc` -- Flux deploys `v1.2.6-rc` first, and then replaces it with `v1.2.5-rc` if we are using `glob:*-rc` filter

globSemver filter  `globsemver:v{1.x.x}-rc` could prevent such an issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ operator is in [./CHANGELOG-helmop.md](./CHANGELOG-helmop.md).
 
 ### Improvements
 
-- ..
+- Add `globsemver:` filter which is useful when you want to do glob matching and semantic versioning for image tags simultaneously.
 
 ## 1.7.1 (2018-09-26)
 
@@ -478,4 +478,3 @@ Initial semver release.
 -   Refactored registry code and improved coverage
 
 See https://github.com/weaveworks/flux/milestone/7?closed=1 for full details.
-
--- a/policy/pattern.go
+++ b/policy/pattern.go
@@ -1,17 +1,20 @@
 package policy
 
 import (
+	"errors"
+	"regexp"
+	"strings"
+
 	"github.com/Masterminds/semver"
 	"github.com/ryanuber/go-glob"
 	"github.com/weaveworks/flux/image"
-	"strings"
-	"regexp"
 )
 
 const (
-	globPrefix   = "glob:"
-	semverPrefix = "semver:"
-	regexpPrefix = "regexp:"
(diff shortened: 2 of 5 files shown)


white-hat:
It's my first commit in Golang, so comments are welcome :)

rndstr:
Hi @white-hat thanks for the contribution, the Go code looks great.

As for your use case, semver has the concept of 'pre-release versions'. The Go library we are using goes a bit into more detail what exactly is matched by constraints: https://github.com/Masterminds/semver#working-with-pre-release-versions

> SemVer comparisons without a pre-release comparator will skip pre-release versions.

So if you include a pre-release in the semver constraint, it will match your examples:
`semver:1.x.x-0` (which is the same as `>=1.0.0-0`) is the same as your suggested `globsemver:v{1.x.x}-rc`. (Also note that the `v` prefix is accepted/ignored by semver)

white-hat:
@rndstr thank you for the explanation. So the only benefit of `globsemver` is that you can have >2 environments to deploy to.
For example, a developer wants to deploy `v1.0.0` first to `dev` environment, then to `stg` and then to `prd`.
Using `v1.0.0-0` and semver allows him to have such a "flag" to distinguish between two environments (`stg`/`prd`), but he cannot deploy to(`dev`).
But I don't think it's worth complication of app anyway, closing the PR

zeeZ:
I think there are valid use cases, for example where semver is only kind of followed.

As an example, what if I want the latest version of redis, but only the alpine images?

* `globsemver:{~5}-alpine`: 5.0.4-alpine
* `semver:~5-alpine`: 5.0.4 > 5.0.4-stretch > 5.0.4-alpine > 5.0.4-32bit if we're unlucky
* `glob:5.*-alpine`: whichever 5.*-alpine image was pushed last

If you want to spin this further, `globsemver:{~5}-alpine{~3}`...

PR #1462, opened 2018-10-23
Add prometheus annotation to Helm chart

### LINKED ISSUE #1461: Add prometheus annotation for Helm chart
We're using the Helm chart to deploy Flux, and want to let Prometheus scrape the metrics endpoint. Right now I'm annotating the pod with `kubectl annotate ...` but ideally there should be an option to add it to the deployment. 

I'd like to either add `prometheus.io/scrape: "true"` to the chart manifest by default, or enable it through `values.yml`. I'd happily submit a PR if you think it's worthwhile.

marcusolsson:
This change adds the `prometheus.io.scrape: "true"` annotation in order for prometheus to start collecting metrics from Flux.

This fixes #1461.

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -15,8 +15,10 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
+      {{- if .Values.prometheus.enabled }}
       annotations:
         prometheus.io.scrape: "true"
+      {{- end }}
       labels:
         app: {{ template "flux.name" . }}
         release: {{ .Release.Name }}
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -15,6 +15,8 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
+      annotations:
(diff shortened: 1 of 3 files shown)


squaremo:
I wonder would it be useful to make it parameterised? Perhaps some people will want to switch scraping off.

squaremo:
First-time contributor high five :hand: by the way!

marcusolsson:
Thanks! I don't mind either way. I imagine the `values.yml` would look something like this then:

```
prometheus:
    enabled: true    # adds prometheus.io.scrape
    path: "/metrics" # adds prometheus.io.path
    port: 3031       # adds promethues.io.port
```

squaremo:
Yeah, something like that, with a matching set for the helm operator. It gets complicated quickly, doesn't it. Let's limit the scope of this PR to just scrape `true` / `false` / omitted -- on because that's what you originally wanted, off for symmetry, and omitted (the default) because that preserves the current behaviour.

marcusolsson:
Sounds good! 👍 I'll fix it tomorrow morning

squaremo:
Looks good to me. wdyt @stefanprodan ?

marcusolsson:
Thanks!

PR #1463, opened 2018-10-23
Uniform language highlight on helm chart README

sfrique:
This is just a formatting change, removed some white spaces and changed the language from console/bash to sh on all code blocks.


dholbach:
Thanks for the PR!

PR #1470, opened 2018-10-25
Helm Chart: Flux 1.8.0 and helm-op 0.3.0

stefanprodan:
- Bump Flux version to 1.8.0
- Bump Flux Helm Operator version to 0.3.0
- Add Flux Helm Operator update chart deps flag to values.yaml


squaremo:
`--registry-cache-expiry` is deprecated; people may still want to use it with older versions of fluxd though. If you don't supply it, will it be omitted?

PR #1472, opened 2018-10-26
Add events for container release

lelenanam:
For `ReleaseEventMetadata` use spec with both types: `ReleaseImageSpec` and `ReleaseContainersSpec`.

Need this for https://github.com/weaveworks/service/issues/2277

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -95,10 +95,12 @@ func (e Event) String() string {
 		if len(strImageIDs) == 0 {
 			strImageIDs = []string{"no image changes"}
 		}
-		for _, spec := range metadata.Spec.ServiceSpecs {
-			if spec == update.ResourceSpecAll {
-				strServiceIDs = []string{"all services"}
-				break
+		if metadata.Spec.Type == ReleaseImageSpecType {
+			for _, spec := range metadata.Spec.ReleaseImageSpec.ServiceSpecs {
+				if spec == update.ResourceSpecAll {
+					strServiceIDs = []string{"all services"}
+					break
+				}
 			}
 		}
 		if len(strServiceIDs) == 0 {
@@ -239,11 +241,27 @@ type ReleaseEventCommon struct {
 	Error string `json:"error,omitempty"`
 }
 
+const (
+	// ReleaseImageSpecType is a type of release spec when there are update.Images
+	ReleaseImageSpecType = "releaseImageSpecType"
+	// ReleaseContainerSpecType is a type of release spec when there are update.Containers
+	ReleaseContainerSpecType = "releaseContainerSpecType"
+)
+
+// ReleaseSpec is a spec for images and containers release
+type ReleaseSpec struct {
+	// Type is ReleaseImagesSpecType or ReleaseContainersSpecType
+	// if empty (for previous version), then use ReleaseImagesSpecType
+	Type                 string
+	ReleaseImageSpec     *update.ReleaseImageSpec
+	ReleaseContainerSpec *update.ReleaseContainersSpec
+}
+
 // ReleaseEventMetadata is the metadata for when service(s) are released
 type ReleaseEventMetadata struct {
 	ReleaseEventCommon
-	Spec  update.ReleaseSpec `json:"spec"`
-	Cause update.Cause       `json:"cause"`
+	Spec  ReleaseSpec  `json:"spec"`
+	Cause update.Cause `json:"cause"`
 }
 
 // AutoReleaseEventMetadata is for when service(s) are released
--- a/event/event.go
+++ b/event/event.go
@@ -95,7 +95,7 @@ func (e Event) String() string {
 		if len(strImageIDs) == 0 {
 			strImageIDs = []string{"no image changes"}
 		}
-		if metadata.Spec.Type == ReleaseImageSpecType {
+		if metadata.Spec.Type == "" || metadata.Spec.Type == ReleaseImageSpecType {
 			for _, spec := range metadata.Spec.ReleaseImageSpec.ServiceSpecs {
 				if spec == update.ResourceSpecAll {
 					strServiceIDs = []string{"all services"}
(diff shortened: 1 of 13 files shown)


event/event.go
```diff
@@ -239,11 +241,27 @@ type ReleaseEventCommon struct {
...
+	ReleaseImageSpec      *update.ReleaseImageSpec
+	ReleaseContainersSpec *update.ReleaseContainersSpec
+}
+
 // ReleaseEventMetadata is the metadata for when service(s) are released
 type ReleaseEventMetadata struct {
 	ReleaseEventCommon
-	Spec  update.ReleaseSpec `json:"spec"`
-	Cause update.Cause       `json:"cause"`
+	Spec  ReleaseSpec  `json:"spec"`
```
squaremo: I am not convinced that this will parse old serialisations of this struct. My reasoning goes like this:

This field used to contain what is now `update.ReleaseImageSpec`, so would get serialised as

```
{
  "revision": ..., "result", ..., "error", ..., // from ReleaseEventCommon
  "spec": {
    "serviceSpecs": ...,
    "imageSpec": ...,
    // and so on, from update.ReleaseSpec
  },
```

Now, the equivalent value would serialise as

```
{
  "revision", // etc. from ReleaseEventCommon
  "spec": {
    "type": "releaseImageSpecType",
    "releaseImageSpec": {
      "serviceSpecs": ..., // etc.
    },
```

and, a value of the first kind would now be decoded as

```
ReleaseEventMetadata{
  ReleaseEventCommon: ReleaseEventCommon{Revision: ...etc},
  Spec: ReleaseSpec{Type: "", ReleaseImageSpec: nil, ReleaseContainerSpec: nil},
}
```

I _think_ the tweak needed is to make `ReleaseImageSpec` and `ReleaseContainerSpec` both embedded, i.e,

```
type ReleaseSpec struct {
   Type string
    *update.ReleaseImageSpec
    *update.ReleaseContainerSpec
}
```

I may have this wrong of course -- the JSON codec often trips me up. The way to check (or demonstrate that I'm wrong!) would be to take a value serialised from the code in master, and decode it with the code here -- as a test, even.

lelenanam:
Right, but if I use embedded structs there is another issue here: `json.Marshal` skips embedded fields with the same names like `Kind` and `Force` (ambiguous selector?):
```
type ReleaseImageSpec struct {
    ServiceSpecs    []ResourceSpec
    ImageSpec    ImageSpec
    Kind        ReleaseKind
    Excludes    []flux.ResourceID
    Force        bool
}

type ReleaseContainersSpec struct {
    Kind        ReleaseKind
    ContainerSpecs    map[flux.ResourceID][]ContainerUpdate
    SkipMismatches    bool
    Force        bool
}
```
So now I tend to use a separate type for `ReleaseContainersSpec`, like:
```
type ReleaseContainersEventMetadata struct {
	ReleaseEventCommon
	Spec  update.ReleaseContainersSpec `json:"spec"`
	Cause update.Cause                 `json:"cause"`
}
```

squaremo:
> `json.Marshal` skips embedded fields with the same names like `Kind` and `Force` (ambiguous selector?):

Argh, I knew there was a catch.

How about something like

```
func (s *ReleaseSpec) UnmarshalJSON([]byte) error {
  alias T ReleaseSpec
  t := (*T)(s)
  if err := json.Unmarshal(b, t); er != nil {
    return err
  }

  switch t.Type {
    case "":
      // it's an old-style inline update.ReleaseSpec
      var r update.ReleaseSpec
      json.Unmarshal(b, &r)
      s.ReleaseImageSpec = &r
      s.Type = ReleaseImageSpecType
    case ReleaseContainersSpecType, ReleaseImageSpecType:
      // great, we got a new style value
    default:
      return errors.New("unknown type")
  }
  return nil
}
```

squaremo:
> and then try to unmarshal it, I have `runtime: goroutine stack exceeds 1000000000-byte limit fatal error: stack overflow` during unmarshal because of recursion calls in `UnmarshalJSON`.

For posterity: the purpose of the type alias is to create a type that has the same fields (and can therefore be unmarshalled into) but not the `UnmarshalJSON` method; it's the latter that leads to stack overflow.

event/event.go
```diff
@@ -239,11 +241,59 @@ type ReleaseEventCommon struct {
...
+	Type                  string
+	ReleaseImageSpec      *update.ReleaseImageSpec
+	ReleaseContainersSpec *update.ReleaseContainersSpec
+}
+
+// UnmarshalJSON for old version of spec (update.ReleaseImageSpec) where Type is empty
+func (s *ReleaseSpec) UnmarshalJSON(b []byte) error {
+	type T ReleaseSpec
+	t := (*T)(s)
+	if err := json.Unmarshal(b, &t); err != nil {
```
squaremo: `t` is already a pointer `*T`, so you don't need to pass its address.

event/event.go
```diff
@@ -239,11 +241,59 @@ type ReleaseEventCommon struct {
...
+		var r update.ReleaseImageSpec
+		if err := json.Unmarshal(b, &r); err != nil {
+			return err
+		}
+		s.ReleaseImageSpec = &r
+		s.Type = ReleaseImageSpecType
+
+	case ReleaseImageSpecType, ReleaseContainersSpecType:
+		r := (*T)(s)
+		if err := json.Unmarshal(b, &r); err != nil {
```
squaremo: `r` is already a pointer; but moreover, this decoding has already been done, in line 264. I think this case is a no-op -- all the fields will have been decoded. (Is there a test that doesn't pass unless this line is present?)

event/event.go
lelenanam: Fixed.

event/event.go
lelenanam: Fixed.

event/event.go
```diff
@@ -239,11 +241,55 @@ type ReleaseEventCommon struct {
...
+	case "":
+		r := &update.ReleaseImageSpec{}
+		if err := json.Unmarshal(b, r); err != nil {
+			return err
+		}
+		s.Type = ReleaseImageSpecType
+		s.ReleaseImageSpec = r
+
+	case ReleaseImageSpecType, ReleaseContainersSpecType:
+		s.Type = t.Type
```
squaremo: This is tautological, since `t === s` :-) But it does make clear what's going on, and will probably be eliminated by the compiler.

PR #1473, opened 2018-10-26
Create README-TR.md

aytekk:
Turkish version of README.md

### DIFF
--- a/README-TR.md
+++ b/README-TR.md
@@ -1,0 +1,113 @@
+Akı
+
+GitOps'a inanıyoruz:
+
+Sisteminizin istediğiniz halini deklare olarak tanımlayın. Buna uygulamalar, yapılandırma, kontrol panelleri, izleme ve diğer her şey dahildir.
+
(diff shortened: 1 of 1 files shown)


dholbach:
Merhaba Meryem! Thanks a lot for the translation into Turkish. This is great.

Unfortunately Flux is not really set up yet to be a localised project. Basically everything is in English: the code, the README you translated, the docs in the [site directory](https://github.com/weaveworks/flux/tree/master/site) and elsewhere, discussions in our Slack, etc.

I'm not sure how others feel about it, but my gut feeling is we're not at a stage right now, where we could easily translate our docs and support them on an ongoing basis, so that if things change, we make sure that everything else is updated at the same time.

Was there demand for a Turkish translation of Flux's docs? Are you part of any Turkish Cloud Native group by any chance?

Another thing that's missing from your Turkish version of the README is markup. Basically all links and formatting are lost in there.

I hope what I said above is not too discouraging. I appreciate your effort and enthusiasm, but would like to see us think this through before creating localised resources. What do you think?

PR #1474, opened 2018-10-26
making compiling work on raspberry pi's

MansM:
Making code compile on my raspberries, will not have any effect on arm64

makefile can be changed as well, but I am not that big of an Makefile expert ( sed -i "s/amd64/arm/g" Makefile did work for me)

fixes: https://github.com/weaveworks/flux/issues/1411


hiddeco:
Thanks a lot for the PR! :1st_place_medal: 

The relevant lines in the `Makefile` are https://github.com/weaveworks/flux/blob/34b2e99b7ba5af18a7e00623789a771bae2410b7/Makefile#L26-L31

The required change is to rewrite this so that an additional `GOOS=$$os GOARCH=$$arch` is run with `$$arch=arm` and `$$os=linux`.

The current for loop structure makes this kind of hard to do, but instead you could do something like

```sh
	for platform in linux/amd64 linux/arm darwin/amd64 windows/amd64; do \
		split=($${platform//\// }); os=$${split[0]}; arch=$${split[1]}; \
		CGO_ENABLED=0 GOOS=$$os GOARCH=$$arch go build -o "build/fluxctl_"$$os"_$$arch" $(LDFLAGS) -ldflags "-X main.version=$(shell ./docker/image-tag)" ./cmd/fluxctl/; \
	done
```

squaremo:
That stanza in the Makefile is for building `fluxctl`, but the code change is in _fluxd_.
(I'm not against that change to the Makefile -- it's clear enough.)

@MansM Have you got to the point of building an image and running it? Or is it sufficient for you to be able to compile the fluxd binary?

I'm happy to accept this code change, if a binary is all you need, so long as we make sure it compiles somewhere in the CI script, so it doesn't rot.

squaremo:
> That stanza in the Makefile is for building `fluxctl`, but the code change is in _fluxd_.

That said, we could do a similar thing for fluxd, if a binary is the desired artefact.

MansM:
Fluxd runs perfectly fine in my pi kube cluster.
Used this on pi3 with hypriot to build the image: 
https://github.com/MansM/vagrant-kubeadm/tree/master/flux-builder
Image can be found: https://hub.docker.com/r/mansm/flux/tags/

hiddeco:
> That stanza in the Makefile is for building fluxctl, but the code change is in fluxd.
(I'm not against that change to the Makefile -- it's clear enough.)

:man_facepalming: I need to stop reviewing stuff before lunch.

---

That being said I think we can do two things: we can either compile both versions for the Flux daemon (which I think would be great) or we can make the `GOARCH` configurable (probably with `amd64` as a default).

MansM:
To be honest, I highly doubt anyone would use an rpi a jumphost, so I think at least creating the fluxd image and/or binary would be plenty

MansM:
During some rainy cycling I thought of the following.
Only the binaries can be build with the go cross compiler, or does circle ci supports building arm images on their platform (as RUN commands exists in the Dockerfile)

squaremo:
> Only the binaries can be build with the go cross compiler, or does circle ci supports building arm images on their platform (as RUN commands exists in the Dockerfile)

Good question! This hacks suggests it might be possible: https://blog.hypriot.com/post/setup-simple-ci-pipeline-for-arm-images/

If you just need the binaries, and you're happy doing the rest yourself for now, let's just make sure they are compiled somewhere in the CI script. I'm OK with having a Makefile target specifically for checking that cross-compilation works.

MansM:
I only need the images, I would be fine with merging this now and keep doing the search and replace for now on the Makefile.

PR #1476, opened 2018-10-26
Fixing link to chart's README

endrec:
Fixed the link. I do not think this worth a Changelog entry, but I'm happy to add one, if you insist.


squaremo:
> I do not think this worth a Changelog entry

Me neither, it's good to go :-)

PR #1480, opened 2018-10-29
Soften the changelog requirement for PRs

squaremo:
I'd rather people wrote their suggestion for the changelog in the PR
description, rather than including it in the commits. I end up
looking through the PRs anyway, and this way we don't have to keep a
"staging area" for the next release in the changelog.


squaremo:
Cool, thanks Daniel, thanks Hidde.

PR #1486, opened 2018-11-01
Changelog entry for Helm Op 0.4.0

squaremo:


squaremo:
(I should add the Thanks section too)

CHANGELOG-helmop.md
```diff
@@ -1,3 +1,40 @@
...
+`FluxHelmRelease` resources by attaching Kubernetes secrets. This
+helps with a few use cases:
+
+ - supplying the same default values to several releases
+ - providing secrets (e.g., a password) to a chart that expects them as values
+ - using values files without inlining them into FluxHelmReleases
+
+### Bug fixes
+
+ - Make `--tiller-tls-verify` work as intended, by giving better
```
hiddeco: Can you add a note about the Tiller version requirement (should be at least `v2.10.0`).

During testing I was able to talk to a lower Tiller version but this is because the Go client does not [check compatibility](https://github.com/helm/helm/blob/master/pkg/version/compatible.go#L40) like the Helm CLI does.

squaremo: Is it accurate to say that Tiller 2.10 is the first that properly supports TLS verify?

hiddeco: No, the `--tls-hostname` flag is introduced in this version.

When you set the `--tls-hostname` instead of validating the CA against the hostname you are connecting with it is validated against the value of this flag. So say you generated a certificate with a very generic hostname (e.g. `tiller`, you can set `--tls-hostname=tiller` and `--tls-verify` will pass although you are connecting with `tiller-deploy.kube-system`.

squaremo: Got it -- I have added a statement about Tiller compatibility, see what you think.

squaremo:
Thanks Hidde :spaghetti:

PR #1489, opened 2018-11-02
re-flow GKE note in get-started doc

dholbach:
The note on GKE somewhat interrupted the flow, re-flow parts of the doc to make things clearer.


dholbach:
Thanks for the review @hiddeco!

PR #1490, opened 2018-11-02
Remove note about --bootstrapper=kubeadm.

dholbach:
Apparently `kubeadm` is the default bootstrapper in `minikube` forabout 11 months now, let's remove the note:

  https://github.com/kubernetes/minikube/blame/ae9f4b20c7fcb7a89fc65e6eab0fb716f4d1a4ed/pkg/minikube/constants/constants.go#L119


dholbach:
Thanks for the review @hiddeco!

PR #1507, opened 2018-11-13
Bump to kubeyaml 0.5.1 to fix (F)HR updating

### LINKED ISSUE #1504: Why is image not updated?
I set up a pod with a deployement yaml file like so:
```
    annotations:
        flux.weave.works/tag.container-name: "master-*"
    spec:
      containers:
      - name: k8s-mon
        image: index.docker.io/larktech/k8s-mon:latest
```

I don't need to do anything more than deploy with the annotations to get it to update when a new docker image appears with a name like `master-*`, right? 
 
Is it a problem that I'm starting from the `latest` tag? If I put in `master-nonexistent` and push that to the git repo, will it fail at first and then see that it's not the latest and update it appropriately? Or do I have to create a `master-latest` tag for that first installation?

squaremo:
kubeyaml 0.5.1 has a fix for updating images that include ports in
(Flux)HelmRelease manifests. So: bump that in the Dockerfile and shim
(`/bin/kubeyaml`), and add a test that it does account for such images
when updating.

Side issue in update_test.go: calling `t.Parallel()` _before_ invoking
the test case meant that the closed-over loop variable could be
reassigned before the test case was run.

Fixes #1504.

### DIFF
--- a/bin/kubeyaml
+++ b/bin/kubeyaml
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker run --rm -i quay.io/squaremo/kubeyaml:0.5.0 "$@"
+docker run --rm -i quay.io/squaremo/kubeyaml:0.5.1 "$@"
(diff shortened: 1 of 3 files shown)


squaremo:
Roli, do you mind having a look at this? Checking that the additional test actually tests something that could (and previously, did) fail would be adequate I reckon.

squaremo:
Thanks @rndstr :shaved_ice:

PR #1515, opened 2018-11-14
Release 1.8.1 changelog

squaremo:
This includes a fast-forward merge from master, since I want to release everything from there anyway.

The salient change is the last commit, the changelog entry.


squaremo:
Re-merged master, so that we get Elena's fix.

PR #1520, opened 2018-11-15
Housekeeping: merge release branch 1.8.x back into master

squaremo:
Bring the changelog up to date by merging 1.8.x (from where releases are made) back into master.


squaremo:
Thanks Hidde (and for the other PRs today)!
For these ones, maybe we could squash the commits, since almost all the time they will be just the changelog commit and one or two cherry-picks -- wdyt?

Issue #1424, opened 2018-10-04
Show rollout progress in `fluxctl release`


squaremo:
As of #1380 and related PRs, flux reports the pod counts (ready, available, etc.) and any conditions indicating a rollout error. It would be great to have an option to `fluxctl release` watch the rollout happen -- for peace of mind, at the least.

I believe this can be done statelessly -- i.e., you figure out which workloads you need to watch, then call ListControllers for those workloads, and report based on what it says, until all the workloads are in a terminal state (either rolled out, or stuck).

squaremo:
Hacktoberfest notes: this is non-trivial, but doesn't need deep knowledge of flux's internals.

kingdonb:
No further enhancements can land in Flux v1,

* #3320

PR #1525, opened 2018-11-16
Add `--watch` option to release command

iandotmartin:
Adds a flag to `fluxctl release` to observe rollout progress during the release. Shows the numbers of pods desired, up-to-date, available, and outdated.

Note: I wasn't sure if it made sense to add a case somewhere in `release_cmd_test` for this flag, since there's no change to the release spec. Happy to add one if desired! 

Addresses #1424

### DIFF
--- a/cmd/fluxctl/release_cmd.go
+++ b/cmd/fluxctl/release_cmd.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/spf13/cobra"
 
 	"github.com/weaveworks/flux"
+	"github.com/weaveworks/flux/api/v11"
 	"github.com/weaveworks/flux/job"
 	"github.com/weaveworks/flux/update"
 )
@@ -23,6 +25,7 @@ type controllerReleaseOpts struct {
 	dryRun         bool
 	interactive    bool
 	force          bool
+	watch          bool
 	outputOpts
 	cause update.Cause
 
@@ -57,6 +60,7 @@ func (opts *controllerReleaseOpts) Command() *cobra.Command {
 	cmd.Flags().BoolVar(&opts.dryRun, "dry-run", false, "Do not release anything; just report back what would have been done")
 	cmd.Flags().BoolVar(&opts.interactive, "interactive", false, "Select interactively which containers to update")
 	cmd.Flags().BoolVarP(&opts.force, "force", "f", false, "Disregard locks and container image filters (has no effect when used with --all or --update-all-images)")
+	cmd.Flags().BoolVarP(&opts.watch, "watch", "w", false, "Watch rollout progress during release")
 
 	// Deprecated
 	cmd.Flags().StringSliceVarP(&opts.services, "service", "s", []string{}, "Service to release")
@@ -89,7 +93,10 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 		fmt.Fprintf(cmd.OutOrStderr(), "Warning: --force will not ignore container image tags when used with --update-all-images\n")
 	}
 
-	var controllers []update.ResourceSpec
+	var (
+		controllers   []update.ResourceSpec
+		controllerIds []flux.ResourceID
+	)
 	if opts.allControllers {
 		controllers = []update.ResourceSpec{update.ResourceSpecAll}
 	} else {
@@ -98,6 +105,7 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 			if err != nil {
 				return err
 			}
+			controllerIds = append(controllerIds, id)
 			controllers = append(controllers, update.MakeResourceSpec(id))
 		}
 	}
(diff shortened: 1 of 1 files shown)


rndstr:
Had a look and it worked great, thanks @iandotmartin!

I'm thinking about how to display the data. It's quite hard to parse for me now and gauge any progress:
```
CONTROLLER                CONTAINER  IMAGE            RELEASE   DESIRED  UP-TO-DATE  AVAILABLE  OUTDATED
default:deployment/nginx  nginx      nginx:1.14-perl  updating  3        2           2          2
```

What about having the column header say `Replicas`, and the cell value:
`<available>/<desired>`
if you pass in `--verbose` it expands 
`<available>/<desired> (<outdated> outdated, <updated> updated)`
?

That would then look like
```
CONTROLLER                CONTAINER  IMAGE            RELEASE   REPLICAS
default:deployment/nginx  nginx      nginx:1.14-perl  updating  2/3 (2 outdated, 2 updated)
```

cmd/fluxctl/release_cmd.go
```diff
@@ -179,6 +187,53 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 
 		opts.dryRun = false
 	}
+	if opts.watch {
+		await(ctx, cmd.OutOrStdout(), cmd.OutOrStderr(), opts.API, jobID, !opts.dryRun, opts.verbosity)
+
+		fmt.Fprintf(cmd.OutOrStderr(), "Monitoring rollout...\n")
```
rndstr: In other places we have whitespace before the triple dots, looks a bit off when appearing after each other
```
Submitting release ...
Monitoring rollout...
```

cmd/fluxctl/release_cmd.go
```diff
@@ -179,6 +187,53 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
 
 		opts.dryRun = false
 	}
+	if opts.watch {
+		await(ctx, cmd.OutOrStdout(), cmd.OutOrStderr(), opts.API, jobID, !opts.dryRun, opts.verbosity)
```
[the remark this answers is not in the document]
rndstr: which i think will prevent the list of services being displayed even though there are no changes.

cmd/fluxctl/release_cmd.go
iandotmartin: at the moment both styles are used in the command - i'll update so that there's always a whitespace before the dots.

iandotmartin:
@squaremo @rndstr thanks for the reviews! this is ready for another look when you get a chance :)

squaremo:
> @squaremo @rndstr thanks for the reviews! this is ready for another look when you get a chance :)

:eyes:

squaremo:
@iandotmartin Please consider cherry-picking https://github.com/weaveworks/flux/commit/ae29a97acd95c0f6b345e6a489ae06f328687684, which I made to tweak how this works. Main thing it addresses: it needs to wait for the commit to be synced before moving on to show rollout progress, otherwise it will (usually) succeed immediately.

squaremo:
@rndstr Can you advise on the best format for showing ready and updated pods?

iandotmartin:
@squaremo cherry-pick'd

cmd/fluxctl/release_cmd.go
```diff
@@ -179,7 +187,68 @@ func (opts *controllerReleaseOpts) RunE(cmd *cobra.Command, args []string) error
...
+		fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%d/%d", service.ID, c.Name, c.Current.ID, service.Status, service.Rollout.Updated, service.Rollout.Desired)
+		if verbosity > 0 {
+			fmt.Fprintf(w, " (%d outdated, %d updated)", service.Rollout.Outdated, service.Rollout.Updated)
+		}
+		fmt.Fprintf(w, "\n")
+		for _, c := range service.Containers[1:] {
+			fmt.Fprintf(w, "\t%s\t%s\t\t\n", c.Name, c.Current.ID)
+		}
+	} else {
+		fmt.Fprintf(w, "%s\t\t\t%s\t%d/%d", service.ID, service.Status, service.Rollout.Ready, service.Rollout.Desired)
```
rndstr: should use `Updated` here too instead of `Ready`.

squaremo: Thanks Roli :tomato:

iandotmartin:
@rndstr yep, seems reasonable! made those updates.

yinzara:
I tried using this with the helm operator and it seems to not work.  For example, I have a custom git helm chart called "openapi-web" that I'm deploying to my cluster via the HelmRelease CRD and the helm-operator.  When I disable automation and use "fluxctl release --all --update-all-images" I get:
CONTROLLER                                               CONTAINER   IMAGE                                                           RELEASE   REPLICAS
feature-fluxwatch:helmrelease/feature-fluxwatch-openapi-web  deployment  us.gcr.io/my-project/openapi-web:feature-fluxwatch-0c9ac47  DEPLOYED  0/0

Repeated over and over again and it never completes.

MilanDasek:
I have exactly same problem. --watch never ends when automation disabled

```
  annotations:
     flux.weave.works/automated: "false"

```

BTW:
Is there any way/solution how to be sure whole set of charts/versions was successful?

PR #1526, opened 2018-11-17
Skip TLS check when using insecure host

### LINKED ISSUE #1497: Insecure registries behave differently than Docker Daemon
Currently, passing flux an insecure registry only toggles the HTTP request to use `http` over `https`.

It also needs to pass `InsecureSkipVerify` to the http client to ignore invalid SSL certificates (e.g. self signed).

ref: `registry/client_factory.go`

Timer:
Fixes #1497

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -106,7 +106,7 @@ func main() {
 		registryRPS          = fs.Float64("registry-rps", 50, "maximum registry requests per second per host")
 		registryBurst        = fs.Int("registry-burst", defaultRemoteConnections, "maximum number of warmer connections to remote and memcache")
 		registryTrace        = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
-		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "let these registry hosts skip TLS host verification, and fall back to HTTP (without basic auth) instead of HTTPS; this allows man-in-the-middle attacks, so use with extreme caution")
+		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "let these registry hosts skip TLS host verification and fall back to using HTTP instead of HTTPS; this allows man-in-the-middle attacks, so use with extreme caution")
 		registryExcludeImage = fs.StringSlice("registry-exclude-image", []string{"k8s.gcr.io/*"}, "do not scan images that match these glob expressions; the default is to exclude the 'k8s.gcr.io/*' images")
 
 		// AWS authentication
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -102,13 +102,12 @@ func main() {
 		memcachedTimeout  = fs.Duration("memcached-timeout", time.Second, "maximum time to wait before giving up on memcached requests.")
 		memcachedService  = fs.String("memcached-service", "memcached", "SRV service used to discover memcache servers.")
 
-		registryPollInterval       = fs.Duration("registry-poll-interval", 5*time.Minute, "period at which to check for updated images")
-		registryRPS                = fs.Float64("registry-rps", 50, "maximum registry requests per second per host")
-		registryBurst              = fs.Int("registry-burst", defaultRemoteConnections, "maximum number of warmer connections to remote and memcache")
-		registryTrace              = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
-		registryInsecure           = fs.StringSlice("registry-insecure-host", []string{}, "use HTTP for this image registry domain (e.g., registry.cluster.local), instead of HTTPS")
-		registryInsecureSkipVerify = fs.Bool("registry-tls-insecure-skip-verify", false, "skip TLS host certificate verification when the registry is listed as an insecure host")
-		registryExcludeImage       = fs.StringSlice("registry-exclude-image", []string{"k8s.gcr.io/*"}, "do not scan images that match these glob expressions; the default is to exclude the 'k8s.gcr.io/*' images")
+		registryPollInterval = fs.Duration("registry-poll-interval", 5*time.Minute, "period at which to check for updated images")
+		registryRPS          = fs.Float64("registry-rps", 50, "maximum registry requests per second per host")
+		registryBurst        = fs.Int("registry-burst", defaultRemoteConnections, "maximum number of warmer connections to remote and memcache")
+		registryTrace        = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
+		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "let these registry hosts skip TLS host verification, and fall back to HTTP (without basic auth) instead of HTTPS; this allows man-in-the-middle attacks, so use with extreme caution")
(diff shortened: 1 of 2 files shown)


squaremo:
@Timer I added that flag -- mind seeing if it works for your setup, when you have a sec?

I was thinking maybe the _right_ thing to do is to act like docker, that is, for an insecure host:

 - try HTTPS, but with TLS insecure skip verify
 - if that fails, use HTTP but don't send credentials

I wonder if that would break some private registry setups ...

donifer:
Any updates on this? I _think_ I'm hitting this issue using an insecure registry when polling images.
>err="Get https://registry.internal.mydomain.com/v2/: x509: certificate is valid for ingress.local, not registry.internal.mydomain.com

squaremo:
@donifer Mayyyybe -- that looks more like you're using a registry host _with HTTPS_, and it has an invalid certificate. That's a closely related problem, but not solved by this PR as it stands -- it would need the alternate formulation I posted above (that is: for insecure hosts, try https:// with TLS_INSECURE_SKIP_VERIFY first, and fall back to HTTP if that doesn't connect).

Nonetheless, I've pushed a build from this branch to quay.io/squaremo/flux:insecure-host-behavior-ad833b9, if you want to try it out. (NB you need to supply the new flag to fluxd: `--registry-tls-insecure-skip-verify`)

cazzoo:
Looking to this fix

squaremo:
@Timer @donifer @cazzoo I've implemented this idea:

> - try HTTPS, but with TLS insecure skip verify
> - if that fails, use HTTP but don't send credentials

i.e., to use your own registry, you now have the option of using a self-signed certificate (or HTTP but no authentication!) so long as you supply the hostname with `--registry-insecure-host`.

If you have the chance to test this out in a controlled environment, I would appreciate feedback.

registry/client_factory.go
```diff
@@ -98,9 +123,14 @@ func (f *RemoteClientFactory) ClientFor(repo image.CanonicalName, creds Credenti
 		f.Logger.Log("repo", repo.String(), "auth", cred.String(), "api", registryURL.String())
 	}
 
-	tokenHandler := auth.NewTokenHandler(tx, &store{cred}, repo.Image, "pull")
-	basicauthHandler := auth.NewBasicHandler(&store{cred})
-	tx = transport.NewTransport(tx, auth.NewAuthorizer(manager, tokenHandler, basicauthHandler))
+	authHandlers := []auth.AuthenticationHandler{}
+	// only send creds over HTTPS
+	if registryURL.Scheme == "https" {
```
2opremio: I think this is too restrictive and opinionated. For instance, one may intentionally want to use http in a sealed environment in which the registry just happens to require credentials for reasons beyond your control.

Also, it is almost equally unsafe to send the creds over HTTPS when the certificate verification is being skipped

registry/client_factory.go
```diff
@@ -40,7 +45,26 @@ func (t *logging) RoundTrip(req *http.Request) (*http.Response, error) {
...
 
 func (f *RemoteClientFactory) ClientFor(repo image.CanonicalName, creds Credentials) (Client, error) {
-	tx := f.Limiters.RoundTripper(http.DefaultTransport, repo.Domain)
+	insecure := false
+	for _, h := range f.InsecureHosts {
+		if repo.Domain == h {
+			insecure = true
+			break
+		}
+	}
```
2opremio: The size of this function is getting out of hand. I would suggest factoring out the challenge obtention (or part of it at least)

registry/client_factory.go
```diff
@@ -52,22 +76,17 @@ func (f *RemoteClientFactory) ClientFor(repo image.CanonicalName, creds Credenti
...
+		Scheme: "https",
 		Host:   repo.Domain,
 		Path:   "/v2/",
 	}
 
 	// Before we know how to authorise, need to establish which
 	// authorisation challenges the host will send. See if we've been
 	// here before.
+	attemptInsecureFallback := insecure
+attempt:
```
2opremio: Maybe make the label more descriptive? (e.g. `schemeAttempt` or `attemptScheme`)

cmd/fluxd/main.go
```diff
@@ -97,14 +97,15 @@ func main() {
...
+		memcachedTimeout  = fs.Duration("memcached-timeout", time.Second, "Maximum time to wait before giving up on memcached requests.")
+		memcachedService  = fs.String("memcached-service", "memcached", "SRV service used to discover memcache servers.")
+
 		registryPollInterval = fs.Duration("registry-poll-interval", 5*time.Minute, "period at which to check for updated images")
 		registryRPS          = fs.Float64("registry-rps", 50, "maximum registry requests per second per host")
 		registryBurst        = fs.Int("registry-burst", defaultRemoteConnections, "maximum number of warmer connections to remote and memcache")
 		registryTrace        = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
-		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "use HTTP for this image registry domain (e.g., registry.cluster.local), instead of HTTPS")
+		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "let these registry hosts skip TLS host verification, and fall back to HTTP (without basic auth) instead of HTTPS; this allows man-in-the-middle attacks, so use with extreme caution")
```
2opremio: Is there something we could break in existing installations using the flag by changing its behavior? 

We are attempting `http` if `https` doesn't work, so, I don't think so, but it's worth giving it some further thought.

2opremio: Alright, there is something which can break: we are now skipping auth, which I am against, but if we really want to skip basic auth with http, we should at least create a new flag and deprecate this one.

registry/client_factory.go
squaremo: Yeah those two points are fair. So I guess the recommendation would be to

 1. only fall back to http if the domain is flagged as insecure
 2. only use TLS_INSECURE_SKIP_VERIFY if the domain is flagged as insecure
 3. always use credentials if you have them

on the basis that cases 1. and 2. are basically equivalent, and if you've flagged the domain as insecure, you are accepting the risk of a MitM attack in either situation -- ?

2opremio: Yep, that sounds good

cmd/fluxd/main.go
squaremo: Update: if we're not skipping auth (as agreed above), I don't think this will be a breaking or even surprising change; if you already had a domain listed as insecure, it's because you wanted it to use HTTP (including sending credentials); instead, it will attempt HTTPS and fallback to HTTP, still sending creds. The change is now you can _also_ use it if you want it to use HTTPS with a self-signed cert.

registry/client_factory.go
squaremo: ^^ @Timer @donifer @cazzoo Just to bring this significant change to your attention, since you've indicated this PR may be of interest to you

cazzoo:
That kind of solution will be enough for my case, I'm actually using self-signed certificate and would like to use HTTPS anyway. Actually using http as hotfix but this is unwanted here.
Looking for merge to next release

squaremo:
@2opremio I've implemented the "always send creds" scheme we discussed, and split the long method up. PTAL!

PR #1536, opened 2018-11-20
Remove CRD hook from chart and update helm-op to v0.5.1

stefanprodan:
- remove the crd-install hook
- add CRD kubectl apply to install docs 

I would wait for a new helm-op release before merging this. Once this is merged and published to our Helm repo I will delete the faulty `0.5.0` release.

### DIFF
--- a/chart/flux/Chart.yaml
+++ b/chart/flux/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.8.1"
 description: Flux is a tool that automatically ensures that the state of a cluster matches what is specified in version control
 name: flux
-version: 0.5.0
+version: 0.5.1
 home: https://weave.works
 sources:
 - https://github.com/weaveworks/flux
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -73,10 +73,19 @@ weaveworks/flux
 
 #### To install Flux with the Helm operator:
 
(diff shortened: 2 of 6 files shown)


sfrique:
Hello,

Only removing the hook and leaving the CRD as it is, is working, even on upgrade. So helm still managing the CRD from templates.

The problem is, if flux needs to create CR, so the hook is necessary, so it creates the CRD first and then creates the CR.

stefanprodan:
The Flux Helm chart has no CRs in it and will never have one.

stefanprodan:
We've updated the CRD fields in the past and Helm upgrade worked so it should be ok as long as we don't use that hook. The only problem with having the CRD in the chart is that on delete the Helm release managed by Flux will also be purged.

stefanprodan:
@squaremo are you ok with deleting the Helm chart package that contains the hook from the Helm repository?

squaremo:
> are you ok with deleting the Helm chart package that contains the hook from the Helm repository?

Can we deprecate it? (and would that be better?)

squaremo:
Hrm, deprecation doesn't work as I thought: https://github.com/helm/helm/blob/master/docs/charts.md#deprecating-charts. It doesn't seem to apply to particular versions of the chart, so much as the chart entirely.

Yes, just delete the buggy version, once there's a replacement.

PR #1538, opened 2018-11-21
Reduce spam log generated by registry cache warmer

### LINKED ISSUE #1529: Reduce spam log generated by registry cache warmer
Due to the fact that Flux and helm-op images are hosted on Quay, every time someone installs Flux will be faced with hundreds of errors. 

These errors will increase with every Flux commit since we log an error per image tag:

```
ts=2018-11-17T20:11:23.752382666Z caller=warming.go:300 component=warmer canonical_name=quay.io/weaveworks/helm-operator auth={map[]} err="error parsing HTTP 429 response body: invalid character '<' looking for beginning of value: \"<html>\\r\\n<head><title>429 Too Many Requests</title></head>\\r\\n<body bgcolor=\\\"white\\\">\\r\\n<center><h1>429 Too Many Requests</h1></center>\\r\\n<hr><center>nginx/1.14.0</center>\\r\\n</body>\\r\\n</html>\\r\\n\"" ref=quay.io/weaveworks/helm-operator:master-821cc2c
```

We could detect the HTTP 429 and log a warning once per image or just ignore them and keep the rate limiter logs:

```
ts=2018-11-17T20:10:35.998697942Z caller=rate_limiter.go:71 component=ratelimiter info="reducing rate limit" host=quay.io limit=50.00
```

On GKE is even worse since Flux can't parse the nvidia-gpu-device-plugin tag and this adds up to the startup error count:

```
ts=2018-11-17T20:16:23.188597011Z caller=warming.go:192 component=warmer canonical_name=gcr.io/google-containers/nvidia-gpu-device-plugin@sha256 auth={map[]} err="requesting tags: mux: variable \"google-containers/nvidia-gpu-device-plugin@sha256\" doesn't match, expected \"^(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\\\\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?$\""
```

Another issue is that we log a git clone error even if the git URL is setup correctly, adding up to the confusion:

```
ts=2018-11-17T20:10:22.299663884Z caller=loop.go:87 component=sync-loop err="git repo not ready: git repo has not been cloned yet"
```

For first time users this is a bad experience and makes it very hard to diagnose a git URL miss configuration see #1528

stefanprodan:
- abort image tags fetching on HTTP 429 errors
- log the rate limit error once per image repository

Fix #1529

### DIFF
--- a/registry/cache/warming.go
+++ b/registry/cache/warming.go
@@ -280,7 +280,7 @@ func (w *Warmer) warm(ctx context.Context, now time.Time, logger log.Logger, id 
 	updates:
 		for _, up := range toUpdate {
 			select {
-			case <-ctxc.Done():
+			case <-ctx.Done():
 				break updates
 			case fetchers <- struct{}{}:
 			}
@@ -306,8 +306,8 @@ func (w *Warmer) warm(ctx context.Context, now time.Time, logger log.Logger, id 
 					if strings.Contains(err.Error(), "429") {
 						once.Do(func() {
 							errorLogger.Log("warn", "aborting image tag fetching due to rate limiting, will try again later")
+							cancel()
 						})
-						cancel()
 					} else {
 						errorLogger.Log("err", err, "ref", imageID)
 					}
--- a/registry/cache/warming.go
+++ b/registry/cache/warming.go
@@ -273,26 +273,20 @@ func (w *Warmer) warm(ctx context.Context, now time.Time, logger log.Logger, id 
 		awaitFetchers := &sync.WaitGroup{}
 		awaitFetchers.Add(len(toUpdate))
 
-		ctxc, cancel := context.WithCancel(context.Background())
+		ctxc, cancel := context.WithCancel(ctx)
 		var once sync.Once
 		defer cancel()
 
 	updates:
 		for _, up := range toUpdate {
 			select {
-			case <-ctx.Done():
+			case <-ctxc.Done():
 				break updates
 			case fetchers <- struct{}{}:
 			}
 
 			go func(update update) {
 				defer func() { awaitFetchers.Done(); <-fetchers }()
-
-				select {
-				case <-ctxc.Done():
-					return // terminate on error
-				default: // avoid blocking
-				}
 
 				imageID := update.ref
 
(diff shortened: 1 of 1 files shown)


registry/cache/warming.go
```diff
@@ -273,6 +273,10 @@ func (w *Warmer) warm(ctx context.Context, now time.Time, logger log.Logger, id
 		awaitFetchers := &sync.WaitGroup{}
 		awaitFetchers.Add(len(toUpdate))
 
+		ctxc, cancel := context.WithCancel(context.Background())
```
squaremo: This should be created with the context passed into the func; otherwise, any deadline or cancellation signal given to that context will be ignored.

squaremo: You should also use this fresh context for the requests, i.e., in L304 below. Otherwise you have a race between the context being cancelled, and the early exit.

registry/cache/warming.go
```diff
@@ -284,6 +288,12 @@ func (w *Warmer) warm(ctx context.Context, now time.Time, logger log.Logger, id
 			go func(update update) {
 				defer func() { awaitFetchers.Done(); <-fetchers }()
 
+				select {
+				case <-ctxc.Done():
+					return // terminate on error
+				default: // avoid blocking
```
squaremo: Having expanded the unchanged lines above, I don't think you need this here, since

 1. it'll exit the loop entirely when the context has been cancelled, thanks to L282-286
 2. any goroutines still underway when the context is cancelled will either complete or immediately return an error from `client.Manifest()`

This early exit will only ever be hit if the goroutine was created before the context was cancelled, and run after it was cancelled. Which will be the case for fewer (probably many fewer) than w.Burst goroutines. I'd rather have less code than a very slightly earlier exit.

registry/cache/warming.go
squaremo: And for the loop-breaking select in L282-286.

stefanprodan:
The automated updates are not working, I'm investigating it...

stefanprodan:
The cancelable context will block forever after the cancel event, so I've used the passed context in the select and the cancelable one for client manifest fetching. The automated updates are working fine now.

stefanprodan:
The blocking was caused by the wait group, that's being fixed in the latest commit. Ready to merge 🔨

Issue #1542, opened 2018-11-22
Instructions for known_hosts don't work when using flux chart


squaremo:
The instructions for constructing a known_hosts in https://github.com/weaveworks/flux/blob/master/site/standalone-setup.md#using-a-private-git-host dodn't work with the flux chart, because it mounts a configmap even if it's empty. Doing so unconditionally is not unreasonable; so, we'll need to change the instructions.

kingdonb:
These linked instructions are gone. Closing.

PR #1544, opened 2018-11-22
Helm chart: Make SSH know hosts volume optional

stefanprodan:
If no know hosts are specified don't create the ConfigMap and the readonly volumes in Flux and helm-op ref #1542


stephenmoloney:
seems like a good idea.

Issue #1179, opened 2018-06-28
Simplify getting the SSH key from the logs


dholbach:
As part of #1130 @squaremo and I talked about how to most easily get the SSH keys from the log.

We currently

 1. advertise `fluxctl identity` (which requires having downloaded the binary and put it into `$PATH`) in `site/using.md`.
 1. extract it from Flux's logs:
 ```sh
[daniel@reef flux ]$ git grep "grep identity.pub"
chart/flux/README.md:kubectl -n flux logs deployment/flux | grep identity.pub | cut -d '"' -f2
chart/flux/templates/NOTES.txt:  kubectl -n {{ .Release.Namespace }} logs $FLUX_POD | grep identity.pub | cut -d '"' -f2
site/helm/installing.md:kubectl -n flux logs $FLUX_POD | grep identity.pub | cut -d '"' -f2
site/standalone/installing.md:kubectl logs deployment/flux | grep identity.pub | cut -d '"' -f2
[daniel@reef flux ]$ 
 ```

The latter can be convenient, but maybe we should agree on one way to go about it.

@squaremo brought up the idea of using something like `kubectl run --image weaveworks/flux -- identity`.

dholbach:
Now that #1212 is landed, shall we cut over our docs to tell people to download `fluxctl` and use `identity` instead?

dholbach:
If yes, we should probably land this before the next release? :)

squaremo:
> Now that #1212 is landed, shall we cut over our docs to tell people to download fluxctl and use identity instead?

Can you elaborate on this suggestion? I thought the objection to "just use fluxctl" is that it means you have to go on an excursion to explain where to download fluxctl and so on.

dholbach:
Where to download was only one step. This was what we had documented before:
```sh
fluxpod=$(kubectl get pod -l name=flux -o name | awk -F / '{ print $2; }')
kubectl port-forward "$fluxpod" 10080:3030 &
export FLUX_URL="http://localhost:10080/api/flux"
fluxctl list-controllers --all-namespaces
```

PR #1548, opened 2018-11-23
Default to fluxctl use for getting git deploy key

dholbach:
For a while we gave people a set of commands to extract the git deploy key from flux logs. Now that it's gotten easier to install fluxctl (and there's lots of benefits to using it anyway), let's default to using

  "fluxctl identity"

  Closes: #1179

### DIFF
--- a/chart/flux/templates/NOTES.txt
+++ b/chart/flux/templates/NOTES.txt
@@ -1,20 +1,10 @@
-1. Get the application URL by running these commands:
-{{- if contains "NodePort" .Values.service.type }}
-  export NODE_PORT=$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ template "flux.fullname" . }})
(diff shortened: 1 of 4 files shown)


dholbach:
I picked Stefan as reviewer as the PR touched on the Helm charts he contributed, but I'd value everyone's else's thoughts too.

chart/flux/templates/NOTES.txt
```diff
@@ -13,7 +13,8 @@
   kubectl -n {{ .Release.Namespace }} port-forward $POD_NAME 8080:3030 &
 {{- end }}
 
-2. Get the Git deploy key by running these commands:
-  export FLUX_POD=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "flux.name" . }},release={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
-  kubectl -n {{ .Release.Namespace }} logs $FLUX_POD | grep identity.pub | cut -d '"' -f2
+2. Get the Git deploy key by installing fluxctl
```
stefanprodan: I would replace option 1. from above with:

```
kubectl -n {{ .Release.Namespace }} logs deployment/{{ .Release.Name }} | grep identity.pub | cut -d '"' -f2
```

PR #1549, opened 2018-11-23
List some popular integrations in main README.md

dholbach:
It'd be good to show off prominently how easily Flux can be integrated in solutions.

I picked some of the integrations I mentioned in the [last longer blog post](https://www.weave.works/blog/weave-increased-our-flux-capacity) about Flux. Happy to add more over time.


stefanprodan:
The istio link is wrong should point to stefanprodan/gitops-istio

dholbach:
Thanks Stefan for noticing and thanks Hidde for fixing!

PR #1552, opened 2018-11-25
Git timeout Helm operator

### LINKED ISSUE #1551: GitTimeout is not been used on helm-op anymore (chartsync)
There was some refactor on upgrading to 0.5 to use git uri per CRD and that change is not using  mirrors at chartsync and not passing the timeout to git.

Here is the line:
https://github.com/weaveworks/flux/blob/master/integrations/helm/chartsync/chartsync.go#L267

hiddeco:
Fixes #1551 -- adds a `Timeout` field to the `HelmRelease` CRD and uses the value (default `20`) while setting up the Git mirror.

Image for testing available as `hiddeco/helm-operator:1551-git-timeout-fix`.

### DIFF
--- a/chart/flux/templates/helm-operator-crd.yaml
+++ b/chart/flux/templates/helm-operator-crd.yaml
@@ -56,6 +56,8 @@ spec:
                     type: string
                   ref:
                     type: string
+                  timeout:
+                    type: int
               - required: ['repository', 'name', 'version']
                 properties:
                   repository:
--- a/chart/flux/templates/helm-operator-deployment.yaml
+++ b/chart/flux/templates/helm-operator-deployment.yaml
@@ -1,5 +1,3 @@
-{{- $gitPollInterval := default .Values.git.pollInterval .Values.helmOperator.git.pollInterval }}
-{{- $gitTimeout := default .Values.git.timeout .Values.helmOperator.git.timeout }}
 {{- if .Values.helmOperator.create -}}
 apiVersion: apps/v1beta2
 kind: Deployment
@@ -83,8 +81,6 @@ spec:
         {{- end }}
         {{- end }}
         args:
-        - --git-poll-interval={{ $gitPollInterval }}
-        - --git-timeout={{ $gitTimeout }}
         - --charts-sync-interval={{ .Values.helmOperator.chartsSyncInterval }}
         - --update-chart-deps={{ .Values.helmOperator.updateChartDeps }}
         - --log-release-diffs={{ .Values.helmOperator.logReleaseDiffs }}
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -49,9 +49,6 @@ var (
 	logReleaseDiffs    *bool
 	updateDependencies *bool
 
-	gitPollInterval *time.Duration
-	gitTimeout      *time.Duration
-
 	listenAddr *string
 )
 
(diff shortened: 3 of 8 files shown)


sfrique:
Hello,

I tested this and didn't work on my env.

I have a few CRs and updated only a few to test it. I also added a few more logs options to my build based on this branch. 

The option when set only to one CR didn't work, as It seems the operator/chartsync will use the timeout for the first one it finds out of CRs.

At logs we can see only one line of  "started mirroring repo" :

`ts=2018-11-27T17:52:09.7072704Z caller=chartsync.go:268 component=chartsync info="started mirroring repo" repo=<snip> timeout=20s path=kubernetes/cert-manager`  
this line only appears once at the log.

When operator gets a CR that doesn't have timeout set, all charts sync fail, at the other hand, if it tries a chart that have the timeout set, all charts sync and it works fine. 

The sync worked fine for all charts on this loop.
`ts=2018-11-27T17:46:35.170039374Z caller=chartsync.go:268 component=chartsync info="started mirroring repo" repo=<snip> timeout=300s path=kubernetes/elasticsearch`

This makes me think that the chartsync loop or some, is only using one value for some reason.

I changed the log line so i could se the timeout for each chart and the path it was trying to sync, but like i said, It only shows up once, and I find that weird.

Hope that helps! 

I'm available to do some testing if needed.

It was a good thing I only updated a few CRs, if I had changed all, it probably would work and I wouldn't notice this.

hiddeco:
I missed the fact that a mirror gets created only once so the first `HelmRelease` wins the race. Given that I see two options:

  1. we make the flag global again and remove the `Timeout` field
  2. we increase the timeout on the mirror each time we find a higher value
  3. ...something else?

@squaremo what would be the best approach from your point of view? Am I missing options?

sfrique:
Any updates on this?
I rather have option 2.

squaremo:
Ah hmm, this is my bad -- I changed a lot of the internal wiring of the helm operator for v0.5, and didn't connect up the git timeout. If just having the `--git-timeout` respected is adequate, I think that would be a lot simpler that having per-HR timeouts -- especially since mirrors are shared among the HRs.

wdyt @hiddeco, @sfrique -- would being able to bump up the timeout for all git operations be good enough?

hiddeco:
It does the job but I can imagine there are people with dozens of different repositories (God may know why) who are not so keen about this solution.

What I understood from @sfrique when talking here (see comment above about option 2) and in private on Slack is that he would like to configure it for each repository. I do not know if this is out of idealism or because of _[insert reason]_.

However from my point of view we should just get this fixed asap. We can think / implement (of) a different / finer solution later on.

squaremo:
:+1: to getting a fix done, and I'm happy to enter into subsequent negotiations :-)

hiddeco:
Will furiously type code and `git push --force` tonight.

squaremo:
FWIW the change was neatly done -- let's keep the revs in our back pocket

sfrique:
My thoughts are if we have distinct configs per repository, timeout should be separated as well.

It could be global, but It makes more sense to me to be per CR. (per repository)

Thanks!

squaremo:
> My thoughts are if we have distinct configs per repository, timeout should be separated as well.
> [...] It makes more sense to me to be per CR. (per repository)

Yep, understood. I do agree that it makes sense to supply git parameters along with the git repo.

Our problem at present is that "per custom resource" is not the same thing as "per git repository", since custom resources can mention the same git repository (by design). A "correct" design would probably have git repositories as their own custom resources with their own config, including timeouts; but that's quite a big change, and is another moving part for people to have to deal with.

Since the parameter in question is specifically a timeout, I think it's reasonable to have a global setting, and people can supply a value that works for their worst case. After all, it's a timeout, so it's already specifying a worst case.

PR #1553, opened 2018-11-25
Add CHANGELOG to Helm chart

hiddeco:


sfrique:
I am not sure if this made the release, nor if it should be mentioned, also this is for chat version 0.5.2
https://github.com/weaveworks/flux/pull/1543

hiddeco:
@sfrique this will be added when we write up the release notes for `0.5.2`.

sfrique:
I mentioned, because chart release and helm-op release are different things. So i wasn't sure how you guys would like to add this.

PR #1556, opened 2018-11-25
Include proxy env variables in git operations if set

### LINKED ISSUE #1535: HTTP Proxy not used when connecting with remote GIT repository
Pods in my environment don't have direct access to Internet resources, they have to use a proxy. So, for `flux-helm-operator`, the deployment manifest contains these environments variables:

```
          env:
            - name: HTTP_PROXY
              value: http://proxy.internalorg:80
            - name: http_proxy
              value: http://proxy.internalorg:80
            - name: HTTPS_PROXY
              value: http://proxy.internalorg:80
            - name: https_proxy
              value: http://proxy.internalorg:80
            - name: NO_PROXY
              value: 127.0.0.1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.internalorg,...
            - name: no_proxy
              value: 127.0.0.1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.internalorg,....
```

Whenever I'm creating a HelmRelease object (using v0.5.0) specifying an external (github) GIT repository, those proxy env variables are not used:

```
---
kind: HelmRelease
metadata:
  name: consul
  namespace: cluster-tools
  labels:
    chart: consul
spec:
  chart:
    git: https://github.com/hashicorp/consul-helm.git
    path: "/"
    ref: "v0.3.0"
  releaseName: consul
```

Logs from `flux-helm-operator`:
```
ts=2018-11-20T10:11:32.776998797Z caller=chartsync.go:243 component=chartsync info="Start of releasesync"
ts=2018-11-20T10:11:32.803799336Z caller=chartsync.go:318 component=chartsync info="chart repo not ready yet" releaseName=consul resource=cluster-tools:HelmRelease/consul status=new err="git clone --mirror: fatal: unable to access 'https://github.com/hashicorp/consul-helm.git/': Could not resolve host: github.com"
```

The `Could not resolve host: github.com` is an indicator that the proxy is not used since DNS resolution should not occur on the client side, but on the proxy. When I manually call `git clone...` from within the pod (using `kubectl exec -ti ... sh`), it works fine since environment variables are set.

My guess is that the `git` command is launched within the operator without retrieving environment variables:
https://github.com/weaveworks/flux/blob/8f09ae615bc9e12e34567d5bd6e1505998c1f608/git/operations.go#L268

Maybe something like `c.Env = os.Environ()` might work ?

hiddeco:
Fixes #1535.

Includes proxy env variables (`http_proxy`, `https_proxy` and `no_proxy`) in Git operations if they are set.

### DIFF
--- a/git/operations.go
+++ b/git/operations.go
@@ -20,7 +20,7 @@ import (
 const trace = false
 
 // Env vars that are allowed to be inherited from the os
-var allowedEnvVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+var allowedEnvVars = []string{"http_proxy", "https_proxy", "no_proxy"}
 
 func config(ctx context.Context, workingDir, user, email string) error {
 	for k, v := range map[string]string{
--- a/git/operations.go
+++ b/git/operations.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"os/exec"
 	"strings"
 
@@ -17,6 +18,9 @@ import (
 
 // If true, every git invocation will be echoed to stdout
 const trace = false
+
+// Env vars that are allowed to be inherited from the os
+var allowedEnvVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 
 func config(ctx context.Context, workingDir, user, email string) error {
 	for k, v := range map[string]string{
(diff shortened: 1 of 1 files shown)


git/operations.go
```diff
@@ -289,7 +293,18 @@ func execGitCmd(ctx context.Context, dir string, out io.Writer, args ...string)
...
+	env := []string{"GIT_TERMINAL_PROMPT=0"}
+
+	// include allowed env vars from os
+	for _, k := range allowedEnvVars {
+		v, ex := os.LookupEnv(k)
+		if ex == false {
+			continue
+		}
+		env = append(env, k+"="+v)
+	}
```
rade: The golang idiomatic way to write this would be
```golang
if v, ok := os.LookupEnv(k); ok {
    env = append(env, k+"="+v)
}
```

hiddeco: You are right and I should have thought about it as I was already familiar with it.

PR #1557, opened 2018-11-27
Set flux and helm-op versions in the upgrade doc

stefanprodan:
Helm upgrade with reuse values flag will prevent flux and helm-op images from being updated, this PR adds the two image tags to the upgrade command.


mgazza:
confirmed after running
```text
 kubectl get deployments --namespace flux -o json | jq '.items[].spec.template.spec.containers[].image'
"quay.io/weaveworks/flux:1.8.1"
"quay.io/weaveworks/helm-operator:0.5.1"
"memcached:1.4.25"
```

PR #1559, opened 2018-11-28
Ignore some errors during manifest loading

### LINKED ISSUE #1558: Manifest loading fails when ephemeral file disappears at wrong time
Git garbage collection creates ephemeral file `.git/gc.pid.lock` and we have seen occasional errors while Flux is trying to enumerate Helm chart directories:

```
walking "/tmp/flux-working002532032" for chartdirs: walking "/tmp/flux-working002532032/.git/gc.pid.lock" for charts: lstat /tmp/flux-working002532032/.git/gc.pid.lock: no such file or directory
```
This error aborts `LoadManifests()` early and fails the `ListServices` as well as `ListImages` calls.

### What happens?
The daemon clones the repo where I presume it garbage collects immediately after. Then it loads the manifest which must have a race with the chart tracker that sees the file during listing but upon actually accessing it cannot find it anymore.
https://github.com/weaveworks/flux/blob/bd2d3e04a2865b5451fb8ad6cc3671c462705208/daemon/daemon.go#L78-L82
https://github.com/weaveworks/flux/blob/4a40c66e9c633cf4d6111b3c154f3979d14f7a27/cluster/kubernetes/resource/load.go#L77

### What do I expect to happen?
What we do here is enumerating the directories to _exclude_. If we don't have the permissions to read something or it no longer exists then it will never be included. In that case there is no need to abort the enumeration here but just skip over to the next file/dir.

rndstr:
When loading manifests, we attempt to walk the base file tree twice. Once
while looking for directories with Helm charts in them, and then to load
all the .yaml/.yml files. The former keeps track of those directories
to exclude them wwhile doing the latter since we do not want to load
yamels from directories with Helm charts in them.

While walking those file trees any error aborted the whole loading
process and as a consequence, the API calls `ListServices` and
`ListImages` would return an error. Suddenly disappearing files such as
Git's gc lock file would trigger an error. Since the Git repo is cloned
before manifests are loaded, there was a race sometimes between having
that lock file being enumerated but then disappear while trying to
retrieve information about it.

This PR ignores all errors while enuemrating the Helm chart directories
since permissions are unlikely to change and disappearing files to not
modify the list of excluded directories.

When walking the yamels, only errors for files or directories that we
actually care about will be reported. That means any error for files
that do not have a yamel extension will be just ignored.

---
Follow-up to #1076 
Fixes #1558

### DIFF
--- a/pkg/cluster/kubernetes/resource/load.go
+++ b/pkg/cluster/kubernetes/resource/load.go
@@ -8,9 +8,9 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/pkg/errors"
 	sops "go.mozilla.org/sops/v3"
 	"go.mozilla.org/sops/v3/decrypt"
-	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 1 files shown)


squaremo:
> Since the Git repo is cloned
> before manifests are loaded, there was a race sometimes between having
> that lock file being enumerated but then disappear while trying to
> retrieve information about it.

I guess you must have seen this to report it specifically, but I am puzzled about how it occurs -- has the clone operation not finished before we look at the filesystem?

rndstr:
I'm also not sure how it happens but Git garbage collection or at least its lock file is being reported as existing but then cannot be accessed. Maybe fs caching?

I looked through the usages of `git.Checkout` whether it might be shared somewhere but I couldn't find anything.

squaremo:
Before papering over the problem, I would like to understand why it happens.

rndstr:
@2opremio sorry I haven't yet gotten around to look more into this. the two independent reports of this happening were on GKE, @foot had it occurring occasionally.

stefanprodan:
I've seen reports of this on Slack, one user mentioned this happens once per week.

2opremio:
@stefanprodan @hiddeco are you still hearing from users having this problem?

hiddeco:
I am not aware of any reports about this issue in the last couple of weeks.

emas80:
Hi, can this be related to https://github.com/fluxcd/flux/issues/2927?
We are experiencing it very often, at least once a week, restarting the pod fixes.
I've been asked by the dev team to kill the Flux pods every day, because it is very annoying sometimes you don't get any feedback that something is actually wrong.

It was not happening before 1.18.1, I think that because of many optimizations Flux is not OOM killed anymore by Kubernetes like before so the same pod stays alive for weeks and this problem occurs.

squaremo:
https://git-scm.com/docs/git-gc#Documentation/git-gc.txt-gcautoDetach:
> Make git gc --auto return immediately and run in background if the system supports it. Default is true.

and https://git-scm.com/docs/git-gc#_description:
> When common porcelain operations that create objects are run, they will check whether the repository has grown substantially since the last maintenance, and if so run git gc automatically.

My guess is that in some circumstances, the git clone triggers a GC in the cloned repo, and this is run in the background.

squaremo:
Rebased and removed the (now unnecessary) release note.

PR #1562, opened 2018-11-29
Add tests for updateDependencies

mgazza:
Added tests to ensure that this functions behaviour with helm is always defined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,8 @@ clean:
 realclean: clean
 	rm -rf ./cache
 
-test:
-	PATH=${PATH}:${PWD}/bin go test ${TEST_FLAGS} $(shell go list ./... | grep -v "^github.com/weaveworks/flux/vendor" | sort -u)
+test: build/helm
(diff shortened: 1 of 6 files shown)


mgazza:
Tests pass locally. I'm not sure how to proceed it looks like helm can't update the repository.

squaremo:
I suspect that it's because `updateDependencies` relies on the `helm` binary, and that's not available in the CI environment. The non-existent chart test succeeds there, because it expects an error, and gets one.

Suggestion: we download the helm client as part of the build, so it could just be a dependency of `make test` (and added to the path).

NB when I try this locally, it's the non-existent chart test that fails, i.e., it doesn't return an error. We can see if that's the case in CI ..

squaremo:
Update: added grin for first-time contributor -- thanks for this PR!

squaremo:
@mgazza Are you happy for me to push a couple of commits here?

squaremo:
@hiddeco Mind having a look at this one when you have a chance? I ought to recuse myself on the basis that I've added commits :-)

Issue #1545, opened 2018-11-22
Allow configuration of the helm install timeout in the flux CRD


stephenmoloney:
### Issue:

I've introduced a chart in my cluster config repo that takes `> 5` minutes on a job with is a long set of migrations. 
The job gets shutdown at the 5 minute mark (300s) and so the subsequent helm installation is broken as the migrations do not go to completion.

### Proposed solution

The helm install timeout needs to be extended to allow for extra time to complete.

Allow the `helm.InstallTimeout` or `helm install --timeout` on the command line to be configured in the flux CRD.

PR #1566, opened 2018-12-04
Add helm install timeout option

stefanprodan:
* add timeout field to HelmRelease CRD (defaults to 300s)
* add helm release timeout to install and upgrade operations
* remove dead code from the `release` package and fix install logging

Fix: #1545

Docker Hub image for testing: `stefanprodan/helm-operator:timeout-v5`

### DIFF
--- a/chart/flux/templates/helm-operator-crd.yaml
+++ b/chart/flux/templates/helm-operator-crd.yaml
@@ -107,6 +107,9 @@ spec:
             releaseName:
               type: string
               pattern: "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+            timeout:
+              type: integer
+              format: int64
             chartGitPath:
               type: string
             values:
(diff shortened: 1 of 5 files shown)


stephenmoloney:
Hi @stefanprodan, thanks for this as I raised this issue originally. It might be 2 weeks approx before I get to try this.

chart/flux/templates/helm-operator-crd.yaml
```diff
@@ -107,6 +107,9 @@ spec:
             releaseName:
               type: string
               pattern: "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+            timeout:
```
squaremo: Should this also be in the `HelmRelease` CRD?

stefanprodan: Sorry I misplaced the validation, I've moved it from FHR to HR in the latest commit.

PR #1597, opened 2018-12-11
Make it clear Slack integration uses a service

### LINKED ISSUE #1587: Flux intro page implies slack support
https://github.com/weaveworks/flux/blob/master/site/introduction.md says 

> One final high level feature is that Flux increases visibility of your application. It provides an audit history for your deployments and Slack integration for "ChatOps" style development.

Since this feature is only available in Weave Cloud, this wording should be reworked to clarify that.

squaremo:
Here I am trying to walk the line between giving accurate advice, and
recognising that the notifications are particular to Weave Cloud
(albeit fluxcloud can also do something with them).

Fixes #1587.

### DIFF
--- a/site/introduction.md
+++ b/site/introduction.md
@@ -52,14 +52,14 @@ version.
 
 ## Integrations with other devops tools
 
-One final high level feature is that Flux increases visibility of 
-your application. It provides an audit history for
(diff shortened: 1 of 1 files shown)


squaremo:
Thanks Hidde!

PR #1601, opened 2018-12-13
[WIP] Explain what's need to use a git repo over HTTPS

squaremo:
(needs more exploration to tighten up the instructions, and probably links or paras in other docs)


songsak2299:
Ko

squaremo:
I tried IRL to set things up so I could use an HTTPS git URL, and it is not quite as simple as I've indicated here. To be clear, this isn't wrong -- it's just not enough information to get it working easily. In particular, you need to tell git to use a credentials store, then mount that as well. It is infuriatingly fiddly to get it right, and computerproof instructions are called for.

squaremo:
(marking as blocked as a reminder that it needs more work)

squaremo:
Closing this as I'm not expecting to complete it any time soon.

PR #1606, opened 2018-12-17
Check that chart exists when updating its deps

squaremo:
This PR fixes the CI failure that resulted from #1562 and #1561 interacting: the latter makes `updateDependencies` succeed trivially if the `requirements.yaml` file is missing, but accidentally also succeeds if the entire chart is missing. The latter tests specifically that `updateDependencies` will fail if the chart is missing.

We don't want to fail at updating chart dependencies when a chart doesn't have a requirements.yaml file (which is optional, for a chart). However, we _do_ want to fail if the chart doesn't exist in the filesystem at all.

### DIFF
--- a/integrations/helm/chartsync/deps.go
+++ b/integrations/helm/chartsync/deps.go
@@ -11,6 +11,15 @@ import (
 func updateDependencies(chartDir, helmhome string) error {
 	var hasLockFile bool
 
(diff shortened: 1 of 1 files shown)


squaremo:
Diligent! Thanks Hidde

PR #1615, opened 2018-12-25
show more context for errors from fluxctl

2opremio:
Errors in fluxctl are unwrapped before printing since we want to
format flux-specific errors specially, and they can come wrapped.

However, the unwrapping can lead to losing context when printing
non-flux-specific errors.

For instance, after
https://github.com/justinbarrick/go-k8s-portforward/pulls/3 we display
the following when a kubeconfig file isn't found:

```
Error: invalid configuration: no configuration has been provided
Run 'fluxctl identity --help' for usage.
```

Which is the cause of a much more helpul error when in full:

```
Error: Could not load kubernetes configuration file: invalid configuration: no configuration has been provided
Run 'fluxctl identity --help' for usage.
```


squaremo:
This changes the behaviour when there's a `*fluxerr.Error` -- it'll print the help text, then the error message, then the usage, while before it just printed the help text. I don't think this is wrong, just flagging it up. This is what it looks like:

```
$ FLUX_URL=https://cloud.weave.works/api/flux fluxctl identity
== Error ==

The request failed authentication

This most likely means you have a missing or incorrect token. Please
make sure you supply a service token, either by setting the
environment variable FLUX_SERVICE_TOKEN, or using the argument --token
with fluxctl.


Error: executing HTTP request: request failed authentication
Run 'fluxctl identity --help' for usage.
```

Issue #1591, opened 2018-12-07
DNS resolution of memcached service fails


dilshad18:
Following error is happening in flux installation inside minikube instance:

`component=memcached err="error updating memcache servers: lookup 172-17-0-3.flux-memcached.flux.svc.cluster.local. on 10.96.0.10:53: no such host"`

This is a rather fresh installation and only few files have been applied. We update a single file and it tries to update that and it fails,

squaremo:
It looks like it either can't reach the DNS server, or doesn't get a result back. The hostname there is a bit odd -- should it include the hyphenated IP address like that? I would expect just `flux-memcached.flux.svc.cluster.local`.

johnraz:
I can reach the memcache container from the fluxd container by hitting the hostname passed to fluxd:

`ps aux` gives (truncated by me):

```
fluxd --ssh-keygen-dir=/var/fluxd/keygen --k8s-secret-name=flux-git-deploy --memcached-hostname=flux-memcached ...
```
Telnet session from the fluxd container gives (again truncated by me):

```
/home/flux # telnet flux-memcached 11211
stats
STAT pid 1
STAT uptime 63144
STAT time 1544719316
...
```

squaremo:
> So I would say the memcache server list is properly provisioned with the valid hostname and some "ghost" hostnames are trying to get in and are rejected because they can't resolve...

Yes, that sounds like a good diagnosis. Filling in some details, after looking in the Kubernetes docs re service discovery:

The way it's set up in the example deployment (and chart, and Weave Cloud config, ...) is that memcached has a headless service. According to https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services (also see https://github.com/kubernetes/dns/blob/master/docs/specification.md), in DNS there will be:

 - An A record for memcached.namespace.svc.cluster.local with the IP of each ready pod;
 - An A record for <hostname>.memcached.namespace.svc.cluster.local for each pod, where <hostname> is a generated name;
 - A SRV record which has the auto-generated hostname, and the port, for each pod in the service

With the arguments given in the example deployment of fluxd, it'll query for the SRV records, then the memcache client code (linked above) will query the IP of each host mentioned. So where it's failing is in that second bit -- it can't resolve some or all of the hosts it got from the SRV records.

But the question remains: how did those unresolveable endpoints get there in the first place? That I don't know :-(

squaremo:
[...]

PR #1618, opened 2018-12-27
Use ClusterIP service name for connecting to memcached

stefanprodan:
- add an option to replace memcached headless service with ClusterIP in Helm chart
- make Flux use the memcached ClusterIP name instead of SRV records when the `memcached-service` is set to string empty

Fix: #1591

Test image: stefanprodan/flux:mem-v5

~~PS. This is a breaking change in the Helm chart, using the new image with the old chart will not work since the old chart deploys memcached with a headless service~~

### DIFF
--- a/deploy/flux-deployment.yaml
+++ b/deploy/flux-deployment.yaml
@@ -17,7 +17,7 @@ spec:
       labels:
         name: flux
     spec:
-      serviceAccount: flux
+      serviceAccountName: flux
       volumes:
       - name: git-key
         secret:
@@ -93,7 +93,10 @@ spec:
         # or with a different service name, you can supply these
         # following two arguments to tell fluxd how to connect to it.
         # - --memcached-hostname=memcached.default.svc.cluster.local
-        # - --memcached-service=memcached
+
+        # use the memcached ClusterIP service name by setting the
+        # memcached-service to string empty
+        - --memcached-service=
 
         # this must be supplied, and be in the tmpfs (emptyDir)
         # mounted above, for K8s >= 1.10
--- a/registry/cache/memcached/memcached.go
+++ b/registry/cache/memcached/memcached.go
@@ -14,8 +14,6 @@ package memcached
 import (
 	"encoding/binary"
 	"fmt"
-	"net"
-	"sort"
 	"sync"
 	"time"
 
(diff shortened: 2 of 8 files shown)


registry/cache/memcached/memcached.go
```diff
@@ -158,20 +156,8 @@ func (c *MemcacheClient) updateLoop(updateInterval time.Duration) {
...
-	}
-	var servers []string
-	for _, srv := range addrs {
-		servers = append(servers, fmt.Sprintf("%s:%d", srv.Target, srv.Port))
-	}
-	// ServerList deterministically maps keys to _index_ of the server list.
-	// Since DNS returns records in different order each time, we sort to
-	// guarantee best possible match between nodes.
-	sort.Strings(servers)
+	servers := []string{fmt.Sprintf("%s:11211", c.hostname)}
```
squaremo: There's no need to change this; Use `FixedServerMemcacheClient` instead.

squaremo:
> his is a breaking change in the Helm chart,

.. and probably in most people's deployments, for the same reason. This ought to require a change in arguments, so it doesn't break when the image is updated.

deploy/flux-deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       labels:
         name: flux
     spec:
-      serviceAccount: flux
+      serviceAccountName: flux
```
squaremo: Well caught

deploy/flux-deployment.yaml
```diff
@@ -93,7 +93,10 @@ spec:
         # or with a different service name, you can supply these
         # following two arguments to tell fluxd how to connect to it.
         # - --memcached-hostname=memcached.default.svc.cluster.local
-        # - --memcached-service=memcached
+
+        # use the memcached ClusterIP service name by setting the
+        # memcached-service to string empty
+        - --memcached-service=
```
squaremo: Should the example service have the `clusterIP: None` line removed, in this case?

stefanprodan:
I've installed the chart with ClusterIP on and off, both are working fine. There is one caveat with helm upgrade since the ClusterIP field is immutable.

The upgrade fails with: 
```
Error: UPGRADE FAILED: Service "flux-memcached" is invalid: spec.clusterIP: Invalid value: "": field is immutable
```

In order to fix the upgrade you need to delete the headless service upfront:
```
kubectl -n flux delete svc flux-memcached
```

stefanprodan:
@squaremo I've added a note to the chart changelog with the upgrade instructions.

PR #1619, opened 2018-12-27
Native ECR auth

### LINKED ISSUE #539: Support for AWS ECR?

squaremo:
Taking inspiration (and code) from #1455, this PR builds support for fetching authorization tokens into the registry credentials code.

The design is simple: it's a wrapper for ImageCreds which keeps an authentication token, refreshing when it needs to via the AWS API. This is pretty similar to `ImageCredsWithDefaults`, and is composed with that in fluxd's main.go. The arguments `--registry-ecr-region`, `--registry-ecr-include-id` and `--registry-ecr-exclude-id` control which images it will scan:

 -  region gives the region(s) that will be scanned; if empty, the cluster region will be assumed
 - include-id gives the AWS accounts (registries) to scan; if empty, any image in the allowed regions will be scanned
 - exclude-id excludes an AWS account (registry) from those to scan; it defaults to the account used for EKS "system" images (e.g., coredns), so images from there won't be scanned.

None of the arguments are necessary; the default will be "scan images from any registry in the cluster region, except the EKS system images".

If fluxd is not running on EC2 (or hasn't been rigged so that it can authenticate as an AWS user, to access ECR), the wrapper will fail to be initialised, a message will be logged, and image scanning will proceed as before.

Fixes #539.

(EDIT: update to reflect latest changes)

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -106,6 +106,10 @@ func main() {
 		registryTrace        = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
 		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "use HTTP for this image registry domain (e.g., registry.cluster.local), instead of HTTPS")
 
+		// AWS authentication
+		awsRegion      = fs.String("aws-region", "", "AWS region to authenticate to, e.g., 'us-east-1'; AWS authorization needs both this, and at least one --aws-registry-id to be supplied")
+		awsRegistryIDs = fs.StringSlice("aws-registry-id", nil, "AWS ECR account ID to authenticate with (multiple values allowed)")
+
 		// k8s-secret backed ssh keyring configuration
 		k8sSecretName            = fs.String("k8s-secret-name", "flux-git-deploy", "Name of the k8s secret used to store the private SSH key")
 		k8sSecretVolumeMountPath = fs.String("k8s-secret-volume-mount-path", "/etc/fluxd/ssh", "Mount location of the k8s secret storing the private SSH key")
@@ -262,10 +266,22 @@ func main() {
 		}
 
 		imageCreds = k8sInst.ImagesToFetch
+		if *awsRegion != "" && len(*awsRegistryIDs) > 0 {
+			awsConf := registry.AWSRegistryConfig{
+				Region:      *awsRegion,
+				RegistryIDs: *awsRegistryIDs,
+			}
+			credsWithAWSAuth, err := registry.ImageCredsWithAWSAuth(imageCreds, log.With(logger, "component", "aws"), awsConf)
+			if err != nil {
+				logger.Log("warning", "AWS authorization not used; pre-flight check failed", "err", err)
+			} else {
+				imageCreds = credsWithAWSAuth
+			}
+		}
 		if *dockerConfig != "" {
 			credsWithDefaults, err := registry.ImageCredsWithDefaults(imageCreds, *dockerConfig)
 			if err != nil {
-				logger.Log("msg", "--docker-config not used", "err", err)
+				logger.Log("warning", "--docker-config not used; pre-flight check failed", "err", err)
 			} else {
 				imageCreds = credsWithDefaults
 			}
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -266,7 +266,7 @@ func main() {
 		}
 
 		imageCreds = k8sInst.ImagesToFetch
-		if *awsRegion != "" && len(*awsRegistryIDs) > 0 {
+		{
 			awsConf := registry.AWSRegistryConfig{
 				Region:      *awsRegion,
 				RegistryIDs: *awsRegistryIDs,
@@ -277,15 +277,16 @@ func main() {
 			} else {
 				imageCreds = credsWithAWSAuth
 			}
-		}
-		if *dockerConfig != "" {
-			credsWithDefaults, err := registry.ImageCredsWithDefaults(imageCreds, *dockerConfig)
-			if err != nil {
-				logger.Log("warning", "--docker-config not used; pre-flight check failed", "err", err)
-			} else {
-				imageCreds = credsWithDefaults
+			if *dockerConfig != "" {
+				credsWithDefaults, err := registry.ImageCredsWithDefaults(imageCreds, *dockerConfig)
+				if err != nil {
+					logger.Log("warning", "--docker-config not used; pre-flight check failed", "err", err)
+				} else {
+					imageCreds = credsWithDefaults
+				}
 			}
 		}
+
 		k8s = k8sInst
 		// There is only one way we currently interpret a repo of
 		// files as manifests, and that's as Kubernetes yamels.
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -186,8 +186,8 @@ func main() {
 	var clusterVersion string
 	var sshKeyRing ssh.KeyRing
 	var k8s cluster.Cluster
-	var imageCreds func() registry.ImageCreds
 	var k8sManifests cluster.Manifests
+	var imageCreds func() registry.ImageCreds
 	{
 		restClientConfig, err := rest.InClusterConfig()
 		if err != nil {
@@ -265,32 +265,33 @@ func main() {
 			logger.Log("ping", true)
 		}
 
-		imageCreds = k8sInst.ImagesToFetch
-		{
-			awsConf := registry.AWSRegistryConfig{
-				Regions:    *awsRegions,
-				AccountIDs: *awsAccountIDs,
-			}
-			credsWithAWSAuth, err := registry.ImageCredsWithAWSAuth(imageCreds, log.With(logger, "component", "aws"), awsConf)
-			if err != nil {
-				logger.Log("warning", "AWS authorization not used; pre-flight check failed", "err", err)
-			} else {
-				imageCreds = credsWithAWSAuth
-			}
-			if *dockerConfig != "" {
-				credsWithDefaults, err := registry.ImageCredsWithDefaults(imageCreds, *dockerConfig)
-				if err != nil {
-					logger.Log("warning", "--docker-config not used; pre-flight check failed", "err", err)
-				} else {
-					imageCreds = credsWithDefaults
-				}
-			}
-		}
-
 		k8s = k8sInst
+		imageCreds = k8sInst.ImagesToFetch
 		// There is only one way we currently interpret a repo of
 		// files as manifests, and that's as Kubernetes yamels.
 		k8sManifests = &kubernetes.Manifests{}
+	}
+
+	// Wrap the procedure for collecting images to scan
+	{
+		awsConf := registry.AWSRegistryConfig{
+			Regions:    *awsRegions,
+			AccountIDs: *awsAccountIDs,
+		}
+		credsWithAWSAuth, err := registry.ImageCredsWithAWSAuth(imageCreds, log.With(logger, "component", "aws"), awsConf)
+		if err != nil {
+			logger.Log("warning", "AWS authorization not used; pre-flight check failed")
+		} else {
+			imageCreds = credsWithAWSAuth
+		}
+		if *dockerConfig != "" {
+			credsWithDefaults, err := registry.ImageCredsWithDefaults(imageCreds, *dockerConfig)
+			if err != nil {
+				logger.Log("warning", "--docker-config not used; pre-flight check failed", "err", err)
+			} else {
+				imageCreds = credsWithDefaults
+			}
+		}
 	}
 
 	// Registry components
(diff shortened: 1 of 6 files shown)


squaremo:
Image pushed to quay.io/squaremo/flux:issue-539-native-ecr-auth-b76ac67 should anyone wish to try this out ...

site/daemon.md
```diff
@@ -74,6 +74,8 @@ fluxd requires setup and offers customization though a multitude of flags.
 |--registry-burst        | `125`      | maximum number of warmer connections to remote and memcache|
 |--registry-insecure-host| []         | registry hosts to use HTTP for (instead of HTTPS) |
 |--docker-config         | `""`       | path to a Docker config file with default image registry credentials |
+|--aws-region            | `""`       | AWS region for authentication, when scanning image in ECR |
+|--aws-registry-id       | `[]`       | ECR registry ID(s) for authentication, when scanning images in ECR (multiple values allowed) |
```
errordeveloper: Do all registries have to be in the same region as set by `--aws-region` or not necessary? The reason I asked is that above it say "region for authentication", so it sounds like the region has is only for authentication... Would be good to clarify whichever is the case. I am not sure how auth works, I thought that there may be a way to authenticate in one region and call out to other regions, but not too sure.

errordeveloper: Also, what IDs are these? We should clarify.

errordeveloper: > Also, what IDs are these? We should clarify.

From looking at `aws ecr help`, (which is not particularly clear either), it sounds like these are actually AWS account IDs.

So I believe if neither `--aws-region` nor `--aws-registry-ids` are set, Flux will simply authenticate agains whatever region and whatever account the cluster is in. Is that the case?
If that is the case, I believe the main use-case for `--aws-registry-ids` is to allow looking at images in multiple accounts (which is common). And `--aws-region` could be used when Flux doesn't run in AWS. Perhaps we could somehow  clarify this in the docs, at least to say that one doesn't need to worry about these flags in a most straight-forward setup.

squaremo: Yes, they are account IDs -- the registry URLs are differentiated by account ID, so the account ID becomes the registry ID. If you reckon it's clearer, I could change this argument name (and perhaps make the connection with registries in the help text).

squaremo: Regarding the region, I see the registry URL _also_ contains a region. So you are right to say that we don't need to be supplied these bits of information -- we could just assume that the necessary credentials are available, and try instigating a session (per region).

These arguments serve a secondary purpose, which is to act as an explicit way of saying "yes, use AWS authentication for these". Perhaps it should be the other way though -- a flag to explicitly switch AWS authentication off, and otherwise assume the ambient credentials are there to be used.

squaremo:
> Please don't consider my comments as blockers for the PR, those are just a few thoughts on how we can improve things in the future...

The question is, if we released this as is, would it limit our ability to improve it? Probably not -- we can always deprecate those arguments, or reinterpret them to mean "just auth this region / these account IDs". But it would be good to get it right the first time, too :-)

squaremo:
Two more points:

 - usually we don't know all the registry URLs in play; so we wouldn't be able to authenticate against all of them at once unless we're told them explicitly. So perhaps `--registry-id` could mean "authenticate for these and only these ECR registries".
 - @errordeveloper tells me that using a registry in another region would incur data costs (of course -- everything costs, on AWS). So although, in principle, we could authenticate for whichever region is in the registry URL, it might be desirable to limit it to the region in which the cluster is running, by default.

With those in mind, I propose that I change the arguments:

 - `--aws-region=[]` -- if empty, only authenticate against the cluster region; otherwise, authenticate against all listed regions (if they come up in a registry URL)
 - `--aws-account-id=[]` -- if empty, attempt to authenticate against any account ID that is used in a registry URL; otherwise only authenticate for those given.

squaremo:
Image built from the most recent commit at quay.io/squaremo/flux:issue-539-native-ecr-auth-dd7e7cd

runningman84:
@squaremo I tried the new image but I still get logs like this:
``
flux/flux-79477cfbbf-t5xpx[flux]: ts=2019-01-03T16:04:36.487877004Z caller=warming.go:192 component=warmer canonical_name=602401143452.dkr.ecr.eu-central-1.amazonaws.com/eks/coredns auth="{map[602401143452.dkr.ecr.eu-central-1.amazonaws.com:<registry creds for AWS@602401143452.dkr.ecr.eu-central-1.amazonaws.com, from AWS API>]}" err="requesting tags: denied: User: arn:aws:sts::1234567890:assumed-role/eks-preprod20190102102137043800000008/i-06edfb64ff3d5add4 is not authorized to perform: ecr:ListImages on resource: arn:aws:ecr:eu-central-1:602401143452:repository/eks/coredns"
``
Is this a known problem for eks internal repos?

squaremo:
> @squaremo I tried the new image 

Thank you!

> ... but I still get logs like this:

Yep, this is the problem I was referring to in

> EKS uses images from some other ECR registry for some infrastructural services, which any old IAM user doesn't have full access to

The way to shut it up is to mention your own account ID in the argument `--aws-account-id`, which will mean it restricts scanning to just your own registry and ignores images elsewhere in ECR.

Having said that, it looks like it's the same account ID causing log noise in eu-central-1 as it is in eu-west-1, so I wonder if I could make the default to ignore that account ID ...

runningman84:
Yes please ignore it by default this seams to be an aws account of the eks team.

site/daemon.md
errordeveloper: > If you reckon it's clearer, I could change this argument name (and perhaps make the connection with registries in the help text).

Yes, I think `--aws-account-ids` seems clearer to me.

> Perhaps it should be the other way though -- a flag to explicitly switch AWS authentication off, and otherwise assume the ambient credentials are there to be used.

I think that would make sense, but I would leave that until a clear use-case for turning it off comes up, unless we already have boolean flags for controlling certain features, in which case this would be a matter of being consistent.

registry/aws.go
```diff
@@ -0,0 +1,216 @@
...
+// possible to refer to any registry from any cluster (although, being
+// AWS, there will be a cost incurred).
+
+func ImageCredsWithAWSAuth(lookup func() ImageCreds, logger log.Logger, config AWSRegistryConfig) (func() ImageCreds, error) {
+	awsCreds := NoCredentials()
+
+	if len(config.Regions) == 0 {
+		// this forces the AWS SDK to load config, so we can get the default region
+		sess := session.Must(session.NewSessionWithOptions(session.Options{
+			SharedConfigState: session.SharedConfigEnable,
```
errordeveloper: I think it's a good start, hopefully it will be sufficient long term and we won't need to add too many input variables, as it will always run in a container and won't need to support a host of configuration modes that `eksctl` has to support (environment variables, flags and `~/.aws/config` file with profiles etc).

registry/aws.go
```diff
@@ -0,0 +1,216 @@
...
+		return Credentials{}, time.Time{}, err
+	}
+	auths := make(map[string]creds)
+	expiry := time.Now().Add(defaultTokenValid)
+	for _, v := range ecrToken.AuthorizationData {
+		// Remove the https prefix
+		host := strings.TrimPrefix(*v.ProxyEndpoint, "https://")
+		creds, err := parseAuth(*v.AuthorizationToken)
+		if err != nil {
+			return Credentials{}, time.Time{}, err
```
errordeveloper: What will happen if we fail to refresh, when will it retry?

errordeveloper: I was looking in the wrong place, there are plenty of comments outside of this function.

site/faq.md
```diff
@@ -157,19 +157,18 @@ There are exceptions:
...
-For ECR (Elastic Container Registry), the credentials supplied by the
-environment are rotated, so it's not possible to supply a file with
-credentials ahead of time. See
-[weaveworks/flux#539](https://github.com/weaveworks/flux/issues/539) for
-workarounds.
+   used instead of image pull secrets:
+    - Google Container Registry works this way; Flux will
+      automatically attempt to use platform-provided credentials when
+      scanning images in GCR.
+    - (Amazon) Elastic Container Registry has its own authentication
```
errordeveloper: I would change this to
```
Amazon Elastic Container Registry (ECR) has its own authentication; Flux will attempt to scan for images in ECR, if EC2 instance it is running on has access to ECR.
```
Which also suggest that we might want to consider to support for ECR outside of EC2, but I don't think kubelet supports that kind of mode yet (IIRC, it was mentioned in the comment in kubelet code linked from here).

As far as EC2 itself is concerned, there is a possibility that user may wish to avoid using instance roles and pass credentials to flux directly (but we should check if kubelet supports that also). The reason someone may wish to do that is security, i.e. for the lack of real IAM roles for pods, someone may prefer to disallow access to ECR for any but the chose process in their cluster.

Just a few thoughts, perhaps we could open an issue for now.

site/faq.md
```diff
@@ -247,11 +246,12 @@ happen:
...
    present, it looks at `imagePullSecret`s attached to workloads,
-   service accounts and a Docker config file if you mount one into the fluxd container
-   (see the [command-line usage](./daemon.md)).
- - Flux doesn't know how to obtain registry credentials for ECR. A
-   workaround is described in
-   [weaveworks/flux#539](https://github.com/weaveworks/flux/issues/539#issuecomment-394588423)
+   service accounts, platform-provided credentials on GCP or AWS, and
+   a Docker config file if you mount one into the fluxd container (see
+   the [command-line usage](./daemon.md)).
+ - When using images in ECR, from AWS, the IAM account used to run the
```
errordeveloper: I'd rephrase as:
```
 - When using images in AWS ECR, the EC2 instance role of the node
  fluxd container runs on must have permissions to query the ECR registry
  or registries in question. In turns, this is a requirement for kubelet also,
  so if your cluster can run containers from ECR, Flux should work too.
```

cmd/fluxd/main.go
```diff
@@ -106,6 +106,10 @@ func main() {
 		registryTrace        = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
 		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "use HTTP for this image registry domain (e.g., registry.cluster.local), instead of HTTPS")
 
+		// AWS authentication
```
errordeveloper: LGTM. Plural vs singular is alway the question, and looks like fluxd is sticking to singular, e.g. `--registry-insecure-host` above.

registry/aws.go
squaremo: I've tried to leave space for mounting config files (e.g., by attempting to load config before using the ec2 metadata service). You are right that contexts etc. complicate this :-/ We'll see ..

site/faq.md
squaremo: I _think_ it would be possible (though ungainly) to use credentials on the host to grant kubelet access to ECR when not running in EC2. And, I think it'll be possible (with this PR) to make credentials available to flux too, by providing a `~/.aws/config`.

But I'd expect this to be well off the beaten path. I agree that it's important to set out the caveats early -- I'll try to rephrase.

site/faq.md
squaremo: Is it definitely the EC2 instance role that decides whether you have access? I confess I skipped lightly over the question of how the container gets any IAM permissions at all (and Just Made It Work TM). I'm happy to be more specific about the mechanism.

cmd/fluxd/main.go
squaremo: I was think perhaps I ought to make these more specific -- taken in isolation, the flag names suggest they might pertain to other AWS uses. Maybe:

 - `--registry-ecr-region`
 - `--registry-ecr-account-id`

(those are getting pretty long-winded, though the 80% scenario is you don't have to supply them)

squaremo:
@errordeveloper I've tweaked the wording in the FAQ (two places) -- is it any better, do you think?

cmd/fluxd/main.go
squaremo: After adding a flag for _excluding_ registries, so the EKS system registry can be excluded, I changed this to `--registry-ecr-include-id`, with the exclude flag being `--registry-ecr-exclude-id`.

squaremo:
Many thanks @errordeveloper and @hiddeco !

roffe:
awesome sauce!

PR #1622, opened 2018-12-30
Allow optional memcached.securityContext

isindir:
Allow optional memcached.securityContext in helm chart deployment template to mitigate PSP runAsNonRoot verification failure.

ie, Following value can be added if deploying on PSP enabled cluster:

```
memcached:
  securityContext:
    runAsUser: 65534
```

### DIFF
--- a/chart/flux/templates/memcached.yaml
+++ b/chart/flux/templates/memcached.yaml
@@ -49,6 +49,10 @@ spec:
       tolerations:
 {{ toYaml . | indent 8 }}
     {{- end }}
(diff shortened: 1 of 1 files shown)


stefanprodan:
This will work for memcached but Flux requires root

squaremo:
@isindir Is it still useful to run memcached as a non-root user if you can't do the same for fluxd?

isindir:
@squaremo I've realised it is not possible to run fluxd as non-root, I feel that less privileges exposed - the better. At the same time it is fully optional. Anyway, I don't have strong argument here.

PR #1623, opened 2018-12-30
Suggest --tls and --namespace args for helm when installing flux

u-phoria:


squaremo:
Thanks for this @u-phoria :pineapple: 

@stefanprodan Is this doc current or should we transplant these changes?

PR #1630, opened 2019-01-03
create minimal nsswitch.conf file in Docker image

### LINKED ISSUE #1627: Flux not honouring /etc/hosts file
We have a private repo where we store all the images. When flux is trying to read images from this private repo, what is happening is  that it is resolving the dns to a public ip and is trying to connect there despite specifying hostAliases in the yaml and I have crossverified that under /etc/hosts there is an entry for the docker repo. Could someone please help me out as to understand what is happening here .

2opremio:
As a result, /etc/hosts will be prioritized over DNS queries

This resolves the conflict between:
 * fluxd using netgo for static compilation. netgo reads nsswitch.conf to mimic glibc,
   defaulting to prioritize DNS queries over /etc/hosts if nsswitch.conf is missing:
   https://github.com/golang/go/issues/22846
 * Alpine not including a nsswitch.conf file. Since Alpine doesn't use glibc
   (it uses musl), maintainers argue that the need of nsswitch.conf is a Go bug:
   https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-354316460

Fixes #1627

### DIFF
--- a/docker/Dockerfile.flux
+++ b/docker/Dockerfile.flux
@@ -37,6 +37,15 @@ ENTRYPOINT [ "/sbin/tini", "--", "fluxd" ]
 COPY --from=quay.io/squaremo/kubeyaml:0.5.1 /usr/lib/kubeyaml /usr/lib/kubeyaml/
 ENV PATH=/bin:/usr/bin:/usr/local/bin:/usr/lib/kubeyaml
 
+# Create minimal nsswitch.conf file to prioritize the usage of /etc/hosts over DNS queries.
+# This resolves the conflict between:
+# * fluxd using netgo for static compilation. netgo reads nsswitch.conf to mimic glibc,
+#   defaulting to prioritize DNS queries over /etc/hosts if nsswitch.conf is missing:
+#   https://github.com/golang/go/issues/22846
+# * Alpine not including a nsswitch.conf file. Since Alpine doesn't use glibc
+#   (it uses musl), maintainers argue that the need of nsswitch.conf is a Go bug:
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.flux
```diff
@@ -37,6 +37,15 @@ ENTRYPOINT [ "/sbin/tini", "--", "fluxd" ]
...
 
+# Create minimal nsswitch.conf file to prioritize the usage of /etc/hosts over DNS queries.
+# This resolves the conflict between:
+# * fluxd using netgo for static compilation. netgo reads nsswitch.conf to mimic glibc,
+#   defaulting to prioritize DNS queries over /etc/hosts if nsswitch.conf is missing:
+#   https://github.com/golang/go/issues/22846
+# * Alpine not including a nsswitch.conf file. Since Alpine doesn't use glibc
+#   (it uses musl), maintainers argue that the need of nsswitch.conf is a Go bug:
+#   https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-354316460
+RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
```
errordeveloper: I would write out the entire file as it is, and make this a `COPY ./nsswitch.conf /etc/nsswitch.conf`. Given the context, I doubt it would eventually appear in the Alpine image.

squaremo: That's a fair point; on the other hand it doesn't hurt to be careful.

hiddeco: Imho the current one-liner is great; I doubt we will ever make changes to the `nsswitch.conf` content so there is no reason to maintain a separate file to ease modifications.

PR #1649, opened 2019-01-12
Conditionally create the flux-helm-tls-ca-config configmap

stephenmoloney:
What does this commit/MR/PR do?

- Conditionally create the flux-helm-tls-ca-config configmap
- Only create the configmap if caContent actually exists

Why is this commit/MR/PR needed?

- Useful for a flux upgrade when you don't want to specify the
caContent into `values.yaml` and don't want the existing configmap
value overriden with an empty value.

### DIFF
--- a/CHANGELOG-helmop.md
+++ b/CHANGELOG-helmop.md
@@ -1,3 +1,22 @@
+## Version (Date)
+
+### Bug fixes
+
+  - 
+    
+### Improvements
+
+  - Only create the `flux-helm-tls-ca-config` file if `.Values.helmOperator.tls.caContent` exists.
+    Useful when doing flux upgrades but do not happen to know or want to specify 
+    the `caContent` in `values.yaml`. Otherwise, the existing caContent will be overriden with an
+    empty value. 
+    [weaveworks/flux#1649](https://github.com/weaveworks/flux/pull/1649)
+    
+
+### Thanks
+
+Thanks to
+
 ## 0.5.2 (2018-12-20)
 
 ### Bug fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG-helmop.md
```diff
@@ -1,3 +1,22 @@
+## Version (Date)
+
+### Bug fixes
+
+  - 
+    
+### Improvements
```
stefanprodan: This is the wrong change log, please move it to chart/flux/CHANGELOG.md

stephenmoloney:
I'll rebase this at end if accepted.

chart/flux/templates/helm-tls.yaml
```diff
@@ -1,5 +1,6 @@
 {{- if .Values.helmOperator.tls.enable -}}
 {{- if .Values.helmOperator.tls.verify -}}
+{{- if .Values.helmOperator.tls.caContent }}
```
stefanprodan: Can you please change this to a single if, with `if and `

stephenmoloney: Hmm... 
I've just tested 3 and 4 - both seem to work ok. 

```
helm template --set helmOperator.tls.enable=true --set helmOperator.tls.verify=true --set helmOperator.tls.caContent="test" ./ -x /home/u2/Documents/repos/opensource/flux/chart/flux/templates/helm-tls.yaml
```

stefanprodan: 3 looks ok to me, if that works please make the change

stephenmoloney: Not sure why I thought 3 wouldn't work. I'll make the PR with number 3 as it seems to be working ok as outlined in command above.
@stefanprodan

Issue #1647, opened 2019-01-11
no matches for kind "CustomResourceDefinition"


Tusharsb:
When I try to Apply the Helm Release CRD with following command 

https://github.com/weaveworks/flux/blob/0d1a400387346fea43c278b6e9abccc5df6e51de/site/helm-get-started.md#L61

I get an error stating that 
```
error: unable to recognize "https://raw.githubusercontent.com/weaveworks/flux/master/deploy-helm/flux-helm-release-crd.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

Following is my kubectl version
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.0", GitCommit:"fc32d2f3698e36b93322a3465f63a14e9f0eaead", GitTreeState:"clean", BuildDate:"2018-03-26T16:55:54Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.0", GitCommit:"fc32d2f3698e36b93322a3465f63a14e9f0eaead", GitTreeState:"clean", BuildDate:"2018-03-26T16:44:10Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

And `apiextensions.k8s.io/v1beta1` also seems to suggest that CustomResourceDefinition is there in it
```
$ curl http://localhost:8080/apis/apiextensions.k8s.io/v1beta1 
{
  "kind": "APIResourceList",
  "apiVersion": "v1",
  "groupVersion": "apiextensions.k8s.io/v1beta1",
  "resources": [
    {
      "name": "customresourcedefinitions",
      "singularName": "",
      "namespaced": false,
      "kind": "CustomResourceDefinition",
      "verbs": [
        "create",
        "delete",
        "deletecollection",
        "get",
        "list",
        "patch",
        "update",
        "watch"
      ],
      "shortNames": [
        "crd",
        "crds"
      ]
    },
    {
      "name": "customresourcedefinitions/status",
      "singularName": "",
      "namespaced": false,
      "kind": "CustomResourceDefinition",
      "verbs": [
        "update"
      ]
    }
  ]
}%                                 
```
Am I doing something wrong? Or have I run into a bug ? I was trying to follow https://github.com/weaveworks/flux/blob/master/site/helm-get-started.md

[...]

PR #1650, opened 2019-01-13
min ver of kubernetes 1.12 specified

Tusharsb:
advice user to use version `1.12` or higher. It was discovered during. #1647 that version 1.10 would not work

### DIFF
--- a/site/helm-get-started.md
+++ b/site/helm-get-started.md
@@ -23,9 +23,9 @@ README](../chart/flux/README.md).
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


hiddeco:
> The CRD spec will work with Kubernetes >= 1.11

Even better: the spec also works fine with Kubernetes `1.10.x`. Just applied the CRD on a fresh minikube setup (`minikube start --kubernetes-version=v1.10.0 [additional flags]`) and saw no errors.

PR #1652, opened 2019-01-13
Add instructions for Weave Cloud helm-op install

stefanprodan:
- add helm-op deployment definition customised for Weave Cloud

Tested on GKE with the latest Weave Cloud agents.


site/helm-operator.md
```diff
@@ -221,3 +221,23 @@ metadata:
...
+kubectl apply -f ${REPO}/deploy-helm/weave-cloud-helm-operator-deployment.yaml
+```
+
+Check the operator logs with:
+
+```bash
+kubectl -n weave logs deployment/flux-helm-operator -f
+```
+
+Note that the above instructions are assuming that Tiller is deployed in the `kube-system` namespace without TLS.
```
dholbach:
```suggestion
**Note:** that the above instructions are assuming that Tiller is deployed in the `kube-system` namespace without TLS.
```

dholbach: Might be better to make this bold as it's an important requirement.

PR #1659, opened 2019-01-14
Allow images to be excluded from scanning

### LINKED ISSUE #1509: Add support to disable image tag caching.
This would be a nice to have.

Right now, our logs for Flux contain a lot of error messages from the `warmer` component:

```
ts=2018-11-13T18:56:00.919008379Z caller=warming.go:162 component=warmer canonical_name=[redacted] auth={map[]} err="requesting tags: Get [redacted] no basic auth credentials"
ts=2018-11-13T18:56:00.921823202Z caller=warming.go:162 component=warmer canonical_name=[redacted] auth={map[]} err="requesting tags: Get [redacted] no basic auth credentials"
ts=2018-11-13T18:56:49.291389931Z caller=warming.go:162 component=warmer canonical_name=[redacted] auth={map[]} err="requesting tags: Get [redacted] no basic auth credentials"
```

### LINKED ISSUE #1654: Exclude domains from image metadata scanning
While implementing #1619 I realised it would be quite easy to add a filter to the domains that get scanned for image metadata. This would be useful for

 - stopping fluxd scanning the system images
 - switching scanning off altogether

and no doubt a few requirements specific to individual cases.

stefanprodan:
- add registry-exclude-image flag that accepts a list of glob expressions
- exclude Kubernetes system images by default `k8s.gcr.io/*`
- add exclusion tests
- add usage examples to docs

Fix #1654 images can be excluded by registry domain eg.:
 `--registry-exclude-image=gcr.io/*,quay.io/*`

Fix #1509 it is now possible to disable the image cache completely with:
 `--registry-exclude-image=*`

### DIFF
--- a/cluster/kubernetes/images.go
+++ b/cluster/kubernetes/images.go
@@ -145,7 +145,7 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 // filterImages returns an image list from a pod spec
 // by removing those matching the exclusion list
 func (c *Cluster) filterImages(podTemplate apiv1.PodTemplateSpec) []image.Name {
-	images := []image.Name{}
+	var images []image.Name
 
 	for _, container := range podTemplate.Spec.InitContainers {
 		r, err := image.ParseRef(container.Image)
@@ -177,12 +177,11 @@ func (c *Cluster) filterImages(podTemplate apiv1.PodTemplateSpec) []image.Name {
 		for _, exp := range c.imageExcludeList {
 			if glob.Glob(exp, imageName) {
 				include = false
+				break
 			}
 		}
 		if include {
 			result = append(result, imageID)
-		} else {
-			//c.logger.Log("debug", fmt.Sprintf("image %s excluded", imageName))
 		}
 	}
 
--- a/cluster/kubernetes/images.go
+++ b/cluster/kubernetes/images.go
@@ -2,10 +2,10 @@ package kubernetes
 
 import (
 	"fmt"
-	"github.com/ryanuber/go-glob"
 
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
+	"github.com/ryanuber/go-glob"
 	apiv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -15,7 +15,18 @@ import (
 	"github.com/weaveworks/flux/registry"
 )
 
-func mergeCredentials(log func(...interface{}) error, client extendedClient, namespace string, podTemplate apiv1.PodTemplateSpec, imageCreds registry.ImageCreds, seenCreds map[string]registry.Credentials) {
+func mergeCredentials(log func(...interface{}) error,
+	filterImages func(podTemplate apiv1.PodTemplateSpec) []image.Name,
+	client extendedClient,
+	namespace string, podTemplate apiv1.PodTemplateSpec,
+	imageCreds registry.ImageCreds,
+	seenCreds map[string]registry.Credentials) {
+	// filter the images based on the exclusion list
+	images := filterImages(podTemplate)
+	if len(images) < 1 {
+		return
+	}
+
 	creds := registry.NoCredentials()
 	var imagePullSecrets []string
 	saName := podTemplate.Spec.ServiceAccountName
@@ -82,21 +93,8 @@ func mergeCredentials(log func(...interface{}) error, client extendedClient, nam
 	}
 
 	// Now create the service and attach the credentials
-	for _, container := range podTemplate.Spec.Containers {
-		r, err := image.ParseRef(container.Image)
-		if err != nil {
-			log("err", err.Error())
-			continue
-		}
-		imageCreds[r.Name] = creds
-	}
-	for _, container := range podTemplate.Spec.InitContainers {
-		r, err := image.ParseRef(container.Image)
-		if err != nil {
-			log("err", err.Error())
-			continue
-		}
-		imageCreds[r.Name] = creds
+	for _, image := range images {
+		imageCreds[image] = creds
 	}
 }
 
@@ -126,7 +124,7 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 			imageCreds := make(registry.ImageCreds)
 			for _, podController := range podControllers {
 				logger := log.With(c.logger, "resource", flux.MakeResourceID(ns.Name, kind, podController.name))
-				mergeCredentials(logger.Log, c.client, ns.Name, podController.podTemplate, imageCreds, seenCreds)
+				mergeCredentials(logger.Log, c.filterImages, c.client, ns.Name, podController.podTemplate, imageCreds, seenCreds)
 			}
 
 			// Merge creds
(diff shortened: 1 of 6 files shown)


cluster/kubernetes/images.go
```diff
@@ -140,5 +141,16 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 		}
 	}
 
+	// remove images based on the glob exclusion list
```
squaremo: I wonder if it is possible to exit (continue) early if an image it be excluded, rather than do all the work then undo some of it. The way `mergeCredentials` works is to get all the secrets, then record them against each images used by the workload. Exiting early would mean testing whether _any_ image in the workload is to be included, and only continuing if that's the case. In other words, compile the list of images in a workload that are to be included, first, and if len==0, exit.

This adds a bit of complication to `mergeCredentials`, but not much more than it removes from here; with the benefit that it'll reduce some Kubernetes API use (no use looking up image secrets for images we don't care about).

If it turns out that it's much simpler to do it here, I'd consider putting exclusion in a separate ImageFetch wrapper, like the ECR auth bit, instead.

cmd/fluxd/main.go
```diff
@@ -105,6 +105,7 @@ func main() {
 		registryBurst        = fs.Int("registry-burst", defaultRemoteConnections, "maximum number of warmer connections to remote and memcache")
 		registryTrace        = fs.Bool("registry-trace", false, "output trace of image registry requests to log")
 		registryInsecure     = fs.StringSlice("registry-insecure-host", []string{}, "use HTTP for this image registry domain (e.g., registry.cluster.local), instead of HTTPS")
+		registryExcludeImage = fs.StringSlice("registry-exclude-image", []string{"k8s.gcr.io/*"}, "Do not scan images that match these glob expressions; the default is to exclude the 'k8s.gcr.io/*' images")
```
squaremo: the prevailing convention is to not capitalise the usage message

cluster/kubernetes/images.go
squaremo: In any case, it'd be good to see some tests of the exclusion -- both as a guard against regressions, and as docs of how it's intended to work.

stefanprodan: I've moved the exclusion logic to `mergeCredentials` and added tests.

stephenmoloney:
just wondering, when is it desirable to not scan images in a registry?

stefanprodan:
@stephenmoloney see @alanjcastonguay comment on #1654

stefanprodan:
@squaremo should we exclude `quay.io/weaveworks/flux*` and `quay.io/weaveworks/helm-operator*` by default?

squaremo:
> should we exclude `quay.io/weaveworks/flux*` and `quay.io/weaveworks/helm-operator*` by default?

No; people routinely automate those (e.g., Weaveworks).

stefanprodan:
@squaremo I've addressed your comments in the latest commits and I've also made changes to the docs. Let me know if I've missed something. Thanks

cluster/kubernetes/images.go
```diff
@@ -142,3 +141,50 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 
 	return allImageCreds
 }
+
+// filterImages returns an image list from a pod spec
+// by removing those matching the exclusion list
+func (c *Cluster) filterImages(podTemplate apiv1.PodTemplateSpec) []image.Name {
+	images := []image.Name{}
```
squaremo: `var images []image.Name` is very slightly more idiomatic.

cluster/kubernetes/images.go
```diff
@@ -142,3 +141,50 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
...
+		return images
+	}
+
+	result := []image.Name{}
+	for _, imageID := range images {
+		imageName := imageID.CanonicalName().Name.String()
+		include := true
+		for _, exp := range c.imageExcludeList {
+			if glob.Glob(exp, imageName) {
+				include = false
```
squaremo: There's no point in continuing if it's already false.

cluster/kubernetes/images.go
```diff
@@ -142,3 +141,50 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
...
+			}
+		}
+		if include {
+			result = append(result, imageID)
+		} else {
```
squaremo: Remove this branch if it's not actually going to do anything.

cluster/kubernetes/images_test.go
```diff
@@ -66,11 +67,51 @@ func TestMergeCredentials(t *testing.T) {
 	client := extendedClient{clientset, nil}
 
 	creds := registry.ImageCreds{}
-	mergeCredentials(noopLog, client, ns, spec, creds, make(map[string]registry.Credentials))
+	cluster := NewCluster(clientset, nil, nil, nil, log.NewNopLogger(), []string{}, []string{})
```
squaremo: Hrmm, it's a bit unwieldy that you have to create a cluster in order to filter the images. Maybe construct the filter independently of the cluster.

deploy/memcache-dep.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-# You can optionally deploy memcache, for the Flux daemon to cache
+# The memcached deployment is used by Flux to cache
```
squaremo: This does not seem to be a related change.

site/faq.md
```diff
@@ -29,11 +29,11 @@ menu_order: 60
   * [Can I restrict the namespaces that Flux can see or operate on?](#can-i-restrict-the-namespaces-that-flux-can-see-or-operate-on)
   * [Can I change the namespace Flux puts things in by default?](#can-i-change-the-namespace-flux-puts-things-in-by-default)
   * [Can I temporarily make Flux ignore a deployment?](#can-i-temporarily-make-flux-ignore-a-deployment)
+  * [Can I disable Flux registry scanning?](#can-i-disable-flux-registry-scanning)
 - [Flux Helm Operator questions](#flux-helm-operator-questions)
   * [I'm using SSL between Helm and Tiller. How can I configure Flux to use the certificate?](#im-using-ssl-between-helm-and-tiller-how-can-i-configure-flux-to-use-the-certificate)
-  * [I've deleted a FluxHelmRelease file from Git. Why is the Helm release still running on my cluster?](#ive-deleted-a-fluxhelmrelease-file-from-git-why-is-the-helm-release-still-running-on-my-cluster)
+  * [I've deleted a HelmRelease file from Git. Why is the Helm release still running on my cluster?](#ive-deleted-a-helmrelease-file-from-git-why-is-the-helm-release-still-running-on-my-cluster)
```
squaremo: Unrelated change. Not unwelcome, but belongs in another PR.

cluster/kubernetes/images_test.go
squaremo: That way you can make it a no-op if there are no exclusions, and you can build the list of included images in one step in mergeCredentials (rather than building a list of all of them, then filtering it down).

deploy/memcache-dep.yaml
stefanprodan: Because of that "optionally" we've got users on Slack that thought Flux will work without memcached. But even with `--registry-exclude-image=*` memcached is mandatory.

squaremo: I'm not arguing that the sentence is correct, just that you should keep the PR on point.

stefanprodan:
@squaremo I've removed the docs changes from this PR and I've addressed your comments in the latest commit. As for making the tests work without cluster I'm not really sure how to do that without adding the exclusion list on all functions... I would leave that for another PR if it's ok with you.

cluster/kubernetes/images_test.go
squaremo: What I meant is this: given a predicate `includeImage(imageName) bool`, you can step through a pod template, testing each image extracted against the predicate as you go. There's no need to build the list of images, then run through it again to filter it.

How do you get such a predicate? You can just construct it -- you have the algorithm right here -- and it need not depend on the cluster, since all that's doing is providing the list of exclusions.

cluster/kubernetes/images.go
```diff
@@ -14,7 +15,18 @@ import (
...
 
-func mergeCredentials(log func(...interface{}) error, client extendedClient, namespace string, podTemplate apiv1.PodTemplateSpec, imageCreds registry.ImageCreds, seenCreds map[string]registry.Credentials) {
+func mergeCredentials(log func(...interface{}) error,
+	filterImages func(podTemplate apiv1.PodTemplateSpec) []image.Name,
+	client extendedClient,
+	namespace string, podTemplate apiv1.PodTemplateSpec,
+	imageCreds registry.ImageCreds,
+	seenCreds map[string]registry.Credentials) {
+	// filter the images based on the exclusion list
+	images := filterImages(podTemplate)
```
squaremo: All that happens to `filterImages` is that it's applied to one of the other arguments -- you may as well supply the result.

squaremo:
That last commit is a nice refinement :+1:

PR #1664, opened 2019-01-15
Add option to limit helm-op to a single namespace

### LINKED ISSUE #1406: Feature: Restrict FHR scanning to the namespace flux is installed in
Our observations lead us to believe that flux installed in namespace 'alpha' will pick up FHRs in namespace 'beta'. This is probably by design.

We would like the option to restrict flux from only finding FHRs in the namespace that flux is installed. We would like to configure this option using helm values.

We want to install multiple instances of flux in different namespaces and we don't want each instance of flux to find all FHR resources in all namespaces.

stefanprodan:
- add namespace restriction to Kubernetes informer and chart sync component
- by default all namespaces are used 

Fix #1406

### DIFF
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -21,6 +21,8 @@ helmOperator:
   repository: quay.io/weaveworks/helm-operator
   tag: 0.5.3
   pullPolicy: IfNotPresent
+  # Limit the operator scope to a single namespace
+  namespace:
   # Update dependencies for charts
   updateChartDeps: true
   # Log the diff when a chart release diverges
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -22,7 +22,7 @@ helmOperator:
   tag: 0.5.3
   pullPolicy: IfNotPresent
   # Limit the operator scope to a single namespace
-  namespace:
+  allowNamespace:
   # Update dependencies for charts
   updateChartDeps: true
   # Log the diff when a chart release diverges
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -106,8 +106,8 @@ type clone struct {
 }
 
 type ChartChangeSync struct {
-	logger log.Logger
 	Polling
+	logger     log.Logger
 	kubeClient kubernetes.Clientset
 	ifClient   ifclientset.Clientset
 	release    *release.Release
@@ -117,9 +117,11 @@ type ChartChangeSync struct {
 
 	clonesMu sync.Mutex
 	clones   map[string]clone
+
+	namespace string
 }
 
-func New(logger log.Logger, polling Polling, clients Clients, release *release.Release, config Config) *ChartChangeSync {
+func New(logger log.Logger, polling Polling, clients Clients, release *release.Release, config Config, namespace string) *ChartChangeSync {
 	return &ChartChangeSync{
 		logger:     logger,
 		Polling:    polling,
@@ -129,6 +131,7 @@ func New(logger log.Logger, polling Polling, clients Clients, release *release.R
 		config:     config.WithDefaults(),
 		mirrors:    git.NewMirrors(),
 		clones:     make(map[string]clone),
+		namespace:  namespace,
 	}
 }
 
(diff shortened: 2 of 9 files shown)


2opremio:
I think it would make sense to align the argument name to that of `fluxd` .

In https://github.com/weaveworks/flux/issues/1471 @squaremo suggested using `--kubenernetes-allow-namespace`

squaremo:
Is there a reason for it being no restriction vs exactly one namespace, rather than no restriction vs some namespaces?

stefanprodan:
@squaremo the `NewSharedInformerFactoryWithOptions` takes one namespace only, we could have many informers but that will require greater changes to the operator logic. Anyway most requests I've seen from our users is to be able to run one Flux/helm-op/tiller per namespace. This PR makes it possible.

errordeveloper:
I also feel that it'd be a good idea to match what Flux does, i.e. a list of namespaces, however Tiller can only be used in a context of a single names or the entire cluster, and I don't think making operator support 
 multiple Tillers in different namespaces would make much sense.

chart/flux/values.yaml
```diff
@@ -21,6 +21,8 @@ helmOperator:
   repository: quay.io/weaveworks/helm-operator
   tag: 0.5.3
   pullPolicy: IfNotPresent
+  # Limit the operator scope to a single namespace
+  namespace:
```
errordeveloper: should this be `namespace: ""`?

stefanprodan: If the namespace is not set then the default from the `pflag` is applied. I've added the namespace in the values.yaml to document the option not to provide a default value for it.

stefanprodan:
@2opremio if I use the same flag as Flux `--kubenernetes-allow-namespace` people will think helm-op accepts multiple namespaces. 
@squaremo what's your opinion on this? should I rename `--namespace` to `--kubenernetes-allow-namespace` ?

stefanprodan:
@stephenmoloney if you set helm-op namespace to `dev` then only the HelmReleases in that namespace will be installed/upgraded; all the others will be ignored,. Flux will still sync them all but there will be no actual releases in Tiller except for the ones in that namespace. Same for `adm` namespace.

stephenmoloney:
@stefanprodan  ok, thanks.

So is the motivation for this PR to allow having one flux/helm operator per namespace ?

Also, would it make sense to allow the option of a list of namespaces ?

stefanprodan:
@stephenmoloney the motivation is to be able to run one flux/helm-op/tiller per namespace and since Tiller can't target a list of namespaces it doesn't make sense for the helm-op to do so.

2opremio:
However flux is soon going to support proper namespace filtering soon and it seems that mutiple namespaces are (or were) in the plans for Tiller too  https://github.com/helm/helm/issues/838#issuecomment-228118873

2opremio:
See https://github.com/helm/helm/issues/2060#issuecomment-287164881 as well

stephenmoloney:
> @stephenmoloney the motivation is to be able to run one flux/helm-op/tiller per namespace and since Tiller can't target a list of namespaces it doesn't make sense for the helm-op to do so.

@stefanprodan  thanks.

stefanprodan:
@2opremio Tiller is going away with Helm v3. Once v3 is out and Tiller is gone we can revisit the multi-namespace issue.

integrations/helm/chartsync/chartsync.go
```diff
@@ -403,6 +406,11 @@ func (chs *ChartChangeSync) DeleteRelease(fhr fluxv1beta1.HelmRelease) {
 
 // getNamespaces gets current kubernetes cluster namespaces
 func (chs *ChartChangeSync) getNamespaces() ([]string, error) {
+	if chs.namespace != "" {
```
squaremo: The other branch here asks the API server, whereas this just trusts what it's been told. This might break assumptions made by callers of this code, e.g., that the namespace(s) returned definitely exist.

integrations/helm/status/status.go
```diff
@@ -56,14 +58,23 @@ bail:
 			break bail
 		case <-ticker.C:
 		}
-		// Look up HelmReleases
-		namespaces, err := a.kube.CoreV1().Namespaces().List(metav1.ListOptions{})
-		if err != nil {
-			logErr = err
-			break bail
+		namespaces := []string{}
```
squaremo: `var namespaces []string`

integrations/helm/chartsync/chartsync.go
stefanprodan: This gets called by `getCustomResources` and if the namespace is missing the error is caught when fetching a HR. The error will tell you that there is no HR in that namespace. Also if the namespace is missing, Flux will not apply the HR.

squaremo: That may be how it works now. Can you guarantee it will always be that way?
Actually I have a feeling that there's only one call site for this (`getCustomResources`); if that's the case, you may as well inline it, and my objection won't matter.

stefanprodan: I've deleted `getNamespaces` and moved the code to `getCustomResources`

kraney:
What about this use case - I want a "cluster owner" flux / op / helm that monitors all, and then 'namespace owner' flux / op / helm that watches a specific namespace. I don't see a practical way to make the cluster-level operator ignore that namespace & leave it to the helm that lives there.

Issue #1471, opened 2018-10-25
Extend --kubernetes-namespace-whitelist to all operations


squaremo:
At present, the experimental flag `--kubernetes-namespace-whitelist` controls which workloads appear in ListControllers and ListImages. This is useful for restricting the view of the cluster.

The usual expectation is that fluxd's operations can be entirely limited to the namespaces given; i.e., that the restriction appliees also to the resources that will be synced, workloads that are scanned for images, workloads that can be updated, and so on.

One tricky thing here is what to do with syncing resources that are not namespaced. I can think of a couple of designs:

 - Apply namespaces if they're on the whitelist; otherwise, ignore cluster-level resources _unless_ fluxd has not been given a whitelist. A problem with this is: how do you then sync cluster-level things? A fluxd without a whitelist will try to sync everything ..

 - Apply namespaces if they're on the whitelist; otherwise, always attempt to apply cluster-level resources. This has the benefit of being simple, but you may get repeating errors if you have restricted fluxd with RBAC.

 - Have a flag determining whether fluxd will try to sync cluster-level resources, and succeed or fail based on RBAC. This at least means you can tell one fluxd to do the cluster-level things (and give it permission).

 - More hand-wavingly -- specify the whitelist in a way that is a bit closer to RBAC, so you can carve out exactly what you want to be able to operate on.

2opremio:
I would go for:

> Have a flag determining whether fluxd will try to sync cluster-level resources, and succeed or fail based on RBAC. This at least means you can tell one fluxd to do the cluster-level things (and give it permission).

But I would do it gradually (at least in different PRs):

1. Apply namespace filtering to everything which has one (i.e. always include cluster-level resources)
2. Take RBACs into account to avoid errors
3. Add a flag to filter global resources

However, as far as I recall (I haven't done k8s stuff in a few months, so correct me if I am wrong) there aren't so many global resources, so I wonder if `flux.weave.works/ignore` would be enough to avoid implementing (3 and maybe 2). I am assuming namespace whitelisting is meant to be used as a filtering mechanism and not as a security policy. If the latter were true then the `ignore` label won't be enough.

@squaremo Thoughts?

squaremo:
@2opremio I like that approach -- make it consistent first, then more sensitive to RBAC.

[...]

PR #1668, opened 2019-01-17
Extend namespace filtering to all operations on namespaced resources

2opremio:
Addresses part of #1471 

* Deprecate and rename flag `--kubernetes-namespace-whitelist` to ~`--kubernetes-allow-namespace`~ `--k8s-allow-namespace`
* Honor `--k8s-allow-namespace` for namespaced kubernetes resources in all flux operations.
* cluster-global resources cluster-global resources (i.e. resources without an associated namespace) won't be filtered out by this change (they will keep being synced).

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -27,6 +27,7 @@ type Cluster interface {
 	// Get all of the services (optionally, from a specific namespace), excluding those
 	AllWorkloads(maybeNamespace string) ([]Workload, error)
 	SomeWorkloads([]flux.ResourceID) ([]Workload, error)
+	IsAllowedResource(flux.ResourceID) bool
 	Ping() error
 	Export() ([]byte, error)
 	Sync(SyncSet) error
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -157,7 +157,7 @@ func (c *Cluster) SomeWorkloads(ids []flux.ResourceID) (res []cluster.Workload, 
 // AllWorkloads returns all workloads in allowed namespaces matching the criteria; that is, in
 // the namespace (or any namespace if that argument is empty)
 func (c *Cluster) AllWorkloads(namespace string) (res []cluster.Workload, err error) {
-	namespaces, err := c.getAllowedNamespaces()
+	namespaces, err := c.getAllowedAndExistingNamespaces()
 	if err != nil {
 		return nil, errors.Wrap(err, "getting namespaces")
 	}
@@ -217,7 +217,7 @@ func (c *Cluster) Ping() error {
 func (c *Cluster) Export() ([]byte, error) {
 	var config bytes.Buffer
 
-	namespaces, err := c.getAllowedNamespaces()
+	namespaces, err := c.getAllowedAndExistingNamespaces()
 	if err != nil {
 		return nil, errors.Wrap(err, "getting namespaces")
 	}
@@ -266,11 +266,11 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
 	return publicKey, nil
 }
 
-// getAllowedNamespaces returns a list of namespaces that the Flux instance is expected
-// to have access to and can look for resources inside of.
+// getAllowedAndExistingNamespaces returns a list of existing namespaces that
+// the Flux instance is expected to have access to and can look for resources inside of.
 // It returns a list of all namespaces unless an explicit list of allowed namespaces
 // has been set on the Cluster instance.
-func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
+func (c *Cluster) getAllowedAndExistingNamespaces() ([]apiv1.Namespace, error) {
 	if len(c.allowedNamespaces) > 0 {
 		nsList := []apiv1.Namespace{}
 		for _, name := range c.allowedNamespaces {
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -95,22 +95,22 @@ type Cluster struct {
 	syncErrors   map[flux.ResourceID]error
 	muSyncErrors sync.RWMutex
 
-	nsWhitelist       []string
-	nsWhitelistLogged map[string]bool // to keep track of whether we've logged a problem with seeing a whitelisted ns
+	allowedNamespaces []string
+	loggedAllowedNS   map[string]bool // to keep track of whether we've logged a problem with seeing an allowed namespace
 
 	imageExcludeList []string
 	mu               sync.Mutex
 }
 
 // NewCluster returns a usable cluster.
-func NewCluster(client ExtendedClient, applier Applier, sshKeyRing ssh.KeyRing, logger log.Logger, nsWhitelist []string, imageExcludeList []string) *Cluster {
+func NewCluster(client ExtendedClient, applier Applier, sshKeyRing ssh.KeyRing, logger log.Logger, allowedNamespaces []string, imageExcludeList []string) *Cluster {
 	c := &Cluster{
 		client:            client,
 		applier:           applier,
 		logger:            logger,
 		sshKeyRing:        sshKeyRing,
-		nsWhitelist:       nsWhitelist,
-		nsWhitelistLogged: map[string]bool{},
+		allowedNamespaces: allowedNamespaces,
+		loggedAllowedNS:   map[string]bool{},
 		imageExcludeList:  imageExcludeList,
 	}
 
@@ -264,22 +264,22 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
 
 // getAllowedNamespaces returns a list of namespaces that the Flux instance is expected
 // to have access to and can look for resources inside of.
-// It returns a list of all namespaces unless a namespace whitelist has been set on the Cluster
-// instance, in which case it returns a list containing the namespaces from the whitelist
-// that exist in the cluster.
+// It returns a list of all namespaces unless an explicit list of allowed namespaces
+// has been set on the Cluster instance.
 func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
-	if len(c.nsWhitelist) > 0 {
+	if len(c.allowedNamespaces) > 0 {
 		nsList := []apiv1.Namespace{}
-		for _, name := range c.nsWhitelist {
+		for _, name := range c.allowedNamespaces {
 			ns, err := c.client.CoreV1().Namespaces().Get(name, meta_v1.GetOptions{})
 			switch {
 			case err == nil:
-				c.nsWhitelistLogged[name] = false // reset, so if the namespace goes away we'll log it again
+				c.loggedAllowedNS[name] = false // reset, so if the namespace goes away we'll log it again
 				nsList = append(nsList, *ns)
 			case apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err) || apierrors.IsNotFound(err):
-				if !c.nsWhitelistLogged[name] {
-					c.logger.Log("warning", "whitelisted namespace inaccessible", "namespace", name, "err", err)
-					c.nsWhitelistLogged[name] = true
+				if !c.loggedAllowedNS[name] {
+					c.logger.Log("warning", "cannot access namespace set as allowed",
+						"namespace", name, "err", err)
+					c.loggedAllowedNS[name] = true
 				}
 			default:
 				return nil, err
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/weaveworks/flux"
 	"github.com/weaveworks/flux/cluster"
+	"github.com/weaveworks/flux/cluster/kubernetes/resource"
 	fhrclient "github.com/weaveworks/flux/integrations/client/clientset/versioned"
 	"github.com/weaveworks/flux/ssh"
 )
@@ -119,12 +120,15 @@ func NewCluster(client ExtendedClient, applier Applier, sshKeyRing ssh.KeyRing, 
 
 // --- cluster.Cluster
 
-// SomeWorkloads returns the workloads named, missing out any that aren't
-// accessible in the cluster. They do not necessarily have to be returned
-// in the order requested.
+// SomeWorkloads returns the workloads named, missing out any that don't
+// exist in the cluster or aren't in an allowed namespace.
+// They do not necessarily have to be returned in the order requested.
 func (c *Cluster) SomeWorkloads(ids []flux.ResourceID) (res []cluster.Workload, err error) {
 	var workloads []cluster.Workload
 	for _, id := range ids {
+		if !c.IsAllowedResource(id) {
+			continue
+		}
 		ns, kind, name := id.Components()
 
 		resourceKind, ok := resourceKinds[kind]
@@ -150,7 +154,7 @@ func (c *Cluster) SomeWorkloads(ids []flux.ResourceID) (res []cluster.Workload, 
 	return workloads, nil
 }
 
-// AllWorkloads returns all workloads matching the criteria; that is, in
+// AllWorkloads returns all workloads in allowed namespaces matching the criteria; that is, in
 // the namespace (or any namespace if that argument is empty)
 func (c *Cluster) AllWorkloads(namespace string) (res []cluster.Workload, err error) {
 	namespaces, err := c.getAllowedNamespaces()
@@ -277,7 +281,7 @@ func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
 				nsList = append(nsList, *ns)
 			case apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err) || apierrors.IsNotFound(err):
 				if !c.loggedAllowedNS[name] {
-					c.logger.Log("warning", "cannot access namespace set as allowed",
+					c.logger.Log("warning", "cannot access allowed namespace",
 						"namespace", name, "err", err)
 					c.loggedAllowedNS[name] = true
 				}
@@ -293,6 +297,32 @@ func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
 		return nil, err
 	}
 	return namespaces.Items, nil
+}
+
+func (c *Cluster) IsAllowedResource(id flux.ResourceID) bool {
+	if len(c.allowedNamespaces) == 0 {
+		// All resources are allowed when all namespaces are allowed
+		return true
+	}
+
+	namespace, kind, name := id.Components()
+	namespaceToCheck := namespace
+
+	if namespace == resource.ClusterScope {
+		// All cluster-scoped resources (not namespaced) are allowed ...
+		if kind != "namespace" {
+			return true
+		}
+		// ... except namespaces themselves, whose name needs to be explicitly allowed
+		namespaceToCheck = name
+	}
+
+	for _, allowedNS := range c.allowedNamespaces {
+		if namespaceToCheck == allowedNS {
+			return true
+		}
+	}
+	return false
 }
 
 // kind & apiVersion must be passed separately as the object's TypeMeta is not populated
--- a/cluster/kubernetes/sync.go
+++ b/cluster/kubernetes/sync.go
@@ -54,7 +54,7 @@ func (c *Cluster) Sync(syncSet cluster.SyncSet) error {
 
 	// NB we get all resources, since we care about leaving unsynced,
 	// _ignored_ resources alone.
-	clusterResources, err := c.getResourcesBySelector("")
+	clusterResources, err := c.getAllowedResourcesBySelector("")
 	if err != nil {
 		return errors.Wrap(err, "collating resources in cluster for sync")
 	}
@@ -62,7 +62,11 @@ func (c *Cluster) Sync(syncSet cluster.SyncSet) error {
 	cs := makeChangeSet()
 	var errs cluster.SyncError
 	for _, res := range syncSet.Resources {
-		id := res.ResourceID().String()
+		resID := res.ResourceID()
+		if !c.IsAllowedResource(resID) {
+			continue
+		}
+		id := resID.String()
 		// make a record of the checksum, whether we stage it to
 		// be applied or not, so that we don't delete it later.
 		csum := sha1.Sum(res.Bytes())
@@ -122,7 +126,7 @@ func (c *Cluster) collectGarbage(
 
 	orphanedResources := makeChangeSet()
 
-	clusterResources, err := c.getGCMarkedResourcesInSyncSet(syncSet.Name)
+	clusterResources, err := c.getAllowedGCMarkedResourcesInSyncSet(syncSet.Name)
 	if err != nil {
 		return nil, errors.Wrap(err, "collating resources in cluster for calculating garbage collection")
 	}
@@ -188,7 +192,7 @@ func (r *kuberesource) GetGCMark() string {
 	return r.obj.GetLabels()[gcMarkLabel]
 }
 
-func (c *Cluster) getResourcesBySelector(selector string) (map[string]*kuberesource, error) {
+func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*kuberesource, error) {
 	listOptions := meta_v1.ListOptions{}
 	if selector != "" {
 		listOptions.LabelSelector = selector
@@ -216,14 +220,12 @@ func (c *Cluster) getResourcesBySelector(selector string) (map[string]*kuberesou
 			if !contains(verbs, "list") {
 				continue
 			}
-
 			groupVersion, err := schema.ParseGroupVersion(resource.GroupVersion)
 			if err != nil {
 				return nil, err
 			}
-
-			resourceClient := c.client.dynamicClient.Resource(groupVersion.WithResource(apiResource.Name))
-			data, err := resourceClient.List(listOptions)
+			gvr := groupVersion.WithResource(apiResource.Name)
+			list, err := c.listAllowedResources(apiResource.Namespaced, gvr, listOptions)
 			if err != nil {
 				if apierrors.IsForbidden(err) {
 					// we are not allowed to list this resource but
@@ -233,7 +235,7 @@ func (c *Cluster) getResourcesBySelector(selector string) (map[string]*kuberesou
 				return nil, err
 			}
 
-			for i, item := range data.Items {
+			for i, item := range list {
 				apiVersion := item.GetAPIVersion()
 				kind := item.GetKind()
 
@@ -244,7 +246,7 @@ func (c *Cluster) getResourcesBySelector(selector string) (map[string]*kuberesou
 				}
 				// TODO(michael) also exclude anything that has an ownerReference (that isn't "standard"?)
 
-				res := &kuberesource{obj: &data.Items[i], namespaced: apiResource.Namespaced}
+				res := &kuberesource{obj: &list[i], namespaced: apiResource.Namespaced}
 				result[res.ResourceID().String()] = res
 			}
 		}
@@ -253,18 +255,48 @@ func (c *Cluster) getResourcesBySelector(selector string) (map[string]*kuberesou
 	return result, nil
 }
 
-func (c *Cluster) getGCMarkedResourcesInSyncSet(syncSetName string) (map[string]*kuberesource, error) {
-	allGCMarkedResources, err := c.getResourcesBySelector(gcMarkLabel) // means "gcMarkLabel exists"
+func (c *Cluster) listAllowedResources(
+	namespaced bool, gvr schema.GroupVersionResource, options meta_v1.ListOptions) ([]unstructured.Unstructured, error) {
+	if !namespaced || len(c.allowedNamespaces) == 0 {
+		// The resource is not namespaced or all the namespaces are allowed
+		resourceClient := c.client.dynamicClient.Resource(gvr)
+		data, err := resourceClient.List(options)
+		if err != nil {
+			return nil, err
+		}
+		return data.Items, nil
+	}
+
+	// List resources only from the allowed namespaces
+	var result []unstructured.Unstructured
+	for _, ns := range c.allowedNamespaces {
+		data, err := c.client.dynamicClient.Resource(gvr).Namespace(ns).List(options)
+		if err != nil {
+			return result, err
+		}
+		result = append(result, data.Items...)
+	}
+	return result, nil
+}
+
+func (c *Cluster) getAllowedGCMarkedResourcesInSyncSet(syncSetName string) (map[string]*kuberesource, error) {
+	allGCMarkedResources, err := c.getAllowedResourcesBySelector(gcMarkLabel) // means "gcMarkLabel exists"
 	if err != nil {
 		return nil, err
 	}
-	syncSetGCMarkedResources := map[string]*kuberesource{}
+	allowedSyncSetGCMarkedResources := map[string]*kuberesource{}
 	for resID, kres := range allGCMarkedResources {
-		if kres.GetGCMark() == makeGCMark(syncSetName, resID) {
-			syncSetGCMarkedResources[resID] = kres
+		// Discard disallowed resources
+		if !c.IsAllowedResource(kres.ResourceID()) {
+			continue
 		}
+		// Discard resources out of the Sync Set
+		if kres.GetGCMark() != makeGCMark(syncSetName, resID) {
+			continue
+		}
+		allowedSyncSetGCMarkedResources[resID] = kres
 	}
-	return syncSetGCMarkedResources, nil
+	return allowedSyncSetGCMarkedResources, nil
 }
 
 func applyMetadata(res resource.Resource, syncSetName, checksum string) ([]byte, error) {
--- a/cluster/kubernetes/sync.go
+++ b/cluster/kubernetes/sync.go
@@ -257,8 +257,8 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
 
 func (c *Cluster) listAllowedResources(
 	namespaced bool, gvr schema.GroupVersionResource, options meta_v1.ListOptions) ([]unstructured.Unstructured, error) {
-	if !namespaced || len(c.allowedNamespaces) == 0 {
-		// The resource is not namespaced or all the namespaces are allowed
+	if !namespaced {
+		// The resource is not namespaced
 		resourceClient := c.client.dynamicClient.Resource(gvr)
 		data, err := resourceClient.List(options)
 		if err != nil {
@@ -268,9 +268,13 @@ func (c *Cluster) listAllowedResources(
 	}
 
 	// List resources only from the allowed namespaces
+	namespaces, err := c.getAllowedAndExistingNamespaces()
+	if err != nil {
+		return nil, err
+	}
 	var result []unstructured.Unstructured
-	for _, ns := range c.allowedNamespaces {
-		data, err := c.client.dynamicClient.Resource(gvr).Namespace(ns).List(options)
+	for _, ns := range namespaces {
+		data, err := c.client.dynamicClient.Resource(gvr).Namespace(ns.Name).List(options)
 		if err != nil {
 			return result, err
 		}
@@ -286,11 +290,7 @@ func (c *Cluster) getAllowedGCMarkedResourcesInSyncSet(syncSetName string) (map[
 	}
 	allowedSyncSetGCMarkedResources := map[string]*kuberesource{}
 	for resID, kres := range allGCMarkedResources {
-		// Discard disallowed resources
-		if !c.IsAllowedResource(kres.ResourceID()) {
-			continue
-		}
-		// Discard resources out of the Sync Set
+		// Discard resources whose mark doesn't match their resource ID
 		if kres.GetGCMark() != makeGCMark(syncSetName, resID) {
 			continue
 		}
--- a/cluster/kubernetes/sync.go
+++ b/cluster/kubernetes/sync.go
@@ -258,7 +258,7 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
 func (c *Cluster) listAllowedResources(
 	namespaced bool, gvr schema.GroupVersionResource, options meta_v1.ListOptions) ([]unstructured.Unstructured, error) {
 	if !namespaced {
-		// The resource is not namespaced
+		// The resource is not namespaced, everything is allowed
 		resourceClient := c.client.dynamicClient.Resource(gvr)
 		data, err := resourceClient.List(options)
 		if err != nil {
--- a/cluster/mock.go
+++ b/cluster/mock.go
@@ -10,15 +10,16 @@ import (
 
 // Doubles as a cluster.Cluster and cluster.Manifests implementation
 type Mock struct {
-	AllWorkloadsFunc   func(maybeNamespace string) ([]Workload, error)
-	SomeWorkloadsFunc  func([]flux.ResourceID) ([]Workload, error)
-	PingFunc           func() error
-	ExportFunc         func() ([]byte, error)
-	SyncFunc           func(SyncSet) error
-	PublicSSHKeyFunc   func(regenerate bool) (ssh.PublicKey, error)
-	UpdateImageFunc    func(def []byte, id flux.ResourceID, container string, newImageID image.Ref) ([]byte, error)
-	LoadManifestsFunc  func(base string, paths []string) (map[string]resource.Resource, error)
-	UpdatePoliciesFunc func([]byte, flux.ResourceID, policy.Update) ([]byte, error)
+	AllWorkloadsFunc      func(maybeNamespace string) ([]Workload, error)
+	SomeWorkloadsFunc     func([]flux.ResourceID) ([]Workload, error)
+	IsAllowedResourceFunc func(flux.ResourceID) bool
+	PingFunc              func() error
+	ExportFunc            func() ([]byte, error)
+	SyncFunc              func(SyncSet) error
+	PublicSSHKeyFunc      func(regenerate bool) (ssh.PublicKey, error)
+	UpdateImageFunc       func(def []byte, id flux.ResourceID, container string, newImageID image.Ref) ([]byte, error)
+	LoadManifestsFunc     func(base string, paths []string) (map[string]resource.Resource, error)
+	UpdatePoliciesFunc    func([]byte, flux.ResourceID, policy.Update) ([]byte, error)
 }
 
 func (m *Mock) AllWorkloads(maybeNamespace string) ([]Workload, error) {
@@ -27,6 +28,10 @@ func (m *Mock) AllWorkloads(maybeNamespace string) ([]Workload, error) {
 
 func (m *Mock) SomeWorkloads(s []flux.ResourceID) ([]Workload, error) {
 	return m.SomeWorkloadsFunc(s)
+}
+
+func (m *Mock) IsAllowedResource(id flux.ResourceID) bool {
+	return m.IsAllowedResourceFunc(id)
 }
 
 func (m *Mock) Ping() error {
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -377,6 +377,11 @@ func (d *Daemon) updatePolicy(spec update.Spec, updates policy.Updates) updateFu
 		var anythingAutomated bool
 
 		for workloadID, u := range updates {
+			if d.Cluster.IsAllowedResource(workloadID) {
+				result.Result[workloadID] = update.WorkloadResult{
+					Status: update.ReleaseStatusSkipped,
+				}
+			}
 			if policy.Set(u.Add).Has(policy.Automated) {
 				anythingAutomated = true
 			}
@@ -707,7 +712,7 @@ func policyEvents(us policy.Updates, now time.Time) map[string]event.Event {
 // policyEventTypes is a deduped list of all event types this update contains
 func policyEventTypes(u policy.Update) []string {
 	types := map[string]struct{}{}
-	for p, _ := range u.Add {
+	for p := range u.Add {
 		switch {
 		case p == policy.Automated:
 			types[event.EventAutomate] = struct{}{}
@@ -718,7 +723,7 @@ func policyEventTypes(u policy.Update) []string {
 		}
 	}
 
-	for p, _ := range u.Remove {
+	for p := range u.Remove {
 		switch {
 		case p == policy.Automated:
 			types[event.EventDeautomate] = struct{}{}
--- a/cluster/kubernetes/images.go
+++ b/cluster/kubernetes/images.go
@@ -123,7 +123,7 @@ func mergeCredentials(log func(...interface{}) error,
 func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 	allImageCreds := make(registry.ImageCreds)
 
-	namespaces, err := c.getAllowedNamespaces()
+	namespaces, err := c.getAllowedAndExistingNamespaces()
 	if err != nil {
 		c.logger.Log("err", errors.Wrap(err, "getting namespaces"))
 		return allImageCreds
--- a/cluster/kubernetes/kubernetes_test.go
+++ b/cluster/kubernetes/kubernetes_test.go
@@ -28,7 +28,7 @@ func testGetAllowedNamespaces(t *testing.T, namespace []string, expected []strin
 	client := ExtendedClient{coreClient: clientset}
 	c := NewCluster(client, nil, nil, log.NewNopLogger(), namespace, []string{})
 
-	namespaces, err := c.getAllowedNamespaces()
+	namespaces, err := c.getAllowedAndExistingNamespaces()
 	if err != nil {
 		t.Errorf("The error should be nil, not: %s", err)
 	}
--- a/cluster/kubernetes/sync_test.go
+++ b/cluster/kubernetes/sync_test.go
@@ -290,8 +290,8 @@ metadata:
 			panic(err)
 		}
 
-		// Now check what resources remain in the sync set
-		actual, err := kube.getGCMarkedResourcesInSyncSet("testset")
+		// Now check that the resources were created
+		actual, err := kube.getAllowedGCMarkedResourcesInSyncSet("testset")
 		if err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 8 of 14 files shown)


squaremo:
> * `"default"` indicates _mapped to the default namespace_, either explicitly  (the yaml in git or from the cluster contains it) or implicitly (the yaml in git or the cluster doesn't include a namespace)

`default` is the name of a namespace, so you would not be able to tell whether this meant "implicitly mapped to the default namespace" (which might _not_ be `default`) or "explicitly in `default`".

squaremo:
For those two cases you've put above: in (2), I think the API server will always give you a namespace if the resource is namespaced, and never if it's not. So an empty namespace is a reliable guide to cluster-scoped resources.

But for 1), the only way to know is to check with the API server whether the particular GroupVersionKind is namespaced or not; something which I was able to dodge in #1442 since I don't actually care what the namespace is, so long I can use it to relate resources to manifests (though it would be cleaner if I could give things accurate IDs).

cluster/kubernetes/manifests.go
```diff
@@ -8,18 +8,31 @@ import (
...
+	return resources, err
 }
 
 func (c *Manifests) UpdateImage(def []byte, id flux.ResourceID, container string, image image.Ref) ([]byte, error) {
 	return updatePodController(def, id, container, image)
 }
 
+func (c *Manifests) discardDisallowed(resources map[string]resource.Resource) {
+	for k, r := range resources {
+		if !isInAllowedNamespace(c.AllowedNamespaces, r.ResourceID()) {
```
2opremio: It's a bit ugly doing this when parsing (parsing should be parsing, not filtering) but the only alternative I saw was passing `AllowedNamespaces` all over the place which was uglier and more error-prone (people using new code may forget filtering by namespace where needed).

http/client/client.go
```diff
@@ -31,7 +31,7 @@ type Token string
 
 func (t Token) Set(req *http.Request) {
 	if string(t) != "" {
-		req.Header.Set("Authorization", fmt.Sprintf("Scope-Probe token=%s", t))
+		req.Header.Set("Authorization", fmt.Sprintf("Scoper-Probe token=%s", t))
```
squaremo: Oops

integrations/client/clientset/versioned/clientset.go
```diff
@@ -63,7 +63,7 @@ func (c *Clientset) Helm() helmv1alpha2.HelmV1alpha2Interface {
 	return c.helmV1alpha2
 }
 
-// Discovery retrieves the DiscoveryClient
+// Discovery retrieves the Scoper
```
squaremo: How did this ..?

http/client/client.go
2opremio: Darn, I switched to Goland and it seems automatic refactorings are not woring as I expected. Will fix.

integrations/client/clientset/versioned/clientset.go
2opremio: Same here, thanks for finding it

2opremio:
I am done with the code. I want to do some further high-level testing on a Kubernetes cluster, but it's ready for review.

cluster/kubernetes/scoper.go
```diff
@@ -1,33 +1,112 @@
...
+	"time"
+
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/client-go/discovery"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/weaveworks/flux/cluster"
 )
 
+var (
```
squaremo: Would this suffice? https://github.com/kubernetes/client-go/blob/master/discovery/cached/memcache.go (does not use memcached!)

cluster/kubernetes/manifests.go
squaremo: Yeah I have to say I don't like this either -- these are supposed to be general-purpose. Can I give an example where filtering the return value will screw things up? Not off the top of my head, but I think it's about as likely that will be a problem as the scenario you mention of someone forgetting to filter before an operation.

cluster/kubernetes/resource/resource.go
```diff
@@ -31,7 +34,9 @@ type baseObject struct {
 
 func (o baseObject) ResourceID() flux.ResourceID {
 	ns := o.Meta.Namespace
-	if ns == "" {
+	if o.scope == cluster.NamespaceScope && ns == "" {
+		// namespace-scoped resources with an omitted namespace
+		// implicitly map to the default namespace in the cluster
 		ns = "default"
```
squaremo: This is not necessarily the default namespace in the cluster. Yes, this was a problem before, but we now rely somewhat more on namespaces being correct.

cluster/scoper.go
```diff
@@ -0,0 +1,29 @@
+package cluster
```
squaremo: I don't think this interface is a concern outside the kubernetes package -- namespace scoping is implicit in the Manifests implementation. Granted, resource IDs include a namespace, so there is already some muddling.

cluster/kubernetes/manifests.go
2opremio: Then, the alternative is to filter after loading the manifests (probably systematically after every call). Would you be OK with that?

> Can I give an example where filtering the return value will screw things up? Not off the top of my head

Me neither (it's like the manifests didn't exist on the first place), that's why I chose this approach.

cluster/kubernetes/scoper.go
2opremio: I considered it and discarded it because:

1.  It caches to disk (see `~/.kube/cache` in the environment you normally use for `kubectl`):
       * We only need the `namespaced` bit of every resource, which makes it overkill to store to disk given the small memory requirements
       * Reading from memory is much quicker anyhow
2. It doesn't support TTLs (at least explicitly), which I think are needed in case the resource definitions change (e.g. CRDs being changed or added incrementally).

cluster/kubernetes/resource/resource.go
2opremio: Oh, I didn't now. I am happy to fix it, but after some quick searches I couldn't find how to find that information (I know that `kubectl` can set a a different namespace http://www.kubetips.com/set-a-default-namespace/ , but I don't know if that's what you are referring to, if that's the case we should just provide a way to specify the namespace for flux)

cluster/mock.go
```diff
@@ -10,15 +12,16 @@ import (
...
 	AllServicesFunc    func(maybeNamespace string) ([]Controller, error)
 	SomeServicesFunc   func([]flux.ResourceID) ([]Controller, error)
 	PingFunc           func() error
 	ExportFunc         func() ([]byte, error)
 	SyncFunc           func(SyncDef) error
 	PublicSSHKeyFunc   func(regenerate bool) (ssh.PublicKey, error)
 	UpdateImageFunc    func(def []byte, id flux.ResourceID, container string, newImageID image.Ref) ([]byte, error)
-	LoadManifestsFunc  func(base string, paths []string) (map[string]resource.Resource, error)
-	ParseManifestsFunc func([]byte) (map[string]resource.Resource, error)
+	LoadManifestsFunc  func(base string, paths []string, scoper Scoper) (map[string]resource.Resource, error)
```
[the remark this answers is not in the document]
2opremio: I didn't know this was supposed to be a cluster-independent abstraction, sorry. However, note that it has quite a few leaks (assuming there are manifests, and the namespaces in the resource ids as you said).

However, we only need the scope for namespace filtering, so, if we move that out of `LoadManifests` maybe we can also move scoping out of parsing. Note, however, that then we would need a way to get the original namespace of a resource (for scoping), not the one currently on the resource ID which can be synthesized if empty.

cluster/scoper.go
2opremio: see my answer above

cluster/kubernetes/scoper.go
squaremo: I don't see the bit where it writes to disk (in the code in that file). Is that in the broader caching system?

cluster/kubernetes/resource/resource.go
squaremo: It's a setting specific to a context, given in a kubeconfig. Since the machinery for loading config is somewhat labyrinthine, I reckon the most reliable way to get it (considering we use `kubectl` for applying anyway) is to ask `kubectl`. That's what I've done in https://github.com/weaveworks/flux/pull/1442/commits/b59cef880926501029d24a68d2f603051959ae28#diff-01a1bc681d547c2ba8d68501f424b99fR371 (I'm not proud).

cluster/mock.go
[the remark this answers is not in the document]
squaremo: It totally has leaks. It will probably (at this point) never be extended to cover other types of cluster. 

However, it's still worth trying to contain changes, where we can. The namespace scoping is particular to Kubernetes, so far as we know (and I wouldn't expect to see some other Scoping implementation be combined with a kubernetes cluster).

cluster/kubernetes/scoper.go
2opremio: Ah, you are right. I missed that, sorry. I had just looked at https://github.com/kubernetes/client-go/blob/master/discovery/cached_discovery.go which I had previously looked at, and thought you were referring to that (the file layout is a bit odd).

I guess we can periodically invalidate the whole cache with `Invalidate()` to make up for the lack of explicit TTLs. However, the cache implementation is a bit aggressive, since it's all or nothing (it gets the information of all groups at once) resulting in ~50 requests every time (based on the number of version groups from my local k8s cluster).

cluster/kubernetes/resource/resource.go
2opremio: is the local `kubectl` context exposed to end users?

cluster/kubernetes/scoper.go
squaremo: The newer sync implementation in #1442 relies pretty heavily on the discovery interface -- it uses it to probe for the kinds of resources that the cluster knows about, so it can ask for all of the resources. In anticipation of that, I am less concerned about caching the whole set of endpoints.

Invalidation is trickier to judge. CRDs could change namespacedness, as you say (though this is a niche problem, getting something wrong, like incorrectly deleting a CRD and that cascading to all the custom resources, would be bad). On the other hand, even synchronously requesting all the metadata would not be immune to problems of that nature -- it needs to be solved some other way, if it can be.

cluster/kubernetes/resource/resource.go
squaremo: Yes, in the sense that you can (by design) mount one into the right place in the filesystem and it will be honoured.

cluster/kubernetes/scoper.go
squaremo: If CRDs are the only thing we think could change, we could watch for CRD changes and invalidate on those.

2opremio:
This PR is blocked by #1442 (they have a lot in common and [I agreed with @squaremo](https://github.com/weaveworks/flux/pull/1442#issuecomment-458523547) to get #1442 merged first)

cluster/kubernetes/sync.go
```diff
@@ -238,10 +240,34 @@ func (c *Cluster) getResourcesBySelector(selector string) (map[string]*kuberesou
...
+		data, err := resourceClient.List(options)
+		if err != nil {
+			return nil, err
+		}
+		return data.Items, nil
+	}
+
+	// List resources only from the allowed namespaces
+	var result []unstructured.Unstructured
+	for _, ns := range c.allowedNamespaces {
```
2opremio: Maybe I should be using `getAllowedNamespaces()` to ensure the namespaces are accessible and give a more reasonable error if they aren't?

2opremio:
@squaremo this is ready again, PTAL

cluster/cluster.go
```diff
@@ -27,6 +27,7 @@ type Cluster interface {
 	// Get all of the services (optionally, from a specific namespace), excluding those
 	AllControllers(maybeNamespace string) ([]Controller, error)
 	SomeControllers([]flux.ResourceID) ([]Controller, error)
+	IsAllowedResource(flux.ResourceID) bool
```
squaremo: It's kind of a shame to have a _whoooole_ new method in the interface for this pretty small thing .. could the namespace guard be contained in just the implementations (`kubernetes.Cluster` and `kubernetes.Manifest`)?

daemon/daemon.go
```diff
@@ -377,6 +377,11 @@ func (d *Daemon) updatePolicy(spec update.Spec, updates policy.Updates) updateFu
 		var anythingAutomated bool
 
 		for serviceID, u := range updates {
+			if d.Cluster.IsAllowedResource(serviceID) {
```
squaremo: This may be where my unease at looking at resource IDs comes from -- this is a user-provided value, and we're trusting what it says. Maybe harmless now, but a potential loophole.

cluster/kubernetes/sync.go
2opremio: @squaremo thoughts?

daemon/daemon.go
2opremio: Well, it's comparing against yet another user-provided value (since it comes from the user).

However, for extra assurance, we could be calling or using the output of `getAllowedNamespaces` in `IsAllowedResource()` so that we check against existing namespaces. I don't like this either though, since you may want to restrict the view to namespaces you haven't created yet.

2opremio: In fact, for performance and simplicity, I was pondering to remove the existing calls to `getAllowedNamespaces()` by direct access to the provided allowed namespaces when possible (i.e. everywhere except `Export()`).

cluster/cluster.go
2opremio: Problem is that (at least as is) it needs to be accessed by both the daemon and `kubernetes.Cluster`. The later is needed to filter namespaces when accessing the policies.

squaremo:
@2opremio That commit is what I had in mind (hopefully clearer in code than in comments!). See what you think ..

2opremio:
> The "lowest" we can filter resources by namespace, the better, since it will cover more code paths.

My initial proposal was very similar (to filter out disallowed namespaces when loading the manifests) but it was discarded, see https://github.com/weaveworks/flux/pull/1668#discussion_r250672079 .

I recall you ended up coming with a scenario in which it could be a problem. Now I don't recall and I can't find it, maybe you shared it privately.

At the very least, it will lead to confusing error messages when trying to edit policies from resources belonging to disallowed namespaces. If we remove the manifests when loading the error will be `workload not found` whereas if we don't we can provide a proper message such as `workload belongs to disallowed namespace`.

daemon/daemon.go
squaremo: I see the reasoning here, sometimes you want `{ namespaces named as allowed }`; e.g., when creating a namespace from a manifest. And sometimes you want `{ namespaces allowed and existing }`; e..g, when querying for cluster workloads.

squaremo:
> My initial proposal was very similar (to filter out disallowed namespaces when loading the manifests) but it was discarded, see [#1668 (comment)](https://github.com/weaveworks/flux/pull/1668#discussion_r250672079) .
> 
> I recall you ended up coming with a scenario in which it could be a problem. Now I don't recall and I can't find it, maybe you shared it privately.

Where we ended up on that (in #1442) was that it was an acceptable mid-way point to leave the _parsing_ as generic (that's `cluster/kubernetes/resource/load.go` and friends), but do post-hoc filtering on the result (that's `cluster/kubernetes/manifest.go`). Mind you, the waters are muddy enough now that there may not be much of a distinction to make.

> At the very least, it will lead to confusing error messages when trying to edit policies from workload belonging to disallowed namespaces. If we remove the manifests when loading, the error will be `workload not found` whereas if we don't we can provide a proper message such as `workload belongs to disallowed namespace`.

That's a fair objection. It's the usual trade-off of reliable enforcement versus being helpful to users. In this case, I _don't_ think it's important to hide from a user the fact of namespaces being blocked, and I _do_ think we can do better at closing down opportunities for accidentally side-stepping the check.

What about if the check against the allowed namespaces happened in `cluster/kubernetes/Manifest.UpdateImage` and `cluster/kubernetes/Manifest.UpdatePolicy`? Then it can return a more helpful error, while still being in the narrow bit of the interface.

cluster/kubernetes/kubernetes.go
```diff
@@ -119,12 +120,15 @@ func NewCluster(client ExtendedClient, applier Applier, sshKeyRing ssh.KeyRing,
 
 // --- cluster.Cluster
 
-// SomeWorkloads returns the workloads named, missing out any that aren't
-// accessible in the cluster. They do not necessarily have to be returned
-// in the order requested.
+// SomeWorkloads returns the workloads named, missing out any that don't
```
squaremo: @2opremio Did I accidentally revert this when rebasing?

cluster/kubernetes/kubernetes.go
```diff
@@ -300,6 +304,32 @@ func (c *Cluster) getAllowedNamespaces() ([]apiv1.Namespace, error) {
...
+func (c *Cluster) IsAllowedResource(id flux.ResourceID) bool {
+	if len(c.allowedNamespaces) == 0 {
+		// All resources are allowed when all namespaces are allowed
+		return true
+	}
+
+	namespace, kind, name := id.Components()
+	namespaceToCheck := namespace
+
+	if namespace == resource.ClusterScope {
```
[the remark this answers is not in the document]
squaremo: `ClusterScope` will only be the namespace if the ID is from parsing manifests (cluster/kubernetes/resource/resource.go) or via the sync code (cluster/kubernetes/sync.go). I don't think it can be supplied by a user (as in, you can't construct such a resource ID from user input).

At the minute this doesn't matter, since no operations in the API can be applied to cluster-scope resources (so there's nowhere a user would supply such a thing). But it does mean this check is brittle, if used on user input.

squaremo:
Rebased on master, and removed my extra commit. I've tried out a few scenarios:

 1. Start from scratch with repo and --k8s-allow-namespace=hello
    - only hello namespace and things in hello namespace get created, even though there are manifests for other things

 2. changed --k8s-allow-namespace=default and ran with --sync-garbage-collection
    - new resources are created, old resources are left, as they aren't considered for GC

 3. list-images and list-workloads both show just the namespaces allowed

 4. release --all acts on only the namespaces allowed, even if others use the image in question

So all good so far! Will think of some tricksier tests, next ..

cluster/kubernetes/kubernetes.go
2opremio: I am not sure what you mean. Mind clarifying?

cluster/kubernetes/kubernetes.go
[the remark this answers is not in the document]
2opremio: Yep, I agree. However, I  am not sure how to make this better. 

Regarding the API, even if a user can somehow provide the ID of a cluster-scoped resource, I assume that we will be consistent in the representation of the ID.

Tangentially, I don't like the using a `<cluster>` namespace to indicate cluster-scoped. I think makes much more sense to use an empty string,  but I guess that's besides the point.

cluster/cluster.go
squaremo: (I'm shelving this for the minute, since I couldn't come up with better code inside a modest timebox)

cluster/kubernetes/kubernetes.go
squaremo: There was a conflict in this comment when I was rebasing, and I'm not sure I resolved it correctly. It reads OK, I just wanted to make sure I hadn't undone a change you'd carefully made.

cluster/kubernetes/kubernetes.go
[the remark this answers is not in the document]
squaremo: > Tangentially, I don't like the using a `<cluster>` namespace to indicate cluster-scoped. I think makes much more sense to use an empty string, but I guess that's besides the point.

The trouble with the empty string is that it also means "not supplied". In some cases, "not supplied" lines up with "has no namespace" -- but in other cases, those are distinct.

cluster/kubernetes/sync.go
squaremo: What I gather is that this would mean the difference between returning an error if any namespaces that are allowed happen to not exist, and simply eliding those -- yes, I think you want existing, allowed namespaces here.

(In other places you want "in principle allowed namespaces", i.e., namespaces that are allowed, whether or not they exist).

cluster/kubernetes/kubernetes.go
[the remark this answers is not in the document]
2opremio: gotcha

cluster/kubernetes/kubernetes.go
2opremio: oh, no, it looks fine :)

cluster/kubernetes/sync.go
2opremio: Alright, I will change this.

2opremio:
@squaremo I've dealt with the comments (except for https://github.com/weaveworks/flux/pull/1668#discussion_r267443970  which I don't know how to address).

PTAL

cluster/kubernetes/kubernetes.go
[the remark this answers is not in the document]
squaremo: Let's go with this for now, and give ourselves some room to figure out how to improve it.

2opremio:
@squaremo I tested this PR against https://github.com/2opremio/locked-down-flux to see how it behaved with namespace-RBAC and everything worked as expected (no errors in the logs and the example was created as expected)

```
Flag --k8s-namespace-whitelist has been deprecated, changed to --k8s-allow-namespace, use that instead
ts=2019-03-25T17:00:39.817983756Z caller=main.go:167 version=1471-extend-ns-filtering-cb4ad60c
ts=2019-03-25T17:00:39.882144008Z caller=main.go:297 component=cluster identity=/var/fluxd/keygen/identity
ts=2019-03-25T17:00:39.882313951Z caller=main.go:298 component=cluster identity.pub="ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7OhlwhU0Hwp3kxyUe+FUWIQ3S127QIlQFnj2m9HUDfb1bpkRDK4gRVBM6JiYBTFj4/YAfGcWck4mPt88SgFIPLQzUYbJC1NoKjMVjH/VbAxP1XoJIZMlzHMd2MztcX2/q0QrCFAYhF60OhbTFu3tuZpqSMeZ0DK/KAmCVevXslaKztpayUqatBnmxhm3C5EM22jAwQZ/yxdN9+s8DZNiJS+2g5YwWVJqWB3WisjT9/p6MT/wjW+fiBbuQDpr3xAp7z6tAnr1/yChqmY7RAvTVsyXaNDtTEgXS9+FAgNmo3wf6m4qpT2H42/E6mY4rjDlPYxlf1bgkpHk8eR4I7qbJ"
ts=2019-03-25T17:00:39.882461841Z caller=main.go:299 component=cluster host=https://10.96.0.1:443 version=kubernetes-v1.13.0
ts=2019-03-25T17:00:39.88267732Z caller=main.go:311 component=cluster kubectl=/usr/local/bin/kubectl
ts=2019-03-25T17:00:39.883574849Z caller=main.go:322 component=cluster ping=true
ts=2019-03-25T17:00:45.150716883Z caller=aws.go:69 component=aws warn="no AWS region configured, or detected as cluster region" err="RequestError: send request failed\ncaused by: Get http://169.254.169.254/latest/meta-data/placement/availability-zone: dial tcp 169.254.169.254:80: connect: connection refused"
ts=2019-03-25T17:00:45.151020082Z caller=main.go:347 warning="AWS authorization not used; pre-flight check failed"
ts=2019-03-25T17:00:45.160827588Z caller=main.go:452 url=git@github.com:2opremio/locked-down-flux.git user="Weave Flux" email=support@weave.works signing-key= sync-tag=flux-sync notes-ref=flux set-author=false
ts=2019-03-25T17:00:45.160897014Z caller=main.go:509 upstream="no upstream URL given"
ts=2019-03-25T17:00:45.16101119Z caller=main.go:538 metrics-addr=:3031
ts=2019-03-25T17:00:45.162781302Z caller=loop.go:90 component=sync-loop err="git repo not ready: git repo has not been cloned yet"
ts=2019-03-25T17:00:45.16283077Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2019-03-25T17:00:45.162848493Z caller=images.go:27 component=sync-loop msg="no automated services"
ts=2019-03-25T17:00:45.16403101Z caller=main.go:530 addr=:3030
ts=2019-03-25T17:00:45.728458176Z caller=checkpoint.go:24 component=checkpoint msg="up to date" latest=1.11.0
ts=2019-03-25T17:05:44.941893242Z caller=loop.go:90 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository."
ts=2019-03-25T17:05:44.942947035Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2019-03-25T17:05:44.943062658Z caller=images.go:27 component=sync-loop msg="no automated services"
ts=2019-03-25T17:09:42.250149846Z caller=loop.go:103 component=sync-loop event=refreshed url=git@github.com:2opremio/locked-down-flux.git branch=deploy HEAD=723b91549434575348e4b67affdc857478c5894f
ts=2019-03-25T17:09:42.494560317Z caller=sync.go:416 component=cluster method=Sync cmd=apply args= count=3
ts=2019-03-25T17:09:43.034708957Z caller=sync.go:482 component=cluster method=Sync cmd="kubectl apply -f -" took=539.883683ms err=null output="service/echoserver created\ndeployment.apps/echoserver created\nnetworkpolicy.networking.k8s.io/echoserver created"
ts=2019-03-25T17:09:45.680724239Z caller=warming.go:268 component=warmer info="refreshing image" image=gcr.io/google-samples/hello-app tag_count=2 to_update=2 of_which_refresh=0 of_which_missing=2
ts=2019-03-25T17:09:46.424009606Z caller=warming.go:364 component=warmer updated=gcr.io/google-samples/hello-app successful=2 attempted=2
ts=2019-03-25T17:09:46.424208427Z caller=images.go:17 component=sync-loop msg="polling images"
```

squaremo:
> I tested this PR against https://github.com/2opremio/locked-down-flux

Lovely! What happens if you tell it to allow a namespace it doesn't have RBAC access to?

2opremio:
Good suggestion. After [creating a namespace forbidden by RBAC and included in the whitelist](https://github.com/2opremio/locked-down-flux/commit/c89a1b3362d904300f10a74c6daf3f2cdf900a18) I noticed that flux was creating it anyways.

After quite some digging, it turns out that the Kubernetes installation from Docker For Mac makes all service accounts `cluster-admins`, sigh:
```
$ kubectl get clusterrolebinding -o yaml docker-for-desktop-binding
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2019-03-25T16:29:55Z"
  name: docker-for-desktop-binding
  resourceVersion: "436"
  selfLink: /apis/rbac.authorization.k8s.io/v1/clusterrolebindings/docker-for-desktop-binding
  uid: 38fe7238-4f1b-11e9-b66a-025000000001
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:serviceaccounts
```

I will retest-tomorrow after removing that role-binding.

PS: It's really frustrating that the RBAC groups are second-class citizens and are only documented through examples.

2opremio:
I bit the bullet and everything works as expected:
1. The resources from a RBAC-allowed and whitelisted namespace (`helloworld`) are created just fine
2. The resources from a RBAC-*disallowed* and whitelisted namespace (`helloworkd2`) are not created and it doesn't abort the sync

That said, although we give a sensible error about the not being able to access the `helloworld2`,  subsequent `kubectl apply` are pretty horrible:

```
$ kubectl logs --namespace=flux-system flux-9b5fb66b7-526rz 
Flag --k8s-namespace-whitelist has been deprecated, changed to --k8s-allow-namespace, use that instead
ts=2019-03-26T01:16:01.725052284Z caller=main.go:169 version=1471-extend-ns-filtering-92db72a4
ts=2019-03-26T01:16:01.764536902Z caller=main.go:312 component=cluster identity=/etc/fluxd/ssh/identity
ts=2019-03-26T01:16:01.76731935Z caller=main.go:313 component=cluster identity.pub="ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7OhlwhU0Hwp3kxyUe+FUWIQ3S127QIlQFnj2m9HUDfb1bpkRDK4gRVBM6JiYBTFj4/YAfGcWck4mPt88SgFIPLQzUYbJC1NoKjMVjH/VbAxP1XoJIZMlzHMd2MztcX2/q0QrCFAYhF60OhbTFu3tuZpqSMeZ0DK/KAmCVevXslaKztpayUqatBnmxhm3C5EM22jAwQZ/yxdN9+s8DZNiJS+2g5YwWVJqWB3WisjT9/p6MT/wjW+fiBbuQDpr3xAp7z6tAnr1/yChqmY7RAvTVsyXaNDtTEgXS9+FAgNmo3wf6m4qpT2H42/E6mY4rjDlPYxlf1bgkpHk8eR4I7qbJ"
ts=2019-03-26T01:16:01.767403888Z caller=main.go:314 component=cluster host=https://10.96.0.1:443 version=kubernetes-v1.13.0
ts=2019-03-26T01:16:01.76744863Z caller=main.go:326 component=cluster kubectl=/usr/local/bin/kubectl
ts=2019-03-26T01:16:01.769259121Z caller=main.go:337 component=cluster ping=true
ts=2019-03-26T01:16:07.132247759Z caller=aws.go:69 component=aws warn="no AWS region configured, or detected as cluster region" err="RequestError: send request failed\ncaused by: Get http://169.254.169.254/latest/meta-data/placement/availability-zone: dial tcp 169.254.169.254:80: connect: connection refused"
ts=2019-03-26T01:16:07.132330617Z caller=main.go:362 warning="AWS authorization not used; pre-flight check failed"
ts=2019-03-26T01:16:07.133919183Z caller=main.go:467 url=git@github.com:2opremio/locked-down-flux.git user="Weave Flux" email=support@weave.works signing-key= sync-tag=flux-sync notes-ref=flux set-author=false
ts=2019-03-26T01:16:07.14674934Z caller=main.go:524 upstream="no upstream URL given"
ts=2019-03-26T01:16:07.14715081Z caller=main.go:553 metrics-addr=:3031
ts=2019-03-26T01:16:07.154751956Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2019-03-26T01:16:07.154995676Z caller=images.go:27 component=sync-loop msg="no automated workloads"
ts=2019-03-26T01:16:07.155227009Z caller=loop.go:90 component=sync-loop err="git repo not ready: git repo has not been cloned yet"
ts=2019-03-26T01:16:07.15584015Z caller=main.go:545 addr=:3030
ts=2019-03-26T01:16:07.203392013Z caller=kubernetes.go:284 component=cluster warning="cannot access allowed namespace" namespace=helloworld2 err="namespaces \"helloworld2\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"namespaces\" in API group \"\" in the namespace \"helloworld2\""
ts=2019-03-26T01:16:07.601941976Z caller=checkpoint.go:24 component=checkpoint msg="up to date" latest=1.11.0
ts=2019-03-26T01:16:15.779525806Z caller=loop.go:103 component=sync-loop event=refreshed url=git@github.com:2opremio/locked-down-flux.git branch=master HEAD=c89a1b3362d904300f10a74c6daf3f2cdf900a18
ts=2019-03-26T01:16:15.871514063Z caller=sync.go:455 component=cluster method=Sync cmd=apply args= count=7
ts=2019-03-26T01:16:16.26625243Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=394.659317ms err="running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"/v1, Resource=namespaces\", GroupVersionKind: \"/v1, Kind=Namespace\"\nName: \"helloworld2\", Namespace: \"\"\nObject: &{map[\"apiVersion\":\"v1\" \"kind\":\"Namespace\" \"metadata\":map[\"annotations\":map[\"flux.weave.works/sync-checksum\":\"eff788baf5f8906b3879d666e60b5407f7b9f294\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"flux.weave.works/sync-gc-mark\":\"sha256.cFRnOwOXCB_oNpi52QLlUkjvODSLBIo1ntBRfzoPqp0\"] \"name\":\"helloworld2\" \"namespace\":\"\"]]}\nfrom server for: \"STDIN\": namespaces \"helloworld2\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"namespaces\" in API group \"\" in the namespace \"helloworld2\"\nError from server (Forbidden): error when retrieving current configuration of:\nResource: \"/v1, Resource=services\", GroupVersionKind: \"/v1, Kind=Service\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"apiVersion\":\"v1\" \"kind\":\"Service\" \"metadata\":map[\"labels\":map[\"app\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.3KBCW83dRp2s10Zkls_vLADf8Wl0E9Agzz8sIT3wHkU\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\" \"annotations\":map[\"flux.weave.works/sync-checksum\":\"24d87352fb027d306ce7b03676b6d723dc59a799\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"]] \"spec\":map[\"ports\":[map[\"protocol\":\"TCP\" \"targetPort\":'\\u1f90' \"name\":\"http\" \"port\":'\\u1f90']] \"selector\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"] \"type\":\"ClusterIP\"]]}\nfrom server for: \"STDIN\": services \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"services\" in API group \"\" in the namespace \"helloworld2\"\nError from server (Forbidden): error when retrieving current configuration of:\nResource: \"apps/v1, Resource=deployments\", GroupVersionKind: \"apps/v1, Kind=Deployment\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"apiVersion\":\"apps/v1\" \"kind\":\"Deployment\" \"metadata\":map[\"annotations\":map[\"flux.weave.works/sync-checksum\":\"191df8fec32a1deda61e32f995d8ffb7ea160781\" \"flux.weave.works/tag.echoserver\":\"semver:*\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.b3ah4IFHE0JJUNAUMz8Q7ikY7f9dDREvkAYx40-0FTA\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\"] \"spec\":map[\"replicas\":'\\x01' \"selector\":map[\"matchLabels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"strategy\":map[\"type\":\"RollingUpdate\"] \"template\":map[\"metadata\":map[\"labels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"spec\":map[\"automountServiceAccountToken\":%!q(bool=false) \"containers\":[map[\"securityContext\":map[\"allowPrivilegeEscalation\":%!q(bool=false)] \"image\":\"gcr.io/google-samples/hello-app:1.0\" \"name\":\"echoserver\" \"ports\":[map[\"protocol\":\"TCP\" \"containerPort\":'\\u1f90' \"name\":\"web\"]]]]]]]]}\nfrom server for: \"STDIN\": deployments.apps \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"deployments\" in API group \"apps\" in the namespace \"helloworld2\"\nError from server (Forbidden): error when retrieving current configuration of:\nResource: \"networking.k8s.io/v1, Resource=networkpolicies\", GroupVersionKind: \"networking.k8s.io/v1, Kind=NetworkPolicy\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"apiVersion\":\"networking.k8s.io/v1\" \"kind\":\"NetworkPolicy\" \"metadata\":map[\"annotations\":map[\"kubectl.kubernetes.io/last-applied-configuration\":\"\" \"flux.weave.works/sync-checksum\":\"d98993cf8b5516b417533ebeeb0cfabe5c7ff6fa\"] \"labels\":map[\"app\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.yxyBpR0YQllFHVEBWXu_LxyedQX59j9IYMah3mi6rL4\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\"] \"spec\":map[\"ingress\":[map[\"ports\":[map[\"port\":'\\u1f90' \"protocol\":\"TCP\"]]]] \"podSelector\":map[\"matchLabels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"policyTypes\":[\"Ingress\" \"Egress\"]]]}\nfrom server for: \"STDIN\": networkpolicies.networking.k8s.io \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"networkpolicies\" in API group \"networking.k8s.io\" in the namespace \"helloworld2\"" output="service/echoserver created\ndeployment.apps/echoserver created\nnetworkpolicy.networking.k8s.io/echoserver created"
ts=2019-03-26T01:16:16.523525929Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=256.930326ms err="running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"/v1, Resource=namespaces\", GroupVersionKind: \"/v1, Kind=Namespace\"\nName: \"helloworld2\", Namespace: \"\"\nObject: &{map[\"apiVersion\":\"v1\" \"kind\":\"Namespace\" \"metadata\":map[\"annotations\":map[\"flux.weave.works/sync-checksum\":\"eff788baf5f8906b3879d666e60b5407f7b9f294\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"flux.weave.works/sync-gc-mark\":\"sha256.cFRnOwOXCB_oNpi52QLlUkjvODSLBIo1ntBRfzoPqp0\"] \"name\":\"helloworld2\" \"namespace\":\"\"]]}\nfrom server for: \"STDIN\": namespaces \"helloworld2\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"namespaces\" in API group \"\" in the namespace \"helloworld2\"" output=
ts=2019-03-26T01:16:16.753540835Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=229.928726ms err=null output="service/echoserver unchanged"
ts=2019-03-26T01:16:17.020017775Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=266.312128ms err="running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"/v1, Resource=services\", GroupVersionKind: \"/v1, Kind=Service\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"metadata\":map[\"annotations\":map[\"kubectl.kubernetes.io/last-applied-configuration\":\"\" \"flux.weave.works/sync-checksum\":\"24d87352fb027d306ce7b03676b6d723dc59a799\"] \"labels\":map[\"app\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.3KBCW83dRp2s10Zkls_vLADf8Wl0E9Agzz8sIT3wHkU\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\"] \"spec\":map[\"ports\":[map[\"name\":\"http\" \"port\":'\\u1f90' \"protocol\":\"TCP\" \"targetPort\":'\\u1f90']] \"selector\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"] \"type\":\"ClusterIP\"] \"apiVersion\":\"v1\" \"kind\":\"Service\"]}\nfrom server for: \"STDIN\": services \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"services\" in API group \"\" in the namespace \"helloworld2\"" output=
ts=2019-03-26T01:16:17.187479884Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=167.164138ms err=null output="deployment.apps/echoserver configured"
ts=2019-03-26T01:16:17.414758158Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=226.754625ms err="running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"apps/v1, Resource=deployments\", GroupVersionKind: \"apps/v1, Kind=Deployment\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"spec\":map[\"selector\":map[\"matchLabels\":map[\"component\":\"echoserver\" \"app\":\"echoserver\"]] \"strategy\":map[\"type\":\"RollingUpdate\"] \"template\":map[\"metadata\":map[\"labels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"spec\":map[\"automountServiceAccountToken\":%!q(bool=false) \"containers\":[map[\"image\":\"gcr.io/google-samples/hello-app:1.0\" \"name\":\"echoserver\" \"ports\":[map[\"containerPort\":'\\u1f90' \"name\":\"web\" \"protocol\":\"TCP\"]] \"securityContext\":map[\"allowPrivilegeEscalation\":%!q(bool=false)]]]]] \"replicas\":'\\x01'] \"apiVersion\":\"apps/v1\" \"kind\":\"Deployment\" \"metadata\":map[\"name\":\"echoserver\" \"namespace\":\"helloworld2\" \"annotations\":map[\"flux.weave.works/sync-checksum\":\"191df8fec32a1deda61e32f995d8ffb7ea160781\" \"flux.weave.works/tag.echoserver\":\"semver:*\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.b3ah4IFHE0JJUNAUMz8Q7ikY7f9dDREvkAYx40-0FTA\"]]]}\nfrom server for: \"STDIN\": deployments.apps \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"deployments\" in API group \"apps\" in the namespace \"helloworld2\"" output=
ts=2019-03-26T01:16:17.57742458Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=184.507826ms err="running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"networking.k8s.io/v1, Resource=networkpolicies\", GroupVersionKind: \"networking.k8s.io/v1, Kind=NetworkPolicy\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"kind\":\"NetworkPolicy\" \"metadata\":map[\"annotations\":map[\"flux.weave.works/sync-checksum\":\"d98993cf8b5516b417533ebeeb0cfabe5c7ff6fa\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"app\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.yxyBpR0YQllFHVEBWXu_LxyedQX59j9IYMah3mi6rL4\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\"] \"spec\":map[\"ingress\":[map[\"ports\":[map[\"port\":'\\u1f90' \"protocol\":\"TCP\"]]]] \"podSelector\":map[\"matchLabels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"policyTypes\":[\"Ingress\" \"Egress\"]] \"apiVersion\":\"networking.k8s.io/v1\"]}\nfrom server for: \"STDIN\": networkpolicies.networking.k8s.io \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"networkpolicies\" in API group \"networking.k8s.io\" in the namespace \"helloworld2\"" output=
ts=2019-03-26T01:16:17.764813618Z caller=sync.go:521 component=cluster method=Sync cmd="kubectl apply -f -" took=186.921197ms err=null output="networkpolicy.networking.k8s.io/echoserver unchanged"
ts=2019-03-26T01:16:17.765076931Z caller=loop.go:210 component=sync-loop err="<cluster>:namespace/helloworld2: running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"/v1, Resource=namespaces\", GroupVersionKind: \"/v1, Kind=Namespace\"\nName: \"helloworld2\", Namespace: \"\"\nObject: &{map[\"apiVersion\":\"v1\" \"kind\":\"Namespace\" \"metadata\":map[\"annotations\":map[\"flux.weave.works/sync-checksum\":\"eff788baf5f8906b3879d666e60b5407f7b9f294\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"flux.weave.works/sync-gc-mark\":\"sha256.cFRnOwOXCB_oNpi52QLlUkjvODSLBIo1ntBRfzoPqp0\"] \"name\":\"helloworld2\" \"namespace\":\"\"]]}\nfrom server for: \"STDIN\": namespaces \"helloworld2\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"namespaces\" in API group \"\" in the namespace \"helloworld2\"; helloworld2:service/echoserver: running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"/v1, Resource=services\", GroupVersionKind: \"/v1, Kind=Service\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"metadata\":map[\"annotations\":map[\"kubectl.kubernetes.io/last-applied-configuration\":\"\" \"flux.weave.works/sync-checksum\":\"24d87352fb027d306ce7b03676b6d723dc59a799\"] \"labels\":map[\"app\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.3KBCW83dRp2s10Zkls_vLADf8Wl0E9Agzz8sIT3wHkU\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\"] \"spec\":map[\"ports\":[map[\"name\":\"http\" \"port\":'\\u1f90' \"protocol\":\"TCP\" \"targetPort\":'\\u1f90']] \"selector\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"] \"type\":\"ClusterIP\"] \"apiVersion\":\"v1\" \"kind\":\"Service\"]}\nfrom server for: \"STDIN\": services \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"services\" in API group \"\" in the namespace \"helloworld2\"; helloworld2:deployment/echoserver: running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"apps/v1, Resource=deployments\", GroupVersionKind: \"apps/v1, Kind=Deployment\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"spec\":map[\"selector\":map[\"matchLabels\":map[\"component\":\"echoserver\" \"app\":\"echoserver\"]] \"strategy\":map[\"type\":\"RollingUpdate\"] \"template\":map[\"metadata\":map[\"labels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"spec\":map[\"automountServiceAccountToken\":%!q(bool=false) \"containers\":[map[\"image\":\"gcr.io/google-samples/hello-app:1.0\" \"name\":\"echoserver\" \"ports\":[map[\"containerPort\":'\\u1f90' \"name\":\"web\" \"protocol\":\"TCP\"]] \"securityContext\":map[\"allowPrivilegeEscalation\":%!q(bool=false)]]]]] \"replicas\":'\\x01'] \"apiVersion\":\"apps/v1\" \"kind\":\"Deployment\" \"metadata\":map[\"name\":\"echoserver\" \"namespace\":\"helloworld2\" \"annotations\":map[\"flux.weave.works/sync-checksum\":\"191df8fec32a1deda61e32f995d8ffb7ea160781\" \"flux.weave.works/tag.echoserver\":\"semver:*\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.b3ah4IFHE0JJUNAUMz8Q7ikY7f9dDREvkAYx40-0FTA\"]]]}\nfrom server for: \"STDIN\": deployments.apps \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"deployments\" in API group \"apps\" in the namespace \"helloworld2\"; helloworld2:networkpolicy/echoserver: running kubectl: Error from server (Forbidden): error when retrieving current configuration of:\nResource: \"networking.k8s.io/v1, Resource=networkpolicies\", GroupVersionKind: \"networking.k8s.io/v1, Kind=NetworkPolicy\"\nName: \"echoserver\", Namespace: \"helloworld2\"\nObject: &{map[\"kind\":\"NetworkPolicy\" \"metadata\":map[\"annotations\":map[\"flux.weave.works/sync-checksum\":\"d98993cf8b5516b417533ebeeb0cfabe5c7ff6fa\" \"kubectl.kubernetes.io/last-applied-configuration\":\"\"] \"labels\":map[\"app\":\"echoserver\" \"flux.weave.works/sync-gc-mark\":\"sha256.yxyBpR0YQllFHVEBWXu_LxyedQX59j9IYMah3mi6rL4\"] \"name\":\"echoserver\" \"namespace\":\"helloworld2\"] \"spec\":map[\"ingress\":[map[\"ports\":[map[\"port\":'\\u1f90' \"protocol\":\"TCP\"]]]] \"podSelector\":map[\"matchLabels\":map[\"app\":\"echoserver\" \"component\":\"echoserver\"]] \"policyTypes\":[\"Ingress\" \"Egress\"]] \"apiVersion\":\"networking.k8s.io/v1\"]}\nfrom server for: \"STDIN\": networkpolicies.networking.k8s.io \"echoserver\" is forbidden: User \"system:serviceaccount:flux-system:flux\" cannot get resource \"networkpolicies\" in API group \"networking.k8s.io\" in the namespace \"helloworld2\""
ts=2019-03-26T01:16:17.767684483Z caller=daemon.go:624 component=daemon event="Sync: c89a1b3, <cluster>:namespace/helloworld2, helloworld2:deployment/echoserver, helloworld2:networkpolicy/echoserver, helloworld2:service/echoserver, helloworld:deployment/echoserver, helloworld:networkpolicy/echoserver, helloworld:service/echoserver" logupstream=false
ts=2019-03-26T01:16:20.95148383Z caller=loop.go:441 component=sync-loop tag=flux-sync old= new=c89a1b3362d904300f10a74c6daf3f2cdf900a18
ts=2019-03-26T01:16:22.190745263Z caller=loop.go:103 component=sync-loop event=refreshed url=git@github.com:2opremio/locked-down-flux.git branch=master HEAD=c89a1b3362d904300f10a74c6daf3f2cdf900a18
```

That said, that's something unrelated to this PR.

Issue #1309, opened 2018-08-24
Explain how the annotations work, thoroughly


squaremo:
- lock and lock_msg etc.
 - automated
 - ignore (there's already good material there)
 - tag filters

sahitpj:
I think I would like to work on this issue, if it hasn't been assigned to anyone yet.

squaremo:
> I think I would like to work on this issue, if it hasn't been assigned to anyone yet.

That'd be great! I'm more than happy to discuss and review things.

dholbach:
https://github.com/weaveworks/flux/blob/master/site/helm-integration.md#optional-fields lists two examples for the `automated` annotation (thanks to @stefanprodan).

They might be useful to pull out and explain more generally somewhere else.

dholbach:
https://github.com/weaveworks/flux/issues/1337#issuecomment-421152125 (a duplicate issue) has another example.

tux-00:
Any news on this issue ? A good documentation would ease adoption of flux.

paulmil1:
where the hell is documentation?  are you serious????

2opremio:
@paulmil1 the tone of your comment is unprofessional, uncalled for and against CNCF's code of conduct. Thus, I am locking this conversation. 

However, if you want to improve the documentation, feel free to create a PR for it, it will always be welcome as long as it confirms the code of conduct.

dholbach:
You might be looking for https://docs.fluxcd.io/en/latest/tutorials/driving-flux.html (we moved the docs a little).

PR #1675, opened 2019-01-18
add tutorial which discusses automation, annotations, locks

dholbach:
Closes: #1309

This tutorial aims to show step-by-step to set up Flux and drive it by using locks, annotations and automation. The idea was to keep it simple but have a practical example to look at with clear instructions and clear outcomes.

Feedback is very welcome. Let me know if I missed anything.

We also have instructions for doing the same with Helm, I'm just not sure how to best implement it. Maybe offer 'alternate' setup steps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ Get started by browsing through the documentation below:
 - Get Started with Flux
   - [Standalone Flux](/site/get-started.md)
   - [Flux using Helm](/site/helm-get-started.md)
+  - [Automation: annotations and locks](/site/annotations-tutorial.md)
 - Operating Flux
   - [Using fluxctl to control Flux](/site/fluxctl.md)
   - [Helm Operator](/site/helm-operator.md)
--- a/site/annotations-tutorial.md
+++ b/site/annotations-tutorial.md
@@ -1,0 +1,146 @@
+# Driving Weave Flux - automations, locks and annotations
+
+In this tutorial we want to get a better feel for what we can do with Weave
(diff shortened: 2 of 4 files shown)


stefanprodan:
Installing operators in the `default` namespace is bad practice in my opinion. 

After each fluxctl command I would put a snippet with the annotations it created and I would mention that you don't need fluxctl for all those operations, doing the changes in Git works the same.

dholbach:
> Installing operators in the `default` namespace is bad practice in my opinion.

Mh, ok. I take your point. It's just a bit clunky to have to type `--k8s-fwd-ns` for every single time you call `fluxctl`.

> After each fluxctl command I would put a snippet with the annotations it created and I would mention that you don't need fluxctl for all those operations, doing the changes in Git works the same.

Will do. Good idea.

dholbach:
> > Installing operators in the `default` namespace is bad practice in my opinion.
> 
> Mh, ok. I take your point. It's just a bit clunky to have to type `--k8s-fwd-ns` for every single time you call `fluxctl`.

For this I'd either need to

1. add a step in which the non-Helm setup changes `flux-account.yaml` (then both sections can use `--k8s-fwd-ns` in every step) - which I still feel is a bit clunky.
1. differ in the post-setup steps of the two types of install

I'd obviously prefer to go with 1).

Another option would be to go with `default` on this one. And then in a separate issue discuss where we want the example deploy to go and if `fluxctl` can automatically use that namespace. WDYT?

PR #1680, opened 2019-01-21
Run signal listener in a goroutine instead of deferring.

### LINKED ISSUE #1679: Helm operator does not respond to signals
I ran into this issue running the helm-operator locally. There appears to be a bug in the signal handler not being run because it is in a defer from main. The result is that Ctrl-C and such do not actually signal the process to exit, which can be quite annoying. Sending PR to fix this imminently.

captncraig:
Fixes #1679

### DIFF
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -131,12 +131,6 @@ func main() {
 		errc <- fmt.Errorf("%s", <-c)
 	}()
 
-	go func() {
-		logger.Log("exiting...", <-errc)
-		close(shutdown)
-		shutdownWg.Wait()
-	}()
-
 	mainLogger := log.With(logger, "component", "helm-operator")
 
 	cfg, err := clientcmd.BuildConfigFromFlags(*master, *kubeconfig)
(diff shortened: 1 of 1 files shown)


squaremo:
The existing code is probably ported from fluxd, where it is deferred so that it can appear in the context of setting up the shutdown machinery, but run last -- the idea is to grant the various other goroutines, told to shut down by closing the channel, a chance to gracefully do so before exiting.

The reason is doesn't work in helm-operator is that the last thing to happen in `main()` is synchronous, so the deferred procedure is not reached in the usual course of operation. Running it in a goroutine means it will be reached -- but waiting on the waitgroup in a goroutine will not have the effect of allowing the other tasks to complete.

Since `opr` is given a shutdown channel and waitgroup, I think you could run it in a goroutine, and keep the deferred procedure to handle the graceful shutdown. Or, if you prefer, inline that bit at the end of `main`.

2opremio:
@captncraig Are you interesting in finishing up the PR following @squaremo 's advice?

captncraig:
Sorry, I got sidetracked and lost my dev environment on this one. It would be best if someone else finish this up, or just close it. My bad.

squaremo:
> Sorry, I got sidetracked and lost my dev environment

No worries -- spotting the problem and suggesting a plausible solution gets us most of the way there, so thank you :+1:

squaremo:
Thanks @2opremio :pineapple:

PR #1684, opened 2019-01-23
Publish Flux chart with GitHub actions

stefanprodan:
This PR adds a GitHub workflow that lints, packages and pushes the Flux chart to `gh-pages` branch when tagging the master branch with `chart-*`

For more details on this: [Automate Helm chart repository publishing with GitHub Actions and Pages](https://medium.com/@stefanprodan/automate-helm-chart-repository-publishing-with-github-actions-and-pages-8a374ce24cf4)

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,0 +1,10 @@
+workflow "Publish Helm chart" {
+  on = "push"
+  resolves = ["helm-gh-pages"]
+}
+
+action "helm-gh-pages" {
+  uses = "stefanprodan/gh-actions/helm-gh-pages@master"
+  args = ["chart/flux","https://weaveworks.github.io/flux","chart-"]
+  secrets = ["GITHUB_TOKEN"]
+}
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -8,9 +8,15 @@ action "helm-lint" {
   args = ["lint chart/flux"]
 }
 
-action "helm-gh-pages" {
+action "tag-filter" {
   needs = ["helm-lint"]
+  uses = "actions/bin/filter@master"
+  args = "tag chart-*"
+}
+
+action "helm-gh-pages" {
+  needs = ["tag-filter"]
   uses = "stefanprodan/gh-actions/helm-gh-pages@master"
-  args = ["chart/flux","https://weaveworks.github.io/flux","chart-"]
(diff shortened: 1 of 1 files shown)


.github/main.workflow
```diff
@@ -0,0 +1,16 @@
+workflow "Publish Helm chart" {
+  on = "push"
+  resolves = ["helm-gh-pages"]
```
squaremo: Can the action not be gated on the tag, here?

.github/main.workflow
```diff
@@ -0,0 +1,16 @@
...
+
+action "helm-lint" {
+  uses = "stefanprodan/gh-actions/helm@master"
+  args = ["lint chart/flux"]
+}
+
+action "helm-gh-pages" {
+  needs = ["helm-lint"]
+  uses = "stefanprodan/gh-actions/helm-gh-pages@master"
+  args = ["chart/flux","https://weaveworks.github.io/flux","chart-"]
```
squaremo: What's the last argument -- a prefix for the tag to look for?

stefanprodan: Yes, as per our chart release procedure we tag the master branch with `chart-<VERSION>`. This action will only publish the chart to gh-pages when that prefix is used.

.github/main.workflow
stefanprodan: Sorry I don't understand your question, can you please elaborate? The `tag` gate is inside the `helm-gh-pages` action. The linter will run on every push.

.github/main.workflow
squaremo: OK, so combining with my other question: is it not possible in GitHub Actions to only run this bit if it's tagged, rather than relying on the script to know how to do that?

.github/main.workflow
squaremo: What happens when the linter fails?

stefanprodan: The workflow stops and errors out, marking the PR as invalid or stopping the publishing if we're dealing with git tag.

.github/main.workflow
stefanprodan: Yes that's possible. I've added a tag filter before the the publish action.

squaremo: Neat!

stefanprodan:
I look forward to the day when we'll be able to run the whole CI in GitHub Actions, feels way nicer than CircleCI.

PR #1685, opened 2019-01-23
WIP: Modify insecure hosts to first try HTTPS with no CA verfication.

michaelfig:
Adapted from: https://github.com/weaveworks/flux/commit/096bd478dbe2808d3b3a5699b73f94568b06a9f7

Please do not merge: this is just to document what I did for my own use, and there isn't any new code except what already exists in the above commit.

Thanks,
Michael.


michaelfig:
See #1526 instead.

Issue #1686, opened 2019-01-24
Flux conflicts with pods autoscale


wenbli:
When a controller is set to manual mode for production environment, and have the autoscale enabled for the deployment, fluxctl and horizontalpodautoscaler conflicts causing pods keep deleting/creating to meet the "replicas" decided by fluxctl and autoscale.

wenbli:
Thanks, it works!

PR #1688, opened 2019-01-24
FAQ improvements 

stefanprodan:
- add multi-cluster multi-path option to FAQ
- add HPA replicas workaround to FAQ Fix: #1686
- rename FHR to HR in FAQ
- remove question about HR deletions on uninstall since it doesn't apply anymore
- remove comment about memcached being optional

### DIFF
--- a/site/faq.md
+++ b/site/faq.md
@@ -460,7 +460,17 @@ To avoid this you have to manually delete the Flux Helm Operator with `kubectl -
 
 ### I have a dedicated Kubernetes cluster per environment and I want to use the same Git repo for all. How can I do that?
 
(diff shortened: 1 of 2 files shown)


site/faq.md
```diff
@@ -430,14 +429,19 @@ only returns the kinds of resource mentioned above. So,
 annotating a running resource only works if it's one of those
 kinds; putting the annotation in the file always works.
 
+### How can I prevent Flux overriding the replicas when using HPA?
```
hiddeco: This should be added to the ToC.

stefanprodan:
Thanks for the review @hiddeco

PR #1689, opened 2019-01-24
[WIP] add goreleaser as build/release process

### LINKED ISSUE #1377: Publish fluxctl Linux packages
It'd be great to have `fluxctl` packages for Linux distributed without having to go download a binary (and chmod) or build from source.

A similar discussion came up in the `eksctl` project recently (https://github.com/weaveworks/eksctl/issues/215) and I snaps, as they're supported on a variety of Linux flavours and the build ties in with our CI easily (only one additional yaml required).

Unfortunately are builds for both `fluxctl` and `eksctl` currently blocked on access to .dotfiles, but a strawman solution was just brought up to remedy this: https://github.com/snapcore/snapd/pull/5845

guilhem:
As a first step to fix #1377, I begin to work on `goreleaser`.

First major problem for this to be merged:
* I don't decorrelate `helm-operator` from `fluxctl`/`fluxd`.  
It, maybe, can be done with 2 different files but I suggest slipping this 2 projects into 2 different repo...
* I didn't add CI
* I didn't add snap / deb / rpm building (easy but waiting for your input about this work)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,53 @@
+# This is an example goreleaser.yaml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+before:
+  hooks:
+    - dep ensure -vendor-only
+builds:
+- binary: fluxctl
+  main: ./cmd/fluxctl/
+  goos:
+    - linux
+    - darwin
+    - windows
+  env:
+   - CGO_ENABLED=0
+  goarch:
+    - amd64
+    - arm
+- binary: fluxd
+  main: ./cmd/fluxd/
+  goos:
+    - linux
+  env:
+   - CGO_ENABLED=0
+  goarch:
+    - amd64
+    - arm
+- binary: helm-operator
+  main: ./cmd/helm-operator/
+  goos:
+    - linux
+  env:
+   - CGO_ENABLED=0
+  goarch:
+    - amd64
+    - arm
+archive:
+  # format: binary
+  replacements:
+    darwin: Darwin
+    linux: Linux
+    windows: Windows
+    386: i386
+    amd64: x86_64
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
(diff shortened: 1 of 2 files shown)


2opremio:
I am not sure it makes sense to package the helm operator and fluxd, since they are designed to work inside a kubernetes cluster (and there are already Docker images for that). I would say it only makes sense to package fluxctl. @squaremo would you agree?

guilhem:
goreleaser can also package docker images ^^
But yes, that's not main point.

One advantage I see with goreleaser in process, is to be able to push ARM docker images easily.

Having a release as tar.gz is only a "plus" for people who don't trust image "from the internet", but trust a binary on GH :x

stefanprodan:
I would keep fluxctl only and remove the changelog. We have a curated change log in the repo and I don't think we want goreleaser to publish all the commits on the GitHub release page. As for automating this I would go with GitHub Actions.

stefanprodan:
Also we should not run goreleaser when we tag a helm-op or chart release. This could be done with GitHub Action filters similar to  https://raw.githubusercontent.com/weaveworks/flux/master/.github/main.workflow

.goreleaser.yml
```diff
@@ -0,0 +1,53 @@
+# This is an example goreleaser.yaml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+before:
+  hooks:
+    - dep ensure -vendor-only
+builds:
+- binary: fluxctl
+  main: ./cmd/fluxctl/
```
stefanprodan: Will this inject version at build time? see https://github.com/weaveworks/flux/blob/master/Makefile#L33

guilhem: default is `-s -w -X main.version={{.Version}} -X main.commit={{.ShortCommit}} -X main.date={{.Date}}`

2opremio:
I think that the current process works fine. Let me know if you think otherwise

Issue #1396, opened 2018-09-26
Flux can't check for updated images in a private registry


brantb:
When using a private docker registry, flux doesn't seem to be able to check for updated images using the default ServiceAccount's `imagePullSecrets`.

I tested this out on a brand new [Azure Container Registry][acr] instance and k8s namespace created for this purpose.

```yaml
# Created using:
# kubectl create secret docker-registry registry-secret -o json \
#   --docker-server=branttest.azurecr.io \
#   --docker-email=xyz@example.com \
#   --docker-username=branttest \
#   --docker-password=<redacted>
apiVersion: v1
data:
  .dockerconfigjson: <snip>
kind: Secret
metadata:
  name: registry-secret
  namespace: flux-auth
type: kubernetes.io/dockerconfigjson
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: default
  namespace: flux-auth
imagePullSecrets:
  - name: registry-secret
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app: party-clippy
  name: party-clippy
  namespace: flux-auth
spec:
  selector:
    matchLabels:
      app: party-clippy
  template:
    metadata:
      labels:
        app: party-clippy
    spec:
      containers:
        - image: branttest.azurecr.io/party-clippy:latest
          imagePullPolicy: Always
          name: party-clippy
          args: ["-p", "80"]
```
k8s is able to pull and run the container. However, after killing flux and memcached (to wipe out any cached creds) I still see this in the logs:

`ts=2018-09-26T20:14:29.675347521Z caller=warming.go:162 component=warmer canonical_name=branttest.azurecr.io/party-clippy auth={map[]} err="requesting tags: Get https://branttest.azurecr.io/v2/party-clippy/tags/list: unauthorized: authentication required"`

Adding an `imagePullSecret` directly to the Deployment makes this error go away (presumably because it's working).

[acr]: https://docs.microsoft.com/en-us/azure/container-registry/container-registry-intro

brantb:
It looks like I was actually running an older version of flux in this cluster (v1.5.0). I'll reopen this once I get a chance to test with the latest if it's still an issue.

sc250024:
Getting the same error; adding the ```imagePullSecrets:
  - name: registry-secret``` didn't do anything, and I'm running the latest version of Flux.

Morriz:
[...]

PR #1694, opened 2019-01-29
AKS credentials for ACR

alanjcastonguay:
Azure Kubernetes Service (AKS) nodes' kubelet can use the cluster's
service principal to fetch images from Azure Container Registry (ACR).

If azure.json is projected into the Flux pod, consume the service
principal credentials when authenticating to azure container registry.

Note that this method may eventually be deprecated, possibly replaced
by Managed Identity + OAuth, similar to the GCP implementation.
See https://github.com/kubernetes/kubernetes/issues/58034

Prompted by [discussion](https://weave-community.slack.com/archives/C4U5ATZ9S/p1547928858828100) with @squaremo on slack

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -45,6 +45,12 @@ spec:
       - name: git-keygen
         emptyDir:
           medium: Memory
+      {{- if .Values.registry.acr.enabled }}
+      - name: acr-credentials
+        hostPath:
+          path: "{{ .Values.registry.acr.mountPath }}"
+          type: ""
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
@@ -66,6 +72,11 @@ spec:
             readOnly: true
           - name: git-keygen
             mountPath: /var/fluxd/keygen
+          {{- if .Values.registry.acr.enabled }}
+          - name: acr-credentials
+            mountPath: /etc/kubernetes/azure.json
+            readOnly: true
+          {{- end }}
           env:
           - name: KUBECONFIG
             value: /root/.kubectl/config
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -48,7 +48,7 @@ spec:
       {{- if .Values.registry.acr.enabled }}
       - name: acr-credentials
         hostPath:
-          path: "{{ .Values.registry.acr.mountPath }}"
+          path: "{{ .Values.registry.acr.hostPath }}"
           type: ""
       {{- end }}
       containers:
--- a/registry/azure.go
+++ b/registry/azure.go
@@ -1,0 +1,55 @@
+package registry
+
+import (
+	"encoding/json"
+	"os"
+	yaml "gopkg.in/yaml.v2"
+)
+
+const (
+    // Mount volume from hostpath.
+    azureCloudConfigJsonFile = "/etc/kubernetes/azure.json"
+
+    // List from https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/azure/azure_credentials.go
+    azureContainerRegistryHosts = []string{".azurecr.io", ".azurecr.cn", ".azurecr.de", ".azurecr.us"};
+)
+
+type azureCloudConfig struct {
+    aadClientId string `json:"aadClientId"`
+    aadClientSecret string `json:"aadClientSecret"`
+}
+
+// Fetch Azure Active Directory clientid/secret pair from azure.json, usable for container registry authentication.
+//
+// Note: azure.json is populated by AKS/AKS-Engine script kubernetesconfigs.sh. The file is then passed to kubelet via
+// --azure-container-registry-config=/etc/kubernetes/azure.json, parsed by kubernetes/kubernetes' azure_credentials.go
+// https://github.com/kubernetes/kubernetes/issues/58034 seeks to deprecate this kubelet command-line argument, possibly
+// replacing it with managed identity for the Node VMs. See https://github.com/Azure/acr/blob/master/docs/AAD-OAuth.md
+func GetAzureCloudConfigAADToken(host string) (creds, error) {
+    jsonFile, err := os.Open(azureCloudConfigJsonFile)
+    if err != nil {
+        return creds{}, err
+    }
+    defer jsonFile.Close()
+
+    var token azureCloudConfig
+    decoder := json.NewDecoder(jsonFile)
+    if err := decoder.Decode(&token); err != nil {
+        return creds{}, err
+    }
+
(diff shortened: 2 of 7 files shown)


registry/azure.go
```diff
@@ -0,0 +1,53 @@
...
+    AADClientSecret string `json:"aadClientSecret"`
+}
+
+// Fetch Azure Active Directory clientid/secret pair from azure.json, usable for container registry authentication.
+//
+// Note: azure.json is populated by AKS/AKS-Engine script kubernetesconfigs.sh. The file is then passed to kubelet via
+// --azure-container-registry-config=/etc/kubernetes/azure.json, parsed by kubernetes/kubernetes' azure_credentials.go
+// https://github.com/kubernetes/kubernetes/issues/58034 seeks to deprecate this kubelet command-line argument, possibly
+// replacing it with managed identity for the Node VMs. See https://github.com/Azure/acr/blob/master/docs/AAD-OAuth.md
+func GetAzureCloudConfigAADToken(host string) (creds, error) {
```
squaremo: I don't think this or `HostIsAzureContainerRegistry` need to be exported -- they are just used in the tests and in `credsFor`, both in the same package.

alanjcastonguay: I was matching the pattern of `GetGCPOauthToken` being exported - it's also only used in those two places in the `registry` package. I wasn't sure if that was intentional. I can make this private anyway.

squaremo: Fair point, following the apparent convention ought not earn you demerits :-)

4c74356b41:
I can confirm this is working by adding these changes to the deploy/flux:
```
        volumeMounts:
        xxx
        - mountPath: /etc/kubernetes/azure.json
          name: acr-credentials
        xxx
      volumes:
      xxx
      - hostPath:
          path: /etc/kubernetes/azure.json
          type: ""
        name: acr-credentials
```

squaremo:
> I can confirm this is working

Splendid, thank you very much for trying it out @4c74356b41

alanjcastonguay:
Awesome!

Let me fix the formatting and docs....

timwebster9:
ACR user here...

Just so I understand, does this PR remove the need to provide `imagePullSecrets` along with your k8s resources in order to enable the registry scanning to work?

Oddly enough, that method works for me, unlike what people are reporting in #1396.  But this method looks nicer :-)

Also, will this help with ACR Helm authentication at all?

4c74356b41:
In my experiments using imagePullSecrets doesnt work if you grant AKS permissions to ACR, so you have to strip AKS permissions on ACR and after that it will work with imagePullSecrets.

Not sure what do you mean about helm and ACR. helm is a client side tool

timwebster9:
@4c74356b41 

The SP my cluster is using has AcrPull role assignment, and I'm using a different SP (also with AcrPull) for the credentials used in the `imagePullSecrets`.  It seems to work.

Re: the ACR Helm thing - sorry I wasn't clear.  ACR can also host Helm charts (i.e. it's a Helm repository), but since it's private it requires authentication.  

https://docs.microsoft.com/en-us/azure/container-registry/container-registry-helm-repos

I'm guessing nobody has tried this or looked at it.  I tried briefly using the method described [here](https://github.com/weaveworks/flux/blob/master/site/helm-integration.md#authentication-for-helm-repos), but it didn't work.  It's not an urgent requirement for us at the moment so I haven't made time to investigate further.  The authentication method shown in the docs uses an OAuth token which expires after an hour, so not sure this will work.

However when using ACR as a container registry, you can authenticate directly using a service principal (i.e. with `docker login`), so I was wondering if this would work with the Helm repo authentication somehow.

4c74356b41:
I'm well aware of that, what I dont understand, with flux all your templates are in the repo already, doesnt matter you have them in a private repo as well? also, if you have acr pull for AKS sp it wont even use imagePullSecrets

timwebster9:
> I'm well aware of that, what I dont understand, with flux all your templates are in the repo already, doesnt matter you have them in a private repo as well? also, if you have acr pull for AKS sp it wont even use imagePullSecrets

The reason I needed `imagePullSecrets` was not for AKS to pull images, but for Flux to poll/scan the repositories.  Unless I'm mistaken, Flux doesn't use the AKS SP credentials to do this. 

So back to the my original question: my guess is that this PR will provide those same credentials (via the host mount) to Flux, so we don't need the `imagePullSecrets`.  Does this sound right?

We currently have some internal Helm charts published to ACR, so I wanted to see if Flux could install those.  However we don't need to do this, rather we can just install them from Github instead.

I'm not sure what you mean by `your templates are in the repo already`.  What repo?  The Helm charts are in separate repos, not in the flux one.  The Flux repo only has the `HelmRelease` resources that reference these charts.

4c74356b41:
oh, ok. I didnt knew you can do that. I thought it needed charts in the same repo.

indeed flux doesnt use aks sp credentials (this is why this PR exists), but what I'm telling you is that the solution worked for you because you didnt have ACRpull for the AKS sp, if you did - it wouldnt.

timwebster9:
> oh, ok. I didnt knew you can do that. I thought it needed charts in the same repo.
> 
> indeed flux doesnt use aks sp credentials (this is why this PR exists), but what I'm telling you is that the solution worked for you because you didnt have ACRpull for the AKS sp, if you did - it wouldnt.

I have ACR pull on both the AKS SP and the one I use for `imagePullSecrets' and everything works fine.  I guess what you're saying is you need to use `imagePullSecrets` both for polling the registry *and* pulling images?

Before I added `imagePullSecrets` to my Flux resources AKS could pull images fine using the usual SP role assignment on ACR.  I only had to add it for the registry scan bit.

alanjcastonguay:
@squaremo corrected whitespace (via gofmt) in 3cb2179, un-Exported two methods in df5ca66, added docs to faq.md in cf30277, and plumbed hostPath mount into helm chart (disabled by default, always readOnly) in 0ce0668. I think that's all the requested changes.

4c74356b41:
@alanjcastonguay cool, didnt think about readOnly, makes a lot of sense.
@timwebster9 ok, i guess you are special :)

squaremo:
@alanjcastonguay This is a thorough job, thank you very much!

chart/flux/templates/deployment.yaml
```diff
@@ -45,6 +45,12 @@ spec:
       - name: git-keygen
         emptyDir:
           medium: Memory
+      {{- if .Values.registry.acr.enabled }}
+      - name: acr-credentials
+        hostPath:
+          path: "{{ .Values.registry.acr.mountPath }}"
```
hiddeco:
```suggestion
          path: "{{ .Values.registry.acr.hostPath }}"
```

as per docs and `values.yaml` example.

alanjcastonguay: Thanks for the catch! Looks like this is the only place I made that mistake.

PR #1695, opened 2019-01-30
Warn users about external changes in sync tag

2opremio:
Multiple users have reported an unexpectedly high number of changes in the git
respository's sync tag.

This has been attributed to multiple fluxd instances, running in separate
clusters, using the same git repository and sync tag.

In addition, multiple fluxd instances using the same tag is a bad idea in
general, since it will lead to commits being missed by fluxd.

This change tracks the value of the sync tag and prints a warning when detecting
an external change.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -42,19 +42,17 @@ const (
 // Daemon is the fully-functional state of a daemon (compare to
 // `NotReadyDaemon`).
 type Daemon struct {
-	V                        string
-	Cluster                  cluster.Cluster
-	Manifests                cluster.Manifests
-	Registry                 registry.Registry
-	ImageRefresh             chan image.Name
-	Repo                     *git.Repo
-	GitConfig                git.Config
-	lastKnownSyncTagRev      string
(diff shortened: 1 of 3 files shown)


daemon/daemon.go
```diff
@@ -42,17 +42,19 @@ const (
...
-	EventWriter    event.EventWriter
-	Logger         log.Logger
+	V                        string
+	Cluster                  cluster.Cluster
+	Manifests                cluster.Manifests
+	Registry                 registry.Registry
+	ImageRefresh             chan image.Name
+	Repo                     *git.Repo
+	GitConfig                git.Config
+	lastKnownSyncTagRev      string
```
squaremo: Is there any reason to stick this in the struct, rather than it being local to the loop? (You'd have to pass it to doSync maybe)

2opremio: Fair point, no real reason, no. I will change it.

2opremio: Done, PTAL

2opremio:
Ah, Will fix ASAP and merge.

PR #1702, opened 2019-02-01
Merge existing image credentials into a fresh set

### LINKED ISSUE #1485: Flux uses wrong imagePullSecret
Our clusters are running on GKE under GCP `project-x`. Images are however published to GCP projects dedicated to the product (e.g. GCP project `product-y`).

We add a Docker secret for this GCP project to every product's namespace and use this to pull from the GCR dedicated to this product with a `imagePullSecrets` set on the workload.

Now what Flux tries to do is to pull an image from `product-y` with the credentials from `product-z`.

Say we have a product `subscription` and a product `publish` Flux logs the following:

`ts=2018-10-30T11:10:19.340162325Z caller=warming.go:192 component=warmer canonical_name=eu.gcr.io/subscription/image auth="{map[eu.gcr.io:<registry creds for _json_key@eu.gcr.io, from publish:secret/gcrsecret>]}" err="requesting tags: denied: Failed to read tags for host 'eu.gcr.io', repository '/v2/subscription/image/tags/list'`

---

After some thinkering with @squaremo we came to the conclusion that credential merge happens on registry (e.g. `eu.gcr.io`), which is the same for both secrets although they differ (they have a unique `project_id`).

I think the solution would be for Flux to adhere to the same rules as Kubernetes and only use `imagePullSecrets` from the same namespace as the workload.

hiddeco:
Fixes #1485

With the previous merging strategy the set of credentials which got
mutated could be assigned to other images, resulting in overwrites
where we would (and did) not expect them.

This new approach is the most direct fix to work around mutating the
credentials of other images. It creates a fresh set of (empty)
credentials and merges the others into it before assigning it to the
image.

[Details](https://github.com/weaveworks/flux/pull/1702#issuecomment-460265100)

### DIFF
--- a/cluster/kubernetes/images.go
+++ b/cluster/kubernetes/images.go
@@ -152,7 +152,10 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 			for imageID, creds := range imageCreds {
 				existingCreds, ok := allImageCreds[imageID]
 				if ok {
-					existingCreds.Merge(creds)
+					mergedCreds := registry.NoCredentials()
+					mergedCreds.Merge(existingCreds)
+					mergedCreds.Merge(creds)
+					allImageCreds[imageID] = mergedCreds
 				} else {
 					allImageCreds[imageID] = creds
 				}


cluster/kubernetes/images.go
```diff
@@ -150,10 +150,7 @@ func (c *Cluster) ImagesToFetch() registry.ImageCreds {
 
 			// Merge creds
```
2opremio:
```suggestion
```

hiddeco:
After deploying this to one of our clusters it seems to create a new issue where Flux is unable to pull images tied to `HelmRelease` manifests.

```
ts=2019-02-01T14:30:10.59171206Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2019-02-01T14:32:24.797392145Z caller=warming.go:192 component=warmer canonical_name=eu.gcr.io/<image> auth={map[]} err="requesting tags: denied: Failed to read tags for host 'eu.gcr.io', repository '/v2/<image>/tags/list'"
ts=2019-02-01T14:32:26.38038089Z caller=warming.go:192 component=warmer canonical_name=eu.gcr.io/<image> auth={map[]} err="requesting tags: denied: Failed to read tags for host 'eu.gcr.io', repository '/v2/<image>/tags/list'"
ts=2019-02-01T14:33:31.017196418Z caller=warming.go:192 component=warmer canonical_name=eu.gcr.io/<image> auth={map[]} err="requesting tags: denied: Failed to read tags for host 'eu.gcr.io', repository '/v2/<image>/tags/list'"
```

squaremo:
> Flux is unable to pull images tied to `HelmRelease` manifests because it fails to extract credentials from this resource kind.

fluxd doesn't try to extract secrets from `HelmRelease` resources (they don't have any); it relies on scanning images from resources created as part of the release. So, what is happening is fluxd doesn't pick up credentials for those.

squaremo:
> So, what is happening is fluxd doesn't pick up credentials for those.

No, I'm being silly -- what's happening is that the credentials extracted from HelmRelease (i.e., none) overwrite the credentials from the other workloads.

This is a bug, but it's not clear what the correct solution is. Going back to merging credentials will solve it, but is not obviously correct either; not exclusing HelmRelease objects when scanning for images will give the right result, but seems like an arbitrary special case.

squaremo:
Logically, I'm not sure how this changes the situation: before, credentials were merged meaning an entry for a particular registry can be overwritten (last entry wins); now, whichever workload is scanned first will supply the credentials, and no others can (first entry wins). Either way, if you have two distinct credentials against workloads using the same image, one of them will non-deterministically "win".

squaremo:
> Logically, I'm not sure how this changes the situation: before, credentials were merged meaning an entry for a particular registry can be overwritten (last entry wins); now, whichever workload is scanned first will supply the credentials, and no others can (first entry wins). Either way, if you have two distinct credentials against workloads using the same image, one of them will non-deterministically "win".

.. all of which doesn't go anywhere in countering your actual experience of it _not working_!

We found the problem with the existing code: https://github.com/weaveworks/flux/blob/master/cluster/kubernetes/images.go#L155 mutates a set of credentials which can be assigned to other images. The most direct fix would be to create a fresh set of credentials for each image, and merge the others into it; i.e.,

```
existingCreds, ok := allImageCreds[imageID]
if ok {
    mergedCreds := registry.NoCredentials()
    mergedCreds.Merge(existingCreds)
    mergedCreds.Merge(creds)
    allImageCreds[imageID] = mergedCreds
} else {
    allImageCreds[imageID] = creds
}
```

hiddeco:
Force-pushed the fix mentioned above to solve the issue without leaving a trail of my original approach, which skipped the merge if there were already credentials present for an image.

PR #1712, opened 2019-02-07
add "skipDepUpdate" config parameter to HelmRelease

yinzara:
some helm charts (istio being an example) add helm charts into the "charts" folder directly and do not need "helm dep up" to be called. In fact doing so will cause an error since the requirements have already been fulfilled.  While you can disable this feature system wide it is much more useful to allow individual helm releases to specify this behavior.  this commit adds a skipDepUpdate attribute to the
HelmRelease crd that causes the dependency update step to be skipped for that release.

It defaults to "false" and is backward compatible (as it is not required).

### DIFF
--- a/deploy-helm/flux-helm-release-crd.yaml
+++ b/deploy-helm/flux-helm-release-crd.yaml
@@ -31,6 +31,8 @@ spec:
               format: int64
             resetValues:
               type: boolean
+            skipDepUpdate:
+              type: boolean
             valueFileSecrets:
               type: array
               properties:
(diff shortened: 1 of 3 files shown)


deploy-helm/flux-helm-release-crd.yaml
```diff
@@ -31,6 +31,8 @@ spec:
               format: int64
             resetValues:
               type: boolean
+            skipDepUpdate:
```
squaremo: Since this pertains to charts from git only (published charts _always_ have their dependencies, as I understand it), should it not go in the `.spec.chart.git` section?

yinzara: Technically no.  Published charts don't always have their dependencies.  Sometimes they do. Sometimes they don't.  So it's not really appropriate in the git section.

squaremo: I was under the impression that helm won't let you package a chart without its dependencies, implying that any published chart _must_ include its dependencies. The helm operator doesn't attempt to download the dependencies unless the chart came from git.

PR #1713, opened 2019-02-07
 Release Helm chart v0.6.1

stefanprodan:


stephenmoloney:
missing reference to  improvement
```
# Do not scan images that match these glob expressions
excludeImage:
```

stefanprodan:
Thanks @stephenmoloney I've added it.

stefanprodan:
Yes I forgot about that, the change log should contain the version bump even if it's the same PR

Issue #1716, opened 2019-02-07
Add imagePullSecrets to template


jimangel:
```
spec:
      {{- if .Values.image.pullSecret -}}
      imagePullSecrets:
        - name: {{ .Values.image.pullSecret }}
      {{- end }}
```

This will help companies deploy flux using helm + private internal repository.

jimangel:
Thanks @stefanprodan !

PR #1718, opened 2019-02-08
Add image pull secret options to Helm chart

stefanprodan:
Allows flux, helm-op and memcached images to be pulled from a private container registry.

Fix: #1716


stefanprodan:
Thanks for the review @hiddeco

PR #1737, opened 2019-02-13
Allow pod annotations to be set when using the helm chart

stefansedich:
Currently it is not possible to set pod annotations for the flux pods, this is useful if for example you are using something like kube2iam and need to assign an IAM role to the pods so that ECR image scanning can work.


stefanprodan:
You changes have been published to our Helm repository, the chart version is 0.6.3.

PR #1745, opened 2019-02-15
Use UID of HelmRelease as dry run release name

### LINKED ISSUE #1703: Longer release names fails to update
When using a release name that's 51+ characters long we run into tiller rejecting the the dry-run operations because flux adds -temp to the end (pushing it above helm's self imposed 53 character limit.) 

https://github.com/helm/helm/blob/master/pkg/tiller/release_server.go#L45-L48

Ideally, flux would dry-run with a similar name, to avoid any kind of regex errors. this one was interesting because on an initial install all the validations passed, only when flux tried to run dry-run it failed.

in the weave users slack, we discussed maybe using a uuid for dry run, or reusing some other bit of metadata. 
https://github.com/weaveworks/flux/blob/4270060c4e260ba1052421ae28320685a88cd8ad/integrations/helm/release/release.go#L127

maybe flux needs it's own validator, if it's going to create a release name that doesn't match tillers, then we could emit a message, or change it. (but then what happens when tiller changes their validator, etc)

hiddeco:
Fixes #1703 

This prevents dry run issues when a user has a release name that is 51+
characters long, as our previous strategy was to add '-temp' to the
given release name and Helm has a self imposed release name character
limit of 53 (51 + '-temp' = >53).

### DIFF
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -48,7 +48,7 @@ import (
 	google_protobuf "github.com/golang/protobuf/ptypes/any"
 	"github.com/google/go-cmp/cmp"
 	"github.com/ncabatoff/go-seq/seq"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


stefanprodan:
If the dry run fails how can you tell from the logs what HelmRelease was it? I assume you'll see just a UID

hiddeco:
@stefanprodan see the change to the logger.

stefanprodan:
That logs the info not the error

integrations/helm/chartsync/chartsync.go
```diff
@@ -509,7 +509,7 @@ func (chs *ChartChangeSync) shouldUpgrade(chartsRepo string, currRel *hapi_relea
 
 	// Get the desired release state
 	opts := release.InstallOptions{DryRun: true}
-	tempRelName := currRel.GetName() + "-temp"
+	tempRelName := string(fhr.UID)
 	desRel, err := chs.release.Install(chartsRepo, tempRelName, fhr, release.InstallAction, opts, &chs.kubeClient)
 	if err != nil {
 		return false, err
```
stefanprodan: Maybe append the HelmRelease name to the error

hiddeco: Another thing we could do is use the release name from the spec  (`fhr.Spec.ReleaseName`) for informative errors (just like I did with the log entry) but use the given `releaseName` for running operations.

stefanprodan: sounds good to me

PR #1746, opened 2019-02-18
Record the Revision in the status

paroxp:
## What

At current state, we're unable to track what exact revision of code is
running on our clusters.

This should not be a problem, if we'd rely on chart "version" being the
source of truth, however sadly it is possible to make changes to the
chart without bumping a version. Possibly should be bumped with a CI/CD
tool, but it also has a potential to fail.

Our use case:

```
Merge to master happens
CI picks the changes up
CI runs unit tests
CI marks merge as staging ready by promoting a tag
Flux running on staging cluster/namespace picks up the changes
CI (that is not exposed to the internet - webhooks are a no no) is
   assured the change has been applied
CI runs the integration tests
CI marks merge as staging ready by promoting a tag
Flux running on production cluster/namespace picks up the changes
CI notifies of completion
```

This could be done better. Instead of us passing around a string from
function to function, we could convert the current code obtaining the
revision into it's own function and call it at the very end to gather
this again. Open to interpretation.

## How to review

- Code review and sanity check should be sufficient

### DIFF
--- a/integrations/apis/flux.weave.works/v1beta1/types.go
+++ b/integrations/apis/flux.weave.works/v1beta1/types.go
@@ -105,6 +105,11 @@ type HelmReleaseStatus struct {
 	// managed by this resource.
 	ReleaseStatus string `json:"releaseStatus"`
 
+	// Revision would define what Git hash or Chart version has currently
+	// been deployed.
+	// +optional
+	Revision string `json:"revision,omitempty"`
+
 	// Conditions contains observations of the resource's state, e.g.,
 	// has the chart which it refers to been fetched.
 	// +optional
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -118,20 +118,22 @@ type ChartChangeSync struct {
 	clonesMu sync.Mutex
 	clones   map[string]clone
 
-	namespace string
+	namespace     string
+	statusUpdater *status.Updater
 }
 
-func New(logger log.Logger, polling Polling, clients Clients, release *release.Release, config Config, namespace string) *ChartChangeSync {
+func New(logger log.Logger, polling Polling, clients Clients, release *release.Release, config Config, namespace string, statusUpdater *status.Updater) *ChartChangeSync {
 	return &ChartChangeSync{
-		logger:     logger,
-		Polling:    polling,
-		kubeClient: clients.KubeClient,
-		ifClient:   clients.IfClient,
-		release:    release,
-		config:     config.WithDefaults(),
-		mirrors:    git.NewMirrors(),
-		clones:     make(map[string]clone),
-		namespace:  namespace,
+		logger:        logger,
+		Polling:       polling,
+		kubeClient:    clients.KubeClient,
+		ifClient:      clients.IfClient,
+		release:       release,
+		config:        config.WithDefaults(),
+		mirrors:       git.NewMirrors(),
+		clones:        make(map[string]clone),
+		namespace:     namespace,
+		statusUpdater: statusUpdater,
 	}
 }
 
@@ -335,6 +337,11 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
 			}
 		}
 
+		err := chs.statusUpdater.UpdateReleaseRevision(fhr, chartClone.head)
+		if err != nil {
+			chs.logger.Log("warning", "could not update the release revision", "namespace", fhr.Namespace, "resource", fhr.Name, "err", err)
+			return
+		}
 	} else if fhr.Spec.ChartSource.RepoChartSource != nil { // TODO(michael): make this dispatch more natural, or factor it out
 		chartSource := fhr.Spec.ChartSource.RepoChartSource
 		path, err := ensureChartFetched(chs.config.ChartCache, chartSource)
@@ -345,6 +352,12 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
 		}
 		chs.setCondition(&fhr, fluxv1beta1.HelmReleaseChartFetched, v1.ConditionTrue, ReasonDownloaded, "chart fetched: "+filepath.Base(path))
 		chartPath = path
+
+		err = chs.statusUpdater.UpdateReleaseRevision(fhr, chartSource.Version)
+		if err != nil {
+			chs.logger.Log("warning", "could not update the release revision", "namespace", fhr.Namespace, "resource", fhr.Name, "err", err)
+			return
+		}
 	}
 
 	if rel == nil {
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -118,22 +118,20 @@ type ChartChangeSync struct {
 	clonesMu sync.Mutex
 	clones   map[string]clone
 
-	namespace     string
-	statusUpdater *status.Updater
+	namespace string
 }
 
 func New(logger log.Logger, polling Polling, clients Clients, release *release.Release, config Config, namespace string, statusUpdater *status.Updater) *ChartChangeSync {
 	return &ChartChangeSync{
-		logger:        logger,
-		Polling:       polling,
-		kubeClient:    clients.KubeClient,
-		ifClient:      clients.IfClient,
-		release:       release,
-		config:        config.WithDefaults(),
-		mirrors:       git.NewMirrors(),
-		clones:        make(map[string]clone),
-		namespace:     namespace,
-		statusUpdater: statusUpdater,
+		logger:     logger,
+		Polling:    polling,
+		kubeClient: clients.KubeClient,
+		ifClient:   clients.IfClient,
+		release:    release,
+		config:     config.WithDefaults(),
+		mirrors:    git.NewMirrors(),
+		clones:     make(map[string]clone),
+		namespace:  namespace,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


squaremo:
I am on board with this idea. Could you explain a bit about what the revision in status represents? Intuitively I presumed it would mean "last revision released", but it's set to empty on some occasions for which that hasn't changed, so it must mean something else.

squaremo:
First time contributor high five :wave:

paroxp:
> Intuitively I presumed it would mean "last revision released", but it's set to empty on some occasions for which that hasn't changed, so it must mean something else.

I think you are right. It should mean, "this is the version currently deployed".

I too, don't like the empty string being a possibility, but sadly this has to be the case if the code didn't manage to get the revision SHA yet.

In fairness, if the code fails fails before the SHA is obtained, you as a user have different issues to worry about 🤔 

The solution to that, would be not passing the string around like I have approached it, but by separating the logic that obtains the SHA in the first place to its own function, and then run it at the very end in `setCondition`  function as well as where it runs at the moment.

I thought the current approach is less logic heavy for your reviewal.

paroxp:
Sorry for the spam ☝️ 

FYI. We're running this version in our sandbox and staging environments and it seems to be working as we expect.

integrations/helm/chartsync/chartsync.go
```diff
@@ -48,7 +48,7 @@ import (
 	google_protobuf "github.com/golang/protobuf/ptypes/any"
 	"github.com/google/go-cmp/cmp"
 	"github.com/ncabatoff/go-seq/seq"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
squaremo: This is an unnecessary alias -- consider reconfiguring your editor (is it VC?)

integrations/helm/chartsync/chartsync.go
```diff
@@ -339,22 +339,22 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
 		chartSource := fhr.Spec.ChartSource.RepoChartSource
 		path, err := ensureChartFetched(chs.config.ChartCache, chartSource)
 		if err != nil {
-			chs.setCondition(&fhr, fluxv1beta1.HelmReleaseChartFetched, v1.ConditionFalse, ReasonDownloadFailed, "chart download failed: "+err.Error())
+			chs.setCondition(&fhr, fluxv1beta1.HelmReleaseChartFetched, v1.ConditionFalse, ReasonDownloadFailed, "chart download failed: "+err.Error(), refHead)
```
squaremo: It doesn't make much sense to put the refHead here (or on L346), since the condition pertains to downloading a chart package from a helm repo

integrations/helm/chartsync/chartsync.go
```diff
@@ -385,7 +385,7 @@ func (chs *ChartChangeSync) reapplyReleaseDefs() error {
 	}
 
 	for _, fhr := range resources {
-		chs.reconcileReleaseDef(fhr)
+		chs.reconcileReleaseDef(fhr, "")
```
squaremo: This is going to (potentially) reset the revision, if it re-releases a chart. I don't think that's what you want to do, since it will arbitrarily remove information that's still relevant.

squaremo:
> The solution to that, would be not passing the string around like I have approached it, but by separating the logic that obtains the SHA in the first place to its own function, and then run it at the very end in `setCondition` function as well as where it runs at the moment.

The head revision is not always relevant -- e.g., when a chart comes from a helm repo, there's no git involved. I think this needs to be more accurate in when it supplies the head revision, and when it sets it to `""`. I'll make comments in a review.

integrations/helm/chartsync/chartsync.go
paroxp: I agree it's useless. Both VIM and VCS are doing that... Thanks to you I had to resolve to using `nano` 😂 

Gone.

integrations/helm/chartsync/chartsync.go
paroxp: I agree, I didn't like this approach... That was a quick POC for myself. I have taken a different simpler approach (force pushed).

integrations/helm/chartsync/chartsync.go
paroxp: I feared that would be the case. Changed the approach.

paroxp:
I agree. I've learned a lot about flux and what it does, while digging through that codebase :O 

I have decided to take a different approach and in fact set yet another condition when using git specifically.

I have force pushed my changes. It's way simpler now, makes me happier.

squaremo:
I liked the idea of including the head revision in the conditions, where relevant. Would it still suit your purposes if it were just part of the `message` field of conditions?

paroxp:
Yeah I'd say so. Our need is to have it accessible.

With the latest change, I'm simply creating a new condition which makes it easier to iterate over.

With the latest change, we'd probably run something like that:

```go
for _, condition := range entry.Status.Conditions {
    if condition.Type == chartsync.GitRevision {
        gitCondition = condition
        continue
    }

    if condition.Type == flux.HelmReleaseReleased {
        helmCondition = condition
    }
}

report(releaseEntry{
    Release:  entry.Status.ReleaseName,
    Revision: gitCondition.Message,
    Running:  helmCondition.Status == "True",
    Status:   entry.Status.ReleaseStatus,
})
```

squaremo:
Relying on the message field, which is intended to be human-readable, is not ideal. If there was a status field (rather than a condition) for the released chart revision, you could check both the conditions (to make sure it's transitioned to Released), and the ChartRevision to get the data you need. Does that make sense?

paroxp:
I know what you mean.

Could you tell me, what's the problem with having *yet another* new condition set?

Previously, there were two: `ChartFetched`, `Released` and if I recall correctly when I've added the extra `Revision` field, it wasn't always being updated (perhaps due to the empty string every now and again).

I guess it was caused by a race condition. We are after all running go routine in a loop.

squaremo:
Conditions are on/off/unknown states; the chart was fetched, the chart was not successfully released. You can't rely on any fields in a condition other than the type and the status. The git revision (or chart revision) isn't an on/off/unknown state, so it's not suitable for representation as a condition.

I think it is entirely reasonable to have the last released chart revision as a part of the status -- it will not always be non-empty, but in the expected course of things, it should be mostly non-empty. It's similar to the `replicas` field of a deployment status, say.

paroxp:
Made some changes again.

Now it's part of the `status` and not a `condition`.
Should only "patch" when dealing with the git repository.

integrations/apis/flux.weave.works/v1beta1/types.go
```diff
@@ -105,6 +105,10 @@ type HelmReleaseStatus struct {
 	// managed by this resource.
 	ReleaseStatus string `json:"releaseStatus"`
 
+	// Revision would define what Git hash has currently been deployed.
```
squaremo: This could also contain the chart version, if the chart is from Helm, couldn't it?

integrations/helm/chartsync/chartsync.go
```diff
@@ -202,6 +204,11 @@ func (chs *ChartChangeSync) Run(stopCh <-chan struct{}, errc chan error, wg *syn
 						chs.logger.Log("warning", "could not get revision for ref while checking for changes", "repo", repoURL, "ref", ref, "err", err)
 						continue
 					}
+					err = chs.statusUpdater.UpdateReleaseRevision(fhr, refHead)
```
squaremo: What's the rationale for updating the status here? We haven't even looked at what revision might be released, at this point. If the intention of the ReleaseRevision is to keep track of the most recently released chart revision (/version), then `reconcileReleaseDef` (later in the file, around L350) is probably a better place to do it, since

 - that's where the release is attempted
 - all the chart releasing is funneled through there -- i.e., any change to which version of the chart is released will go through there.

(But I may be wrong about the intention!)

integrations/helm/status/status.go
```diff
@@ -119,3 +119,21 @@ func UpdateReleaseStatus(client v1beta1client.HelmReleaseInterface, fhr v1beta1.
 	}
 	return err
 }
+
+func (a *Updater) UpdateReleaseRevision(fhr v1beta1.HelmRelease, revision string) error {
```
squaremo: It might be easier to do this as a function rather than a method -- that would mean you don't have to pass an extra argument to the chartsync.

squaremo:
Sorry to be leading you along a winding path! I think this is pretty close to what we want. I'm happy to make some finishing touches, if you have other things to do :-)

integrations/helm/status/status.go
paroxp: I'm not sure if it'd be easier.
If it's not a method, we will need to recreate the Updater client at inconvenient place.

integrations/helm/chartsync/chartsync.go
paroxp: You're right :)

integrations/apis/flux.weave.works/v1beta1/types.go
paroxp: Sure! Does `Revision` make sense for chart version? As in, could one field be used for both? Something tells me no...

paroxp:
No worries. Glad to help. :) But don't feel obligated waiting for my response. If you feel like you can take it to the finish line in 2 minutes, go for it :D

hiddeco:
CircleCI is having some issues with running `dep ensure` (probably due to the 'partially degraded service' GitHub is reporting). I restarted the build (once again) hoping on good luck -- the PR itself looks clean to me.

Thanks @paroxp for your work, always appreciated :1st_place_medal:

PR #1749, opened 2019-02-19
Take ignore policy into account when working with automated resources

### LINKED ISSUE #1748: `automated` and `ignore` annotations clash
Having a resource annotated as both automated and ignored appears to be treated as both and seems to break automation of all cluster resources.

In our case, as long as we had some HelmRelease manifests annotated with both, none of the automated resources in our cluster were getting their images updated. Note that we were seeing the latest images available (via `fluxctl list-images`), they were just never used.

Upon removing one of the two, everything started flowing as normal.

One of our HelmReleases looked like this:

```yaml
---
kind: HelmRelease
metadata:
  name: my-webapi
  namespace: default
  labels:
    chart: webapi
  annotations:
    flux.weave.works/tag.chart-image: glob:release-*
    flux.weave.works/ignore: true
...
```

With both of these annotations in place, we'd have log entries that seem to indicate it being both ignored and automated:

```shell
ts=2019-02-18T22:02:34.277718708Z caller=sync.go:66 component=sync-loop resource=default:helmrelease/my-webapi ignore=apply
ts=2019-02-18T22:07:27.170619954Z caller=images.go:33 component=sync-loop error="checking services for new images: helmreleases.flux.weave.works \"my-webapi\" not found"
```

After doing some digging, I think this occurs due to an issue in [pollForNewImages](https://github.com/weaveworks/flux/blob/bd2d3e04a2865b5451fb8ad6cc3671c462705208/daemon/images.go#L33).

It looks like [GetUnlockedAutomatedResources](https://github.com/weaveworks/flux/blob/bd2d3e04a2865b5451fb8ad6cc3671c462705208/daemon/images.go#L109) doesn't take the `policy.Ignore` into account, so the image check fails when comparing with a non-existent flux controller on line 33. Since that error then causes a return, it short circuits image updates for everything else.

hiddeco:
Fixes #1748

Before this change having a HelmRelease annotated with both 'automated' and 'ignore' put automated releases to a halt for all resources known to Flux.

The reason for this was the image check failing while comparing with a non-existent controller, resulting in an early error return; which short circuited image updates for everything else.

### DIFF
--- a/daemon/images.go
+++ b/daemon/images.go
@@ -18,7 +18,7 @@ func (d *Daemon) pollForNewImages(logger log.Logger) {
 
 	ctx := context.Background()
 
-	candidateServices, err := d.getUnlockedAutomatedResources(ctx)
+	candidateServices, err := d.getAllowedAutomatedResources(ctx)
 	if err != nil {
 		logger.Log("error", errors.Wrap(err, "getting unlocked automated resources"))
 		return
@@ -95,9 +95,10 @@ func (r resources) IDs() (ids []flux.ResourceID) {
 	return ids
 }
 
-// getUnlockedAutomatedServices returns all the resources that are
-// both automated, and not locked.
-func (d *Daemon) getUnlockedAutomatedResources(ctx context.Context) (resources, error) {
+// getAllowedAutomatedResources returns all the resources that are
+// automated but do not have policies set to restrain them from
+// getting updated.
+func (d *Daemon) getAllowedAutomatedResources(ctx context.Context) (resources, error) {
 	resources, _, err := d.getResources(ctx)
 	if err != nil {
 		return nil, err
@@ -106,7 +107,7 @@ func (d *Daemon) getUnlockedAutomatedResources(ctx context.Context) (resources, 
 	result := map[flux.ResourceID]resource.Resource{}
 	for _, resource := range resources {
 		policies := resource.Policy()
-		if policies.Has(policy.Automated) && !policies.Has(policy.Locked) {
+		if policies.Has(policy.Automated) && !policies.Has(policy.Locked) && !policies.Has(policy.Ignore) {
 			result[resource.ResourceID()] = resource
 		}
 	}
--- a/update/filter.go
+++ b/update/filter.go
@@ -8,6 +8,7 @@ import (
 
 const (
 	Locked               = "locked"
+	Ignore               = "ignore"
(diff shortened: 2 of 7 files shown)


daemon/images.go
```diff
@@ -96,7 +96,7 @@ func (r resources) IDs() (ids []flux.ResourceID) {
 }
 
 // getUnlockedAutomatedServices returns all the resources that are
-// both automated, and not locked.
+// both automated, and not locked or ignored.
 func (d *Daemon) getUnlockedAutomatedResources(ctx context.Context) (resources, error) {
```
2opremio: I would rename this to something more consistent/descriptive.

Perhaps `getAllowedAutoResources`, `filterAutomatedResources`, `getEffectivelyAutomated` ...?

daemon/images.go
```diff
@@ -96,7 +96,7 @@ func (r resources) IDs() (ids []flux.ResourceID) {
 }
 
 // getUnlockedAutomatedServices returns all the resources that are
-// both automated, and not locked.
+// both automated, and not locked or ignored.
```
2opremio: Can you rephrase it so that the reader doesn't need to think about logic operator precedence?

squaremo:
Arguably, we should not bother updating resources when they have the ignore annotation _either_ in the file, or in the cluster resource, since the latter still gives surprising results -- the resource looks like it should have been updated, and visibly isn't. To do that, we'll also need to skip ignored resources when calculating the commit to make.

squaremo:
> Arguably, ...

Having put forward that argument, I think this PR unambiguously improves things as it is.

hiddeco:
> I think this PR unambiguously improves things as it is.

I like cherries on my cake (or _icing on the cake_, as you English people would say) so I added a filter.

update/filter.go
```diff
@@ -89,3 +90,16 @@ func (f *LockedFilter) Filter(u ControllerUpdate) ControllerResult {
 	}
 	return ControllerResult{}
 }
+
+type IgnoreFilter struct {
+}
+
+func (f *IgnoreFilter) Filter(u ControllerUpdate) ControllerResult {
```
squaremo: I reckon this should be filtering based on the `Controller` field -- the `Resource`* field has the definition from a file, while we want to filter out workloads that are ignored in the cluster.

*I know I know, shifting terminology. Ugh

PR #1752, opened 2019-02-20
Log skipped Charts directories

marccarre:
We [occassionally get cases](https://weave-community.slack.com/archives/C4U5ATZ9S/p1550650777411900) where it is difficult to understand why Helm integration doesn't work well, and in particular, which directories are taken into account by Flux or not. Extra logging might help there?

N.B.: I noticed loggers are only created in `main`s, and propagated down to where needed. However, this consequently introduces many small signature / field changes in various places. Please let me know if we'd rather do this differently and I'll happily update this PR.

### DIFF
--- a/cluster/kubernetes/manifests.go
+++ b/cluster/kubernetes/manifests.go
@@ -1,6 +1,7 @@
 package kubernetes
 
 import (
+	"github.com/go-kit/kit/log"
 	"github.com/weaveworks/flux"
 	kresource "github.com/weaveworks/flux/cluster/kubernetes/resource"
 	"github.com/weaveworks/flux/image"
@@ -8,10 +9,11 @@ import (
 )
 
 type Manifests struct {
+	Logger log.Logger
 }
 
 func (c *Manifests) LoadManifests(base string, paths []string) (map[string]resource.Resource, error) {
-	return kresource.Load(base, paths)
+	return kresource.Load(base, paths, c.Logger)
 }
 
 func (c *Manifests) ParseManifests(allDefs []byte) (map[string]resource.Resource, error) {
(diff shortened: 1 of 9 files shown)


squaremo:
> I noticed loggers are only created in `main`s, hence propagated these down to where needed.

For my part, this is because _how on earth are you_ supposed _to propagate loggers?_. So they variously get handed down through procedure calls, hidden in structs, and so on. Which all come with the problems you identify. To some extent this would be mitigated by putting `Log` calls behind an API (or APIs). But there would still be the problem of how to get a reference to a logger.

squaremo:
I am in two minds about whether it's worth all this trouble for `info` logs. What was the original confusion over?

marccarre:
[Here](https://weave-community.slack.com/archives/C4U5ATZ9S/p1550654261413700?thread_ts=1550652548.412900&cid=C4U5ATZ9S) was the original confusion. I didn't follow this issue until the very end, but it initially seemed like some manifests/pieces weren't discovered by Flux. When @hiddeco and I started to look into it, we realised Flux was never explicitly logging which directories were being ignored. I believe having extra logging could help people self-diagnose (or rule out) path issues. This may also help us figure out what's going on when people ask support questions in Slack.

What is your concern with such extra logging @squaremo?

squaremo:
> What is your concern with such extra logging @squaremo?

That it's extra complexity, and a lot of extra logging, since it'll run every time the repo is consulted, which will average out at several times a minute.

Looking in the Slack conversation, I don't see the bit where there was significant confusion over what was a chart and what wasn't; specifically, I don't think the user had put manifests they intended to be applied into a chart directory. It looks more like the problem turned out to be that fluxd hadn't got as far as applying the manifests at all.

I do see that it might have helped rule out a red herring cause, and I am sympathetic to the idea of being transparent about what a program is doing. Being able to get more information back from `fluxctl sync`, like which files were included, and which weren't, might be a better place for that transparency - -wdyt? (not asking you to implement that :-)

marccarre:
Sure -- closing this then.

PR #1753, opened 2019-02-20
Updates Docs (wording, typos, formatting)

dimitropoulos:
I did a thorough once-over on all of the documentation I could find and made a few minor changes I saw along the way.  I made the changes in atomic commits so that **(A)** the rationale for the individual change is always clear and **(B)** individual changes can easily be rejected.

### DIFF
--- a/site/fluxctl.md
+++ b/site/fluxctl.md
@@ -437,8 +437,8 @@ When building images it is often useful to tag build images by the branch that t
 quay.io/weaveworks/helloworld:master-9a16ff945b9e
 ```
 
-Indicates that the "helloworld" image was built against master
-commit "9a16ff945b9e".
+Indicates that the `helloworld` image was built against master
+commit `9a16ff945b9e`.
 
 When automation is turned on flux will, by default, use whatever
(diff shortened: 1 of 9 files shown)


2opremio:
This is awesome, thanks!

site/fluxctl.md
```diff
@@ -188,13 +176,24 @@ Using an SSH key allows you to maintain control of the repository. You
 can revoke permission for `flux` to access the repository at any time
 by removing the deploy key.
 
-```sh
+```
```
hiddeco: I think we may need to rethink about including this output in the docs (or automate it?) as it is likely to change over time.

dimitropoulos: I was thinking the same thing.  In the past I have used https://github.com/danger/danger which basically would allow us to write code that enforces that when the help text changes that the documentation will/should also change.  That's the best option I know of (and, there are ways in which it isn't great) that meets the minimum requirement of curbing some of that stuff getting out of date.

Another option would be to just have a quarterly health-and-wellness checkup list that would include updating documentation of this sort.  Just an idea.

PR #1757, opened 2019-02-22
Annotate resources outside of HelmRelease NS

### LINKED ISSUE #1293: helm-operator:  does not match namespace
In some charts (e.g. coreos/kube-prometheus) a few resource is created in the "kube-system" namespace, and not in the release namespace.
helm-operatot logs this error:

```
ts=2018-08-17T12:56:44.566927601Z caller=release.go:296 component=release output="secret \"alertmanager-monitoring\" annotated\n[...]servicemonitor \"monitoring-prometheus\" annotated\nthe namespace from the provided object \"kube-system\" does not match the namespace \"monitoring\". You must pass '--namespace=kube-system' to perform this operation.\nthe namespace from the provided object \"kube-system\" does not match the namespace \"monitoring\". You must pass '--namespace=kube-system' to perform this operation.\nthe namespace from the provided object \"kube-system\" does not match the namespace \"monitoring\". You must pass '--namespace=kube-system' to perform this operation.\n" err="exit status 1"
```

Despite of this error, the release seems working fine.

hiddeco:
Fixes #1293

### DIFF
--- a/integrations/helm/release/release.go
+++ b/integrations/helm/release/release.go
@@ -1,7 +1,6 @@
 package release
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"os"
@@ -11,6 +10,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/go-kit/kit/log"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/helm/pkg/chartutil"
 	k8shelm "k8s.io/helm/pkg/helm"
@@ -19,6 +19,7 @@ import (
 	"github.com/weaveworks/flux"
 	fluxk8s "github.com/weaveworks/flux/cluster/kubernetes"
 	flux_v1beta1 "github.com/weaveworks/flux/integrations/apis/flux.weave.works/v1beta1"
+	helmutil "k8s.io/helm/pkg/releaseutil"
 )
 
 type Action string
@@ -197,7 +198,7 @@ func (r *Release) Install(chartPath, releaseName string, fhr flux_v1beta1.HelmRe
 			return nil, err
 		}
 		if !opts.DryRun {
-			err = r.annotateResources(res.Release, fhr)
+			r.annotateResources(res.Release, fhr)
 		}
 		return res.Release, err
 	case UpgradeAction:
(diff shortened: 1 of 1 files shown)


integrations/helm/release/release.go
```diff
@@ -296,3 +298,49 @@ func mergeValues(dest, src chartutil.Values) chartutil.Values {
...
+	}
+	return objs
+}
+
+func namespacedResourceMap(objs []unstructured.Unstructured, defaultNamespace string) map[string][]string {
+	resources := make(map[string][]string)
+	for _, obj := range objs {
+		namespace := obj.GetNamespace()
+		if namespace == "" {
+			namespace = defaultNamespace
```
squaremo: The name `defaultNamespace` misled me into thinking this code was fallible, since it had to be given a hard-wired default namespace. But it's OK in fact, because it is the release namespace that is passed, which is indeed the namespace given to anything that needs a namespace and hasn't got one. Would calling the argument `releaseNamespace` make this clearer do you think, or am I just obsessed with default namespaces after #1442?

integrations/helm/release/release.go
```diff
@@ -296,3 +298,49 @@ func mergeValues(dest, src chartutil.Values) chartutil.Values {
...
+	var objs []unstructured.Unstructured
+	for _, manifest := range manifests {
+		json, err := yaml.YAMLToJSON([]byte(manifest))
+		if err != nil {
+			logger.Log("err", err)
+			continue
+		}
+
+		var u unstructured.Unstructured
+		if err := u.UnmarshalJSON(json); err != nil {
```
squaremo: This made me pause for thought -- the usual formulation is

```
var u unstructured.Unstructured
if err := json.Unmarshal(bytes, &u); err != nil {
 ...
}
```

.. though this code will work fine, so far as I know (it's just eliding the dispatch which json.Unmarshal would do after checking `&u` implemented `json.Unmarshaler`)

integrations/helm/release/release.go
hiddeco: It misled me too now that I hadn't seen this code in a couple of days, which is a sign. Will adjust it.

integrations/helm/release/release.go
hiddeco: The reason I choose for `u.UnmarshalJSON` and not the common `json.Unmarshal` is that it makes use of `UnstructuredJSONScheme`, which to my knowledge does some sanity checks; it will for example fail if there is no `kind` key set.

This may however be over engineered and I am happy to change it to a simpler approach.

squaremo: Additional checks are good :+1: `json.Unmarshal` will end up calling `UnmarshalJSON` for anything that implements it, so you will still get those.

PR #1760, opened 2019-02-25
Add Helm force upgrade option

stefanprodan:
When upgrading a release that's not backwards compatible (changes in volumes, selectors, etc) Tiller will mark the release as failed and helm-op users have to delete the HR with kubectl  or force the upgrade using Helm CLI. This PR adds `forceUpgrade` option to the HR CRD to avoid manually interventions on the cluster.

### DIFF
--- a/chart/flux/templates/helm-operator-crd.yaml
+++ b/chart/flux/templates/helm-operator-crd.yaml
@@ -40,6 +40,10 @@ spec:
               format: int64
             resetValues:
               type: boolean
+            skipDepUpdate:
+              type: boolean
+            forceUpgrade:
+              type: boolean
             valueFileSecrets:
               type: array
               properties:
--- a/deploy-helm/flux-helm-release-crd.yaml
+++ b/deploy-helm/flux-helm-release-crd.yaml
@@ -33,6 +33,8 @@ spec:
               type: boolean
             skipDepUpdate:
               type: boolean
+            forceUpgrade:
+              type: boolean
             valueFileSecrets:
               type: array
               properties:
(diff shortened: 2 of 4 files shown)


squaremo:
This seems like it would be a property of a _particular_ upgrade (e.g., from whatever chart, version 2.1 to 2.2), and not of the release in general. So the mode of use would be something like

 - find out (perhaps through it failing) that the upgrade won't work
 - put the extra field in the `HelmRelease` manifest, along with the new chart version
 - let it be rolled forward
 - remove the field from the HelmRelease manifest, so it doesn't keep using `--force`

I guess this would work, but it is pretty awkward. I admit I do not have a counter-proposal, mind you.

hiddeco:
> I guess this would work, but it is pretty awkward.

Users will only need it in particular scenarios, mostly due to faulty chart releases that do not maintain (correct) compatibility (one that comes to mind is the `stable/nginx-ingress` chart that had a [backwards and forwards incompatible breaking change](https://github.com/helm/charts/pull/10993)).

For this particular scenario it is the only tool that is missing right now to make the operation possible with the Helm operator. If it ends up to be too awkward for our user base I expect that we get feedback fast enough to iterate over the solution.

squaremo:
> If it ends up to be too awkward for our user base I expect that we get feedback fast enough to iterate over the solution.

OK. Bear in mind that adding things to the CRD is hard to back out of. Though, we already have a few additions that alter the flags given to helm (e.g., `resetValues`), so we've already taken that road.

hiddeco:
> OK. Bear in mind that adding things to the CRD is hard to back out of.

As the flag is a valid config flag for all `helm install|upgrade` operations I do not see it getting removed from the CRD (as users may permanently need it for whatever chart setup they have). I would rather expect us to add some mechanics in the future to make the operation you described above more seamless.

PR #1762, opened 2019-02-25
Helm Chart allow mounting of docker credentials file

shaneramey:
Fixes https://github.com/weaveworks/flux/issues/1310 as an alternative solution to https://github.com/weaveworks/flux/pull/1314

Enables fluxd to connect to private registries via the Helm Chart. This is done by mounting a custom docker config file from a Kubernetes Secret. The credentials are not placed in the Helm values or overrides, but instead must be side-loaded.

This is compatible with https://github.com/bzon/ecr-k8s-secret-creator per https://github.com/weaveworks/flux/pull/1314#issuecomment-418668316


stefanprodan:
Flux has native support for EKS/ECR authentication since 1.9.0 see https://github.com/weaveworks/flux/pull/1619

PR #1765, opened 2019-02-25
Improve testing in local environments other than linux-amd64

2opremio:
Unit tests rely on the `helm` binary and assume that the executing environment
is `linux-amd64`.

This makes tests fail in other environments like `darwin`, making development
unpleasant.

This change improves the situation by downloading and running test-specific
binaries from the local OS/architecture.

Apart from running the `helm` binary this change also downloads `kubectl` in
preparation for testing the parsing of `kubectl config current-context` in
https://github.com/weaveworks/flux/pull/1442 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ testdata/sidecar/sidecar-linux-amd64
 testdata/sidecar/.sidecar
 docker/fluxy-dumbconf.priv
 test/profiles
+test/bin/kubectl
+test/bin/helm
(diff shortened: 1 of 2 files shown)


hiddeco:
@2opremio having checked out to a different branch: `test/bin/` should probably be added to `.gitignore` with an exclusion for `test/bin/test-flux` to prevent committing binaries.

2opremio:
> @2opremio having checked out to a different branch: `test/bin/` should probably be added to `.gitignore` with an exclusion for `test/bin/test-flux` to prevent committing binaries.

Good one, will fix

PR #1768, opened 2019-02-26
Provide descriptive error when chart dir is absent while updating deps

### LINKED ISSUE #1764: Ambiguous log error when the chart path is wrong
There is a top level flat to skip dependency updates.  If you change the flag in a repo, it appears that the new value of the flag is ignored.

hiddeco:
Fixes #1764 

Explicitly check if the given chart directory does exist before trying to stat the `requirements.yaml` file and return a descriptive error if this is not the case. This provides users more guidance than the stat error we returned in the past.

```
ts=2019-02-26T17:58:36.52551703Z caller=chartsync.go:247 component=chartsync info="Start of releasesync"
ts=2019-02-26T17:58:36.566542344Z caller=chartsync.go:333 component=chartsync warning="Failed to update chart dependencies" namespace=default name=ghost error="chart path /tmp/flux-working475910449/charts/ghost does not exist"
ts=2019-02-26T17:58:36.566698882Z caller=chartsync.go:252 component=chartsync info="End of releasesync"
```

### DIFF
--- a/integrations/helm/chartsync/deps.go
+++ b/integrations/helm/chartsync/deps.go
@@ -1,6 +1,7 @@
 package chartsync
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -12,8 +13,13 @@ func updateDependencies(chartDir, helmhome string) error {
 	var hasLockFile bool
 
 	// sanity check: does the chart directory exist
+	if chartDir == "" {
+		return errors.New("empty path to chart supplied")
+	}
 	chartInfo, err := os.Stat(chartDir)
 	switch {
+	case os.IsNotExist(err):
+		return fmt.Errorf("chart path %s does not exist", chartDir)
 	case err != nil:
 		return err
 	case !chartInfo.IsDir():


stefanprodan:
Can we have a similar message when the dep update is disable?

hiddeco:
@stefanprodan this is already the case: https://github.com/weaveworks/flux/blob/193dff6ee957e12f48428edd586472793d366bc3/integrations/helm/release/release.go#L134

integrations/helm/chartsync/deps.go
```diff
@@ -14,6 +14,8 @@ func updateDependencies(chartDir, helmhome string) error {
 	// sanity check: does the chart directory exist
 	chartInfo, err := os.Stat(chartDir)
 	switch {
+	case os.IsNotExist(err):
```
squaremo: .. although, checking a possibly `nil` value here makes me nervous. We're relying on the implementation of `os.IsNotExist` being adequately defensive.

hiddeco: I could apply the same safety net for this like we do during installation.

https://github.com/weaveworks/flux/blob/193dff6ee957e12f48428edd586472793d366bc3/integrations/helm/release/release.go#L128-#L130

hiddeco: This should be more to your liking.

squaremo: I meant using `os.IsNotExist(err)` before checking whether `err != nil`

hiddeco: Misread your initial comment...

We also do this in the `release.go` linked above and the example for `os.IsNotExist` (https://golang.org/pkg/os/#IsNotExist) uses the same approach, the latter giving me the impression that this is fine.

PR #1771, opened 2019-02-27
Improve Docs (focusing on local development)

dimitropoulos:
This is a draft pull request for some miscellaneous blips and bloops found while getting started with local development.

### DIFF
--- a/site/get-started-developing.md
+++ b/site/get-started-developing.md
@@ -1,6 +1,6 @@
 # Get Started Developing Flux
 
-This tutorial is going to walk you through one possible workflow for making a small (actually, tiny) change to flux and building and testing that change locally.
+This tutorial is going to walk you through one possible workflow for making a small (actually, tiny) change to Flux and building and testing that change locally.
 
 > ### TL;DR
 >
@@ -17,7 +17,7 @@ This tutorial is going to walk you through one possible workflow for making a sm
 >    1. make a change to the code
 >    1. see your code changes have been deployed
 >    1. repeat
-> 1. Use `helm` and `skaffold` together to deploy changes to the flux helm chart.
+> 1. Use `helm` and `skaffold` together to deploy changes to the Flux helm chart.
 >    1. `make`
 >    1. make a change to the code
 >    1. see your code changes have been deployed
@@ -64,7 +64,7 @@ Now that we know everything is working with `flux-getting-started`, we're going 
     eval $(minikube docker-env)
     ```
 
-    Now, be warned. These are local variables.  This means that if you run this `eval` in one terminal and then switch to another for later when we build the flux project, you're gonna hit some issues.  For one, you'll know it isn't working because Kubernetes will tell you that it can't pull the image when you run `kubectl get pods`:
+    Now, be warned. These are local variables.  This means that if you run this `eval` in one terminal and then switch to another for later when we build the Flux project, you're gonna hit some issues.  For one, you'll know it isn't working because Kubernetes will tell you that it can't pull the image when you run `kubectl get pods`:
 
     ```
     NAME                        READY   STATUS         RESTARTS   AGE
@@ -77,7 +77,7 @@ Now that we know everything is working with `flux-getting-started`, we're going 
 
 1. Clone `git@github.com:<YOUR-GITHUB-USERNAME>/flux.git` replacing `<YOUR-GITHUB-USERNAME>` with your GitHub username.
 
-    In the same terminal you ran `eval $(minikube docker-env)`, run `dep ensure` followed by `make` from the root directory of the flux repo.  You'll see docker's usual output as it builds the image layers.  Once it's done, you should see something like this in the middle of the output:
+    In the same terminal you ran `eval $(minikube docker-env)`, run `dep ensure` followed by `make` from the root directory of the Flux repo.  You'll see docker's usual output as it builds the image layers.  Once it's done, you should see something like this in the middle of the output:
     ```
     Successfully built 606610e0f4ef
     Successfully tagged quay.io/weaveworks/flux:latest
@@ -118,7 +118,7 @@ Now that we know everything is working with `flux-getting-started`, we're going 
     secret/flux-git-deploy configured
     ```
 
-    Congrats you just deployed your local flux to your default namespace.  Check that everything is running:
+    Congrats you just deployed your local Flux to your default namespace.  Check that everything is running:
 
     ```sh
     kubectl get pods --selector=name=flux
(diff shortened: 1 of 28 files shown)


site/how-it-works.md
```diff
@@ -23,9 +23,9 @@ with an orchestrator (which is a common source of accidental failure) or
 with the systems that ensure that the orchestrator is in a working
 state.
 
-Flux also provides a CLI and a UI (in Weave Cloud) to perform these
-operations manually. Flux is flexible enough to fit into any development
-process.
+Flux provides a CLI (`fluxctl`) and a UI (as a compoent of Weave Cloud)
```
rndstr: `component`

dimitropoulos: :)

site/helm-get-started.md
```diff
@@ -140,7 +140,7 @@ deploying the workloads of `flux-get-started`. After a while you
 will be able to see the Helm releases listed like so:
 
 ```sh
-helm list --namespace demo
+helm list --namespace flux
```
stefanprodan: This should be `--namespace demo` since we want to show the charts deployed by Flux, not Flux itself.

site/helm-get-started.md
```diff
@@ -180,7 +180,7 @@ you can see when the change landed in the cluster.
 Confirm the change landed with:
 
 ```sh
-kubectl describe -n demo deployment/mongodb | grep Image
+kubectl describe -n flux deployment/mongodb | grep Image
```
stefanprodan: Same here, please undo this change.

site/standalone-setup.md
```diff
@@ -62,7 +62,19 @@ fluxctl identity
 Create a Kubernetes Secret from a private key:
 
 ```sh
-kubectl create secret generic flux-git-deploy --from-file=identity=/path/to/private_key
+kubectl create secret generic flux-git-deploy --from-file=identity=/full/path/to/private_key
+```
+
+this will result in a secret that has the structure:
```
squaremo: I'm not against this change, but I am curious what prompted it.

dimitropoulos: it was prompted by the other instruction `kubectl create secret generic flux-git-deploy --from-file /path/to/identity` in site/fluxctl.md not naming the argument (which then depends on the path being called `identity` so that the data key will be named `identity`.  I didn't catch this when I was following the instructions  and it caused  a lot of unnecessary frustration.  Eventually, while debugging I started inspecting the raw secret and the obvious question came forward: what data key is flux even looking for?  fast forwarding: I eventually figured it out because I saw other commands (like this one, actually) that were naming the argument to prevent confusion. I made sure to then go and make any other references to this action in the docs consistent with each other and that's where this one comes from.

site/helm-get-started.md
dimitropoulos: gotcha.    I misunderstood.  I'll add a line to make that more obvious.  After all, "the Helm releases" could be referring to the helm release of flux itself (which is how I read it)

site/standalone-setup.md
squaremo: I see, in `--from-file=/path/identity` it's not clear that the base name of the file _must_ be `identity` for it to work. Giving the example of the result (to some extent) reinforces that constraint.

dimitropoulos: exactly

dimitropoulos:
@stefanprodan thanks for the approval! it's a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) meaning I'm going to keep it open for a little bit (as a work-in-progress PR might) and add to it as I find things to add.  I suspect that will include some changes to the [building](https://github.com/weaveworks/flux/blob/master/site/building.md) file to say a little more about the development workflow.

rade:
> I'm going to keep it open for a little bit

Does the PR as it stands improve on the status quo? If so, merge it!

dimitropoulos:
@rade I catch your drift (and the answer is yes) but the thing I'm trying to avoid is to have to go through the `branch creation - pull request - review - merge` cycle for every individual commit (not that you're suggesting to go that granular, but in fairness, there's almost nothing of substance yet in this PR).  I expect to have this merged by the end of the week and I don't want to spend time creating pull requests when I could spend that same time coding.  if the trade-off of that is such that it will take a few extra days for these things to land on master then I think that's an acceptable trade-off.

dimitropoulos:
but more generally @stefanprodan @squaremo if you would like me to merge it, just let me know and I'm happy to push up another PR at the end of the week.  I'm actually surprised to find that it's possible to review (or approve) a draft PR before I've signaled that it's "Ready for review".  here is what it looks like on my end:

screenshot_20190227_162020

site/how-it-works.md
```diff
@@ -23,9 +23,9 @@ with an orchestrator (which is a common source of accidental failure) or
 with the systems that ensure that the orchestrator is in a working
 state.
 
-Flux also provides a CLI and a UI (in Weave Cloud) to perform these
-operations manually. Flux is flexible enough to fit into any development
-process.
+Flux provides a CLI (`fluxctl`) and a UI (as a component of Weave Cloud)
```
dholbach: Maybe link to our `fluxctl.md` doc?

dimitropoulos: great suggestion! done.

site/get-started-developing.md
```diff
@@ -0,0 +1,154 @@
...
+    - [Go](https://golang.org/doc/install)
+
+1. We want to make sure we're starting fresh.  Tell Minikube to clear any previously running clusters.
+    ```sh
+    minikube delete
+    ```
+
+1. The [Minikube addon](https://github.com/kubernetes/minikube/blob/master/docs/addons.md) called [freshpod](https://github.com/GoogleCloudPlatform/freshpod) that will be very useful to us later.  You'll see.  It's gonna be cool.
+    ```sh
+    minikube addons enable freshpod
```
stefanprodan: Can you please replace Freshpod with [Skaffold](https://github.com/GoogleContainerTools/skaffold), Freshpod is no longer maintained.

dimitropoulos: Yes.  @squaremo had suggested I used freshpod.  That, and it being super easy (with the one-liner) to integrate was the reason for using it.

site/get-started-developing.md
```diff
@@ -0,0 +1,154 @@
+# Get Started Developing Flux
+
+This tutorial is going to walk you through one possible workflow for making a small (actually, tiny) change to flux and building and testing that change locally.  Before we get started there are some baseline prerequisites you're gonna need:
+
+## Run `flux-getting-started`
+
+1. We're going to make some changes soon enough, but just to get a good baseline please follow the [Getting Started](site/get-started.md) guide and run the `flux-getting-started` repo through its normal paces.
```
dholbach: Minor: Probably doesn't need the `1. ` as it's just one step.

dimitropoulos:
> Maybe we can even agree on the tools we use for development? :-D

That is the current issue.  @stefanprodan has rightly pointed out that it's not super-duper responsible to put `freshpod` in here since it's on its way to the grave (with `skaffold` as the official successor).  The problem is that it's very much still supported by minikube and does what it does well.  All the same, I'm trying my hand at getting skaffold going in place of freshpod.

rade:
> freshpod vs skaffold

I wouldn't hold up this PR for that. Go with freshpod and change the recommendation to skaffold once that has been shown to work well.

dimitropoulos:
ok.  will do.

squaremo:
Thanks for sticking at this @dimitropoulos :bento:

PR #1776, opened 2019-02-27
Provide hook to instruct Helm operator to refresh git mirrors

### LINKED ISSUE #1430: Hook for telling helm-operator about git changes
For the flux daemon we have (in the RPC, and possibly in the HTTP API) a hook to tell it when there's new commits in git. The helm operator would also benefit from a hook, since it also monitors a git repository.

NB though, this most likely requires some architectural changes, or at least other components to be employed, since there still needs to be something to listen to a public IP and convey the webhook to the operator.

hiddeco:
Fixes #1430

This adds a webhook to the Helm operator as described in the issue above. The webhook starts a goroutine which instructs every mirror to refresh.

### DIFF
--- a/integrations/helm/http/daemon/server.go
+++ b/integrations/helm/http/daemon/server.go
@@ -73,7 +73,7 @@ type APIServer struct {
 // has been started.
 // TODO(hidde): in the future we may want to give users the option to
 // request the status after it has been started. The Flux (daemon) API
-// archives this by working with jobs whos IDs can be tracked.
+// achieves this by working with jobs whos IDs can be tracked.
 func (s APIServer) SyncGit(w http.ResponseWriter, r *http.Request) {
 	go s.server.SyncMirrors()
 
--- a/integrations/helm/http/daemon/server.go
+++ b/integrations/helm/http/daemon/server.go
@@ -9,6 +9,7 @@ import (
 	"github.com/weaveworks/flux/integrations/helm/api"
 	transport "github.com/weaveworks/flux/integrations/helm/http"
 	"net/http"
+	"sync/atomic"
 	"time"
 )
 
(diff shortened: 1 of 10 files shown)


stefanprodan:
I had the impression that `fluxctl sync` does a git sync and applies the manifest on the cluster. If that's the case I would expect that `sync` will also trigger helm upgrades for all releases after the git refresh.

hiddeco:
@stefanprodan this is already how it works, see:

https://github.com/weaveworks/flux/blob/dde522e0bf5a60f510aa4936757ab4d18a68064b/git/mirrors.go#L75
https://github.com/weaveworks/flux/blob/dde522e0bf5a60f510aa4936757ab4d18a68064b/integrations/helm/chartsync/chartsync.go#L159

squaremo:
> Some method of authentication

In general it's desirable to defer to existing means of authentication, i.e., credentials for the Kubernetes API. However that requires either more work on the part of the user (to construct a port forward or proxy) or set up a reverse proxy/ingress (with its own authentication); or a command-line tool for the helm operator, which we don't have.

I guess we could have `fluxctl helm notify` or some such, which would use the port-forwarding already in fluxctl.

hiddeco:
> I guess we could have `fluxctl helm notify` or some such, which would use the port-forwarding already in fluxctl.

I am somewhat opposed to this as it tightens the relationship between the 3 tools currently living in this repository.

`fluxctl` and the daemon already have a strong relationship but with our plans for the operator to actually move away from this repository I am afraid this would be a step into the wrong direction.

stefanprodan:
I think most users will be happy with:

```
kubectl -n flux port-forward deployment/helm-op 3030:3030 &
curl -XPOST http://localhost:3030/api/v1/sync-git
```

squaremo:
> `fluxctl` and the daemon already have a strong relationship but with our plans for the operator to actually move away from this repository I am afraid this would be a step into the wrong direction.

Yeah, it is a weak suggestion. Leaving it to people to construct a port-forward themselves is probably acceptable.

integrations/helm/http/daemon/server.go
```diff
@@ -46,3 +56,27 @@ func ListenAndServe(listenAddr string, logger log.Logger, stopCh <-chan struct{}
...
+type APIServer struct {
+	server api.Server
+}
+
+// SyncGit starts a goroutine in the background to sync all git mirrors
+// and writes back a HTTP 200 status header and 'OK' body to inform it
+// has been started.
+// TODO(hidde): in the future we may want to give users the option to
+// request the status after it has been started. The Flux (daemon) API
+// archives this by working with jobs whos IDs can be tracked.
```
squaremo: "achieves"

integrations/helm/http/daemon/server.go
```diff
@@ -46,3 +56,27 @@ func ListenAndServe(listenAddr string, logger log.Logger, stopCh <-chan struct{}
...
+func (s APIServer) SyncGit(w http.ResponseWriter, r *http.Request) {
+	go s.server.SyncMirrors()
```
squaremo: In general it's undesirable to start longish-running goroutines in response to user requests, without any backpressure or rate limiting (granted, there are precedents elsewhere in this code). A way to limit its effects here would be to succeed trivially if a sync is already in progress.

PR #1777, opened 2019-02-28
Mega-rename of service(s)/controller(s)/podController(s) to workload(s)

2opremio:
There was a mixture of the term _service_ and _controller_ throughout the whole codebase meaning the same thing. It was confusing for the newcomer and unpleasant to read.
    
I don't know if the term _service_ was supposed to be an abstraction agnostic to Kubernetes for _things which can be updated_ in a cluster. If so, it was leaking considerably, including `fluxctl` and data structure names.
    
I left "service(s)" strings/identifiers as is were unavoidable, to preserve backwards compatibility (e.g. query parameters, serialized data structure fields exposed to the http client etc ...).


squaremo:
Apparently even "controller" is wrong -- it refers to a program that interprets resources and e.g., starts pods. We eventually started naming the Deployments etc. "workloads", though not very visibly.

api/v6/api.go
```diff
@@ -68,7 +68,7 @@ type NotDeprecated interface {
 	Export(context.Context) ([]byte, error)
 
 	// v6
-	ListServices(ctx context.Context, namespace string) ([]ControllerStatus, error)
+	ListControllers(ctx context.Context, namespace string) ([]ControllerStatus, error)
```
squaremo: The API methods get called by RPC via reflection. Annoyingly.

2opremio: Where does that happen?

2opremio: I recall renaming the strings making the calls

2opremio:
> Apparently even "controller" is wrong -- it refers to a program that interprets resources and e.g., starts pods. We eventually started naming the Deployments etc. "workloads", though not very visibly.

I am happy to rename everything to workload if that works better (even in fluxctl, deprecating the `controller` argument)

api/v6/api.go
squaremo: That ... probably is good enough ... though the client part of the remote package runs in weave cloud, and we have to be sure that changing the name won't break compatibility.

squaremo:
> I am happy to rename everything to workload if that works better

A mass rename of Service and Controller to Workload would be _mostly_ correct, I think. Maybe we should start a glossary.

I'll go through all the actual changes and see whether each one makes sense.

2opremio:
Upon further thinking I am not sure I like _Workload_ either. It's a familiar term, though not very precise and it doesn't go well with the fact that we explicitly exclude pods which is counterintuitive (although ... who uses them in isolation anyways).

_Controller_ is more specific (although maybe too much since it's a k8s term, but I don't see it as a huge problem at this point). I guess the reason for moving towards _Workload_ is that we now support _HelmReleases_ which aren't controllers?

I don't have a better suggestion at this point but I think it's worth spending some time thinking about it. I don't want to go through another mega-rename anytime soon (particularly since I plan to update `fluxctl`'s flags accordingly).

@hiddeco @stefanprodan Thoughts?

hiddeco:
> _Controller_ is more specific (although maybe too much since it's a k8s term, but I don't see it as a huge problem at this point).

If I read the [docs](https://kubernetes.io/docs/concepts/workloads/controllers/) correctly the term _Controller_ is a _ReplicaSet_ in Kubernetes. All other type of objects are _Workloads_.

Given that with `fluxctl` we are able to list various types of objects (with an exclusion for Pods), I would say the term _Workload_ is correct.

stefanprodan:
I vote for workload, is what we use in Weave Cloud. Also Istio and other projects are using `workload` to define a group of pods owned by a deployment, statefulse or daemonset

squaremo:
> _Controller_ is more specific (although maybe too much since it's a k8s term, but I don't see it as a huge problem at this point).

I have had Kubernetes-embedded people express confusion over the use of "controller" to refer to Deployments, etc., rather than the programs that interpret them. The consensus terminology seems to be that "workload" means a definition of something for the cluster to run, and a "controller" is a particular kind of program.

squaremo:
.. though I should also note that within Kubernetes docs and code and so on, it's not especially consistent and has changed over time, too.

2opremio:
_Workload_ it is then

2opremio:
Thanks a lot for all the comments! Again, I'm really enjoying working with you guys!

2opremio:
> _Workload_ it is then

It took forever but it's done. I think I am going to take me some time to gain courage for doing this sort of thing again :)

I also removed `fluxctl`'s `--service` flags and `list-services` command (they were just printing an error anyways). Please let me know if it is too early and I will revert it.

daemon/daemon_test.go
```diff
@@ -35,14 +35,14 @@ import (
 
 const (
 	// These have to match the values in cluster/kubernetes/testfiles/data.go
-	svc               = "default:deployment/helloworld"
+	ctrl              = "default:deployment/helloworld"
```
2opremio: Ops, this should be `wl` now

daemon/daemon_test.go
```diff
@@ -35,14 +35,14 @@ import (
...
 	container         = "greeter"
 	ns                = "default"
 	oldHelloImage     = "quay.io/weaveworks/helloworld:3" // older in time but newer version!
 	newHelloImage     = "quay.io/weaveworks/helloworld:2"
 	currentHelloImage = "quay.io/weaveworks/helloworld:master-a000001"
 
-	anotherSvc       = "another:deployment/service"
+	anotherCtrl      = "another:deployment/service"
```
2opremio: And this should be `anotherWl`

2opremio:
I rebased and addressed the comments

remote/rpc/clientV11.go
```diff
@@ -29,16 +29,16 @@ func NewClientV11(conn io.ReadWriteCloser) *RPCClientV11 {
...
+	var resp ListWorkloadsResponse
+	for _, svc := range opts.Workloads {
+		if err := requireWorkloadIDKinds(svc, supportedKindsV8); err != nil {
 			return resp.Result, remote.UnsupportedResourceKind(err)
 		}
 	}
 
-	err := p.client.Call("RPCServer.ListServicesWithOptions", opts, &resp)
-	listServicesRolloutStatus(resp.Result)
+	err := p.client.Call("RPCServer.ListWorkloadsWithOptions", opts, &resp)
```
squaremo: This depends on the server (which is the daemon) having the changed method name. Before we update client code -- used in Weave Cloud, for instance -- newer daemons will not respond correctly. More seriously, _after_ we update client code, older daemons will not respond correctly. In other words, this breaks backward compatibility, in a way that would be quite difficult to even paper over. I think it's probably better to let the existing versioned APIs remain as they are.

update/automated.go
```diff
@@ -15,26 +15,26 @@ type Automated struct {
 }
 
 type Change struct {
-	ServiceID flux.ResourceID
-	Container resource.Container
-	ImageID   image.Ref
+	WorkloadID flux.ResourceID
```
squaremo: (NB to myself: this looks like it might be used in the API, but isn't)

update/release_image.go
```diff
@@ -45,18 +45,18 @@ func ParseReleaseKind(s string) (ReleaseKind, error) {
...
 
 // NB: these get sent from fluxctl, so we have to maintain the json format of
 // this. Eugh.
 type ReleaseImageSpec struct {
-	ServiceSpecs []ResourceSpec
-	ImageSpec    ImageSpec
-	Kind         ReleaseKind
-	Excludes     []flux.ResourceID
-	Force        bool
+	WorkloadSpecs []ResourceSpec `json:"serviceSpecs"`
```
squaremo: Changing the struct tags makes me nervous -- in my experience, unexpected serialisations can make things tricky to track down. Mind you, the reason this is a problem is that this struct is used in the API -- what we should really do is have a versioned type for the API, and convert to the "timeless" internal types here. But perhaps better to leave this one for now.

cluster/cluster.go
```diff
@@ -100,13 +100,13 @@ func (s Controller) ContainersOrError() ([]resource.Container, error) {
...
-	ErrServiceHasNoSelector = errors.New("service has no selector")
-	ErrNoMatching           = errors.New("no matching replication controllers or deployments")
-	ErrMultipleMatching     = errors.New("multiple matching replication controllers or deployments")
-	ErrNoMatchingImages     = errors.New("no matching images")
+	ErrEmptySelector         = errors.New("empty selector")
+	ErrWrongResourceKind     = errors.New("new definition does not match existing resource")
+	ErrNoMatchingWorkload    = errors.New("no matching workload")
+	ErrWorkloadHasNoSelector = errors.New("workload has no selector")
+	ErrNoMatching            = errors.New("no matching replication workloads or deployments")
+	ErrMultipleMatching      = errors.New("multiple matching replication workloads or deployments")
```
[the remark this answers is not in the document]
2opremio: They aren't I will just remove them

2opremio:
> I think changing the RPC methods will break compatibility in a way that's not intended, so can we leave those for now;

I think this will lead to reverting a big part of the renaming. Do you mean the method names or also the type names? (I don't know the details of `net/rpc` and how arguments are passed)

squaremo:
Owps, looks like we angered pflag:

```
$ fluxctl --help
panic: unlock flag redefined: controller

goroutine 1 [running]:
github.com/weaveworks/flux/vendor/github.com/spf13/pflag.(*FlagSet).AddFlag(0xc0003fa200, 0xc000214d20)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/vendor/github.com/spf13/pflag/flag.go:836 +0x870
github.com/weaveworks/flux/vendor/github.com/spf13/pflag.(*FlagSet).VarPF(0xc0003fa200, 0x1466c60, 0xc0002b87d0, 0x12b3d85, 0xa, 0x12af52f, 0x1, 0x12bd141, 0x14, 0xc000214be0)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/vendor/github.com/spf13/pflag/flag.go:819 +0x10b
github.com/weaveworks/flux/vendor/github.com/spf13/pflag.(*FlagSet).VarP(0xc0003fa200, 0x1466c60, 0xc0002b87d0, 0x12b3d85, 0xa, 0x12af52f, 0x1, 0x12bd141, 0x14)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/vendor/github.com/spf13/pflag/flag.go:825 +0x8e
github.com/weaveworks/flux/vendor/github.com/spf13/pflag.(*FlagSet).StringVarP(0xc0003fa200, 0xc0002b87d0, 0x12b3d85, 0xa, 0x12af52f, 0x1, 0x0, 0x0, 0x12bd141, 0x14)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/vendor/github.com/spf13/pflag/string.go:42 +0xaa
main.(*workloadUnlockOpts).Command(0xc0002b8780, 0xc0002b8780)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/cmd/fluxctl/unlock_cmd.go:39 +0x2d1
main.(*rootOpts).Command(0xc0000f1860, 0xc0000f1860)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/cmd/fluxctl/root_cmd.go:87 +0x6d8
main.run(0xc00000c090, 0x1, 0x1, 0xc0000b4058)
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/cmd/fluxctl/main.go:12 +0x4d
main.main()
        /home/mikeb/space/cloud/go/src/github.com/weaveworks/flux/cmd/fluxctl/main.go:29 +0x62
```

squaremo:
> looks like we angered pflag:

It's just a near-duplicate line in `cmd/fluxctl/unlock_cmd.go`

2opremio:
> Owps, looks like we angered pflag:

Sigh, sorry for this. I did test `fluxctl` at some point bot obviously not recently enough.

It's fixed now.

PR #1789, opened 2019-03-04
Wait for shutdown before returning from main()

### LINKED ISSUE #1788: Not re-querying DNS
Looking in to an issue similar to #1591, I find the goroutine which is supposed to re-query DNS is not running.

Which I suspect is because the [`defer memcachedClient.Stop()`](https://github.com/weaveworks/flux/blob/8c82f10a77e5d2adc69120b50172761203705e8c/cmd/fluxd/main.go#L391) gets run very soon after the beginning of the program.

bboreham:
Fixes #1788

Currently, `memcachedClient.Stop()` is getting called at the beginning of the run, on return from `main()`, which breaks things if the cache moves address.

~Personally I find this style too confusing and would just wait at the end of `main()`, but this looks like a trivial fix.~

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -222,15 +222,6 @@ func main() {
 		errc <- fmt.Errorf("%s", <-c)
 	}()
 
-	// This means we can return, and it will use the shutdown
-	// protocol.
-	defer func() {
-		// wait here until stopping.
-		logger.Log("exiting", <-errc)
-		close(shutdown)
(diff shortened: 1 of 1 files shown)


squaremo:
In the Helm operator main.go, I recanted this defer thing and just moved the blocking code to the end of the procedure. I'd be happy with that here as well.

bboreham:
I updated my PR to just wait at the end of `main()`.

PR #1791, opened 2019-03-04
Verify git commit signatures

### LINKED ISSUE #1683: git signature verification
Git supports signing commits/tags with PGP. If flux could be configured with a set of allowed PGP keys, it would add another layer of defense.

As it stands, if someone can manage to commit (eg. insider, previous employee with unrevoked access, unpatched git sever, stolen domain creds etc) they can compromise the entire cluster. You'll have an audit log and hopefully someone should notice which is way better than the status quo, but this could further improve the already great security story of flux.

hiddeco:
Fixes #1683 

This feature adds the `--git-verify-signatures` flag to the daemon.
When this flag is set the daemon will verify the signatures of the tag
and all commits it is working with, ensuring no unauthorized
modifications are synchronized.

To ensure the daemon always synchronizes a verified state a ratchet
mechanism was introduced to the loop. This mechanism moves the sync
HEAD to the latest valid revision after repository refreshes. During
sync runs this revision gets checked out and is applied on to the
cluster.

During modification actions, either automated or instructed by fluxctl
commands, the HEAD of the working clone is compared to the latest valid
revision. If these mismatch the commit is blocked and an error is
returned as the daemon can not be sure it is committing on top of a
verified state.


daemon/loop.go
```diff
@@ -199,6 +199,52 @@ func (d *Daemon) doSync(logger log.Logger, lastKnownSyncTagRev *string, warnedAb
 		return err
 	}
 
+	// Retrieve commits we are going to work with. We do this early as a signature validation
+	// error may occur, resulting in a change to the newTagRev.
+	var initialSync bool
+	var commits []git.Commit
+	{
```
2opremio: Sorry if this is uncalled for (the PR is still a draft) but I think it would be way cleaner if we factored this code out of `doSync()` which is already humongous in size.

hiddeco: Oh no, early reviews are more than welcome and I fully agree the size of `doSync()` is outrageous.

The reason I did (not yet) factored it out is that this part of Flux is relatively new to me and I would like to have the logic (and order of it) working first before deciding how we should shrink the size.

daemon/sync.go
```diff
@@ -0,0 +1,484 @@
...
+func (d *Daemon) NewSync(logger log.Logger) Sync {
+	s := Sync{logger: logger, Daemon: d}
+
+	return s
+}
+
+// Run starts the sync.
+func (s *Sync) Run(ctx context.Context) (retErr error) {
+	started := time.Now().UTC()
+	defer func() {
```
squaremo: I reckon do this around the method call, rather than in the method call, if possible.

daemon/sync.go
```diff
@@ -0,0 +1,484 @@
...
+// Sync holds the data we are working with during a sync.
+type Sync struct {
+	logger      log.Logger
+	working     *git.Checkout
+	oldTagRev   string
+	newTagRev   string
+	initialSync bool
+	commits     []git.Commit
+
+	*Daemon
```
squaremo: Including the Daemon here is convenient -- but brittle, on a couple of counts:

 - because it's typed as *Daemon, the code accessing it could depend on anything in that struct
 - because it's embedded, it is hard to see in the code which method or field access are dependencies

I think providing the *git.Working to the entry point (either as an argument to `Run` or when initialising a `struct Sync`) would obviate much of the need for `*Daemon`. Going through other implicit uses of *Daemon I can find:

 - `GitConfig`, or the relevant fields, can be supplied up front too
 - The syncSetName can be calculated up front and supplied
 - `lastKnownSyncTagRev`/`warnAboutSyncTagChange` -- either encode the required interactions into an interface, and supply it at the start, or take the values (as a struct) in an arg, and return the new values at the end.

No doubt I've missed some uses and underestimated the complexity. If these are just too difficult to disentangle, another way to advance would be to replace the `*Daemon` field with an interface containing only the dependencies.

daemon/sync.go
```diff
@@ -0,0 +1,484 @@
...
+			fluxmetrics.LabelSuccess, fmt.Sprint(retErr == nil),
+		).Observe(time.Since(started).Seconds())
+	}()
+
+	if err := s.checkout(ctx); err != nil {
+		return err
+	}
+	defer s.working.Clean()
+
+	if err := s.loadTagRevisions(ctx); err != nil {
```
squaremo: The dependencies between the different operations might be easier to follow if `{commits, oldTagRev, newTagRev}` are returned by this, and passed into the other methods (or functions, if they can be made stateless).

daemon/sync.go
hiddeco: Took most of your comments into account and succeeded (with some pain along the way) to turn everything into stateless functions.

I reckon the `Run()` method needs some in-code docs / annotations to help someone navigate through the function calls, and some log and error messages need a more descriptive message.

jimmyjones2:
Hey @hiddeco, thanks for your work on this. I presume the admin would configure the operator with their trusted GPG keys?

How do you plan on handling the case when trust is revoked, eg. when someone leaves or their key is compromised? However you'd still want to trust their commits up to a certain point. In normal operation I guess this wouldn't be an issue, as flux would have checked for valid signatures at the time and moved passed that commit and wouldn't ever check again retrospectively. However if you wanted to redeloy from scratch, it would try and verify all commits from the first and fail. Could there be an option to only verify commits from a certain point to handle this and the scenario where someone want to migrate to enforcing signatures without rewriting their history to be all signed?

hiddeco:
> I presume the admin would configure the operator with their trusted GPG keys?

Correct, any key imported to Flux (or configured by setting the `GNUPGHOME` environment variable) is seen as trusted.

> How do you plan on handling the case when trust is revoked, eg. when someone leaves or their key is compromised?

This is a good question, and I see two possible solutions

1. Is letting a human apply the insecure state and create a signed git tag, the daemon will then start synchronization from this point and all is well.

   I like this approach because it is secure and leaves no space for attack vectors. I dislike this approach because it requires manual interaction (although this process could be automated).
2. Is adding a flag like you suggested, to instruct the daemon to start verifying commits after a certain revision. This is super friendly to our users but also way easier to abuse.

If you put those two on a scale I am not sure what outweighs as the best option. Although the purist in me is telling me to go for option 1.

2opremio:
> refactor the _gigantic_ `doSync()` method in to smaller elements

❤️

jimmyjones2:
>1. Is letting a human apply the insecure state and create a signed git tag, the daemon will then start synchronization from this point and all is well.
>
>I like this approach because it is secure and leaves no space for attack vectors. I dislike this approach because it requires manual interaction (although this process could be automated).

I like this too. Considering people will already in the business of signing commits, doesn't seem that much more to ask to tag where to start from, especially as it'll be rare thing to need to do.

hiddeco:
Gave my head some rest from this after having seen 3 revisions of the solution and I think we are almost there.

The ratchet mechanism introduced detaches the sync itself from the verification part and gives us one point of truth for validated state we are working with, which is really nice (thanks @squaremo for your input on this).

Besides a rebase, there are two things I would like to tackle before this feature is complete:

* Some listing commands need to work with the verified `HEAD` instead of the `HEAD` from the repository.
* Documentation, and especially documentation about disaster recovery and working with revoked and/or compromised keys. The outline for this is point 1 in [my comment above](https://github.com/weaveworks/flux/pull/1791#issuecomment-473857904).

site/git-gpg.md
```diff
@@ -0,0 +1,235 @@
...
+                defaultMode: 0400
+          containers:
+          - name: flux
+            volumeMounts:
+            - name: gpg-keys
+              mountPath: /root/gpg-import
+              readOnly: true
+            args:
+            - --git-gpg-key-import=/root/gpg-import
+            - --git-signing-key=FBDA462F10B05292 # key id
```
squaremo: This is another ID again ...

site/git-gpg.md
```diff
@@ -0,0 +1,235 @@
...
+   or set the `gpgKeys.secretName` in your Helm `values.yaml` to
+   `gpg-keys`, and `signingKey` to your `<key id>`.
+
+> **Note:** Flux *does not* recursively scan a given directory but does
+understand symbolic links to files.
+
+> **Note:** Flux will automatically add any imported key to the GnuPG
+  trustdb. This is required as git will otherwise not trust signatures
+  made with the imported keys.
+
```
squaremo: It'd be good to explain how to check that the signatures are working. I used `git log --show-signature` after getting fluxd to make a commit to the repo.

site/git-gpg.md
```diff
@@ -0,0 +1,235 @@
...
+   with a signature, enable verification.
+
+b. Sign the sync tag by yourself, with a key that is imported, to point
+   towards the first commit with a signature (or the current `HEAD`).
+   Flux will then start synchronizing the changes between the sync tag
+   revision and `HEAD`.
+
+   ```sh
+   $ git tag --force --local-user=<key id> -a -m "Sync pointer" <tag name> <revision>
+   ```
```
squaremo: and push it?

hiddeco:
Tests are failing due to the Git version in the `circleci/golang:1.10` image being too old (`--format` was introduced in Git `2.12`) :disappointed:

squaremo:
> When I got to the end and wanted to kick start syncing with verified signatures. I signed the sync tag and pushed that, but the sync process still seems stuck at the first ever commit. I'm not clear on whether I did something wrong, or have the wrong expectation.

This time I started fluxd with `--verify-signatures` and it synced up to HEAD, which was signed and had a signed sync tag pointing at it (from last time).

Then,

 - I added a commit via github, which got a signature but an invalid one (since I've not added github's key to fluxd) -- fluxd refused to sync that commit, and stayed where it was
 - I moved the signed tag and force-pushed that -- fluxd synced to (the new, unsigned) HEAD

Is there a one-off error in evidence here? It seems like fluxd will check commits `tag < n <= HEAD`, which means if the sync tag _is_ HEAD, it'll sync to there whether it's signed or not. I may be misunderstanding.

hiddeco:
> Is there a one-off error in evidence here? It seems like fluxd will check commits `tag < n <= HEAD`, which means if the sync tag is HEAD, it'll sync to there whether it's signed or not. I may be misunderstanding.

It falls back to the revision of the signed tag in case there is no history (https://github.com/weaveworks/flux/pull/1791/files#diff-1a1f3e7ad9b1d7584e2d3e7d0c4c3db9R800), this is based on the assumption that the revision of the signed tag is a signed commit (as the daemon has seen it and thus must have been signed). This assumption may be wrong and/or causing unwanted behaviour in the edge-case you are describing.

squaremo:
> this is based on the assumption that the revision of the signed tag is a signed commit (as the daemon has seen it and thus must have been signed). This assumption may be wrong and/or causing unwanted behaviour in the edge-case you are describing.

I see -- so this is less a matter of correctness, and perhaps more about practicality. I think I would lean towards making sure the tag points at a signed commit, and balking otherwise. There would still be a way to jump-start a hitherto unsigned git repo, by making a signed commit then pointing the (signed) tag at it. wdyt?

hiddeco:
>  I think I would lean towards making sure the tag points at a signed commit, and balking otherwise. [...] wdyt?

Way better, see added commit and log entry below.

```
ts=2019-04-02T22:39:02.617389217Z caller=loop.go:96 component=sync-loop url=ssh://git@192.168.39.131:2222/home/git/repo err="failed to verify signature of sync tag revision: failed to verify commit ea3384d991f1258a7f09994ceece404139c29df3"
```

site/git-gpg.md
```diff
@@ -0,0 +1,247 @@
...
+
+In case a signature can not be verified, Flux will sync state up to the
+last valid revision it can find _before_ the unverified commit was
+made, and lock on this revision.
+
+## Importing trusted GPG keys and enabling verification
+
+1. Collect the public keys from all trusted git authors.
+
+2. Add the collected keys to your secret with GPG keys.
```
squaremo: It is not super obvious how to do this step. Would it make it easier if `--gpg-import-keys` was slice-valued, then one could create a new secret from a directory of public keys, and mount it in its own directory. This would be easier to explain how to do, and easier to do, I reckon.

hiddeco: And the public keys could live in a `ConfigMap` instead of a `Secret` which also makes more sense. Will make adjustments.

squaremo: I tried this after your latest commit, and it works very nicely

squaremo:
For the record: I've happy with how this works and is explained in the docs, but haven't been through the implementation with a fine-toothed comb yet

daemon/daemon.go
```diff
@@ -353,12 +353,21 @@ func (d *Daemon) sync() jobFunc {
...
 		}
-		head, err := d.Repo.Revision(ctx, d.GitConfig.Branch)
-		if err != nil {
+		var latestVerifiedRev string
+		if latestVerifiedRev, _, err = latestValidRevision(ctx, d.Repo, d.GitConfig); err != nil {
 			return result, err
-		}
-		result.Revision = head
-		return result, nil
+		} else if head, err := d.Repo.Revision(ctx, "heads/"+d.GitConfig.Branch); err != nil {
```
squaremo: This is done in a couple of places, I assume because those calls need to unambiguously check a branch head. Could it be put in a method `BranchHead`?

daemon/sync.go
```diff
@@ -0,0 +1,399 @@
...
+	repo      *git.Repo
+	gitConfig git.Config
+	manifests cluster.Manifests
+	cluster   cluster.Cluster
+	eventLogger
+}
+
+type SyncTag interface {
+	Revision() string
+	SetRevision(oldRev, NewRev string)
```
squaremo:
```suggestion
	SetRevision(oldRev, newRev string)
```

daemon/sync.go
```diff
@@ -0,0 +1,399 @@
...
+	if err := s.working.MoveSyncTagAndPush(ctx, tagAction); err != nil {
+		return err
+	}
+	cancel()
+	return nil
+}
+
+// refresh refreshes the repository, notifying the daemon we have a new
+// sync head.
+func refresh(ctx context.Context, s *Sync) error {
```
squaremo: Is this used? I couldn't find where

2opremio: What IDE are you using? GoLand grays out  orphan private symbols. I highly recommend it. (Sorry for the not so useful comment, I am AFK and can't check it right now).

squaremo: Emacs. Is GoLand also an NNTP client? I think not.

2opremio: /me drops his case

hiddeco: /offtopic

I have started trying out GoLand again (used GogLand in the past when it was still slow and unreliable) and it is much better than VSC.

hiddeco: It is used here: https://github.com/weaveworks/flux/pull/1791/files/4536f7840682032c19bbf07b5ba6403ef941fa80#diff-48e0783dedd59969ac55a5997f8c2ee9R125

hiddeco:
@squaremo addressed all your comments and verified (and fixed) my own theory in https://github.com/weaveworks/flux/pull/1791/commits/494113825f6ed43cf622c857f9165e45688552c3.

daemon/daemon.go
```diff
@@ -751,3 +768,77 @@ func policyEventTypes(u policy.Update) []string {
...
 }
+
+// latestValidRevision returns the latest valid revision for the
+// configured branch when the verification of GPG signatures for Git
+// is enabled _or_ the HEAD revision of the configured branch when it
+// is not. Signature validation happens for commits between the sync
+// tag revision and the HEAD, after the signature of the sync tag
+// itself has been validated, as the branch can not be trusted when
+// the tag originates from an unknown source.
+func latestValidRevision(ctx context.Context, repo *git.Repo, gitConfig git.Config) (string, git.Commit, error) {
```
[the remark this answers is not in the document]
squaremo: What's the `git.Commit` return value?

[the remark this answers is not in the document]
hiddeco: Sorry about that, quite sure I included something about that in the comment but apparently it got lost. See updated commit.

daemon/sync.go
```diff
@@ -0,0 +1,385 @@
...
+	"github.com/weaveworks/flux"
+	"github.com/weaveworks/flux/cluster"
+	"github.com/weaveworks/flux/event"
+	"github.com/weaveworks/flux/git"
+	"github.com/weaveworks/flux/resource"
+	fluxsync "github.com/weaveworks/flux/sync"
+	"github.com/weaveworks/flux/update"
+)
+
+type SyncTag interface {
```
squaremo: 1. I don't think this needs to be exported -- it's defining the interface consumed here, rather than an interface to be mentioned elsewhere
 2. From the type name and the methods, I'd expect this to handle the sync tag (i.e., make the tag and push it) -- but that's done independently. Does it make any sense to put the sync tag update in the implementation too?
 3. I'm not sure `.Revision()` is used (but I have been wrong about this kind of thing before)

hiddeco: All your comments made sense (and you were actually correct about the unused `.Revision()` this time :wink:).

I am not too happy about the amount of parameters `SetRevision` ended up with, but I see no other way at the moment, I may do a next gen refactor in a month or so.

squaremo: > I am not too happy about the amount of parameters `SetRevision` ended up with

I didn't think of that when making the suggestion -- I don't think it's too bad though. There will surely be some refactoring of this bit in the "read-only git" PR, so it'll get attention soon in any case.

PR #1792, opened 2019-03-05
If applied, this commit will allow flux to work with different git ssh ports

OskarFromSpain:
Hello,

these changes will help flux to get work with some private hosted gits, which are not using the default ssh port.
Otherwise flux will always try to access the git with port 22 and ignoring the added port to the url.


OskarFromSpain:
delete request

PR #1796, opened 2019-03-05
update developing docs to remind to `make test`

### LINKED ISSUE #1790: kubeyaml should not be a prerequisite for running the tests
currently when running tests, there are errors unless you have [`kubeyaml`](https://github.com/squaremo/kubeyaml) in your path:

```
$ go test github.com/weaveworks/flux/daemon
--- FAIL: TestDaemon_Release (0.21s)
        daemon_test.go:838: applying changes: exec: "kubeyaml": executable file not found in $PATH
--- FAIL: TestDaemon_PolicyUpdate (0.22s)
        daemon_test.go:838: exec: "kubeyaml": executable file not found in $PATH
--- FAIL: TestDaemon_SyncStatus (0.26s)
        daemon_test.go:838: applying changes: exec: "kubeyaml": executable file not found in $PATH
--- FAIL: TestDaemon_JobStatusWithNoCache (0.13s)
        daemon_test.go:838: exec: "kubeyaml": executable file not found in $PATH
--- FAIL: TestDaemon_Automated (10.17s)
        daemon_test.go:821: Waiting for image tag: "2"
--- FAIL: TestDaemon_Automated_semver (10.11s)
        daemon_test.go:821: Waiting for image tag: "3"
event="Sync: 9d04903, default:daemonset/init, default:deployment/helloworld, default:deployment/list-deploy, default:deployment/locked-service, default:deployment/multi-deploy, default:deployment/semver, default:deployment/test-service, default:service/list-service, default:service/multi-service" logupstream=true
event="Sync: cba0b77, default:deployment/helloworld" logupstream=true
FAIL
FAIL    github.com/weaveworks/flux/daemon       23.027s
```

It seems reasonable enough to me that this can/should be automated out.  Seems like we're close to that already anyway.  There's a `/bin` folder in the Flux repo that contains a dockerfile for running `kubeyaml`:
```dockerfile
#!/bin/sh
docker run --rm -i quay.io/squaremo/kubeyaml:0.5.1 "$@"
```

The code in question is in `kubeyaml.go`:

```
func execKubeyaml(in []byte, args []string) ([]byte, error) {
	cmd := exec.Command("kubeyaml", args...)
	...
	err := cmd.Run()
}
```

I don't have a solution off hand, but it seems reasonable that this is something you shouldn't need to install in your PATH before running tests.  I could happily be persuaded otherwise.  If this is intended, then I will update the documentation to list `kubeyaml` being in your path a requirement for the path.

---

For reference, here is what the `Dockerfile.flux` does:
```dockerfile
# Get the kubeyaml binary (files) and put them on the path
COPY --from=quay.io/squaremo/kubeyaml:0.5.1 /usr/lib/kubeyaml /usr/lib/kubeyaml/
ENV PATH=/bin:/usr/bin:/usr/local/bin:/usr/lib/kubeyaml
```

dimitropoulos:
closes #1790

### DIFF
--- a/site/get-started-developing.md
+++ b/site/get-started-developing.md
@@ -1,6 +1,6 @@
 # Get Started Developing Flux
 
-This tutorial is going to walk you through one possible workflow for making a small (actually, tiny) change to Flux and building and testing that change locally.
+This guide shows a workflow for making a small (actually, tiny) change to Flux, building and testing that change locally.
 
 > ### TL;DR
 >
@@ -38,6 +38,11 @@ Now that we know everything is working with `flux-getting-started`, we're going 
     - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
     - [Docker](https://docs.docker.com/install/)
     - [Go](https://golang.org/doc/install)
+
+1. Configure your enviroment so you can run tests.  Run:
+    ```sh
+    make test
+    ```
 
 1. We want to make sure we're starting fresh.  Tell Minikube to clear any previously running clusters.
     ```sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,7 @@ This is a rough outline of how to prepare a contribution:
 Refer to the [building doc](site/building.md) to find out how to build from
 source.
 
-Refer to the [up-and-running with Flux Development](site/get-started-developing.md) doc for a walkthrough on developing Flux locally.
+Refer to the [Get Started Developing](site/get-started-developing.md) guide for a walkthrough on developing Flux locally.
 
 ### How to run the test suite
 


site/get-started-developing.md
```diff
@@ -39,6 +39,11 @@ Now that we know everything is working with `flux-getting-started`, we're going
     - [Docker](https://docs.docker.com/install/)
     - [Go](https://golang.org/doc/install)
 
+1. Configure your enviroment so you can run tests.  Run:
+    ```sh
+    make test
+    ```
+
 1. We want to make sure we're starting fresh.  Tell Minikube to clear any previously running clusters.
```
hiddeco:
```suggestion
2. We want to make sure we're starting fresh.  Tell Minikube to clear any previously running clusters.
```

dimitropoulos: all of the bullets are `1.`.  this practice prevents any chance of misnumbering the individual bullets in the future when edits are made.

hiddeco: Oh, upon further inspection it looks like all list items start with `1`. Is there a reason behind this?

dimitropoulos: in other words: this is a markdown best-practice I am following.  Markdown ignores your numbering, actually, anyway:

1. this is numbered with a `1`
2. this is numbered with a `2`
4. this is numbered with a `4`
3. this is numbered with a `3`

check the raw text on the above to see what I mean ^

hiddeco: Ah, yeah I guess that makes sense as GitHub corrects them.

It would however surprise me as I am a fervent `cat *.md` user.

dimitropoulos: it's not GitHub, this is the spec-compliant behavior of markdown

dimitropoulos: well, the GitHub markdown spec.  I won't get into the various messes with markdown standards.  but, to my knowledge, this is how all markdown parsers work

hiddeco: Of course parsers will display it just fine. The whole idea behind Markdown however is that it also reads as well formatted text without parsing.

> Philosophy
> Markdown is intended to be as easy-to-read and easy-to-write as is feasible.
> Readability, however, is emphasized above all else. A Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions. 

I do however expect people to read this doc from within GitHub so I suppose ease of use > strict correctness applies.

dimitropoulos: I guess I don't see what's not well formatted.  That's the syntax, right..  It reads well to me.  Maybe that's just because I'm used to it.

Like, to use your quote:
> without looking like it’s been marked up with tags or formatting instructions
to me, specifying `1.`, `2.`, `3.` is a formatting instruction - especially once you consider that that formatting instruction is _ignored_ when the file is rendered.

PR #1798, opened 2019-03-06
Make garbage-collection labelling resource-specific

### LINKED ISSUE #1787: garbage collection deleting children resources
Hi friends!

I'm testing the garbage collection feature, and i'm noticing that some operators' children are being garbage collected! 

example:
Prometheus operator creates a statefulset, and the statefulset gets cleaned up when gc runs

we've identified that this is because the statefulset is mirroring the labels from the `Prometheus` object

both the `Prometheus` object and the `StatefulSet` get this
```metadata:
  annotations:
    flux.weave.works/sync-checksum: ca0b295832f409e52084f1db0f0185d097603642
```

We had talked in #flux about some of the ways we could handle this 

Proposed ideas:
1.) use git repo history for gc @hiddeco pointed out that this may be problematic
2.) use a combination of labels to identify that a hash belongs to a type, for example that hash above belongs to a `Prometheus` so fluxd won't garbage collect a `StatefulSet`

questions? 
what if a `Foo` creates a `Foo`? maybe the labels would have a hash of some kind to identify the resource it belongs to?

2opremio:
This prevents unintended delitions from happening if/when cluster object labels
are copied around.

I also changed some names and tweaked comments a bit.

Fixes #1787 

Disclaimer: I haven't tested it (yet)

### DIFF
--- a/cluster/kubernetes/sync.go
+++ b/cluster/kubernetes/sync.go
@@ -3,6 +3,8 @@ package kubernetes
 import (
 	"bytes"
 	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"fmt"
 	"io"
@@ -28,7 +30,12 @@ import (
 )
 
 const (
-	syncSetLabel       = kresource.PolicyPrefix + "sync-set"
+	// We use mark-and-sweep garbage collection to delete cluster objects.
+	// Marking is done by adding a label when creating and updating the objects.
+	// Sweeping is done by comparing Marked cluster objects with the manifests in Git.
+	gcMarkLabel = kresource.PolicyPrefix + "sync-gc-mark"
+	// We want to prevent garbage-collecting cluster objects which haven't been updated.
+	// We annotate objects with the checksum of their Git manifest to verify this.
 	checksumAnnotation = kresource.PolicyPrefix + "sync-checksum"
 )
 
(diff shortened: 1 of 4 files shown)


2opremio:
Sure, I will add one tomorrow

cluster/kubernetes/sync.go
```diff
@@ -275,6 +292,15 @@ func applyMetadata(res resource.Resource, set, checksum string) ([]byte, error)
 	return bytes, nil
 }
 
+func makeGCMark(syncSetName, resourceID string) string {
+	hasher := sha256.New()
+	hasher.Write([]byte(syncSetName))
+	// To prevent deleting objects with copied labels
+	// an object-specific mark is created (by including its identifier).
+	hasher.Write([]byte(resourceID))
+	return "sha256:" + base64.RawURLEncoding.EncodeToString(hasher.Sum(nil))
```
rndstr: the colon `:` is not allowed in the k8s version we have running:

```
ts=2019-03-07T20:07:59.874086773Z caller=sync.go:482 component=cluster method=Sync cmd="kubectl apply -f -" took=1.260555818s err="running kubectl: The Secret \"my-secret\" is invalid: metadata.labels: Invalid value: \"sha256:XmUQTWgJCpTt_vgSxwB113djqIlbulw_bNcdVL9zNmk\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')" output=
```

will create an issue with more information

dimitropoulos: @rndstr so were you debuggin this for the last hour too? https://github.com/weaveworks/flux/issues/1805

rndstr: seems like someone beat me to the issue :) https://github.com/weaveworks/flux/issues/1805

PR #1804, opened 2019-03-07
Replace ':' with '.' in GC mark

hiddeco:
A valid Kubernetes label may only contain alphanumeric characters, '-',
'_' or '.', and must start and end with an alphanumeric character (e.g.
'MyValue',  or 'my_value',  or '12345'.


hiddeco:
Replaced by #1806 as this branch name is going to be confusing if it shows up in the merge commit.

PR #1823, opened 2019-03-14
Move skipDepUpdate to GitChartSource part of CRD

### LINKED ISSUE #1782: Helm operator should either update deps for chart packages, _or_ skipDepUpdate should move
In https://github.com/weaveworks/flux/pull/1712, a field was introduced into the HelmRelease CRD to instruct the helm operator to skip dependency updates.

Since `helm package` won't complete unless the chart dependencies are present in `charts/`, the helm operator assumes[1] all charts from a Helm repo include their dependencies, and doesn't attempt to update them (anyway that would involve unpacking the chartball).

Either the assumption at [1] is wrong, and the helm operator should be unpacking and updating the dependencies for all charts; or, it's correct, in which case `skipDepUpdate` pertains only to charts from git and should be moved into that section in the CRD.

hiddeco:
Fix #1782

### DIFF
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -329,7 +329,7 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
 		chartPath = filepath.Join(chartClone.export.Dir(), chartSource.Path)
 		chartRevision = chartClone.head
 
-		if chs.config.UpdateDeps && !fhr.Spec.SkipDepUpdate {
+		if chs.config.UpdateDeps && !fhr.Spec.ChartSource.GitChartSource.SkipDepUpdate {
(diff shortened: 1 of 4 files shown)


integrations/helm/chartsync/chartsync.go
```diff
@@ -329,7 +329,7 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
 		chartPath = filepath.Join(chartClone.export.Dir(), chartSource.Path)
 		chartRevision = chartClone.head
 
-		if chs.config.UpdateDeps && !fhr.Spec.SkipDepUpdate {
+		if chs.config.UpdateDeps && !fhr.Spec.ChartSource.GitChartSource.SkipDepUpdate {
```
squaremo: this could use `chartSource`, assigned above, I think

PR #1826, opened 2019-03-14
Correct misconception about GC only deleting namespaced resources

2opremio:


2opremio:
I've already corrected the 1.11.0 release text https://github.com/weaveworks/flux/releases/tag/1.11.0

PR #1828, opened 2019-03-14
Bump Helm API package and binary to v2.13.0

squaremo:
There was a breaking change in 2.13 (see
https://github.com/weaveworks/flux/issues/1774 and the linky therein).

It appears that the Helm operator _as it is_ will work OK with the
newer Tiller (including TLS). But it's about time to make the leap to
a newer version, since 2.9.1 is just about a year old.

I have also verified that a helm operator built with this commit works
in a cluster with Tiller 2.9.1, i.e., with existing installations.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -33,12 +33,28 @@
   version = "v0.3.1"
 
 [[projects]]
+  digest = "1:debe400440582cfe5580591afd0f7fdef5b17a25945f06bd3087bc67b4b8fd03"
+  name = "github.com/Masterminds/goutils"
+  packages = ["."]
+  pruneopts = ""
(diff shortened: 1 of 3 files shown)


squaremo:
Ops, didn't notice this was ready to merge! I guess it's good that we got a release out before doing this, in any event.

hiddeco:
> I guess it's good that we got a release out before doing this, in any event.

I exactly thought you were waiting for your dopamine shot because of said release.

squaremo:
I don't _quite_ get why `dep` removed the stanza to do with k8s.io/kubernetes from Gopkg.lock.

davidkarlsen:
Will there be a soonish release with this included?

PR #1829, opened 2019-03-14
Desambiguate revisions and files when invoking `git`

2opremio:
User Git repositories may use the same name for:
* A file and a branch used by Flux
* A file and the tag used by Flux for syncing

For instance:
```
ts=2019-03-13T19:36:40.432679542Z caller=loop.go:90 component=sync-loop err="fatal: ambiguous argument 'staging': both revision and filename"
```

(See https://weave-community.slack.com/archives/C4U5ATZ9S/p1552505808477700?thread_ts=1552504490.476100&cid=C4U5ATZ9S )

This change removes that ambiguity by appending `--` at the right places when invoking `git`.

In a few cases we were only appending it when adding files at the end, which wasn't correct.


alanjcastonguay:
Good catch, @dcondomitti. A lot of people probably got lucky with their choice of filenames vs refs here.

dcondomitti:
Thanks for the quick fix! :+1:

PR #1831, opened 2019-03-14
Require creation metadata during auto release

### LINKED ISSUE #1817: Don't auto release when we don't have metadata for the current image tag
In the following circumstances:

* A registry experiences a fault where it fails to enumerate all tags when asked without signalling an error condition
* The tag for an image in an automated workload is _not_ returned by the registry API
* Older image tags _are_ returned

Flux will roll back the image to an older version, because it can't tell the current image is newer. Instead, Flux should refuse to automate a release unless it has timestamp information for the current tag as well as the candidate target.

This is not a hypothetical situation, it has happened to us recently.

hiddeco:
Fixes #1817 

In the following circumstances:

* A registry experiences a fault where it fails to enumerate all
  tags when asked without signalling an error condition
* The tag for an image in an automated workload is not returned
  by the registry API
* Older image tags are returned

Flux would roll back the image to an older version, because it
couldn't tell the current image is newer. Instead, Flux now refuses
to automate unless it has timestamp information for the current and
new version, as this assures the made comparison is valid.

### DIFF
--- a/daemon/daemon_test.go
+++ b/daemon/daemon_test.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


squaremo:
Nice -- less code, and better safeguard. Can we have a test case, please?

daemon/daemon_test.go
```diff
@@ -620,6 +621,62 @@ func TestDaemon_Automated_semver(t *testing.T) {
 	w.ForImageTag(t, d, resid.String(), container, "3")
 }
 
+func TestDaemon_Automated_no_metadata(t *testing.T) {
```
squaremo: A bit tricky verifying that something _doesn't_ happen, isn't it :-/ Is it possible to verify that the update is not processed, by faking `UpdateManifests`, or otherwise getting the result of the `pollForNewImages` calculation? (It occurs to me that factoring the calculation out might be a useful tidy anyway)

hiddeco: Not sure if faking `UpdateManifests` and testing if there are no calls made to it makes this test more reliable as you still can't verify why the call hasn't been made (although while I am typing this it occurred to me that combined with the log output it should give you a pretty good idea).

I will look into factoring out the calculation and see if this eases the situation.

hiddeco: I factored out the image change calculation. PTAL.

daemon/images.go
```diff
@@ -113,3 +75,37 @@ func (d *Daemon) getAllowedAutomatedResources(ctx context.Context) (resources, e
...
 }
+
+func calculateChanges(logger log.Logger, workloads []cluster.Workload, imageRepos update.ImageRepos) *update.Automated {
+	changes := &update.Automated{}
+
+	for _, workload := range workloads {
+	containers:
+		for _, container := range workload.ContainersOrNil() {
+			currentImageID := container.Image
+			pattern := policy.GetTagPattern(workload.Policies, container.Name)
```
hiddeco: Before I factored this out the policies were collected from the `candidateWorkloads`, which would have required me to include the `candidateWorkloads` as a parameter to this function and did not look right to me as it collected information from two sources.

Removing it broke the semver test but that was due to how the daemon mock is constructed, fixed that by adding the `policy.Set{}` to the workload mock.

If there was a specific reason we collected the policies from the candidates and not the workload itself, please tell and I will revert the change.

squaremo: Let's see, so the previous formulation was

```
1. find all the manifests in the repo that might qualify for an update
2. find the resources in the cluster that correspond to those
3. calculate which of the latter can be updated, looking at the policies from the manifests
```

the new formulation is

```
1. find all the manifests in the repo that might qualify for an update
2. find the resources in the cluster that correspond to those
3. calculate which of the latter can be updated, looking at the policies they have
```

The difference is that the resources in the cluster might not be up to date with respect to the policies -- if they have not been synced yet, for instance.

It's difficult to argue that the original formulation is entirely correct. On balance, though, it's better to preserve the behaviour when it's not clear a change improves on it.

squaremo: Looking again, the existing behaviour is pretty weird. It looks at the image from the cluster, but ultimately it will be making changes to the manifests. Empirically, this mostly works fine -- the problem is more that it's hard to reason about, therefore hard to change :thinking:

Anyway I don't think we need to address that right now; the simplest change would just be to include the manifests (`candidateWorkloads`) in the calculation.

hiddeco: > On balance, though, it's better to preserve the behaviour when it's not clear a change improves on it.

I am not sure if I agree with you that the change does not improve behaviour.

We now rely on one source for decision making, which is _applied state_ and not _desired state_. As we are in the process of possibly making modifications these should, in my opinion, be made based upon what we are looking at now and not what we may be looking at in the future.

squaremo: > We now rely on one source for decision making

`workloads` is obtained by asking the cluster for all the workloads that are marked as automated _in the git repo_. So as it stands, this code decides what qualifies for an update (the automation annotation) based on the file in the repo, and decides how to update them (the tag filter annotation) based on the cluster.

hiddeco: Gotcha, would it make sense to change this to collect automated workloads from the cluster instead (in a new PR) to ensure we are only making modifications to things the user is actually running?

squaremo: If I had to chose a direction to take things, I would say treat the git repo as canonical and refer to the cluster to determine if operations are (likely to be) safe. Does that make sense from your point of view?

PR #1836, opened 2019-03-18
Support loading Helm values from different source types

### LINKED ISSUE #1809: Helm Chart: Allow use of alternative values file from remote chart
Some helm charts include multiple value file options (example: https://github.com/istio/istio/tree/master/install/kubernetes/helm/istio). It would be valuable to be able to select a specific values file to use rather than the default. If a user were to download the preferred file and add it to a secret, it could become stale.

hiddeco:
Fixes #1809 

`valuesFrom` supports different types of sources: `configMapKeyRef`,
`secretKeyRef` and `externalSourceRef`. Where the latter can be used to
retrieve a values file from an URL by setting the
`externalSourceRef.URL`. The array of sources is merged in specified
order, with the `.spec.values` getting merged last.

For backwards compatiblity the old `valuesFromSecrets` are transformed
into `secretKeyRef`s and lined up in front of any specified
`valuesFrom` sources.

On all source types the `<sourceRef>.Optional` can be set to `true` to
allow the release of a Helm chart to continue if the values file could
not be found.


squaremo:
Reading values from a file is a bit odd -- the path is given in the HelmRelease, but refers to a file in the filesystem of the operator. This is a trapeze act too far, in my opinion.

squaremo:
I'm not against fetching values from URLs -- though it's not very "GitOps", I think on balance it's reasonable to give people some flexibility on where the values come from. (Or to put it another way: it's not something I'd do myself, but who am I to judge ...).

hiddeco:
I do agree with you that reading values from a file is a bit odd, but at the same time I think it is also _oddly specific_ to create a `valueFileURLs` array in the CRD.

deploy-helm/flux-helm-release-crd.yaml
```diff
@@ -70,4 +102,4 @@ spec:
                   chartPullSecret:
                     properties:
                       name:
-                        type: string
+                        type: string
```
stefanprodan: New line here

chart/flux/templates/helm-operator-crd.yaml
```diff
@@ -44,13 +44,45 @@ spec:
...
+                type: object
+                required: ['name']
+                properties:
+                  name:
+                    type: string
+            valuesFrom:
+              type: array
+              items:
+                type: object
+                properties:
```
squaremo: Should there be a `oneOf` somewhere around here, to restrict each item to be _either_ a `secretKeyRef` or a `configMapKeyRef` or an `externalSourceRef`?

integrations/helm/release/release.go
```diff
@@ -272,8 +273,97 @@ func fhrResourceID(fhr flux_v1beta1.HelmRelease) flux.ResourceID {
...
+			}
+			d, ok := secret.Data[key]
+			if !ok {
+				if optional {
+					continue
+				}
+				return result, fmt.Errorf("could not find key %s in Secret %s/%s", key, ns, name)
+			}
+			if err := yaml.Unmarshal(d, &valueFile); err != nil {
+				return result, fmt.Errorf("unable to yaml.Unmarshal %v from %s in ConfigMap %s/%s", d, key, ns, name)
```
squaremo:
```suggestion
				return result, fmt.Errorf("unable to yaml.Unmarshal %v from %s in Secret %s/%s", d, key, ns, name)
```

site/helm-integration.md
```diff
@@ -10,8 +10,10 @@ menu_order: 90
     + [What the Helm Operator does](#what-the-helm-operator-does)
   * [Supplying values to the chart](#supplying-values-to-the-chart)
     + [`.spec.values`](#specvalues)
-    + [`.spec.valueFileSecrets`](#specvaluefilesecrets)
-      - [Example of `spec.valueFileSecrets`](#example-of-specvaluefilesecrets)
+    + [`.spec.valuesFrom`](#specvaluesfrom)
+      * [Config maps](#config-maps)
+      * [Secrets](#secrets)
+      * [External sources](#externals-ources)
```
squaremo:
```suggestion
      * [External sources](#external-sources)
```

site/helm-integration.md
```diff
@@ -151,63 +153,65 @@ spec:
...
-  name: uses-secret
-  namespace: dev
 spec:
   # chart: ...
-  valueFileSecrets:
-  - name: default-values
+  valuesFrom:
+  - externalSourceRef:
+      # URL of the values.yaml
+      url: https://flux.dev/weaveworks/example/values.yaml # mandatory
```
squaremo: `flux.dev` is owned by someone (and it's not Weaveworks) -- maybe use a github.com address, or example.com.

chart/flux/templates/helm-operator-crd.yaml
hiddeco: I ended up having quite a fight yesterday with `oneOf` and could not get it to work properly in the end. Will make a second attempt...

hiddeco:
@squaremo I think this is now to your liking.

chart/flux/templates/helm-operator-crd.yaml
```diff
@@ -44,13 +44,49 @@ spec:
...
+                        type: boolean
+                  externalSourceRef:
+                    type: object
+                    required: ['url']
+                    properties:
+                      url:
+                        type: string
+                      optional:
+                        type: boolean
+                oneOf:
```
squaremo: Ah! I didn't know the directive could go here, but this works.

PR #1844, opened 2019-03-19
Enhances git command tracing

dimitropoulos:
There is currently a very basic mechanism built in to `execGitCmd` that is essential in certain kinds of debugging for tracing the git commands that the application is using.  Recently in work on a separate PR, I needed to view the output of the git commands.  On top of that, some frequent commands (such as the two `git config` lines that occur almost every time a flux action occurs) make it hard to quickly zero in on what you're looking for.

This PR adds:
1. functionality such that setting `const trace = true` (as before) will now output a more parse-friendly log line that also includes the output of the git command.
2. an optional blacklist that can be used whilst developing

### DIFF
--- a/git/operations.go
+++ b/git/operations.go
@@ -16,8 +16,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-// If true, every git invocation will be echoed to stdout
+// If true, every git invocation will be echoed to stdout (with the exception of those added to `exemptedTraceCommands`)
 const trace = false
+
+// Whilst debugging or developing, you may wish to filter certain git commands out of the logs when tracing is on.
+var exemptedTraceCommands = []string{
+	// To filter out a certain git subcommand add it here, e.g.:
+	// "config",
+}
 
 // Env vars that are allowed to be inherited from the os
 var allowedEnvVars = []string{"http_proxy", "https_proxy", "no_proxy", "HOME", "GNUPGHOME"}
@@ -276,15 +282,37 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
 	return splitList(out.String()), nil
 }
 
-func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
-	if trace {
-		print("TRACE: git")
-		for _, arg := range args {
-			print(` "`, arg, `"`)
+// traceGitCommand returns a log line that can be useful when debugging and developing git activity
+func traceGitCommand(args []string, config gitCmdConfig, stdout string, stderr string) string {
+	for _, exemptedCommand := range exemptedTraceCommands {
+		if exemptedCommand == args[0] {
+			return ""
 		}
-		println()
 	}
 
+	prepare := func(input string) string {
+		output := strings.Trim(input, "\x00")
+		output = strings.TrimSuffix(output, "\n")
+		output = strings.Replace(output, "\n", "\\n", -1)
+		return output
+	}
+
+	command := `git ` + strings.Join(args, " ")
+	out := prepare(stdout)
+	err := prepare(stderr)
+
+	return fmt.Sprintf(
+		"TRACE: command=%q out=%q err=%q dir=%q env=%q",
+		command,
+		out,
+		err,
+		config.dir,
+		strings.Join(config.env, ","),
+	)
+}
+
+// execGitCmd runs a `git` command with the supplied arguments.
+func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
 	c := exec.CommandContext(ctx, "git", args...)
 
 	if config.dir != "" {
@@ -298,6 +326,13 @@ func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
 	errOut := &bytes.Buffer{}
 	c.Stderr = errOut
 
+	traceStdout := &bytes.Buffer{}
+	traceStderr := &bytes.Buffer{}
+	if trace {
+		c.Stdout = io.MultiWriter(c.Stdout, traceStdout)
+		c.Stderr = io.MultiWriter(c.Stderr, traceStderr)
+	}
+
 	err := c.Run()
 	if err != nil {
 		msg := findErrorMessage(errOut)
@@ -305,6 +340,13 @@ func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
 			err = errors.New(msg)
 		}
 	}
+
+	if trace {
+		if traceCommand := traceGitCommand(args, config, traceStdout.String(), traceStderr.String()); traceCommand != "" {
+			println(traceCommand)
+		}
+	}
+
 	if ctx.Err() == context.DeadlineExceeded {
 		return errors.Wrap(ctx.Err(), fmt.Sprintf("running git command: %s %v", "git", args))
 	} else if ctx.Err() == context.Canceled {
--- a/git/operations_test.go
+++ b/git/operations_test.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/weaveworks/flux/cluster/kubernetes/testfiles"
 )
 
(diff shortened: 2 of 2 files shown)


git/operations.go
```diff
@@ -276,25 +275,60 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
...
 }
 
-func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
-	if trace {
-		print("TRACE: git")
-		for _, arg := range args {
-			print(` "`, arg, `"`)
+// Whilst debugging or developing, you may wish to filter certain git commands out of the logs.
+func exemptedTraceGitCommand(command string) bool {
+	blacklistedGitCommands := []string{
```
squaremo:
```suggestion
	exemptedCommands := []string{
```

squaremo: etc.

git/operations.go
```diff
@@ -305,6 +339,11 @@ func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
 			err = errors.New(msg)
 		}
 	}
+
+	if trace && !exemptedTraceGitCommand(args[0]) {
+		println(traceGitCommand(args, config, traceOutput.String()))
```
squaremo: I would just inline the yes/no logic of `exemptedTraceGitCommand`, `traceGitCommand` and the `println` into one func, and call that here. That would remove some noise from this func.

git/operations.go
```diff
@@ -276,25 +275,60 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
...
+
+	traceOutput := &bytes.Buffer{}
+	c.Stdout = traceOutput
 	if config.out != nil {
-		c.Stdout = config.out
+		c.Stdout = io.MultiWriter(config.out, traceOutput)
 	}
+
 	errOut := &bytes.Buffer{}
 	c.Stderr = errOut
```
squaremo: Much of the time it's going to be stderr that has the interesting information -- I'd include that in the trace, as well.

git/operations.go
```diff
@@ -276,25 +275,60 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
...
 		c.Dir = config.dir
 	}
 	c.Env = append(env(), config.env...)
-	c.Stdout = ioutil.Discard
+
+	traceOutput := &bytes.Buffer{}
+	c.Stdout = traceOutput
 	if config.out != nil {
-		c.Stdout = config.out
+		c.Stdout = io.MultiWriter(config.out, traceOutput)
```
squaremo: Shouldn't this also be conditional on `trace`?

squaremo: It might be clearer to set the writers up as before, then have a stanza conditional on `trace` that swaps them out.

git/operations.go
dimitropoulos: I considered doing that but the result of that could mean that someone would set `trace` to `true` and then see:

```go
if trace {
  traceGitCommand(...)
}
```
but in fact that command could full well do nothing.

On second thought, since (as it stands now, but not as it was then when I made that decision) there's nothing in the exempted list, then I think it makes sense to assume that if someone added something to that list (temporarily) then that person will fully understand what's going on.

I'll make that change.

dimitropoulos: as for inlining the println.. how would I then test that?  otherwise I would need to do

```go
		traceCommand = traceGitCommand(args, config, traceOutput.String())
		if traceCommand != "" {
			println(traceCommand)
		}
```

git/operations.go
dimitropoulos: ok.  I avoided doing this because it's functionally equivalent to what was there before (`c.Stdout = ioutil.Discard`).  Writing to an writer that's never read from is the same as writing to ioutil.Discard.

I'll make this change now

git/operations.go
dimitropoulos: :heavy_check_mark:

git/operations.go
squaremo: You can keep `traceGitCommand` factored out to be able to test it. I was more concerned about `exemptedGitCommand` standing on its own -- it's too simple to justify factoring it out, but distracting when inlined into `execGitCmd` directly.

dimitropoulos: ok, I will keep it like this then:

```
	if trace {
		traceCommand := traceGitCommand(
			args,
			config,
			traceStdout.String(),
			traceStderr.String(),
		)
		if traceCommand != "" {
			println(traceCommand)
		}
	}
```

squaremo: That seems reasonable yup. You could shorten that a little to

```
if trace {
		if traceCommand := traceGitCommand(
			args,
			config,
			traceStdout.String(),
			traceStderr.String(),
		); traceCommand != "" {
			println(traceCommand)
		}
	}
```

(also there's no need to have each arg on its own line -- for better or worse, it's not the Go Way to limit line length (or at least, not to limit it to 72 chars ..))

dimitropoulos:
@squaremo I updated to satisfy all the changes you requested.  Let me know if anything else jumps out at you. :)

git/operations.go
```diff
@@ -276,35 +276,94 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
...
+// traceGitCommand returns a log line that can be useful when debugging and developing git activity
+func traceGitCommand(args []string, config gitCmdConfig, stdout string, stderr string) string {
+	if exemptedTraceGitCommand(args[0]) {
+		return ""
+	}
+
+	prepare := func(input string) string {
+		output := strings.Trim(input, "\x00")
+		scanner := bufio.NewScanner(strings.NewReader(output))
+		lines := []string{}
```
squaremo: Unless you know the length of a slice before you allocate it, usually you'd let it be a zero value since `append` and `len` treat those the same as a zero-length slice.

    var lines []string

git/operations.go
```diff
@@ -276,35 +276,94 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
...
+
+// traceGitCommand returns a log line that can be useful when debugging and developing git activity
+func traceGitCommand(args []string, config gitCmdConfig, stdout string, stderr string) string {
+	if exemptedTraceGitCommand(args[0]) {
+		return ""
+	}
+
+	prepare := func(input string) string {
+		output := strings.Trim(input, "\x00")
+		scanner := bufio.NewScanner(strings.NewReader(output))
```
squaremo: `return strings.Replace(output, "\n", "\\n", -1)`

(the `-1` means replace all; there is a ReplaceAll, but it seems to be in a more recent Go stdlib)

git/operations.go
```diff
@@ -276,35 +276,94 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
 	return splitList(out.String()), nil
 }
 
-func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
-	if trace {
-		print("TRACE: git")
-		for _, arg := range args {
-			print(` "`, arg, `"`)
+// Whilst debugging or developing, you may wish to filter certain git commands out of the logs.
+func exemptedTraceGitCommand(command string) bool {
```
squaremo: I still don't think this needs to be on its own. It's not complicated enough to warrant having its own function, or used in more than one place, or tested (and needn't be).

You could move the `exemptedCommands` up to the package scope and put them beside `trace` (and rename them, since they will be out of context), since people might want to alter them at the same time as `trace`.

git/operations.go
squaremo: (immaterial if you replace all this with strings.Replace)

git/operations.go
dimitropoulos: just to be clear @squaremo you told me to extract it into a func, that's why I did.

Screenshot_20190326_094534

If we're on the topic I would prefer to just use a simple `contains` function, but that has already been added to [`aws.go`](https://github.com/weaveworks/flux/blob/1521741297657c38b9319dd1deb8f88cb1b72bcf/registry/aws.go#L38) and [`compat.go`](https://github.com/weaveworks/flux/blob/1521741297657c38b9319dd1deb8f88cb1b72bcf/remote/rpc/compat.go#L66) and this would make the third place this is copied in the project, that's why it's its own thing in the first place.  I think the semantics of
```go
traceExcemption := contains([]string{
        "config",
    }, args[0])
```
are very clear, but anyway... I'll happily change to match your latest request, no problemo.

git/operations.go
dimitropoulos: thanks for mentioning.  great suggestion.  I'm going to use strings.Replace like you said but that's helpful information to have.

git/operations.go
squaremo: > you told me to extract it into a func

I meant the whole machinery, not just the trivial string-in-slice predicate.

git/operations.go
dimitropoulos: I had used a scanner because on a call when I presented this issue to you (the issue of what method to opt for for replacing the newlines) you had said to use a scanner and scan by line.

I like the strings.Replace option, thanks :)

Since we're taking the "use the standard lib" approach more, I also replaced the other for loop to `strings.TrimSuffix(output, "\n")`.

git/operations.go
dimitropoulos: ok, well no worries.  I just mentioned it because I want you to understand that the only reason it was like that was because I was following your instructions as closely as possible.

thanks :)

git/operations.go
```diff
@@ -276,15 +282,43 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
...
+func traceGitCommand(args []string, config gitCmdConfig, stdout string, stderr string) string {
+	for _, exemptedCommand := range exemptedTraceCommands {
+		if exemptedCommand == args[0] {
+			return ""
 		}
-		println()
 	}
 
+	prepare := func(input string) string {
+		// remove any null characters
```
squaremo: These comments just (p)recapitulate what's obvious from the code, inaccurately. Better to let the code speak for itself.

dimitropoulos: kk.

git/operations.go
```diff
@@ -276,15 +282,37 @@ func changed(ctx context.Context, workingDir, ref string, subPaths []string) ([]
 	return splitList(out.String()), nil
 }
 
-func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
-	if trace {
-		print("TRACE: git")
-		for _, arg := range args {
-			print(` "`, arg, `"`)
+// traceGitCommand returns a log line that can be useful when debugging and developing git activity
+func traceGitCommand(args []string, config gitCmdConfig, stdout string, stderr string) string {
```
squaremo: BTW you can collapse adjacent variables or arguments of the same type: `..., stdout, stderr string`

dimitropoulos: Although the standard library and the Go source code does that a lot I tend to not opt to because if you ever need to refactor something I find it a problem that if you have: a signature like `func example(firstArg, secondArg string)` that if you should want to change `secondArg` to something else that you would need to make a change to firstArg.  This happened to me a lot,[ for example](https://github.com/dimitropoulos/flux/commit/d3d6248d1da783a1df76b8fca11063060282d31d), when I was refactoring the usages of git refs to better understand how they're used in the application.  Although that was mostly for my own understanding it was extremely tedious to have to constantly undo this "shortcut".

Also I find it unnecessarily complicated to not be able to see what the type of something without reading far down the line (e.g. `func clone(ctx context.Context, workingDir, repoURL, repoBranch string)`).  My attitude is that source code should be optimized for reading not writing since reading code is where the vast majority of time is spent - that's why I don't do it.

I'll keep that in mind in the future.

**Would you like me to make that change before merging?**  I only ask because you made this comment after approving.

squaremo: > Would you like me to make that change before merging?

Nahp, no need to change it.

PR #1865, opened 2019-03-26
Detect changes made to git source in HelmRelease

### LINKED ISSUE #1850: Changing repository url of `HelmRelease` doesn't sync
We were changing the URL of the git repository of a `HelmRelease`.
Unfortunately, the Helm Operator didn't pick that up and didn't trigger a sync.

Deleting the pod of `helm-operator`, thus restarting it, helped. We would expect that it automatically picks up the new url without a restart.

We also tried deleting the `HelmRelease` resource. Flux was re-creating the `HelmRelease` resource but still the old repository URL was used by `helm-operator`. Is this a known issue?

hiddeco:
Fixes #1850 

Before this change, changes made to the git chart source in a `HelmRelease` did go unnoticed. Now, in case the URL or ref in the `HelmRelease` changes, the clone is removed so a new one can be made that matches the updated URL and ref. As a bonus, in case a `HelmRelease` is removed, the clone will now be removed too.

### DIFF
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -102,6 +102,8 @@ func (c Config) WithDefaults() Config {
 // revision), so we can keep track of when it needs to be updated.
 type clone struct {
 	export *git.Export
+	remote string
+	ref    string
 	head   string
 }
 
@@ -230,7 +232,7 @@ func (chs *ChartChangeSync) Run(stopCh <-chan struct{}, errc chan error, wg *syn
 							chs.logger.Log("warning", "could not clone from mirror while checking for changes", "repo", repoURL, "ref", ref, "err", err)
 							continue
 						}
-						newCloneForChart := clone{head: refHead, export: newClone}
+						newCloneForChart := clone{remote: repoURL, ref: ref, head: refHead, export: newClone}
 						chs.clonesMu.Lock()
 						chs.clones[releaseName] = newCloneForChart
 						chs.clonesMu.Unlock()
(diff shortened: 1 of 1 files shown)


squaremo:
This is my understanding of the cause of the problem and the solution here:

 - if the git URL in a HelmRelease changes (to something that is not already being mirrored) ...
   - it won't be noticed in the mirror changes (L156 then L172 which checks if a HelmRelease uses a mirror that's changed)
   - when the ticker ticks (L244) the release will be considered in `reconcileReleaseDef`, then ...
     - it will find a clone against the release name `clones[releaseName]` at L307
     - that will be used to calculate the release -- but it's from the old git URL still, so nothing will change

This PR changes how clones are recorded, such that they use the git URL as well as the release name. That means that at (the equivalent of) [[L307]], the clone is _not_ found, a new mirror is set up, and that will trigger a change at [[L172]] when it is ready. At [[L209]], the clone for the release won't be found, and it'll be exported again. From that point, things will continue as before.

One question that occurred to me was "what if the git branch (ref) in the HelmRelease changes, but not the URL?".

This takes a different path: when there's a change in the repo, the check at [[L172]] will succeed, and the clone will be present at [[L209]]. The check for new commits at [[L214]] may fail, if git can't construct a history between the old branch and the new branch; and this problem will likely persist. (This might be solved by just checking if the head is the same; anyway it's a side issue.) Assuming it gets past [[L214]], it'll export the repo and continue as before.

Another thing that struck me is "why not check if the git URL and ref match the clone before using it?". I.e., in `reconcileReleaseDef`, if the clone is found at L307, compare the HelmRelease values to the clone, and remove the clone if they don't match (then proceed with the "clone not found" branch). If it's correct, it would have the advantage of not requiring additional clean-up code.

hiddeco:
@squaremo your comment made a lot of sense and made the solution a lot simpler. PTAL whenever you feel like it.

hpurmann:
Thanks to both of you!

PR #1875, opened 2019-03-28
Use FQRN to prevent errors when branch = label

### LINKED ISSUE #1845: If branch = label then new commits are only pulled when flux is started
Following option 2) https://github.com/weaveworks/flux/blob/master/site/faq.md#i-have-a-dedicated-kubernetes-cluster-per-environment-and-i-want-to-use-the-same-git-repo-for-all-how-can-i-do-that

If you happen to set the label to the same as the branch name, then the newest commit (abc...) on startup of flux is fetched and applied. But then from there on out every time it polls it will say that HEAD=abc... and not apply changes even if you've since pushed a new commit. Restarting flux will then cause it to fetch the actual newest commit and start over being stuck at that commit.

Changing the label to something else, doing `git push --delete origin tags/<label>` and then restarting flux fixes this.

hiddeco:
Fixes #1845 

If the label and branch were set to the same name, Flux would not
create the sync tag as the revision lookup for `HEAD` and `<tag>`
returned the same revision.

This is due to how git looks up symbolic ref names, giving priority
to heads over tags. Using the (almost) fully qualified ref name will
tell git to only look for matches of that type.

https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrefnamegtemegemmasterememheadsmasterememrefsheadsmasterem

### DIFF
--- a/git/working.go
+++ b/git/working.go
@@ -141,7 +141,7 @@ func (c *Checkout) CommitAndPush(ctx context.Context, commitAction CommitAction,
 	}
 
 	if note != nil {
-		rev, err := refRevision(ctx, c.dir, "HEAD")
+		rev, err := c.HeadRevision(ctx)
 		if err != nil {
 			return err
 		}
@@ -174,7 +174,7 @@ func (c *Checkout) HeadRevision(ctx context.Context) (string, error) {
 }
 
 func (c *Checkout) SyncRevision(ctx context.Context) (string, error) {
-	return refRevision(ctx, c.dir, c.config.SyncTag)
+	return refRevision(ctx, c.dir, "tags/"+c.config.SyncTag)
 }
 
 func (c *Checkout) MoveSyncTagAndPush(ctx context.Context, tagAction TagAction) error {


git/working.go
```diff
@@ -170,11 +170,11 @@ func (c *Checkout) GetNote(ctx context.Context, rev string, note interface{}) (b
 }
 
 func (c *Checkout) HeadRevision(ctx context.Context) (string, error) {
-	return refRevision(ctx, c.dir, "HEAD")
+	return refRevision(ctx, c.dir, "heads/"+c.config.Branch)
```
squaremo: This changes the meaning of `HeadRevision` from "head of checked out branch" to "head of branch configured". I don't _think_ these can differ, as things stand, because there's no way to change the checked out branch of a clone. So I suppose this is a note to our future selves ..

2opremio: I would actually rename the function to make it more clear.

hiddeco: I only changed it to create a clear usage pattern of `refRevision`.

Reverting the change to just lookup the `HEAD` will actually do us no harm as this will always be the revision of what is checked out at that moment.

squaremo: > Reverting the change to just lookup the `HEAD` will actually do us no harm as this will always be the revision of what is checked out at that moment.

I have a _slight_ preference for this, but it's pretty slight.

2opremio: I wouldn't totally revert it though, I guess a branch or a tag can also be named HEAD

hiddeco: A branch can't be named `HEAD`, a tag can but `rev-list` will prioritize the real `HEAD` over the tag.

PR #1876, opened 2019-03-28
Obtain scope of CRD instances from its manifest as a fallback

### LINKED ISSUE #1825: Cope when syncing a CRD and resources using the CRD
If you add a CRD as well as resources that use the CRD, in one commit (or in commits since the last sync), fluxd will stop being able to sync.

What I think is happening is that the namespacer will attempt to look up the scope of the custom resources, and since the definition has not yet been created, it will fail. At present this will stop the sync, and fluxd will not be able to make progress.

One solution that suggests itself is to treat a namespace lookup as a failure for the resource in question only, and otherwise proceed. That way, the CRD will still be created, and in the next sync, the lookup will succeed.

2opremio:
Before this, we were only obtaining the scope of CRD instances from the cluster.

However, this wasn't good enough if the Git repository contained both the CRD manifest and instance manifests.

Fixes #1825

### DIFF
--- a/cluster/kubernetes/manifests.go
+++ b/cluster/kubernetes/manifests.go
@@ -1,11 +1,18 @@
 package kubernetes
 
 import (
+	"gopkg.in/yaml.v2"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	"github.com/weaveworks/flux"
 	kresource "github.com/weaveworks/flux/cluster/kubernetes/resource"
 	"github.com/weaveworks/flux/image"
 	"github.com/weaveworks/flux/resource"
 )
+
+// ResourceScopes maps resource definitions (GroupVersionKind) to whether they are namespaced or not
+type ResourceScopes map[schema.GroupVersionKind]v1beta1.ResourceScope
 
 // namespacer assigns namespaces to manifests that need it (or "" if
 // the manifest should not have a namespace.
@@ -14,7 +21,7 @@ type namespacer interface {
 	// the manifest to be applied. This may be "", indicating that it
 	// should not have a namespace (i.e., it's a cluster-level
 	// resource).
-	EffectiveNamespace(kresource.KubeManifest) (string, error)
+	EffectiveNamespace(manifest kresource.KubeManifest, knownScopes ResourceScopes) (string, error)
 }
 
 // Manifests is an implementation of cluster.Manifests, particular to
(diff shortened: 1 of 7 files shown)


cluster/kubernetes/namespacer.go
```diff
@@ -68,7 +70,24 @@ func (n *namespaceViaDiscovery) EffectiveNamespace(m kresource.KubeManifest) (st
 	return m.GetNamespace(), nil
 }
 
-func (n *namespaceViaDiscovery) lookupNamespaced(groupVersion, kind string) (bool, error) {
+func (n *namespaceViaDiscovery) lookupNamespaced(groupVersion string, kind string, knownScopes ResourceDefinitionScopes) (bool, error) {
+	namespaced, clusterErr := n.lookupNamespacedInCluster(groupVersion, kind)
+	if clusterErr == nil || knownScopes == nil {
```
squaremo: If `clusterErr != nil`, you'd usually consider `namespaced` to be a suspect value.

cluster/kubernetes/manifests.go
```diff
@@ -26,11 +33,39 @@ type Manifests struct {
...
+				}
+				result[gvk] = crd.Spec.Scope
+			}
+		}
+	}
+	return result
+}
+
 func postProcess(manifests map[string]kresource.KubeManifest, nser namespacer) (map[string]resource.Resource, error) {
+	knownScopes := getCRDScopes(manifests)
```
squaremo: Is it possible to use the fallback lookup here instead of putting it within `EffectiveNamespace`? It jars a little to have some of the special case code here (collecting the CRD scopes) and some of it there -- it feels like they should just be in the same place.

cluster/kubernetes/namespacer.go
squaremo: Not to mention, if `clusterErr != nil` and `knownScopes` is empty, the result should be an error signalling the scope cannot be determined.

2opremio: Yep, I should be returning err. Good catch, I forgot to restore that after a refactoring.

cluster/kubernetes/manifests.go
2opremio: Uhm, that's how I intended to implement it initially, but I though this would be cleaner. I will give it another go.

squaremo: Oh dear am I too unpredictable as a reviewer :-S

cluster/kubernetes/namespacer.go
2opremio: Addressed

cluster/kubernetes/manifests.go
2opremio: Hehehe, no worries. I gave it a try but I don't see a clean way to do it without replicating `namespaceViaDiscovery.EffectiveNamespace` in `manifests.go`.

I concur in that splitting the obtention of the scope - as per the cluster and as per the git manifests - in separate files is ugly. However for the later you need all the resources.

I could merge `namespacer.go` into `manifests.go` and remove the namespacer interface if you think that's better.

squaremo: I'm OK with having that split, so long as it's clean: that is, namespacer.go does what it can to answer with respect to the cluster, and if it cannot, postProcess in manifests.go looks it up in the map of scopes it has just compiled.

Maybe there's a reason it won't work as I imagine it -- I'm fine with how it is, really.

2opremio: Then let's merge it as is. I am ready to push the button after you've taken another look

squaremo: I worked out how I'm being dim.

It's not enough to have a record of which custom resources are cluster scoped and which are not, to answer the question "which namespace should this go in?"  -- you also need to know the fallback namespace. So it can't be done in manifests.go without further changes to make that latter available, and it wouldn't be any simpler.

2opremio: Exactly :)

2opremio:
I case you are wondering about the last commit amendment, I just renamed `ResourceDefinitionScopes` to `ResourceScopes`

AmberAttebery:
K

PR #1884, opened 2019-04-01
Support Git >=2.21

hiddeco:
In git <=2.20 the error started with an uppercase, in 2.21 this was
changed to be consistent with all other die() and error() messages,
casting it to lowercase enables support for both versions.

Ref: https://github.com/git/git/commit/0b9c3afdbfb62936337efc52b4007a446939b96b


2opremio:
Awesome! Thanks @hiddeco !!

PR #1888, opened 2019-04-01
Release Flux v1.11.1

2opremio:


2opremio:
@hiddeco Yes, but the bumping of the deployment is scripted as a later step.

Sorry for always forgetting.

PR #1892, opened 2019-04-01
Update helm chart to use 1.11.1 as well

timfpark:
Commit https://github.com/weaveworks/flux/commit/ba3e794a59e588eaaafa7bee07c8a0fb322e166e updated the flux/deploy static yaml files, but not the helm template.

This PR adjusts the helm chart's value.yaml file to also use tag 1.11.1.

### DIFF
--- a/chart/flux/Chart.yaml
+++ b/chart/flux/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.11.0"
+appVersion: "1.11.1"
 version: 0.7.0
 kubeVersion: ">=1.9.0-0"
 name: flux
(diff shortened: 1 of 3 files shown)


timfpark:
@stefanprodan Thanks for the review - done.

chart/flux/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.11.0"
+appVersion: "1.11.1"
 version: 0.7.0
```
stefanprodan:
```suggestion
version: 0.7.1
```

stefanprodan: @timfpark you've disabled the push access to your branch. Can you commit the above?

Issue #1862, opened 2019-03-26
Allow custom repositories for dependent helm charts.


bricef:
When using helm charts, The dependencies of the helm chart are not available if they are in a custom repository.

Maybe related to #1278?

Reported by @tmestdagh

PR #1893, opened 2019-04-03
Define custom Helm repositories in the Helm chart

srueg:
This simplifies using custom Helm repositories and generates a
`repositories.yaml` for the Helm operator.

Relates to #1862

Signed-off-by: Simon Rüegg <simon.ruegg@vshn.ch>

### DIFF
--- a/chart/flux/templates/_helpers.tpl
+++ b/chart/flux/templates/_helpers.tpl
@@ -41,3 +41,22 @@ Create the name of the service account to use
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Create a custom repositories.yaml for Helm
+*/}}
+{{- define "flux.customRepositories" -}}
+apiVersion: v1
+generated: 0001-01-01T00:00:00Z
+repositories:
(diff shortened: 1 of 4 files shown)


squaremo:
@stefanprodan The BGB (Big Green Button) is yours

chart/flux/templates/_helpers.tpl
```diff
@@ -41,3 +41,22 @@ Create the name of the service account to use
...
+generated: 0001-01-01T00:00:00Z
+repositories:
+{{- range .Values.helmOperator.configureRepositories.repositories }}
+- name: {{ .name }}
+  url: {{ .url }}
+  cache: /var/fluxd/helm/repository/cache/{{ .name }}-index.yaml
+  caFile: ""
+  certFile: ""
+  keyFile: ""
+  password: {{ .password }}
```
stefanprodan: Does this work with empty string? Shouldn't be `"{{ .password }}"` and `"{{ .username }}"` ?

srueg: You're totally right. I also made the URL and name required.

PR #1895, opened 2019-04-03
[WIP] Adapt e2e scripts so they can be used more generally (and locally)

squaremo:
So far:

 - make them all respect a base directory, to avoid stomping on
   /usr/local
 - add some notes in each about what they depend on

### DIFF
--- a/test/e2e/e2e-golang.sh
+++ b/test/e2e/e2e-golang.sh
@@ -2,18 +2,17 @@
 
 set -o errexit
 
+source $(dirname $0)/e2e-paths.env
+
 GO_VERSION=1.11.4
 
-echo ">>> Installing go ${GO_VERSION}"
-curl -O https://storage.googleapis.com/golang/go${GO_VERSION}.linux-amd64.tar.gz
+echo ">>> Installing go ${GO_VERSION} to $GOBASE/go"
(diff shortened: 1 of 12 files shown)


test/e2e/e2e-golang.sh
```diff
@@ -2,18 +2,17 @@
...
 
+source $(dirname $0)/e2e-paths.env
+
 GO_VERSION=1.11.4
 
-echo ">>> Installing go ${GO_VERSION}"
-curl -O https://storage.googleapis.com/golang/go${GO_VERSION}.linux-amd64.tar.gz
+echo ">>> Installing go ${GO_VERSION} to $GOBASE/go"
+if ! [ -f "go${GO_VERSION}.linux-amd64.tar.gz" ]; then
+    curl -O https://storage.googleapis.com/golang/go${GO_VERSION}.linux-amd64.tar.gz
```
2opremio: This won't work for other architectures (e.g. darwin)

I think that we should simply run e2e tests as a continuation from building through the Makefile.

As an optimization, e2e CircleCI task could gather the artifacts from the build stage https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs

2opremio:
Superseded by https://github.com/weaveworks/flux/pull/2058

PR #1901, opened 2019-04-04
Fix `tag_all` pseudo policy on default-namespaced wokloads

2opremio:
This fixes a regression introduced by #1442 in which applying the `tag_all` pseudo policy failed for workload manifests in which the namespace is omitted.

The effective namespace of the workload wasn't set after parsing, causing a mismatch in the resource identifier (the parsed resource identifier was cluster-scoped due to the lack of explicit namespace in the manifest).

### DIFF
--- a/cluster/kubernetes/policies.go
+++ b/cluster/kubernetes/policies.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
 
 	"github.com/weaveworks/flux"
 	kresource "github.com/weaveworks/flux/cluster/kubernetes/resource"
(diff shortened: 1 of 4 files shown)


cluster/kubernetes/policies.go
```diff
@@ -49,25 +48,17 @@ func (m *Manifests) UpdatePolicies(def []byte, id flux.ResourceID, update policy
...
-	} `yaml:"metadata"`
-}
-
-func extractAnnotations(def []byte) (map[string]string, error) {
-	var m manifest
-	if err := yaml.Unmarshal(def, &m); err != nil {
-		return nil, errors.Wrap(err, "decoding manifest for annotations")
-	}
-	if m.Metadata.Annotations == nil {
-		return map[string]string{}, nil
```
2opremio: This is dead code, I couldn't bother to create a separate PR for this.

squaremo: All good :+1:

PR #1902, opened 2019-04-04
Exclude Helm manifests in get-started guide

### LINKED ISSUE #1898: error looking up API resources for HelmRelease.flux.weave.works/v1beta1
Now that #1896 is resolved I'm hitting another error after a sync of the flux-get-started.git repo from my private host (I had to follow https://github.com/weaveworks/flux/blob/master/site/standalone-setup.md#using-a-private-git-host to get that working but I think the clone is working ok).

I'm seeing the following error just after the close and no sign of the demo namespace or deployment/podinfo

```
ts=2019-04-03T16:54:50.865047078Z caller=loop.go:103 component=sync-loop event=refreshed url=git@github.example.com:xxx-xxx/flux-get-started.git branch=master HEAD=b87b0319407f3f05cdd44a9058599d183f011e20
ts=2019-04-03T16:54:50.938893382Z caller=loop.go:90 component=sync-loop err="loading resources from repo: error looking up API resources for HelmRelease.flux.weave.works/v1beta1: not found"
ts=2019-04-03T16:55:45.37419714Z caller=images.go:17 component=sync-loop msg="polling images"

```

2opremio:
The get started guide uses repo github.com/weaveworks/flux-get-started
which contains HelmRelease manifests, but the guide didn't instruct
the reader to deploy the Helm operator.

Fixes #1898

### DIFF
--- a/deploy/flux-deployment.yaml
+++ b/deploy/flux-deployment.yaml
@@ -105,6 +105,9 @@ spec:
         # replace or remove the following URL
         - --git-url=git@github.com:weaveworks/flux-get-started
         - --git-branch=master
+        # include this if you want to restrict the manifests considered by flux
+        # to those under the following relative paths in the git repository
+        # - --git-paths=subdir1,subdir2
 
         # include these next two to connect to an "upstream" service
         # (e.g., Weave Cloud). The token is particular to the service.
--- a/site/get-started.md
+++ b/site/get-started.md
@@ -63,7 +63,10 @@ $EDITOR deploy/flux-deployment.yaml
 In our example we are going to use
 [flux-get-started](https://github.com/weaveworks/flux-get-started). If you
 want to use that too, be sure to create a fork of it on GitHub and
-add the git URL to the config file above.
+add the git URL to the config file above. After that, set the `--git-path` 
+flag to `--git-path=namespaces,workloads`, this is meant to exclude Helm 
+manifests. Again, if you want to get started with Helm, please refer to the 
+[Helm section](./helm-get-started.md).
 
 ## Deploying Flux to the cluster
 
@@ -146,4 +149,5 @@ very straight-forward and are a quite natural work-flow.
 As a next step, you might want to dive deeper into [how to
 control Flux](./fluxctl.md), check out [more sophisticated
 setups](./standalone-setup.md) or go through our hands-on
-tutorial about driving Flux, e.g. [automations, annotations and locks](annotations-tutorial.md).
+tutorial about driving Flux, e.g. 
+[automations, annotations and locks](annotations-tutorial.md).


hiddeco:
I think we should make it more clear to the user that installation of the Helm operator is optional (and only required because the `flux-get-started` repository makes use of it).

If the user now follows these steps but for example, has not installed Tiller on the cluster (because we did not tell them to), the apply of the `HelmRelease` will succeed but the roll-out of the release itself will fail (and the operator will only be logging errors).

2opremio:
That's true. Then, we either change the example repository to one without HelmReleases or we add instructions to install Tiller and indicate that helm is optional. Thoughts?

2opremio:
BTW, how could the guide have possiblly ever worked? I'll do some digging

hiddeco:
From the `git log` in my head; there was a commit a while ago that just changed the repository URLs without matching the content.

hiddeco:
> Then, we either change the example repository to one without HelmReleases or we add instructions to install Tiller and indicate that helm is optional.

I think the latter would be good, with a note saying that for demo purposes we install the operator too (and thus Tiller must be installed) but it is completely up to the user to only make use of the Flux daemon.

dholbach:
Hum. Maybe I'm missing something, but AFAICS the [get started tutorial](https://github.com/weaveworks/flux/blob/master/site/get-started.md) is about "flux standalone" (cf [this note](https://github.com/weaveworks/flux/blame/master/site/get-started.md#L23), whereas the [helm get started tutorial](https://github.com/weaveworks/flux/blob/master/site/helm-get-started.md) is about helm: the need to install Tiller is discussed there.

hiddeco:
> AFAICS the get started tutorial is about "flux standalone"

Correct, the problem is that the repository linked in this doc (and used as an example) contains `HelmRelease` resources, confusing users (see linked issue).

2opremio:
We can also change the repository.

dholbach:
This is unfortunate. It'd be great if we could 

 1. maintain just one example
 1. have a straight-forward example with clear instructions that you can easily tailor to your use-cases

It looks like we can't easily have it all this time.

My inclination would be to merge this PR, but make a note that we at least document why this is necessary in the future, and even better: make the use of the example repo here more "obvious"?

2opremio:
> It looks like we can't easily have it all this time.

We can also restrict the repo to directories not including `HelmReleases` by configuring the flux deployment. I think I will do that.

2opremio:
@hiddeco @dholbach PTAL

dholbach:
Thanks for the fix - this looks clean. (Even if it adds another step to the "get started" experience - I guess I'll file a separate issue about this.)

Have you tried this locally? In my case the new podinfo gets deployed, but for some reason the color does not change.

2opremio:
>  In my case the new podinfo gets deployed, but for some reason the color does not change.

It does work for me. Do the logs show anything odd?

I don't see why any of the changes I made would cause flux not to update the deployment correctly.

dholbach:
Just started with a fresh minikube, same thing. The correct revision is deployed. Colour stays green.

2opremio:
I will take a deeper look. However I doubt that's caused by this PR. Does it work when deploying helm and the helm op?

dholbach:
The tutorials follow different paths:
 - standalone get started asks people to change `PODINFO_UI_COLOR` to `blue` in `workloads/podinfo-dep.yaml` - this seems not to work in either of the variants.
 - helm get started asks people to update the image rev of mongodb, this works.

dholbach:
Maybe we can just copy https://github.com/weaveworks/flux/blame/master/site/helm-get-started.md#L147 (147 and following) over to `get-started.md`?

hiddeco:
I think the `podinfo` application no longer listens to the `PODINFO_UI_COLOR` environment variable (https://github.com/stefanprodan/k8s-podinfo/search?q=PODINFO_UI_COLOR&unscoped_q=PODINFO_UI_COLOR).

2opremio:
Can me merge this and create a separate issue for the color?

2opremio:
> It does work for me.

I retried and it didn't, I would had sworn it did. ¯\_(ツ)_/¯

Sorry @dholbach

dholbach:
Let's copy over the Helm Get Started bits into the other tutorial in a separate issues.

squaremo:
Is this ready to merge or is it blocked on something -- I'm not clear after reading the comments :sweat_smile:

dholbach:
Yes, let's do it, and fix the second part of it in a separate PR.

PR #1903, opened 2019-04-05
Add spaces to comment lines to fix helm chart lint

benhartley:
`ct lint` fails at present because of missing spaces at the start of comment lines

### DIFF
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -202,7 +202,7 @@ kube:
     kind: Config
     preferences: {}
     users: []
-# for https://github.com/justinbarrick/fluxcloud/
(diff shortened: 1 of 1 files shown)


chart/flux/values.yaml
```diff
@@ -202,8 +202,8 @@ kube:
     kind: Config
     preferences: {}
     users: []
-#for https://github.com/justinbarrick/fluxcloud/
-#additionalArgs:
+# for https://github.com/justinbarrick/fluxcloud/
```
hiddeco:
```suggestion
# For https://github.com/justinbarrick/fluxcloud/
```

To be in line with all other comments in this file. Makes everything a tiny bit prettier.

PR #1908, opened 2019-04-08
Verify scanned keys in same build step as scan

squaremo:
ssh-keyscan can get keys which then fail verification (e.g., if it misses out a host for some reason), but nonetheless Docker will cache the intermediate image, which will cause subsequent builds to fail. This is a pain to recover from.

This commit changes the Dockerfiles such the verification is done in the same step as the keyscan -- so if it fails, the intermediate image won't be cached, and subsequent builds will do the keyscan again.


2opremio:
> ssh-keyscan can get keys which then fail verification

I wonder why ssh-keyscan doesn't exit with non-zero when that happens. [It seems to be by design](https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1661745)

PR #1909, opened 2019-04-08
Support loading values from alternative files in chart

arturo-c:
…d from git.

Addition to https://github.com/weaveworks/flux/pull/1836/files


CHANGELOG-helmop.md
```diff
@@ -1,3 +1,11 @@
+## 0.7.2
```
hiddeco: We write our changelog (and determine the next version) just before we make the release so adding this is not necessary.

integrations/helm/release/release.go
```diff
@@ -275,7 +276,7 @@ func fhrResourceID(fhr flux_v1beta1.HelmRelease) flux.ResourceID {
 
 // values tries to resolve all given value file sources and merges
 // them into one Values struct. It returns the merged Values.
-func values(corev1 k8sclientv1.CoreV1Interface, ns string, valuesFromSource []flux_v1beta1.ValuesFromSource, values chartutil.Values) (chartutil.Values, error) {
+func values(corev1 k8sclientv1.CoreV1Interface, ns string, valuesFromSource []flux_v1beta1.ValuesFromSource, values chartutil.Values, chartPath string) (chartutil.Values, error) {
```
hiddeco:
```suggestion
func values(corev1 k8sclientv1.CoreV1Interface, ns, chartPath string, valuesFromSource []flux_v1beta1.ValuesFromSource, values chartutil.Values) (chartutil.Values, error) {
```

integrations/helm/release/release.go
```diff
@@ -352,6 +353,23 @@ func values(corev1 k8sclientv1.CoreV1Interface, ns string, valuesFromSource []fl
 				}
 				return result, fmt.Errorf("unable to yaml.Unmarshal %v from URL %s", b, url)
 			}
+		case v.ChartFileRef != nil:
+			cf := v.ChartFileRef
+			filePath := cf.Path
+			optional := cf.Optional != nil && *cf.Optional
+			f, err := readLocalFile(fmt.Sprintf("%s/%s", chartPath, filePath))
```
hiddeco: I think it would be cleaner to change this to `readLocalChartFile(chartPath, filePath)`, and move the construction of the path there (so it actually does what the comment tells it does).

`// readLocalFile attempts to read a file from the chart path.`

squaremo:
```suggestion
			f, err := readLocalFile(filepath.Join(chartPath, filePath))
```

squaremo:
.oO(If we unpacked repository charts, we could support this for those as well; no need to do that in this PR though)

hiddeco:
You just officially became a Flux contributor :tada:

Thanks a lot (again) for your contribution, I think quite some people will find it useful.

PR #1912, opened 2019-04-09
Fix instructions for observing colour change

dholbach:
In #1902 it was discovered that the part of the tutorial where
	PODINFO_UI_COLOR was changed didn't actually work. So we're
	using the same instructions as in the Helm tutorial now, modulo
	the namespace (default vs flux) change.

### DIFF
--- a/site/get-started.md
+++ b/site/get-started.md
@@ -133,9 +133,10 @@ run:
 
 ```sh
 kubectl -n demo port-forward deployment/podinfo 9898:9898 &
+curl localhost:9898
 ```
 
-Open your browser and navigate to `http://localhost:9898`.
+Notice the updated `color` value in the JSON reply.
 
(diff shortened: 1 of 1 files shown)


dholbach:
cc @stefanprodan

stefanprodan:
@dholbach how did you test the color change? If you use curl it should return a json with the right color

dholbach:
> @dholbach how did you test the color change? If you use curl it should return a json with the right color

I opened it in a browser. It was still green.

dholbach:
> @dholbach how did you test the color change? If you use curl it should return a json with the right color

I opened it in a browser. It was still green.

https://github.com/weaveworks/flux/pull/1912/files#diff-15622267c2ec3db7451cdecfe25640b9L138

PR #1915, opened 2019-04-10
Make `regex` an alias of `regexp` tag filter type

hiddeco:
The word `regex` seems to be a common alias of `regexp`, so people often use this and are left with confusion when Flux does not adhere to what they configured.

Make it an alias to better live up to people their expectations.

### DIFF
--- a/policy/pattern.go
+++ b/policy/pattern.go
@@ -1,17 +1,19 @@
 package policy
 
 import (
+	"regexp"
+	"strings"
+
 	"github.com/Masterminds/semver"
 	"github.com/ryanuber/go-glob"
 	"github.com/weaveworks/flux/image"
-	"strings"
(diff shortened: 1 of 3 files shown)


policy/pattern.go
```diff
@@ -1,17 +1,19 @@
...
 	"github.com/weaveworks/flux/image"
-	"strings"
-	"regexp"
 )
 
 const (
 	globPrefix   = "glob:"
 	semverPrefix = "semver:"
 	regexpPrefix = "regexp:"
+	regexPrefix  = "regex:"
```
squaremo: These are so similar it's easy to mistake them, and that might lead to mistakes down the road -- maybe `regexpPrefix` and `regexpAltPrefix`?

PR #1919, opened 2019-04-11
Fix hardcoded docker config file

marchmallow:
This is a small change to add flexibility to the flux chart template.

The deployment template in the flux chart currently hardcodes the file name of the docker config that contains the docker secret to:

` --docker-config=/dockercfg/config.json`

When running this on the IBM Cloud Kubernetes service that config filename ends up being different which requires to manually change the yaml post deployment.

This change creates a value parameter for it which defaults to current hardcoded value but allows changing this when we installing flux as per other value parameters.

Tested this by deploying flux to a Kubernetes cluster in IBM Cloud with kubernetes version 1.12, I did not manage to run 'make test' successfully not sure where/if there are docs or tips on running it.


marchmallow:
Thanks for the review @stefanprodan just updated it.

PR #1931, opened 2019-04-12
Harden and simplify yaml stream decoding/encoding

### LINKED ISSUE #1925: Weave cloud deploy page error with '---' at EOF
It appears that when flux syncs a yaml that has a '---' at the end of the file, the application works correctly, and the yaml is applied, but afterwards the Weave Cloud interface for deployments has the following error: 
```parsing YAML doc from "{pathToYaml}": yaml: did not find expected node content'```

When I remove the '---' at the end of the file, this error goes away.

The only effect this has is that it prevents me from doing manual deployments, and we could add validation for this, BUT this seems like a Weave specific problem that should be resolved, since it is valid kubernetes syntax.

### PR DESCRIPTION
**Pull request by 2opremio**
Flux choked on end-of-document markers (`...`).

To avoid complicating the existing multidoc parser (stolen from kubernetes) I abused `go-yaml`'s`Decoder` to obtain the raw documents from the stream by unmarshalling to an `interface{}` and marshalling again.

Also, I replaced our ad-hoc stream parsing and generation by a `go-yaml` `Decoder` and `Encoder`. 

I, however, left the encoding part of `fluxctl save` untouched since `yaml.Encoder` generates different output (no start-of-document prefix) and users may be relying on it.

During testing I made sure that changes I made in `fluxd`'s `Export()` are compatible with `fluxctl save` prior to this change.

Fixes #1925

### DIFF
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -137,6 +137,8 @@ func ParseMultidoc(multidoc []byte, source string) (map[string]KubeManifest, err
 		// from the stream, we decode generically and encode again.
 		// The result is the raw document from the stream
 		// (pretty-printed and without comments)
+		// NOTE: gopkg.in/yaml.v3 supports round tripping comments
+		//       by using `gopkg.in/yaml.v3.Node`.
 		var val interface{}
 		if err := decoder.Decode(&val); err != nil {
 			break
--- a/cluster/kubernetes/resource/load.go
+++ b/cluster/kubernetes/resource/load.go
@@ -1,14 +1,15 @@
 package resource
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 )
 
 // Load takes paths to directories or files, and creates an object set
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: cluster/kubernetes/resource/load.go [hunk 1]
```diff
@@ -159,45 +158,8 @@ func ParseMultidoc(multidoc []byte, source string) (map[string]KubeManifest, err
...
 	return objs, nil
 }
-
-// ---
-// Taken directly from https://github.com/kubernetes/apimachinery/blob/master/pkg/util/yaml/decoder.go.
-
-const yamlSeparator = "\n---"
-
-// splitYAMLDocument is a bufio.SplitFunc for splitting YAML streams into individual documents.
-func splitYAMLDocument(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
hiddeco: Although not related to this bugfix, another copy of this method can be found here: https://github.com/weaveworks/flux/blob/546ea7169eaede3da0b2193f09f263a35bec6474/cmd/fluxctl/save_cmd.go#L220

2opremio:
Uhm, I think I got something wrong. The test seems to pass without the code changes ...

cluster/kubernetes/resource/load.go
2opremio: yikes, that sucks

hiddeco: Also totally not related to this PR, we probably want to include our GC labels and annotations here: https://github.com/weaveworks/flux/blob/546ea7169eaede3da0b2193f09f263a35bec6474/cmd/fluxctl/save_cmd.go#L102

2opremio: I will fix the first part. I am not sure about the second, can you create a separate issue for it?

squaremo:
Q: Does this mean comments are no longer preserved when files are updated by flux?

squaremo:
A: no, because the kubeyaml-using code still just reads a whole file in, and writes a whole file out.

2opremio:
Yep, exactly, sorry I didn't get to answer the question in time.

PR #1935, opened 2019-04-14
Split Warmer.warm() into smaller functions

2opremio:
`Warmer.warming()` was a huge function (~250 loc), making it really hard to read.

This change splits the warmer into smaller functions, improving readability.

After this change `Warmer.warming()` is still 100 loc, but reads much better.

### DIFF
--- a/registry/cache/repocachemanager.go
+++ b/registry/cache/repocachemanager.go
@@ -1,0 +1,268 @@
+package cache
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
+
+	"github.com/weaveworks/flux/image"
+	"github.com/weaveworks/flux/registry"
+)
+
+type imageToUpdate struct {
+	ref             image.Ref
+	previousDigest  string
+	previousRefresh time.Duration
+}
+
+// repoCacheManager handles cache operations for a repository
+type repoCacheManager struct {
+	now         time.Time
+	repoID      image.Name
+	burst       int
+	trace       bool
+	logger      log.Logger
+	cacheClient Client
+	sync.Mutex
+}
+
+func newRepoCacheManager(now time.Time, repoId image.Name, burst int, trace bool, logger log.Logger,
+	cacheClient Client) *repoCacheManager {
+	return &repoCacheManager{
+		now:         now,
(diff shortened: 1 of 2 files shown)


registry/cache/repocachemanager.go
```diff
@@ -0,0 +1,268 @@
...
+			break updates
+		case fetchers <- struct{}{}:
+		}
+		awaitFetchers.Add(1)
+		go func() {
+			defer func() { awaitFetchers.Done(); <-fetchers }()
+			entry, err := c.updateImage(ctxc, registryClient, upCopy)
+			if err != nil {
+				if err, ok := errors.Cause(err).(net.Error); ok && err.Timeout() {
+					// This was due to a context timeout, don't bother logging
```
hiddeco: Does this mean that in cases of repeating failure (e.g. the image registry is down), the daemon simply will not log anything? If so, I am not sure if this is desired (in cases of repeating failure).

2opremio: I think you are right, but I think fixing that falls out of this refactoring.

hiddeco: Sorry, I noticed later that this was moved code and not new code, I agree.

2opremio: No need to apologize, let's fix that later.

registry/cache/repocachemanager.go
```diff
@@ -0,0 +1,268 @@
...
+		imagesToUpdateMissingCount: missing,
+	}
+
+	return result, nil
+}
+
+// updateImages, refreshes the cache entries for the images passed. It may not succeed for all images.
+// It returns the values stored in cache, the number of images it succeeded for and the number
+// of images whose manifest wasn't found in the registry.
+func (c *repoCacheManager) updateImages(ctx context.Context, registryClient registry.Client, images []imageToUpdate) (map[string]image.Info, int, int) {
```
hiddeco: Just a note; I am not a fan of the `.., int, int` return, but wrapping it in a struct to give it more context does not seem very meaningful either, the comment makes up for it a bit though.

2opremio: Yeah, I am not too pleased with `fetchImagesResult` either

Happy to look into any suggestions you may have.

Maybe computing the containers outside of the functions? I gave that a short chance but it was hairy. I will give it another try.

2opremio: I didn't find a cleaner way. At least not with a much bigger refactoring.

PR #1937, opened 2019-04-15
Warn user about non-existing configured branch

### LINKED ISSUE #904: Empty config repo with no commits gives git.status = new and leaves user confused.
This is pretty edge casey but confusing if you're following a tutorial and don't completely understand what you're doing yet and so create a blank repo and end up in a bit of un-helpful-error-message state.

```sh
$ git clone git@github.com:foot/emptier-repo.git
Cloning into 'emptier-repo'...
warning: You appear to have cloned an empty repository.
$ echo $?
0
$ git tag zing
fatal: Failed to resolve 'HEAD' as a valid ref.
$ echo $?
128
```

- Anywho, I guess ideally we should create a commit we can tag.
- Or ... an error to the user saying empty repo's are no good.

### LINKED ISSUE #1885: Fluxd does not use configured branch for write check
When fluxd performs a checkout for a new repository it does a [write check](https://github.com/weaveworks/flux/blob/bd7a91e25f99dedaa70420ed4131ff394c066738/git/operations.go#L81) at a [certain step](https://github.com/weaveworks/flux/blob/267e0115d1a2a74956c94d37e844ffa531ce9f9c/git/repo.go#L266). This write check creates, pushes and removes a tag pointed at the default branch and not the one configured by the user with `--git-branch`.

I think it would be better if we would perform the write check against the configured branch. As we now (temporary) touch something we are not explicitly told to look at and it also has a chance of (unrelated) failure, for example, when the default branch contains no commits and Flux will be unable to get a `HEAD` revision (`caller=loop.go:90 component=sync-loop err="git repo not ready: tag for write check: fatal: Failed to resolve 'HEAD' as a valid ref."`).

hiddeco:
When a user tried to configure Flux against an empty repository or a
branch that did not exist we would report a bogus status and left the
user confused.

This commit makes it possible to configure a branch that should be
checked for existence on creation of a new repository. In case there
is no HEAD found for this branch, it does not exist in git terms, and
we report back a meaningful error message to the user.

As a bonus we also use this configured branch when we create a tag for
our write check, so we do not touch anything the user has not told us
to.

Fixes #904, bonus fixes #1885

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -485,7 +485,7 @@ func main() {
 		SkipMessage: *gitSkipMessage,
 	}
 
-	repo := git.NewRepo(gitRemote, git.PollInterval(*gitPollInterval), git.Timeout(*gitTimeout))
+	repo := git.NewRepo(gitRemote, git.PollInterval(*gitPollInterval), git.Timeout(*gitTimeout), git.Branch(*gitBranch))
 	{
 		shutdownWg.Add(1)
 		go func() {
--- a/git/operations.go
+++ b/git/operations.go
@@ -78,9 +78,12 @@ func checkout(ctx context.Context, workingDir, ref string) error {
 // checkPush sanity-checks that we can write to the upstream repo
 // (being able to `clone` is an adequate check that we can read the
 // upstream).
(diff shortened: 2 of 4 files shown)


2opremio:
Thanks!

PR #1943, opened 2019-04-16
Exclude resources if we cannot find their scope

### LINKED ISSUE #1941: Cope when syncing a helm-provided CRD and resources using the CRD
#1825 describes an issue where a commit cannot be synced if it contains both a CRD and a resource that uses that CRD. A fix was implemented in #1876 to look in the git repo for the manifest, however, this does not work if the CRD is installed via helm.

To reproduce:

* Have a git repository containing a `HelmRelease` for [rbac-manager](https://github.com/reactiveops/rbac-manager) and an `RBACDefinition` manifest
* Create a new Kubernetes cluster
* Install flux (v1.12.0) and flux-helm-operator

Expected result:

Flux applies the `HelmRelease`, which creates the `RBACDefinition` CRD, which allows Flux to apply the `RBACDefinition` manifest. In flux <= 1.10.1 this would take multiple sync passes, but the cluster would eventually reach the desired state.

Actual result:

After cloning the git repo, flux writes the following message to the log, and does not sync anything:

`s=2019-04-15T19:43:02.307951117Z caller=images.go:24 component=sync-loop error="getting unlocked automated resources: error looking up API resources for RBACDefinition.rbacmanager.reactiveops.io/v1beta1: not found"`

Since the CRD is [part of the rbac-manager helm chart](https://github.com/reactiveops/charts/blob/master/stable/rbac-manager/templates/customresourcedefinition.yaml) the code path added in #1876 doesn't seem to find it.

2opremio:
Fixes #1941 

This is aiming at custom resources whose scope is unknown because its CRD is included in a helm chart Flux won't inspect (since it's the helm operator's responsibility to do it).

This allows Flux to move forward and create the `HelmRelease` supplying the CRD. In a subsequent sync, once we know its scope, the custom resources will be created.

I was on the fence on whether I should revert #1876 but I decided against it, since it's not a lot of code (mostly reusable test-code) and it's less confusing for the user to obtain the custom resource scope locally if we have it.

### DIFF
--- a/cluster/kubernetes/manifests.go
+++ b/cluster/kubernetes/manifests.go
@@ -1,6 +1,10 @@
 package kubernetes
 
 import (
+	"fmt"
+	"strings"
+
+	"github.com/go-kit/kit/log"
 	"gopkg.in/yaml.v2"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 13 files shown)


cluster/kubernetes/manifests.go
```diff
@@ -60,31 +74,48 @@ func getCRDScopes(manifests map[string]kresource.KubeManifest) ResourceScopes {
 	return result
 }
 
-func setEffectiveNamespaces(manifests map[string]kresource.KubeManifest, nser namespacer) (map[string]resource.Resource, error) {
+func (m *manifests) setEffectiveNamespaces(manifests map[string]kresource.KubeManifest) (map[string]resource.Resource, error) {
 	knownScopes := getCRDScopes(manifests)
 	result := map[string]resource.Resource{}
 	for _, km := range manifests {
-		if nser != nil {
```
2opremio: I am pretty sure this check was here for the tests, so I simply made sure that we never used a `nil` namespacer.

hiddeco: Tried to confirm this by going back through history but it has seen too many shapes in the past weeks for me to get a clear view. I will trust your 'pretty sure' on this one as you have been living in this part of Flux for quite a while.

2opremio:
> Do we want to make a patch release out of this?

Yeah, I think we do.

@brantb Before we merge, can you test this image to double-check the problem is gone for you? `quay.io/fons/flux:1941-fix-crd-syncing-6f2eba28`

brantb:
@2opremio 
It's working!

2opremio:
Heheh, great

PR #1945, opened 2019-04-16
Initialize klog flags so it logs to stderr

hiddeco:
This got not caught in CircleCI as it does not perform a rebase
on-top of latest master before testing.

We bumped client-go a while ago (#1829) which included a switch to klog,
the klog flags need to be intialized before they can be set, as we did
not have this initialization in place, the operator was unable to
start. Simply initialize the flags and we are good to go.

---

I am tempted to add the following to the CircleCI config in a separate PR to prevent this in the future:

`if [[ "$CIRCLE_BRANCH" != "master"  && "$CIRCLE_BRANCH" != release*  ]]; then git fetch origin master && git rebase origin/master; fi`

### DIFF
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"os/signal"
@@ -14,6 +13,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
 
 	"github.com/weaveworks/flux/checkpoint"
 	clientset "github.com/weaveworks/flux/integrations/client/clientset/versioned"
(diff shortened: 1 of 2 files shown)


cmd/helm-operator/main.go
```diff
@@ -103,7 +104,10 @@ func init() {
 }
 
 func main() {
-	// set glog output to stderr
+	// Ensure klog flags are initialized so they can be updated,
+	// set logging of klog to stderr so it's visible and does not
+	// log to /tmp, and finally parse our own flags.
+	klog.InitFlags(nil)
```
2opremio: are we also doing this for  the other commands?

hiddeco: Not that I am aware, the only other `klog` related thing I can find in our repository is this (made by you): https://github.com/weaveworks/flux/commit/25995d57b64f7cd68c281c401bd9468dc321913c

2opremio: `klog` is just a variable name there  (i.e. it's not using the library) so it's only tangentially related.

Then, I think we should probably do it for all the commands, since `klog` is what kubernetes uses for internal error reporting, which can happen for unhandled errors. The change you mentioned makes sure we use our own logging for that in `fluxd` but I think it's still sane to initialize `klog` properly for all commands.

hiddeco: Updated, also discovered logging to `stderr` is now the default.

stefanprodan:
Hmm I wander why is this working since `InitFlags` should crash when used with `clientcmd` since the client will call it.

hiddeco:
@2opremio I am going to merge this to get master going again, if you have any additional comments I will be happy address them in another PR.

PR #1949, opened 2019-04-17
Move images to Docker Hub

hiddeco:
I choose to use the Fully Qualified Registry Name everywhere (`docker.io/weaveworks/flux`) instead of using the shorthand (`weaveworks/flux`) so it gets picked up when someone is for example excluding `docker.io` images using `--registry-exclude-image=docker.io/*`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,9 +19,9 @@ jobs:
           name: Maybe push master image
           command: |
             if [ -z "${CIRCLE_TAG}" -a "${CIRCLE_BRANCH}" == "master" ]; then
-              docker login -u "$DOCKER_REGISTRY_USER" -p "$DOCKER_REGISTRY_PASSWORD" quay.io
-              docker push "quay.io/weaveworks/flux:$(docker/image-tag)"
-              docker push "quay.io/weaveworks/helm-operator:$(docker/image-tag)"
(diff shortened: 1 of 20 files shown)


.circleci/config.yml
```diff
@@ -19,9 +19,9 @@ jobs:
           name: Maybe push master image
           command: |
             if [ -z "${CIRCLE_TAG}" -a "${CIRCLE_BRANCH}" == "master" ]; then
-              docker login -u "$DOCKER_REGISTRY_USER" -p "$DOCKER_REGISTRY_PASSWORD" quay.io
-              docker push "quay.io/weaveworks/flux:$(docker/image-tag)"
-              docker push "quay.io/weaveworks/helm-operator:$(docker/image-tag)"
+              echo "$DOCKER_REGISTRY_PASSWORD" | docker login --username "$DOCKER_REGISTRY_USER" --password-stdin
```
bboreham: `--password-stdin` is new in Docker 1.18; do you know if this image has a version that new?

hiddeco:
```
❯ docker run -it --rm circleci/golang:1.10 sh
$ docker login --help

Usage:	docker login [OPTIONS] [SERVER]

Log in to a Docker registry

Options:
  -p, --password string   Password
      --password-stdin    Take the password from stdin
  -u, --username string   Username
```

:heavy_check_mark:

PR #1965, opened 2019-04-23
chart: add support for deploying a global Git configuration.

bashims:
Hi,

The following PR adds the ability to deploy a custom Git configuration from a ConfigMap that flux and helm-operator will use during any Git operations. This PR is part of forthcoming PR that adds AWS CodeCommit IAM auth support via a configured Git credential helper. Aternatively it might make more sense to store the configuration as secret.

As this is my first time submitting a PR for this project I am not entirely sure how these changes are meant to be tested. I looked for Helm tests but did not find any. Any pointers in this regard are appreciated.

Thanks,

-ben


stefanprodan:
Hi @bashims 

The Helm chart is being tested with Kubernetes KinD see https://github.com/weaveworks/flux/blob/master/test/e2e/e2e-flux-chart.sh

I would use a secret instead of a config map for git config. Please take a look at how the docker config is being setup https://github.com/weaveworks/flux/blob/master/chart/flux/values.yaml#L165

PR #1967, opened 2019-04-23
Fix insecure-host-checking for repos with an explicit port

2opremio:
Before this change we included the repository port when comparing against the list of insecure hosts.

### DIFF
--- a/registry/client_factory.go
+++ b/registry/client_factory.go
@@ -99,8 +99,8 @@ func (f *RemoteClientFactory) ClientFor(repo image.CanonicalName, creds Credenti
 	}
 	insecure := false
 	for _, h := range f.InsecureHosts {
+		// allow the insecure hosts list to contain hosts with or without the port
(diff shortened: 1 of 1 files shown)


registry/client_factory.go
```diff
@@ -92,9 +93,14 @@ attemptChallenge:
...
 func (f *RemoteClientFactory) ClientFor(repo image.CanonicalName, creds Credentials) (Client, error) {
+	repoHost, _, err := net.SplitHostPort(repo.Domain)
+	if err != nil {
+		return nil, err
+	}
 	insecure := false
 	for _, h := range f.InsecureHosts {
-		if repo.Domain == h {
+		if repoHost == h || repo.Domain == h {
+			// allow host with out without the port in the insecure hosts list
```
hiddeco:
```suggestion
			// allow host with or without the port in the insecure hosts list
```

PR #1973, opened 2019-04-24
Bump kubeyaml to version 0.5.2

2opremio:
Needed for https://github.com/weaveworks/flux/issues/1925


hiddeco:
I think https://github.com/weaveworks/flux/blob/master/bin/kubeyaml also needs an update.

2opremio:
Yep, I will rebuild once that's done

2opremio:
Oh, true

squaremo:
> I will rebuild once that's done

The script in bin/kubeyaml needs the version updated as well

PR #1989, opened 2019-04-25
RBAC improvements

rocktavious:
Remove the need for defining a custom cluster role and just bind to the k8s existing RBAC role for cluster admin

### DIFF
--- a/deploy/flux-account.yaml
+++ b/deploy/flux-account.yaml
@@ -9,28 +9,15 @@ metadata:
   name: flux
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRole
-metadata:
-  labels:
-    name: flux
-  name: flux
-rules:
-  - apiGroups: ['*']
-    resources: ['*']
-    verbs: ['*']
-  - nonResourceURLs: ['*']
-    verbs: ['*']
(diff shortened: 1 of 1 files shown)


rocktavious:
I don't know helm so i'm not sure where to make the change for the helm specific deployment.

rocktavious:
I beleive the source of this cluster role lives here - https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml#L17

stefanprodan:
In my opinion this is bad practice from a security perspective, each operator should define its own role so that it can be audited and restricted easily. We already have restrictions in place when deploying with Helm, for example you can limit Flux and the Helm Operator role to a single namespace.

rocktavious:
@stefanprodan the reason i propose this is because the current k8s manifest do not apply when you throw them at a GKE cluster because you cannot make a cluster role with * for all permissions.  If flux's default cluster role could be written to be better scoped then i would agree with you - but right now it doesn't seem like thats the case (or ever would be the case)

So if flux truely needs * permissions to k8s API - then you need to hook it up against the existing ADMIN cluster role.

Issue #1799, opened 2019-03-06
Use image build time label if present


squaremo:
To work around situations like the one in #1797, and mitigate things like #746 and #891, we could prefer the image build time label, if present, rather than looking at the created time.

hiddeco:
Standard (community) labels:

http://label-schema.org/rc1/#build-time-labels
https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

Issue #746, opened 2017-09-18
Confusing display of image age


bboreham:
Quay says the image was last modified 12 minutes ago; Flux says 26 minutes ago.

![image](https://user-images.githubusercontent.com/8125524/30537566-0e3bd594-9c62-11e7-8607-b1bfb12873d9.png)

![image](https://user-images.githubusercontent.com/8125524/30537593-2d8c9c26-9c62-11e7-94e7-562ee2402e77.png)

I suspect one is the time the image was stamped and the other is the time it was uploaded, since those events are some time apart in the build of this particular image.

squaremo:
We don't in general have access to the time the image was uploaded.

PR #1992, opened 2019-04-29
Make it possible to overwrite image TS with label

### LINKED ISSUE #891: Image timestamps are incorrect
I'm not sure if this is a Docker registry issue or a Flux issue, but the "created" timestamp seems to be based off of the image layers. This means that if a newer build ends up containing the same content as an older build, it will appear to be an older image and Flux will not upgrade.

hiddeco:
This adds support for overwriting the image created at timestamp with labels which are set during build. Supported labels (for now) are the [Open Container Image (OCI) spec](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys) and the [(legacy) Label Schema (LS) spec](http://label-schema.org/rc1/#build-time-labels).

We prioritize OCI over LS, with a fallback to the CreatedAt from the registry.

Fixes #891, addresses #1799, #1797, #746

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -223,6 +223,52 @@ func (i Ref) WithNewTag(t string) Ref {
 	return img
 }
 
+// Labels has all the image labels we are interested in for an image
+// ref, the JSON struct tag keys should be equal to the label.
+type Labels struct {
+	// BuildDate holds the Label Schema spec 'build date' label
+	// Ref: http://label-schema.org/rc1/#build-time-labels
+	BuildDate time.Time `json:"org.label-schema.build-date,omitempty"`
+	// Created holds the Open Container Image spec 'created' label
+	// Ref: https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+	Created   time.Time `json:"org.opencontainers.image.created,omitempty"`
+}
+
+// MarshalJSON returns the Labels value in JSON (as bytes). It is
+// implemented so that we can omit the time values when they are
+// zero, which would otherwise be tricky for e.g., JavaScript to
+// detect.
+func (l Labels) MarshalJSON() ([]byte, error) {
+	var bd, c string
+	if !l.BuildDate.IsZero() {
+		bd = l.BuildDate.UTC().Format(time.RFC3339Nano)
+	}
+	if !l.Created.IsZero() {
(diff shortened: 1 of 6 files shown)


hiddeco:
As this PR changes behaviour (by prioritizing labels over the `CreatedAt` date from the registry) and is more a feature than a bugfix, I marked this as a `v1.13.0` milestone.

image/image.go
```diff
@@ -223,6 +223,52 @@ func (i Ref) WithNewTag(t string) Ref {
...
+		c = l.Created.UTC().Format(time.RFC3339Nano)
+	}
+	encode := struct {
+		BuildDate string `json:"org.label-schema.build-date,omitempty"`
+		Created   string `json:"org.opencontainers.image.created,omitempty"`
+	}{BuildDate: bd, Created: c}
+	return json.Marshal(encode)
+}
+
+// UnmarshalJSON populates a Labels from JSON (as bytes). It's the
```
2opremio:
```suggestion
// UnmarshalJSON populates Labels from JSON (as bytes). It's the
```

image/image.go
```diff
@@ -281,6 +329,19 @@ func (im *Info) UnmarshalJSON(b []byte) error {
 	return err
 }
 
+// CreatedTS returns the created at timestamp, prioritizing timestamps
+// from labels (set during build) over the created at from the Docker
+// registry.
```
2opremio: I think it would be good to explain why we are choosing that priority

image/image.go
```diff
@@ -223,6 +223,52 @@ func (i Ref) WithNewTag(t string) Ref {
 	return img
 }
 
+// Labels has all the image labels we are interested in for an image
+// ref, the JSON struct tag keys should be equal to the label.
+type Labels struct {
```
2opremio: No strong opinion, but wouldn't it be more flexible to make this a map? (in case we end up adding more labels in the future).

I guess we can always make it a map in the future without breaking the JSON representation, right?

hiddeco: > I guess we can always make it a map in the future without breaking the JSON representation, right?

This is just a JSON object with key/values, so yeap, no worries.

2opremio: Makes sense

2opremio:
LGTM

PR #1994, opened 2019-04-29
Fix error shadowing when parsing manifests

2opremio:
This caused parsing errors in YAML files to go unnoticed (the files where simply
not included in the result)


2opremio:
Introduced in #1931

2opremio:
Uhm, I should probably add a test. Will do it tomorrow

PR #2005, opened 2019-05-01
Small sync fixes Helm operator

hiddeco:
This PR does two things:

- it fetches `HelmRelease`s before applying status updates
- it detects and ignores `HelmRelease` status updates while determining if it should proceed with the upgrade of a `HelmRelease`

### DIFF
--- a/integrations/helm/operator/operator.go
+++ b/integrations/helm/operator/operator.go
@@ -282,8 +282,19 @@ func (c *Controller) enqueueUpdateJob(old, new interface{}) {
 		return
 	}
 
-	log := []string{"info", "enqueuing release upgrade"}
-	if diff := cmp.Diff(oldFhr.Spec, newFhr.Spec); diff != "" && c.logDiffs {
+	diff := cmp.Diff(oldFhr.Spec, newFhr.Spec)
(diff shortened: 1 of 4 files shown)


integrations/helm/operator/operator.go
```diff
@@ -282,8 +282,17 @@ func (c *Controller) enqueueUpdateJob(old, new interface{}) {
 		return
 	}
 
-	log := []string{"info", "enqueuing release upgrade"}
-	if diff := cmp.Diff(oldFhr.Spec, newFhr.Spec); diff != "" && c.logDiffs {
+	diff := cmp.Diff(oldFhr.Spec, newFhr.Spec)
+
+	// If the spec did not change, but the status did, we are not
+	// interested in doing an expensive dry-run to detect changes,
+	// as there probably are none.
```
squaremo: Why do we can that the status _did_ change?

If it's to filter out update notifications that are due to status updates _without_ also filtering out updates that are from the periodic refresh (which is my guess), I would say that in the comment.

hiddeco: Reworded.

PR #2006, opened 2019-05-02
Provide optional rollback support for HelmReleases

### LINKED ISSUE #1960: Please provide "helm rollback" option in the HelmRelease
When rolling out new versions of a HelmRelease, there is a case that a release might be marked as `FAILED` by Helm due to some error. In that case, Flux cannot deploy new versions of the `HelmRelease` file, even if the error has been corrected, as the following error appears in the logs:

```
ts=2019-03-15T16:54:32.871180927Z caller=release.go:186 component=release error="Chart release failed: staging-test: &status.statusError{Code:2, Message:\"a release named staging-test already exists. nRun: helm ls --all staging-test; to check the status of the release nOr run: helm del --purge staging-test; to delete it\", Details:[]*any.Any(nil)}"
```

In order to overcome this issue, there are two options without downtime that I am aware of:
1. Create a new HelmRelease with a different name, let it deploy, delete the old one.
2. Do a `helm rollback`

Given that the last revision of Helm will be in `FAILED` state, the one that will be running in the cluster will be the last successful one. The `helm rollback <release_name> <last_successful_revision>` will just make Helm point to the last successful revision so that we overcome this issue.

Regarding the two available options above, the first one involves some manual labor, where the second one can be automated. The downside of the second option is that `helm rollback` is not stable for stateful applications. 

Could we please implement it as an option that one has to explicitly enable for Flux to do the `helm rollback` if this issue is hit? This will be backwards compatible, not create a problem to stateful workloads and allow someone that wants to force this rollback to do so. This will be very handy for stateless application where this issue is hit.

Thanks!

hiddeco:
Fixes #1960 

image


2opremio:
@hiddeco Won't the rollback mechanism change based on how https://github.com/weaveworks/flux/issues/1960#issuecomment-488664697 is resolved? If it can differ, then I think it's better to review it once we have a design.

> Given the feature is opt-in, maybe we can live with it for a while?

Uhm, I am not sure an infinite loop is scary

hiddeco:
@2opremio @squaremo took me some time to get the mechanics sorted, but here it is.

The operator now keeps track of what generation of a resource it has seen, this is done by updating the `ObservedGeneration` status after it has performed a non-deleting operation, it also adds a status condition for rollbacks, and keeps track of both condition change _and_ update timestamps.

In case a rollback status condition is set, there are three options:

1. If the `generation > observedGeneration`, we attempt an upgrade as the `values` may have changed.
2. If the chart has been rolled back, but we have observed an update of the Helm chart source after this happened (`LastUpdateTime`), we attempt an upgrade as the contents of the chart source may have changed (or the failure is old).
3. We bail and log a warning to the user that we skipped the release.

I am still on the fence about two scenarios:

1. In case the Helm operator is restarted, it will attempt an upgrade and rollback afterwards. I _think_ this is because it gets added to the queue through the `AddFunc` event handler on the informer, which does not perform any state checks at the moment, need to investigate...
2. If the rollback was due to the contents of any `valuesFrom` source, we are unable to determine this has been fixed, as any changes in those sources do not result in a new resource generation and we resolve and merge those values JIT for the upgrade we want to perform, which is much later in the process.

2opremio:
> 2\. If the rollback was due to the contents of any `valuesFrom` source, we are unable to determine this has been fixed, as any changes in those sources do not result in a new resource generation and we resolve and merge those values JIT for the upgrade we want to perform, which is much later in the process.

Can't we store the `valuesFrom` in the status?

hiddeco:
We could store a SHA hash of the merged values in a status (someone on Slack suggested something like this too). But to be able to make a comparison and detect a change, this would require us to resolve and merge all values before we schedule an upgrade (in `controller.go`), which does not look like a good solution to me.

integrations/helm/chartsync/chartsync.go
```diff
@@ -404,13 +406,31 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
...
+	if !fhr.Spec.Rollback.Enable {
+		return
+	}
+
+	name := release.GetReleaseName(fhr)
+	err := chs.release.Rollback(name, fhr.Spec.Rollback.GetTimeout(), fhr.Spec.Rollback.Force,
+		fhr.Spec.Rollback.Recreate, fhr.Spec.Rollback.DisableHooks, fhr.Spec.Rollback.Wait)
+	if err != nil {
+		chs.logger.Log("warning", "unable to rollback chart release", "resource", fhr.ResourceID().String(), "release", name, "err", err)
+		chs.setCondition(fhr, fluxv1beta1.HelmReleaseRolledBack, v1.ConditionFalse, ReasonRollbackFailed, err.Error())
```
2opremio: Maybe, only maybe, create variables for the reason and message and call `chs.setCondition` only once?

integrations/helm/release/release.go
```diff
@@ -112,11 +112,29 @@ func (r *Release) GetUpgradableRelease(name string) (*hapi_release.Release, erro
 	}
 }
 
+func (r *Release) mustRollback(name string) (bool, error) {
```
2opremio: I would call it `shoulRollback` since, by convention, `must` is used as a prefix in Golang for something which will panic if doesn't succeed.

2opremio:
Code looks good, but I don't know enough about the Helm Operator to rubber-stamp this.

I am missing tests though (which I guess you planned to add later)

hiddeco:
> I don't know enough about the Helm Operator to rubber-stamp this

It may be an idea to pick up some of the `FIXME`/`TODO`s in ` integrations/helm/chartsync/chartsync.go` and `integrations/helm/release/release.go` in the (near) future to prevent it from becoming heavy opinionated by my views and to spread some knowledge amongst maintainers.

The operator is not _that_ complex and quite small as a whole, so you should be able to fit it into your head rapidly. (I was able to do this and I am not the fastest person on earth in Golang code bases at the moment).

2opremio:
Thanks @hiddeco , I will take a look at those once I am done with the `e2e` test refactoring.

integrations/helm/release/release.go
```diff
@@ -171,43 +172,33 @@ func (r *Release) canDelete(name string) (bool, error) {
 // TODO(michael): cloneDir is only relevant if installing from git;
 // either split this procedure into two varieties, or make it more
 // general and calculate the path to the chart in the caller.
-func (r *Release) Install(chartPath, releaseName string, fhr flux_v1beta1.HelmRelease, action Action, opts InstallOptions, kubeClient *kubernetes.Clientset) (*hapi_release.Release, error) {
+func (r *Release) Install(chartPath, releaseName string, fhr flux_v1beta1.HelmRelease, action Action, opts InstallOptions, kubeClient *kubernetes.Clientset) (*hapi_release.Release, string, error) {
```
hiddeco: My initial thought was to just calculate the checksum based on the `*hapi_release.Release` we return here, during my tests I discovered however that it is possible to receive a `nil, err` while the release has been rolled out (but it failed and will be rolled back). We still need the checksum of this release that is going to be rolled back as otherwise our state check will fail.

integrations/helm/release/release.go
```diff
@@ -254,16 +245,16 @@ func (r *Release) Install(chartPath, releaseName string, fhr flux_v1beta1.HelmRe
 
 		if err != nil {
 			r.logger.Log("error", fmt.Sprintf("Chart upgrade release failed: %s: %#v", fhr.Spec.ReleaseName, err))
-			return nil, err
+			return nil, ValuesChecksum(rawVals), err
```
hiddeco: As described above, we may return an error here while the release has been moved forward.

hiddeco:
@2opremio @squaremo it does now detect changes to external values files, _how_ it does this is up for debate.

I personally am not very happy with it, i.e. using the `*ChartChangeSync` in the `enqueueUpdateJob` method feels wrong (decision making should be fast), but given that we need to resolve all external files this was the only entrypoint I could think of, and as we only do this for HelmReleases with rollbacks enabled the overhead should be minimal.

integrations/helm/status/status.go
```diff
@@ -101,25 +85,109 @@ bail:
...
+	}
+
+	rolledBack := GetCondition(hr.Status, v1beta1.HelmReleaseRolledBack)
+	if rolledBack == nil {
+		return false
+	}
+
+	chartFetched := GetCondition(hr.Status, v1beta1.HelmReleaseChartFetched)
+	if chartFetched != nil {
+		return !(chartFetched.Status == v1.ConditionTrue && rolledBack.LastUpdateTime.Before(&chartFetched.LastUpdateTime))
```
squaremo: Can you explain (in a comment) why `LastUpdateTime` is necessary here, rather than `LastTransitionTime`?

integrations/helm/status/status.go
```diff
@@ -101,25 +85,109 @@ bail:
...
+	}
+
+	return rolledBack.Status == v1.ConditionTrue
```
squaremo: This func reads to me as

 1. If we haven't processed the resource since it was updated, let that happen because things might change
 2. If there's no RolledBack condition at all, it can't have been attempted, so not in a rolled back state
 3. If the chart was fetched more recently than the rollback was attempted, then things might change; but if not, it's definitely in a rollback state
 4. Otherwise, it's in a rollback state if the last rollback succeeded

In other words, the purpose is to detect if there's any reason to think an upgrade could succeed.

If my reading is accurate, there's a problem: in 3.), this depends _only_ on whether the fetch condition is newer than the rollback condition; if it's _not_ newer, I would expect the rollback condition to be consulted (as in 4.)

hiddeco: Your observation was correct and I have adjusted the logic.

I have a draft of a state machine diagram, will add this to the PR once it's finished.

adrian:
I've been testing this branch and came across a problem with StatefulSets.

If an upgrade involves recreating the pods in a StatefulSet then Kubernetes will, as normal, start replacing the pods starting at the last ordinal, e.g.
```
acme-statefulset-0                 1/1     Running            0          5m26s
acme-statefulset-1                 1/1     Running            0          5m21s
acme-statefulset-2                 0/1     ContainerCreating   0          1s
```
If that pod fails to come up (e.g. the image is missing) the release is rolled back but the problem pod is left as-is, i.e.
```
acme-statefulset-0                 1/1     Running            0          5m26s
acme-statefulset-1                 1/1     Running            0          5m21s
acme-statefulset-2                 0/1     ImagePullBackOff   0          33s
```
The problem is addressed by including `recreate: true` in the HelmRelease as all pods in the release are deleted as part of the rollback. However, this seems heavy handed as it means all pods in the StatefulSet and any other workloads created by the helm chart are also deleted.

Is there a way to address this?

Could the `--cleanup-on-fail` flag in Helm v2.14.0 be a solution?

hiddeco:
@adrian given that StatefulSets are known to be problematic to rollback (as mentioned in the issue linked in this PR), and the fact that we mimic `helm upgrade --atomic`. I am afraid this has little to do with the Helm operator itself, but more with Helm.

I am however interested in any solution you may come across which solves the problem for `helm`, so I can include the required flag(s) and/or add the solution to our docs. My advice to you would be to see if you can make it work with `helm upgrade --atomic <maybe --cleanup-on-fail> <or --recreate-pods>` (or maybe a different strategy).

adrian:
In relation to my question earlier about dealing with failed pods in StatefulSets...

`--cleanup-on-fail` isn't an option as it only deletes net new pods, i.e. pods for new workloads rather than updated pods.

`--recreate` recreates all pods belonging to all workloads deployed by the helm chart. This works, but leads to service downtime as the pods are being recreated.

One option we've implemented (not in an operator) is to delete all broken StatefulSet pods after issuing the  helm rollback command. Maybe that's something that the operator could do?

hiddeco:
> FWIW I've been testing the operator on this branch for the past few days and it's working fine.

Awesome, going to rebase and tidy things up and bother @squaremo with reviewing again so it lands in `master`.

> One option we've implemented (not in an operator) is to delete all broken StatefulSet pods after issuing the helm rollback command. Maybe that's something that the operator could do?

I am not very keen on implementing 'hacks' like this in the operator, the reason for this is that it may work for you but breaks things for other people. Instead, and this is probably not to your likening, I would just make note of that it is likely to give issues with some `StatefulSet` setups, due to shortcomings in Helm.

semyonslepov:
@hiddeco this is awesome, thanks a lot for implementing this feature!
Would you mind also updating the docs on Helm integration: https://github.com/weaveworks/flux/blob/master/site/helm-integration.md ?

hiddeco:
@semyonslepov thanks for your enthusiasm and interest in this feature, I merged a PR yesterday which [documents the added fields to the `HelmRelease`](https://github.com/weaveworks/flux/blob/8292179855e15370fb3d3b03135a61b54f00ae42/site/helm-integration.md#rollbacks).

Last, if you do not want to wait on the next Helm operator release, there is a [prerelease build](https://hub.docker.com/r/fluxcd/helm-operator-prerelease/tags) available which also adds experimental support for running multiple workers processing your releases (`--workers=<num>`).

PR #2007, opened 2019-05-02
Add helm tag filters examples

stefanprodan:


hiddeco:
I think the section you edited and the one above it should be merged into one, as they contain (parts of) the same information.

stefanprodan:
@hiddeco I've merged the two sections

2opremio:
Nice! This will hopefully make it more clear for users. Thanks @stefanprodan !!!

PR #2010, opened 2019-05-03
Add a line to docs

xtellurian:
This is a very small PR, but one that I hope saves others.

* Adding an explicit warning to users that Flux ignores JSON files.


squaremo:
@xtellurian Can you rebase onto the latest head of master, and `git push --force-with-lease` please? There shouldn't be any conflicts. (Usually I would just do this and push to the source branch of a PR, but you've created this from your master, so I'm reluctant to do so in this case.)

PR #2014, opened 2019-05-07
Mount sub path of repositories secret

srueg:
To prevent the emptyDir for the repositories cache from overwriting the
repositories.yaml since it's mounted on the same path.
Courtesy of https://blog.sebastian-daschner.com/entries/multiple-kubernetes-volumes-directory


chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.12.1"
-version: 0.9.2
+version: 0.9.3
```
stefanprodan: Please undo this change, the version bump will come later.

PR #2018, opened 2019-05-07
Fix (Flux)HelmRelease cluster lookups

### LINKED ISSUE #2011: incorrect workload names in flux
After some researches and testing I figured out that all helm releases are named the same in flux and I can’t figure out why.
```
default:helmrelease/redis                     application
default:helmrelease/redis                     application
default:helmrelease/redis                     application
default:helmrelease/redis                     application
default:helmrelease/redis                     application
default:helmrelease/redis                     application
default:helmrelease/redis                     chart-image
```
the container `application` is obvious, because all image keys are under application key. Thats correct.
But depending on what deploys last the workload is named `default:helmrelease/redis`. When I delete redis the workload is `default:helmrelease/fluxcloud`.

```
---
kind: HelmRelease
metadata:
  name: some-app
  namespace: default
  annotations:
    flux.weave.works/tag.chart-image: glob:master-*
spec:
  releaseName: some-app
  chart:
    git: git@1.2.3.4:/repository/repo.git
    path: charts/some-chart
    ref: master
  values:
    application:
      image: eu.gcr.io/project-name/some-app:master-123
    other:
      settings: {}
```

This is how a typical helm release looks like in our case. Any ideas what is wrong or did I find a bug in flux? (edited) 

**Setup: **
Kubernetes 1.12.7-gke.10
Helm v2.12.2
flux 1.12.1
fluxctl 1.12.0
flux-helm-operator 0.8.0

**Reproduce:**
1. create 2 helmReleases
1.1. Call one faulty the other faulty2
1.2. Add them to the repository
1.3. Do `$ fluxctl sync`
2. Do `$ fluxctl list-workloads | grep helmrelease`

**Error**
The output of `$ fluxctl list-workloads | grep helmrelease`
```
default:helmrelease/faulty2                                                                          
default:helmrelease/faulty2
```

**Expected behavior**
The workload names should match the release names. `$ fluxctl list-workloads | grep helmrelease` should produce something like
```
default:helmrelease/faulty                                                                         
default:helmrelease/faulty2
```

2opremio:
Fixes #2011 (Introduced in https://github.com/weaveworks/flux/pull/1111)

The value obtained by `_, value := range slice` is overwritten on every iteration. Thus, saving a pointer to `value` is dangerous since its content will be overwritten in the next iteration.

From https://golang.org/ref/spec#For_range :

> For each entry it assigns iteration values to corresponding
> iteration variables if present and then executes the block.

But the iteration variables are the same on each iteration.

More explicitly, from [the gccgo source code](https://github.com/golang/gofrontend/blob/e387439bfd24d5e142874b8e68e7039f74c744d7/go/statements.cc#L5593-L5604):

```
  // The loop we generate:
  //   len_temp := len(range)
  //   range_temp := range
  //   for index_temp = 0; index_temp < len_temp; index_temp++ {
  //           value_temp = range_temp[index_temp]
  //           index = index_temp
  //           value = value_temp
  //           original body
  //   }
```

### DIFF
--- a/cluster/kubernetes/resourcekinds.go
+++ b/cluster/kubernetes/resourcekinds.go
@@ -397,8 +397,8 @@ func (fhr *fluxHelmReleaseKind) getWorkloads(c *Cluster, namespace string) ([]wo
 	}
 
 	var workloads []workload
-	for _, f := range fluxHelmReleases.Items {
-		workloads = append(workloads, makeFluxHelmReleaseWorkload(&f))
+	for i, _ := range fluxHelmReleases.Items {
+		workloads = append(workloads, makeFluxHelmReleaseWorkload(&fluxHelmReleases.Items[i]))
 	}
 
 	return workloads, nil
@@ -459,8 +459,8 @@ func (hr *helmReleaseKind) getWorkloads(c *Cluster, namespace string) ([]workloa
 	}
 
 	var workloads []workload
-	for _, f := range helmReleases.Items {
-		workloads = append(workloads, makeHelmReleaseWorkload(&f))
+	for i, _ := range helmReleases.Items {
+		workloads = append(workloads, makeHelmReleaseWorkload(&helmReleases.Items[i]))
 	}
 
 	return workloads, nil


2opremio:
> Fix looks good to go to me (and looking for this must have been fun).

It didn't take that long, I had a similar problem in #1848 and went directly to the loop.

> I am now wondering if we do this in other places though, guess it is a matter of time before we find out.

I checked the other resources and are ok. I wonder if `go vet` and friends check this kind of stuff (it's similar to  the `loopclosure` check).

squaremo:
I agree that using the address of a loopvar is often going to lead to problems. Can you explain what was the problem in this specific case? It looks to me like the make<Whatever>Workload procedures don't hold on to the pointer, they just extract what they need from it before returning -- so presumably the intent was to avoid passing a large struct by value (and the safest fix would be to pass them by value).

squaremo:
> Can you explain what was the problem in this specific case?

Ah, the intermediate `workload` struct has an embedded field, and that holds on to the pointer. Ugh

hiddeco:
Looks like this did not fix the issue described in #2011. Output from `fluxctl` build fresh from current `master`:

```console
$ fluxctl list-workloads
WORKLOAD                               CONTAINER           IMAGE                               RELEASE  POLICY
<snip>
default:helmrelease/faulty2                                                                             
default:helmrelease/faulty2
$ kubectl get helmrelease
NAME      AGE
faulty    5d21h
faulty2   5d19h

**Comment by 2opremio** [comment]
Interesting, I did test the fix and `fluxctl list-worloads` fixed the issue. @hiddeco can you double-check?

**Comment by 2opremio** [comment]
> Ah, the intermediate workload struct has an embedded field, and that holds on to the pointer. Ugh

Exactly, I should had explained that instead of elaborating on why you should save a pointer to a loop variable.

**Comment by squaremo** [comment]
I had a look at some `git blame` and this is the only change I thought might be relevant: https://github.com/2opremio/flux/commit/abe9db8febf88e926350c78099e852fe52e0c943#diff-18f3f72b1b5287be3ecab6e5884622b3R191 (but even so, I think it's actually OK).

**Comment by hiddeco** [comment]
@2opremio see below

```
❯ which fluxctl
/home/hidde/go/bin/fluxctl

~/go/src/github.com/weaveworks/flux master
❯ rm -rf /home/hidde/go/bin/fluxctl

~/go/src/github.com/weaveworks/flux master
❯ make clean realclean
go clean
rm -rf ./build
rm -f test/bin/kubectl test/bin/helm
rm -rf ./cache

~/go/src/github.com/weaveworks/flux master
❯ which fluxctl
fluxctl not found

~/go/src/github.com/weaveworks/flux master
❯ git pull
Already up to date.

~/go/src/github.com/weaveworks/flux master
❯ make
go install ./cmd/fluxctl
<snip>

~/go/src/github.com/weaveworks/flux master
❯ fluxctl list-workloads
WORKLOAD                               CONTAINER           IMAGE                               RELEASE  POLICY
default:helmrelease/faulty2                                                                             
default:helmrelease/faulty2   

~/go/src/github.com/weaveworks/flux master
❯ kubectl get hr
NAME      AGE
faulty    5d21h
faulty2   5d19h
```

2opremio:
@hiddeco can you triple-check that you are running the image from master?

hiddeco:
@2opremio err, crap, you are right, my image was fixed on something else. Nevermind.

2opremio:
No worries, it happens. Thanks for checking the fox works!

jan-schumacher:
@hiddeco unfortunately this bug is not fixed. I have again a list of helmReleases called helmRelease/redis when I do `fluxctl list-workloads`
Flux version 1.12.2
Helm operator 0.9.1

Screenshot 2019-05-17 at 15 47 55

PR #2029, opened 2019-05-09
Prevent auto-upgrade before lock in annotations tutorial

### LINKED ISSUE #1983: Workload autoupgraded before locking in annotations tutorial
Currently if you follow along with the https://github.com/weaveworks/flux/blob/master/site/annotations-tutorial.md example between the time you release 1.4.1 and you lock the deployment gets auto-upgraded to latest 1.4.x tag and therefore when you lock it locks on latest version. 

I would suggest therefore allowing the "lock" to be passed as a flag to release.

Something similar to:
```
fluxctl release --workload demo:deployment/podinfo -i stefanprodan/podinfo:1.4.1 --lock --lock-msg="1.4.2 does not work for us"
```

2opremio:
Fixes #1983

### DIFF
--- a/site/annotations-tutorial.md
+++ b/site/annotations-tutorial.md
@@ -204,10 +204,33 @@ In our case this is `1.4.2` (it could be a later image too). Let's say an
 engineer found that `1.4.2` was faulty and we have to go back to `1.4.1`.
 That's easy.
 
-Rollback to `1.4.1`:
+Lock deployment with a message describing why:
 
 ```sh
-fluxctl release --workload demo:deployment/podinfo -i stefanprodan/podinfo:1.4.1
+fluxctl lock -w demo:deployment/podinfo -m "1.4.2 does not work for us"
+```
+
+The resulting diff should look like this
+
+```diff
+--- a/workloads/podinfo-dep.yaml
++++ b/workloads/podinfo-dep.yaml
+@@ -10,6 +10,7 @@ metadata:
+     app: podinfo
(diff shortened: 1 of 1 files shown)


squaremo:
Rebasing on master should fix the CI failure

2opremio:
@squaremo done!

squaremo:
You could just reverse the order of the lock / rollback release such that it's locked first (then use `--force` with the rollback release, to make it ignore the lock for that command).

dholbach:
Will https://github.com/weaveworks/flux/commit/0ca2796c9c872e3f081fbeb137a99c07a4c9ae7c#diff-15622267c2ec3db7451cdecfe25640b9 need to be applied as well here?

dholbach:
Apart from that :+1:

2opremio:
> You could just reverse the order of the lock / rollback release such that it's locked first (then use --force with the rollback release, to make it ignore the lock for that command).

Good idea! I wasn't aware of the `--force` flag

2opremio:
> Will [0ca2796#diff-15622267c2ec3db7451cdecfe25640b9](https://github.com/weaveworks/flux/commit/0ca2796c9c872e3f081fbeb137a99c07a4c9ae7c#diff-15622267c2ec3db7451cdecfe25640b9) need to be applied as well here?

I think so, but I will do it in a separate PR.

2opremio:
@squaremo I redid it using your suggestion.

PR #2034, opened 2019-05-09
Log warning re not applying resource by namespace

### LINKED ISSUE #2012: Better logging when a resource does not apply due to k8s-allow-namespace filtering
**Describe the feature**
I scratched my head for some time when I saw a resource being handled by flux - but could not find it in the k8s ap:

```
[et2448@Davids-Work-MacBook-Pro ~ (⎈ |icp-global-context:kube-system)]$ kubectl -n devops logs flux-devops-5c4fbc4f8-rk9l6  |grep -i kube-eagle
ts=2019-05-06T00:14:41.985654952Z caller=daemon.go:627 component=daemon event="Sync: 25e67c1, kube-eagle:helmrelease/kube-eagle" logupstream=true
ts=2019-05-06T00:18:19.617252849Z caller=daemon.go:627 component=daemon event="Sync: 40efa2a, kube-eagle:helmrelease/kube-eagle" logupstream=true
[et2448@Davids-Work-MacBook-Pro ~ (⎈ |icp-global-context:kube-system)]$ 
```

which was due to the resource being put into a namespace not in `k8s-allow-namespace` - it would be nice if such a case would be logged more clearly to make the user aware of that.

**Expected behavior**
Log at warning level, "resource <id> not applied due to no match in k8s-allow-namespace filter"

squaremo:
We log when a resource in a manifest is dropped from sync because it's
marked as ignored; would also help troubleshooting if we log when a
resource is dropped because it's not in the allowed namespaces.

Fixes #2012.

### DIFF
--- a/cluster/kubernetes/sync.go
+++ b/cluster/kubernetes/sync.go
@@ -62,11 +62,12 @@ func (c *Cluster) Sync(syncSet cluster.SyncSet) error {
 
 	cs := makeChangeSet()
 	var errs cluster.SyncError
+	var excluded []string
 	for _, res := range syncSet.Resources {
 		resID := res.ResourceID()
 		id := resID.String()
 		if !c.IsAllowedResource(resID) {
-			logger.Log("warning", "not applying resource; excluded by namespace constraints", "resource", id, "source", res.Source())
+			excluded = append(excluded, id)
 			continue
 		}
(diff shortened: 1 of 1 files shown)


squaremo:
@hiddeco I changed it so it logs all the excluded resources in one line, if there are any. Can you have a quick log and make sure you're still happy with it?

PR #2042, opened 2019-05-13
default run as non-root user. Fixes #2041

### LINKED ISSUE #2041: add security-context to memcached container
**Describe the feature**
The memcached container cannot be installed into a namespace with a strict pod-security-policy due to user it runs as.
The container will refuse to start:
```
  Warning  Failed     3s (x12 over 2m8s)  kubelet, alt-aot-r-kwo01  Error: container has runAsNonRoot and image has non-numeric user (memcache), cannot verify user is non-root
  Normal   Pulled     3s (x11 over 2m8s)  kubelet, alt-aot-r-kwo01  Container image "fsnexus.evry.com:8085/memcached:1.4.25" already present on machine
```

0. Add a default security-context via values for the memcached container. Let it run as non-root user.

**Expected behavior**
Memcached should start without errors.

davidkarlsen:
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>
Adds a default securityContext for the memcached container.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/flux/2042)

### DIFF
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -186,6 +186,10 @@ memcached:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  securityContext:
+    runAsUser: 999
+    runAsGroup: 999
+    fsGroup: 999
   resources: {}
     # If you do want to specify resources, uncomment the following
     # lines, adjust them as necessary, and remove the curly braces after 'resources:'.


stefanprodan:
Having this as default will break the installation on Istio since the iptables init container must run as root. If you need to set a specific security context for the memcached pod do it at install time using `-set`.

PR #2050, opened 2019-05-15
Fix e2e tests

2opremio:
~It seems there are connectivity problems when starting the Kubernetes cluster in Circle. Hopefully they are transient, and this update (which includes https://github.com/stefanprodan/gitsrv/pull/7/ ) will fix the end to end tests.~

External connectivity broke in kind/master. The solution was to pin it to version 0.2.1

Also, on the way I  bumped `gitsrv` so that it retries its initialization indefinitely if there is a connectivity problem.


test/e2e/e2e-kind.sh
```diff
@@ -13,7 +13,7 @@ sudo mv kubectl /usr/local/bin/
 
 echo ">>> Building sigs.k8s.io/kind"
 cd $HOME
-GO111MODULE="on" go get -u sigs.k8s.io/kind@master
```
hiddeco: `0.2.1` is from before they switched to Go modules iirc.

2opremio: Ahh, true. Thanks, I just amended the commit.

2opremio:
@hiddeco PTAL

PR #2053, opened 2019-05-16
Allow to enable git signature verification in Helm chart

hiddeco:


hiddeco:
Labeled as blocked to ease the patch release of the chart next week (planned `1.12.3` release will not include the feature).

PR #2054, opened 2019-05-16
Give full output of git command on errors

2opremio:
As a side effect, the tracing of commands also prints the full output (instead of separating stderr and stdout), which I think is more useful.

Implementing this was triggered by #2050 (`git clone` was failing due to `known_hosts` being updated, but I had to exec into the pod in order to figure it out).

### DIFF
--- a/git/operations.go
+++ b/git/operations.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 )
@@ -327,6 +328,23 @@ func traceGitCommand(args []string, config gitCmdConfig, stdOutAndStdErr string)
 	)
 }
 
+type threadSafeBuffer struct {
+	bytes.Buffer
+	sync.Mutex
+}
+
+func (b *threadSafeBuffer) Write(p []byte) (n int, err error) {
+	b.Lock()
+	defer b.Unlock()
+	return b.Buffer.Write(p)
+}
+
+func (b *threadSafeBuffer) ReadFrom(r io.Reader) (n int64, err error) {
+	b.Lock()
+	defer b.Unlock()
+	return b.Buffer.ReadFrom(r)
+}
+
 // execGitCmd runs a `git` command with the supplied arguments.
 func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
 	c := exec.CommandContext(ctx, "git", args...)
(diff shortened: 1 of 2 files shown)


squaremo:
> Give full output of git command on errors

I restricted the error message to the first line of output deliberately. But who knows where that impulse came from, it was a while ago. It's certainly useful to have the full error from git.

> the tracing of commands also prints the full output (instead of separating stderr and stdout), which I think is more useful.

Because they will be interleaved?

2opremio:
> Because they will be interleaved?

Yep

2opremio:
@squaremo I have finally finished fixing the race conditions PTAL

git/operations.go
```diff
@@ -339,30 +353,26 @@ func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
...
 	}
 
 	err := c.Run()
 	if err != nil {
-		msg := findErrorMessage(errOut)
-		if msg != "" {
-			err = errors.New(msg)
+		if len(stdOutAndStdErr.Bytes()) > 0 {
+			err = errors.New(stdOutAndStdErr.String())
+			msg := findErrorMessage(stdOutAndStdErr)
```
squaremo: I can think of a theoretical problem here, but I don't think it's a practical problem. It is this: `findErrorMessage` is now given both stdout and stderr, and if there's output to stdout which accidentally matches what it's looking for (`"fatal: `), that might be used as the error message.

git/operations.go
```diff
@@ -339,30 +353,26 @@ func execGitCmd(ctx context.Context, args []string, config gitCmdConfig) error {
...
+			if msg != "" {
+				err = fmt.Errorf("%s, full output:\n %s", msg, err.Error())
```
squaremo: I'm not convinced that having stdout as well here is useful, and potentially it could be a lot of unnecessary output. Can you give an example of where having only stderr is inadequate?

2opremio: I can dig for it, but all in all the more context we give, the easier it will be to diagnose. As a user I would appreciate the full command output.

I don't think any of the commands we use can give a long output.

git/operations.go
2opremio: If that could happen (which I doubt) you would still have the full output.

git/operations.go
squaremo: > I don't think any of the commands we use can give a long output.

or I suppose, more to the point: commands which _might_ have a lot of output on stdout won't do so if there's an error.

Issue #1857, opened 2019-03-25
sync-loop context deadline exceeded while pushing into git a lot of changes


dananichev:
Once in a while we are forced to recreate all manifests inside git repository (limits tweaks, new pods and containers and so on). This leads to performing automated release by flux based on tag annotation for every manifest inside git repository (i believe we have ~1000 manifests or so on; maybe less). Which in turn leads to GIT sync. Which fails.

`git commit: running git command: git [commit --no-verify -a -m Auto-release multiple images\n\n - gitlab.***.ru:4567/***:latest-4985eb31bb68b72c377f44dd3905c4213eba2f01\n - gitlab.***.ru:4567/***:latest-e92483f1319148a21624249f472d8b885fbedbe4\n - gitlab.***.ru:4567/***:latest-10a671033011403ce0ebad6701ccb96d9a5a0fb6\n - gitlab.***.ru:4567/***:latest-dee8e41f3f2a1413954050356cfc3c68112715aa\n - gitlab.***.ru:4567/***:latest-ed6b6773ae051e9808ddd837b80a3e4177973cec\n - gitlab.***.ru:4567/***:latest-eb816707b76f8589dc4b667331e1c655f4966696\n - gitlab.***.ru:4567/***:latest-34eb94841504daa85785910e60c1e2e7cdae439b\n]: context deadline exceeded`

So far the only workaround i found: 

- mark every deployment as ignored/not automated 
- wait for flux to sync this
- enable for some deployments automation
- wait for sync and release
- repeat previous 2 steps until there are no ignored/not automated deployments left

This workaround takes a lot of time to "fix" this issue. 

Is there anything i can do to permamently fix it?

Thanks,
Dmitry

squaremo:
So it's a `git` invocation that's timing out? You can try setting `--git-timeout` to a value which is long enough to admit the problematic operation. That's `git.timeout` for the chart. I think you'll need fluxd 1.11.0 for it to affect all operations.

dananichev:
@squaremo i tried setting it up to 2-4 minutes with 1.10 and it wasn't working (i mean i still got `context deadline exceeded`). A week or so ago i updated flux to 1.11 but haven't tried `--git-timeout` again. I'll give it a try.

hiddeco:
I think this may not work as to my knowledge all releases are made through jobs and the job provides its own context with 60 second timeout.

[...]

PR #2056, opened 2019-05-16
Add `--timeout` flag to `fluxctl`

suvl:
In our cluster, there's a few operations that almost always take more than 60 seconds. `sync` is the most sensitive for us, as we use it from tools like Jenkins and Azure DevOps to assert that our changes were applied to the cluster.

The hard-coded timeout prevented us from having a nice deterministic pipeline, so we need to change it: we added the `--timeout <int>` argument and it'll also be looking to the `FLUX_TIMEOUT` env variable.

### DIFF
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -23,6 +23,7 @@ type rootOpts struct {
 	Namespace string
 	Labels    map[string]string
 	API       api.Server
+	Timeout   int
 }
 
 func newRoot() *rootOpts {
@@ -50,12 +51,13 @@ Workflow:
 `)
 
 const (
+	defaultURLGivenToken  = "https://cloud.weave.works/api/flux"
 	envVariableURL        = "FLUX_URL"
 	envVariableNamespace  = "FLUX_FORWARD_NAMESPACE"
 	envVariableLabels     = "FLUX_FORWARD_LABELS"
 	envVariableToken      = "FLUX_SERVICE_TOKEN"
 	envVariableCloudToken = "WEAVE_CLOUD_TOKEN"
-	defaultURLGivenToken  = "https://cloud.weave.works/api/flux"
+	envVariableTimeout    = "FLUX_TIMEOUT"
 )
 
 func (opts *rootOpts) Command() *cobra.Command {
@@ -75,6 +77,8 @@ func (opts *rootOpts) Command() *cobra.Command {
 		fmt.Sprintf("Base URL of the Flux API (defaults to %q if a token is provided); you can also set the environment variable %s", defaultURLGivenToken, envVariableURL))
 	cmd.PersistentFlags().StringVarP(&opts.Token, "token", "t", "",
 		fmt.Sprintf("Weave Cloud authentication token; you can also set the environment variable %s or %s", envVariableCloudToken, envVariableToken))
+	cmd.PersistentFlags().IntVar(&opts.Timeout, "timeout", 60,
+		fmt.Sprintf("Global command timeout, in seconds; you can also set the environment variable %s", envVariableTimeout))
 
 	cmd.AddCommand(
 		newVersionCommand(),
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -40,3 +40,4 @@ docker/fluxy-dumbconf.priv
 test/profiles
 test/bin/kubectl
 test/bin/helm
+.idea
```
hiddeco: This belongs in your personal `.gitignore` as it is environment specific (e.g. `~/.gitignore`).

cmd/fluxctl/root_cmd.go
```diff
@@ -15,6 +15,7 @@ import (
 	transport "github.com/weaveworks/flux/http"
 	"github.com/weaveworks/flux/http/client"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
hiddeco:
```suggestion
```

squaremo:
This is a nice idea :+1:

Is there an advantage to having the `timeout` in `fluxctl`, rather than using e.g., `timeout 2m fluxctl sync`?

I can think of one: it'll be available in environments which don't have GNU coreutils, e.g., Windows -- and since you mentioned Azure, perhaps that's an important advantage for you :-)

cmd/fluxctl/root_cmd.go
suvl: yeah, u'r righ... sorry 😩

.gitignore
suvl: yeah u'r right... sorry 😩

suvl:
@squaremo well, doin' it with `timeout` is a quite nice idea, but since the timeout is currently hard-coded, and as you said it can be advantageous in some cases, I still would like to see a setting/env var on this.

hiddeco:
This PR will probably solve #1857.

cmd/fluxctl/root_cmd.go
```diff
@@ -56,6 +58,7 @@ const (
 	envVariableToken      = "FLUX_SERVICE_TOKEN"
 	envVariableCloudToken = "WEAVE_CLOUD_TOKEN"
 	defaultURLGivenToken  = "https://cloud.weave.works/api/flux"
+	envVariableTimeout    = "FLUX_TIMEOUT"
```
hiddeco: I prefer these sorted by prefix (all `env*` constants below each other) but I may have OCD.

suvl: done, I took the liberty of sorting the entire thing 😆

suvl:
@hiddeco I've accepted all proposed changes.

squaremo:
> This PR will probably solve #1857.

That would be nice, but I don't think it will happen this time -- this PR changes how long `fluxctl` will keep polling for a job to complete, but the job timeout internal to `fluxd` remains the same.

squaremo:
> I still would like to see a setting/env var on this.

Fair enough -- no objection here.

hiddeco:
> That would be nice, but I don't think it will happen this time -- this PR changes how long fluxctl will keep polling for a job to complete, but the job timeout internal to fluxd remains the same.

Ah, you are right :disappointed: 

---

@suvl looking good, tests are failing due to the introduced timeout var not being set in `TestReleaseCommand_CLIConversion`. You can run the tests on your own machine by running `make test` (to confirm you fixed the issue), in addition, can you smash your commits together once this has been fixed?

2opremio:
@suvl are you planning to continue working on this? If that's the case, please revise on master since we have recently moved to go modules, which is causing a conflict

suvl:
@2opremio yeah sorry I was away for a time, but yeah planning on fixing the tests and now on fixing the merge from master.

squaremo:
> fixing the merge from master.

Please rebase on master rather than merging. Usually there will be no conflicts. If you need a hand though, let us know.

hiddeco:
Can't review this myself again as I have touched the code, @stefanprodan / @squaremo PTAL.

PR #2062, opened 2019-05-18
Add Avisi to production users

thojkooi:
We've been successfully using Flux in production for ~5 months now.


2opremio:
Thanks a lot!

PR #2063, opened 2019-05-18
Garbage collection dry run

### LINKED ISSUE #1990: Garbage collection dry-run
This is a flag you can pass to fluxd that makes it log what it _would_ delete if garbage collection were done, but not actually delete anything.

jan-schumacher:
I added a new `syncGarbageCollection.dry` option for helm chart and deployment. With this option activated it is irrelevant how `syncGarbageCollection.enabled` is set, the garbage collection will run in a trace mode. Only logs appear in the console but no resources are touched. 
I created a Test for this behaviour in the style of the existing tests.

Fixes #1990

### DIFF
--- a/cluster/kubernetes/sync.go
+++ b/cluster/kubernetes/sync.go
@@ -107,8 +107,8 @@ func (c *Cluster) Sync(syncSet cluster.SyncSet) error {
 	}
 	c.muSyncErrors.RUnlock()
 
-	if c.GC {
-		deleteErrs, gcFailure := c.collectGarbage(syncSet, checksums, logger)
+	if c.GC || c.DryGC {
+		deleteErrs, gcFailure := c.collectGarbage(syncSet, checksums, logger, c.DryGC)
 		if gcFailure != nil {
 			return gcFailure
 		}
@@ -129,7 +129,8 @@ func (c *Cluster) Sync(syncSet cluster.SyncSet) error {
 func (c *Cluster) collectGarbage(
 	syncSet cluster.SyncSet,
 	checksums map[string]string,
-	logger log.Logger) (cluster.SyncError, error) {
+	logger log.Logger,
+	dryRun bool) (cluster.SyncError, error) {
 
 	orphanedResources := makeChangeSet()
 
@@ -144,10 +145,12 @@ func (c *Cluster) collectGarbage(
 
 		switch {
 		case !ok: // was not recorded as having been staged for application
-			c.logger.Log("info", "cluster resource not in resources to be synced; deleting", "resource", resourceID)
-			orphanedResources.stage("delete", res.ResourceID(), "<cluster>", res.IdentifyingBytes())
+			c.logger.Log("info", "cluster resource not in resources to be synced; deleting", "dry-run", dryRun, "resource", resourceID)
+			if !dryRun {
+				orphanedResources.stage("delete", res.ResourceID(), "<cluster>", res.IdentifyingBytes())
+			}
 		case actual != expected:
-			c.logger.Log("warning", "resource to be synced has not been updated; skipping", "resource", resourceID)
+			c.logger.Log("warning", "resource to be synced has not been updated; skipping", "dry-run", dryRun, "resource", resourceID)
 			continue
 		default:
 			// The checksum is the same, indicating that it was
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -127,6 +127,7 @@ func main() {
 		// syncing
 		syncInterval = fs.Duration("sync-interval", 5*time.Minute, "apply config in git to cluster at least this often, even if there are no new commits")
 		syncGC       = fs.Bool("sync-garbage-collection", false, "experimental; delete resources that were created by fluxd, but are no longer in the git repo")
+		dryGC        = fs.Bool("sync-garbage-collection-dry", false, "experimental; only log what would be garbage collected, rather than deleting. Implies --sync-garbage-collection")
 
 		// registry
 		memcachedHostname = fs.String("memcached-hostname", "memcached", "hostname for memcached service.")
@@ -370,6 +371,7 @@ func main() {
 		allowedNamespaces := append(*k8sNamespaceWhitelist, *k8sAllowNamespace...)
 		k8sInst := kubernetes.NewCluster(client, kubectlApplier, sshKeyRing, logger, allowedNamespaces, *registryExcludeImage)
 		k8sInst.GC = *syncGC
+		k8sInst.DryGC = *dryGC
 
 		if err := k8sInst.Ping(); err != nil {
 			logger.Log("ping", err)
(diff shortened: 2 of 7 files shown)


cmd/fluxd/main.go
```diff
@@ -127,6 +127,7 @@ func main() {
 		// syncing
 		syncInterval = fs.Duration("sync-interval", 5*time.Minute, "apply config in git to cluster at least this often, even if there are no new commits")
 		syncGC       = fs.Bool("sync-garbage-collection", false, "experimental; delete resources that were created by fluxd, but are no longer in the git repo")
+		dryGC        = fs.Bool("dry-run-garbage-collection", false, "experimental; log deletion process for resources that were created by fluxd, but are no longer in the git repo")
```
squaremo: If you make this start with `sync-garbage-collection-` it will sort next to the closely-related `sync-garbage-collection` in usage and docs and so on.

cluster/kubernetes/sync.go
```diff
@@ -145,7 +150,11 @@ func (c *Cluster) collectGarbage(
 		switch {
 		case !ok: // was not recorded as having been staged for application
 			c.logger.Log("info", "cluster resource not in resources to be synced; deleting", "resource", resourceID)
-			orphanedResources.stage("delete", res.ResourceID(), "<cluster>", res.IdentifyingBytes())
+			if dryRun {
```
squaremo: This might be a bit neater in logs if you just alter the message output above, rather than printing another message -- which might get interleaved with something unrelated. Just adding a `dry-run=true` to the logger (once, at the top of the procedure) would be enough I reckon.

cmd/fluxd/main.go
squaremo: The usage text implies that deletion will still happen. Perhaps something like `"experimental; only log what would be garbage collected, rather than deleting. Implies --sync-garbage-collection"`

jan-schumacher: > If you make this start with `sync-garbage-collection-` it will sort next to the closely-related `sync-garbage-collection` in usage and docs and so on.

You mean in the sorting of names for options to start with? Didn't know that. How would you like `sync-garbage-collection-dry`

> The usage text implies that deletion will still happen. Perhaps something like `"experimental; only log what would be garbage collected, rather than deleting. Implies --sync-garbage-collection"`

You're absolutely right. I would just take your suggestion then.

cluster/kubernetes/sync.go
jan-schumacher: That would also serve the purpose, I will do that :)

jan-schumacher: Does this also apply yo line 135?

jan-schumacher: What I usually do in Java, especially in a multithreaded env, is to prefix messages with a context, if needed.

In this case I would do something like:

```
switch {
case !ok && !dryRun:
	c.logger.Log("info", "Garbage-Collection: dry-run, no deletion; cluster resource not in resources to be synced", "resource", resourceID)
	orphanedResources.stage("delete", res.ResourceID(), "<cluster>", res.IdentifyingBytes())
case !ok && dryRun: // was not recorded as having been staged for application
	c.logger.Log("info", "Garbage-Collection: deleting resources; cluster resource not in resources to be synced", "resource", resourceID)
case actual != expected:
	c.logger.Log("warning", "Garbage-Collection: resource to be synced has not been updated; skipping", "resource", resourceID)
	continue
default:
	// The checksum is the same, indicating that it was
	// applied earlier. Leave it alone.
}
```

it's a bit tedious to do this but the benefits are much more clear logs in the end.

jan-schumacher:
> ci/circleci: e2e-testing — Your tests failed on CircleCI

Unfortunately I can’t run the e2e tests because I miss the app kind. Is that an alias?

hiddeco:
@jan-schumacher [kind](https://github.com/kubernetes-sigs/kind) is Kubernetes in Docker, it should also work with minikube.

2opremio:
> Unfortunately I can’t run the e2e tests because I miss the app kind. Is that an alias?

@jan-schumacher You will be able to run them locally if you rebase on master (thanks to #2063 ) by running `make e2e`

Regardless, the current error is:

```
Error: render error in "flux/templates/deployment.yaml": template: flux/templates/deployment.yaml:194:24: executing "flux/templates/deployment.yaml" at <.Values.syncGarbageCollection.enabled>: enabled is not a method but has arguments
Exited with code 1
```

jan-schumacher:
> > Unfortunately I can’t run the e2e tests because I miss the app kind. Is that an alias?
> 
> @jan-schumacher You will be able to run them locally if you rebase on master (thanks to #2063 ) by running `make e2e`
> 
> Regardless, the current error is:
> 
> ```
> Error: render error in "flux/templates/deployment.yaml": template: flux/templates/deployment.yaml:194:24: executing "flux/templates/deployment.yaml" at <.Values.syncGarbageCollection.enabled>: enabled is not a method but has arguments
> Exited with code 1
> ```

@hiddeco 
Does that come from the change of the equation? 
```
{{- if .Values.syncGarbageCollection.enabled and not .Values.syncGarbageCollection.dry}}
- --sync-garbage-collection={{ .Values.syncGarbageCollection.enabled }}
{{- else if .Values.syncGarbageCollection.dry }}
- --sync-garbage-collection-dry={{ .Values.syncGarbageCollection.dry }}
{{- end }}
```
Should I rather do it like this?
```
{{- if .Values.syncGarbageCollection.enabled }}
{{- if not .Values.syncGarbageCollection.dry }}
- --sync-garbage-collection={{ .Values.syncGarbageCollection.enabled }}
{{- end }}
{{- else if .Values.syncGarbageCollection.dry }}
- --sync-garbage-collection-dry={{ .Values.syncGarbageCollection.dry }}
{{- end }}
```

hiddeco:
@jan-schumacher please take a look at the Helm documentation about operators https://helm.sh/docs/chart_template_guide/#operators-are-functions

jan-schumacher:
> @jan-schumacher please take a look at the Helm documentation about operators https://helm.sh/docs/chart_template_guide/#operators-are-functions

I see, thanks mate ;)

hiddeco:
@jan-schumacher this is shaping up well, can you smash your commits together into one?

cluster/kubernetes/sync.go
```diff
@@ -142,12 +143,17 @@ func (c *Cluster) collectGarbage(
 		actual := res.GetChecksum()
 		expected, ok := checksums[resourceID]
 
+		collectGarbage := !ok && !dryRun
```
hiddeco: Michael already mentioned this, but I think adding `dry-run` to the log would be sufficient. E.g.

```go
switch {
case !ok: // was not recorded as having been staged for application
  c.logger.Log("info", "cluster resource not in resources to be synced; deleting", "dry-run", dryRun, "resource", resourceID)
  if !dryRun {
    orphanedResources.stage("delete", res.ResourceID(), "<cluster>", res.IdentifyingBytes())
  }
case actual != expected: 
  c.logger.Log("warning", "resource to be synced has not been updated; skipping", "dry-run", dryRun, "resource", resourceID)
  continue
}
```

jan-schumacher: sure, I can do that. I thought he meant something else.

jan-schumacher: @hiddeco done

PR #2083, opened 2019-05-23
Migrate dependency management to Go modules

2opremio:
Also:
* Make `$GOBIN` work
* Cache go modules

`$GOPATH` is not required anymore. @hiddeco @squaremo @stefanprodan make sure to move your Flux outside of `$GOPATH` once this is merged (the alternative is to prepend every single `go` invocation by `GO111MODULE=on` which I ended up discarding)


2opremio:
Close, but no cigar :( https://github.com/kubernetes/kubernetes/issues/67566

squaremo:
> Close, but no cigar :( [kubernetes/kubernetes#67566](https://github.com/kubernetes/kubernetes/issues/67566)

lolk8s :-(

2opremio:
Huhm, there seems to be a solution https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

2opremio:
The build time goes down to 8.5 minutes since, unlike `dep`, `go mod` caches dependencies.

2opremio:
Great, you can actually stop defining GOPATH altogether

Issue #2075, opened 2019-05-22
Alternate formats for image created time


squaremo:
Some images (at least the library CentOS images) use a non-standard format for created date:

```
ts=2019-05-22T13:08:17.050122049Z caller=repocachemanager.go:226 component=warmer canonical_name=index.docker.io/library/centos auth={map[]} err="parsing time \"20180804\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"0804\" as \"-\"" ref=centos:latest
```

The spec they are following, https://label-schema.org/rc1/, says

> This label contains the Date/Time the image was built. The value SHOULD be formatted according to RFC 3339.

which is not exactly difficult to comply with, but there you are. As a matter of practicality, we could attempt alternate formats to see if we get a usable datetime.

hiddeco:
I first thought we were at fault by not following the spec, but they are.

I am opposed to support something different than what the spec defines, as people should just follow the rules. Instead, I think we should silence the error, ignore the value, and fallback to the timestamp we get from the image registry.

hiddeco:
I implemented what I described above in the PR linked to this issue (and also stated a couple of additional reasons to _not_ support more formats there). Feel free to re-open if you disagree.

PR #2084, opened 2019-05-23
Do not fail on non RFC3339 labels but log warning

hiddeco:
Some image vendors have not read the spec properly and pass along
a timestamp in a different format than RFC3339. Before this change
the unmarshal of the JSON would fail for those labels and the affected
images would be excluded from our cache.

We now include those images in our cache and collect all labels that
failed parsing in a custom error, which we detect and log (instead
of bailing out).

Addresses #2075, but in a stricter way.

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -223,6 +223,17 @@ func (i Ref) WithNewTag(t string) Ref {
 	return img
 }
 
+type LabelTimestampFormatError struct {
+	Labels []string
+}
+
+func (e *LabelTimestampFormatError) Error() string {
+	return fmt.Sprintf(
+		"failed to parse %d timestamp label(s) as RFC3339 (%s); ask the repository administrator to correct this as it conflicts with the spec",
+		len(e.Labels),
+		strings.Join(e.Labels, ","))
+}
+
 // Labels has all the image labels we are interested in for an image
 // ref, the JSON struct tag keys should be equal to the label.
 type Labels struct {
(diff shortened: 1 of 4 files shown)


image/image.go
```diff
@@ -223,6 +223,17 @@ func (i Ref) WithNewTag(t string) Ref {
 	return img
 }
 
+type LabelParseError struct {
```
2opremio: should we make the error more explicit? e.g. `TimestampFormatError`

hiddeco: `LabelTimestampFormatError` even maybe?

2opremio:
> Some image vendors have not read the spec properly and pass along a timestamp in a different format than RFC3339

Do we know what wrong formats they typically use? As @squaremo mentioned in Slack, we could maybe try to accommodate those other formats as well, by trying to parse them as a fallback.

hiddeco:
@2opremio CentOS images use a custom `yyyymmdd` format (a datestamp basically, if that would be a thing). I do not know what other images use the labels with a wrong format.

I am personally opposed to support anything different than what the spec defines, as it would stimulate people not following rules in a way, the specs we support clearly state:

> This label contains the Date/Time the image was built. The value SHOULD be formatted according to RFC 3339.

I also do not think this will be an issue for any user in case they do not have (direct) control over what is defined in the labels, as we safely fallback to the timestamp the registry gives us.

image/image.go
2opremio: Sure!

2opremio:
> @2opremio CentOS images use a custom `yyyymmdd` format (a datestamp basically, if that would be a thing). I do not know what other images use the labels with a wrong format.
>
> I am personally opposed to support anything different than what the spec defines, as it would stimulate people not following rules in a way, and the specs we support clearly state:

Makes sense. It may be worth reporting it to (or even creating a PR against) https://github.com/CentOS/CentOS-Dockerfiles

2opremio:
Actually, this is the repo https://github.com/CentOS/sig-cloud-instance-images

hiddeco:
It has been mentioned in an issue there about a year ago, just did not find the time yet to submit a PR.

Some other reasons that came to mind about not supporting different timestamps:

* we prioritize labels over the date from the registry, we can only do this if the given timestamp is _more reliable_, which is something we can not assume when someone does not follow the spec
* it adds complexity to our codebase to support something that’s essentially wrong

2opremio:
Here's the culprit https://github.com/CentOS/sig-cloud-instance-build/blob/master/docker/containerbuild.sh#L74

2opremio:
Upstream PR https://github.com/CentOS/sig-cloud-instance-build/pull/151

2opremio:
LGTM

PR #2085, opened 2019-05-23
Fix flakey test due to mismatching error

2opremio:
The expected error is `client timeout (2ms) exceeded`, but sometimes, due to rate limiting, it's `Wait(n=1) would exceed context deadline`


2opremio:
See https://circleci.com/gh/weaveworks/flux/6074 for an instance of the failing test

PR #2086, opened 2019-05-23
Remove blocking ReadFrom in threadSafeBuffer

### LINKED ISSUE #2082: `git notes` operation times out
~To recover the set of events represented by a sync, fluxd looks at the notes made in git. With a large enough git repo, this can take longer than the (pretty generous) default git timeout.~

Scratch that: it takes a fraction of a second to run that command, in the pod. I suspect instead there is a deadlock in execGitCmd that leads to the operation timing out -- the problem started on the day that #2054 was merged :-S

squaremo:
Because ReadFrom will block on the src io.Reader it is passed, and
thereby hold the lock indefinitely, we don't want to have it as part
of the threadSafeBuffer.

Instead, don't embed the bytes.Buffer so as not to inherit ReadFrom,
and implement just the minimum needed for io.Reader and io.Writer.

Fixes #2082.

### DIFF
--- a/git/operations.go
+++ b/git/operations.go
@@ -329,20 +329,32 @@ func traceGitCommand(args []string, config gitCmdConfig, stdOutAndStdErr string)
 }
 
 type threadSafeBuffer struct {
-	bytes.Buffer
-	sync.Mutex
+	buf bytes.Buffer
+	mu  sync.Mutex
 }
 
 func (b *threadSafeBuffer) Write(p []byte) (n int, err error) {
-	b.Lock()
-	defer b.Unlock()
-	return b.Buffer.Write(p)
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.Write(p)
 }
 
-func (b *threadSafeBuffer) ReadFrom(r io.Reader) (n int64, err error) {
-	b.Lock()
-	defer b.Unlock()
-	return b.Buffer.ReadFrom(r)
+func (b *threadSafeBuffer) Read(p []byte) (n int, err error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.Read(p)
+}
+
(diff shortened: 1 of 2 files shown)


git/operations.go
```diff
@@ -329,20 +329,32 @@ func traceGitCommand(args []string, config gitCmdConfig, stdOutAndStdErr string)
...
+	defer b.mu.Unlock()
+	return b.buf.Write(p)
 }
 
-func (b *threadSafeBuffer) ReadFrom(r io.Reader) (n int64, err error) {
-	b.Lock()
-	defer b.Unlock()
-	return b.Buffer.ReadFrom(r)
+func (b *threadSafeBuffer) Read(p []byte) (n int, err error) {
+	b.mu.Lock()
```
2opremio: Could this be an `RLock()`?

2opremio:
Thanks @squaremo 🤦‍♂

2opremio:
@squaremo For extra points, can you convert you repro _gist_ into a test?

git/operations.go
squaremo: We could use a RWMutex, yes. For this particular situation, I don't think it would make a difference, because the concurrency is on the `Write`s (stderr and stdout writing to the same buffer), and `Read` and `Bytes` and `String` are not used until those are finished.

(Why was there a race between `Write` and `ReadFrom`, but not concurrency of `Write` and `Read`? Because `ReadFrom` is used when copying data to the buffer, i.e., it's a write operation)

squaremo:
> can you convert you repro _gist_ into a test?

A bit fiddly but far from impossible. The main hurdle would be to not take seconds to fail (or worse, seconds to succeed). Secondary hurdle: finding a reasonable, cross-platform executable to test with.

squaremo:
> finding a reasonable, cross-platform executable to test with.

On second thought, I can't think of a reason just using `git` isn't adequate.

Issue #2067, opened 2019-05-21
Doesn't update & trigger deployment of HelmRelease


dminca:
**Describe the bug**
_Regression bug:_ Flux is not able to match the RegEx pattern of the `HelmRelease` image label, thus not triggering the update (in git repo) & deployment of the `HelmRelease`.

**To Reproduce**
Steps to reproduce the behaviour:
0. Given a Helm Chart with the following `values.yaml` file:
```yaml
<snip>
image:
  repository: marcel-registry.com/myapp
  tag: master-231kdsj92
<snip>
```

1. Create a `HelmRelease` with the following values
```yaml
apiVersion: flux.weave.works/v1beta1
kind: HelmRelease
metadata:
  name: myapp
  namespace: myapp-ns
  annotations:
    flux.weave.works/automated: "true"
    flux.weave.works/tag.chart-image: glob:master-*
spec:
  releaseName: myapp
  chart:
    git: ssh://git@marcel-repo.com/myapp.git
    path: helm/myapp
    ref: master
  values:
    image:
      registry: marcel-registry.com/myapp
      tag: master-3d30e14fd7
```
2. Update & Release of `HelmRelease` actions not firing
3. No errors reported in the logs (expecting `RegEx pattern mismatch` thrown)

**Expected behavior**
Flux detected a new image in the registry and triggered an update of the image label (in git repo) and triggered a deployment of the `HelmRelease`

**Logs**

**Additional context**
Add any other context about the problem here, e.g 
- Flux version: `1.12.2`
- Helm Operator version: `0.9.2`
- Kubernetes version: `v1.11.7`
- Git provider: `Atlassian Bitbucket`
- Container registry provider: `AWS ECR`

Suigi:
AFAIK you have to provide the image name in the `HelmRelease` YAML for flux to be able to list the available tags and auto-release new images. 

If you list the images using `fluxctl list-images` you should see that there are no images listed for the HelmRelease:myapp resource. It will probably list them for the `Deployment`, but the `Deployment` does not have the automation annotations, the` HelmRelease` does.

hiddeco:
Thanks for your comment @Suigi, support like this from the community is highly appreciated :1st_place_medal:

@dminca the comment from Suigi sums it all up pretty well, Flux needs both the `image.repository` and the `image.tag` to be able to automate the workload, once you add the repository I expect it to update the image without any problems.

dminca:
[...]

PR #2089, opened 2019-05-23
update Helm docs with latest behaviour

dminca:
Update readme to address regression bug that doesn't automatically
trigger HelmRelease deployment if
-> the full Docker Registry URL + image name + image label are not provided
  in a Helm Chart as a single variable (not splitted)
-> which in turn is read by the HelmRelease

Resolves:
Related: #2067
Signed-off-by: Daniel Andrei Minca

### DIFF
--- a/site/helm-get-started.md
+++ b/site/helm-get-started.md
@@ -160,9 +160,7 @@ change the `tag:` line to the following:
 
 ```yaml
   values:
-    image:
-      repository: bitnami/mongodb
-      tag: 4.0.6
+    image: bitnami/mongodb:4.0.6
 ```
 
 Commit the change to your `master` branch. It will now get


hiddeco:
The split format should work fine, the issue with the format in #2067 is that

```
image:
  registry: marcel-registry.com/myapp
  tag: master-3d30e14fd7
```

should be

```
image:
  repository: marcel-registry.com/myapp
  tag: master-3d30e14fd7
```

as documented [here](https://github.com/weaveworks/flux/blob/5e17cd52c9b2430ac9caf729e8bd39dc89780f3f/site/helm-integration.md#upgrading-images-in-a-helmrelease-using-flux).

dminca:
Ok, now it makes sense. I'll test and if it doesn't work out well I'll reopen this one.

Thanks 👍

PR #1906, opened 2019-04-05
Make all Helm releases through the operator

hiddeco:
Fixes #1873, fixes #1820

This PR changes a couple of things in the Helm operator (details are commented in each commit), but in one sentence: it ensures all releases are made by the operator.

(If someone has a better one sentence recap of this PR; be my guest to change it)

PR #2103, opened 2019-05-28
Reorder boot and verify informer cache sync

hiddeco:
Before this change the operator would hang at random on "waiting for
informer caches to sync", without anything else happening or it timing
out.

After an extensive search in recently merged PRs, this seems to be
related to an change of order in #1906, where the informer would be
initialized before the operator had registered its event handlers.

In addition to reverting this change and ensuring the informer cache is
synced before it is used, reorder everything so that the operator is
started before anything else that may depend on it, so it can start
processing changes from the queue as soon as possible.

### DIFF
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/spf13/pflag"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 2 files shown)


hiddeco:
@squaremo @2opremio added 30 builds to be 100% sure this is a permanent fix, but my previous tests on CircleCI and on my own machine (about 100 cycles all together) did not run into any issues. https://circleci.com/gh/weaveworks/workflows/flux/tree/helm%2Finformer-sync

2opremio:
LGTM, but again, I am not super familiar with the operator.

PR #2107, opened 2019-05-29
Explicitly configure memcached user

stefansedich:
This is a "fix" for https://github.com/weaveworks/flux/issues/2106 by getting more explicit with the memcached container securityContext and also while there adding some sensible defaults and dissalowing privilege escalation.

As I found some discrepancies with the uid/gid between versions of memcached, it appears as though there was some work in the later versions to settle on the well known value of `11211` (https://github.com/docker-library/memcached/pull/41) so I upgraded to the latest version which so far appears to be working fine.

While this fix is probably not required it ties in with my work of running as non-root and getting explicit about the users running containers and removing capabilities that are not required.


squaremo:
Looks reasonable to me! @stefanprodan, one for you when you are back :palm_tree:

PR #2123, opened 2019-06-03
Confirm release belongs to HR before upgrading

### LINKED ISSUE #2101: Helm Operator will try to apply multiple times when two of more releases have the same name
**Describe the bug**
If someone accidentally uses a helm release name twice.  The helm operator will keep trying to apply the change back and forth between the two releases.  This results in many releases and could take down someones app.

**To Reproduce**
1. Create two different helm releases with different settings apart from the ReleaseName

**Expected behavior**
The operator should back off and fail when there are two of the same release names in the queue.

** Additional info 
This happend  on our test clusters where we don't require PRs.  A PR would likely of avoided this as someone would of checked it, but I think the helm operator should have some logic to fail when this happens.  I will attempt to fix this with a PR and just wanted to log this in case someone else is already working on the issue.

**Additional context**
- Flux version: 1.12.2
- Helm Operator version: 0.9.1

hiddeco:
Before this change, when a user would create multiple HelmReleases
with the same release name configured, the operator would attempt
to upgrade the release indefinitely.

~To overcome this issue, the operator now injects the resource ID
of the HelmRelease into the description of the release and confirms
this value equals the resource ID of the HelmRelease it is going to
run an upgrade for. If the value diverges, the status of the
HelmRelease is updated to make it visible to the user and a message
is logged.~

To overcome this issue, we now check if the resources created by the
release are patched with the correct antecedent annotation, which we
create ourselves and have full control over (and survives rollbacks).

For backwards compatibility, and to be able to migrate existing
releases to a HelmRelease, we see empty descriptions as a positive.

Fixes #2101

### DIFF
--- a/integrations/helm/chartsync/chartsync.go
+++ b/integrations/helm/chartsync/chartsync.go
@@ -377,8 +377,15 @@ func (chs *ChartChangeSync) reconcileReleaseDef(fhr fluxv1beta1.HelmRelease) {
 		}
 		chs.setCondition(fhr, fluxv1beta1.HelmReleaseReleased, v1.ConditionTrue, ReasonSuccess, "helm install succeeded")
 		if err = status.UpdateReleaseRevision(chs.ifClient.FluxV1beta1().HelmReleases(fhr.Namespace), fhr, chartRevision); err != nil {
-			chs.logger.Log("warning", "could not update the release revision", "namespace", fhr.Namespace, "resource", fhr.Name, "err", err)
+			chs.logger.Log("warning", "could not update the release revision", "resource", fhr.ResourceID().String(), "err", err)
 		}
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


hiddeco:
> A minor suggestion, that need not hold this up: to make the description more readable, in case a human looks at it, could that be "Owned by: "+fhrResourceID perhaps.

After running some tests I came to the conclusion it needs further tweaking, as in case no description is given, Helm will write its own description (`<task> complete`) and to make it even more fun, we will also have issues when a user performs a rollback as this will result in a `Rollback to <version>` description.

As these messages are simple strings in the Helm code base (and not public constants as I hoped they would be), writing (and maintaining) logic based on this does not seem preferable to me.

The other option I could think of to fix this problem is the `flux.weave.works/antecedent` annotation we patch all resources with after a successful install/upgrade.

PR #2126, opened 2019-06-04
Refactor (and rename) releasestore package

squaremo:
Four refactors (but see the commit messages):

 - separate the treatment of repos with bare YAMLs from repos with `.flux.yaml`, rather than having one implementation with two different behaviours. To avoid duplicating code, the config-aware implementation delegates to the bare YAMLs implementation when a bare YAML is what is being updated.

A strong motivation here is that further refinements to the config generation can be done without touching the bare YAMLs code.

 - put AppendManifestToBuffer into the Manifests interface, so it's alongside the other methods that assume a particular format (YAML)

 - move the Manifests interface to be with the procedures that depend on it

 - rename things to elide repetitions

### DIFF
--- a/cluster/kubernetes/patch.go
+++ b/cluster/kubernetes/patch.go
@@ -16,7 +16,6 @@ import (
 	k8sscheme "k8s.io/client-go/kubernetes/scheme"
 
 	"github.com/weaveworks/flux"
-	"github.com/weaveworks/flux/cluster"
 	"github.com/weaveworks/flux/cluster/kubernetes/resource"
 )
 
@@ -54,7 +53,7 @@ func createManifestPatch(originalManifests, modifiedManifests []byte, originalSo
 			// Avoid outputting empty patches
 			continue
 		}
-		if err := cluster.AppendManifestToBuffer(patch, buf); err != nil {
+		if err := appendYAMLToBuffer(patch, buf); err != nil {
 			return nil, err
 		}
 	}
@@ -99,7 +98,7 @@ func applyManifestPatch(originalManifests, patchManifests []byte, originalSource
 			}
 			resourceBytes = patched
 		}
-		if err := cluster.AppendManifestToBuffer(resourceBytes, buf); err != nil {
+		if err := appendYAMLToBuffer(resourceBytes, buf); err != nil {
 			return nil, err
 		}
 	}
--- a/daemon/sync.go
+++ b/daemon/sync.go
@@ -13,8 +13,8 @@ import (
 	"github.com/weaveworks/flux/cluster"
 	"github.com/weaveworks/flux/event"
 	"github.com/weaveworks/flux/git"
+	"github.com/weaveworks/flux/manifests"
 	"github.com/weaveworks/flux/resource"
-	"github.com/weaveworks/flux/resourcestore"
 	fluxsync "github.com/weaveworks/flux/sync"
 	"github.com/weaveworks/flux/update"
 )
@@ -58,7 +58,7 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, revision string, s
 
 	// Run actual sync of resources on cluster
 	syncSetName := makeGitConfigHash(d.Repo.Origin(), d.GitConfig)
-	resourceStore, err := d.getResourceStore(working)
+	resourceStore, err := d.getManifestStore(working)
 	if err != nil {
 		return errors.Wrap(err, "reading the respository checkout")
 	}
@@ -139,9 +139,9 @@ func getChangeSet(ctx context.Context, working *git.Checkout, repo *git.Repo, ti
 
 // doSync runs the actual sync of workloads on the cluster. It returns
 // a map with all resources it applied and sync errors it encountered.
-func doSync(ctx context.Context, resourceStore resourcestore.ResourceStore, clus cluster.Cluster, syncSetName string,
+func doSync(ctx context.Context, manifestsStore manifests.Store, clus cluster.Cluster, syncSetName string,
 	logger log.Logger) (map[string]resource.Resource, []event.ResourceError, error) {
-	resources, err := resourceStore.GetAllResourcesByID(ctx)
+	resources, err := manifestsStore.GetAllResourcesByID(ctx)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "loading resources from repo")
 	}
@@ -168,7 +168,7 @@ func doSync(ctx context.Context, resourceStore resourcestore.ResourceStore, clus
 // getChangedResources calculates what resources are modified during
 // this sync.
 func getChangedResources(ctx context.Context, c changeSet, timeout time.Duration, working *git.Checkout,
-	resourceStore resourcestore.ResourceStore, resources map[string]resource.Resource) (map[string]resource.Resource, error) {
+	manifestsStore manifests.Store, resources map[string]resource.Resource) (map[string]resource.Resource, error) {
 	if c.initialSync {
 		return resources, nil
 	}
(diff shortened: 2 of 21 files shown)


cluster/kubernetes/patch.go
```diff
@@ -16,7 +16,6 @@ import (
 	k8sscheme "k8s.io/client-go/kubernetes/scheme"
 
 	"github.com/weaveworks/flux"
-	"github.com/weaveworks/flux/cluster"
```
2opremio: This dependency didn't make sense, much better now.

daemon/sync.go
```diff
@@ -139,7 +139,7 @@ func getChangeSet(ctx context.Context, working *git.Checkout, repo *git.Repo, ti
 
 // doSync runs the actual sync of workloads on the cluster. It returns
 // a map with all resources it applied and sync errors it encountered.
-func doSync(ctx context.Context, resourceStore resourcestore.ResourceStore, clus cluster.Cluster, syncSetName string,
+func doSync(ctx context.Context, resourceStore manifests.Store, clus cluster.Cluster, syncSetName string,
```
2opremio:
```suggestion
func doSync(ctx context.Context, manifestsStore manifests.Store, clus cluster.Cluster, syncSetName string,
```

daemon/sync.go
```diff
@@ -168,7 +168,7 @@ func doSync(ctx context.Context, resourceStore resourcestore.ResourceStore, clus
 // getChangedResources calculates what resources are modified during
 // this sync.
 func getChangedResources(ctx context.Context, c changeSet, timeout time.Duration, working *git.Checkout,
-	resourceStore resourcestore.ResourceStore, resources map[string]resource.Resource) (map[string]resource.Resource, error) {
+	resourceStore manifests.Store, resources map[string]resource.Resource) (map[string]resource.Resource, error) {
```
2opremio:
```suggestion
	manifestsStore manifests.Store, resources map[string]resource.Resource) (map[string]resource.Resource, error) {
```

manifests/configaware.go
```diff
@@ -23,35 +22,30 @@ type resourceWithOrigin struct {
...
-type fileResourceStore struct {
-	manifests       cluster.Manifests
-	baseDir         string
-	rawManifestDirs []string
-	configFiles     []*ConfigFile
-	resourcesByID   map[string]resourceWithOrigin
-	sync.RWMutex
-}
+type configAware struct {
+	rawfiles    *rawFiles
```
2opremio:
```suggestion
	rawFiles    *rawFiles
```

manifests/configaware.go
```diff
@@ -141,39 +135,22 @@ func (frs *fileResourceStore) SetWorkloadContainerImage(ctx context.Context, res
...
-}
-
-func (frs *fileResourceStore) setConfigFileWorkloadContainerImage(ctx context.Context, cf *ConfigFile, r resource.Resource,
+func (ca *configAware) setConfigFileWorkloadContainerImage(ctx context.Context, cf *ConfigFile, r resource.Resource,
 	container string, newImageID image.Ref) error {
 	if cf.PatchUpdated != nil {
-		return frs.updatePatchFile(ctx, cf, func(previousManifests []byte) ([]byte, error) {
-			return frs.manifests.SetWorkloadContainerImage(previousManifests, r.ResourceID(), container, newImageID)
+		return ca.updatePatchFile(ctx, cf, func(previousManifests []byte) ([]byte, error) {
+			return ca.rawfiles.manifests.SetWorkloadContainerImage(previousManifests, r.ResourceID(), container, newImageID)
```
2opremio: Shouldn't we be invoking `SetWorkloadContainerImage()` directly on rawfiles? Otherwise I think we should keep our own manifests copy.

manifests/configaware.go
```diff
@@ -195,13 +172,13 @@ func (frs *fileResourceStore) setConfigFileWorkloadContainerImage(ctx context.Co
...
-func (frs *fileResourceStore) updatePatchFile(ctx context.Context, cf *ConfigFile,
+func (ca *configAware) updatePatchFile(ctx context.Context, cf *ConfigFile,
 	updateF func(previousManifests []byte) ([]byte, error)) error {
 
 	patchUpdated := *cf.PatchUpdated
-	generatedManifests, patchedManifests, patchFilePath, err := frs.getGeneratedAndPatchedManifests(ctx, cf, patchUpdated)
+	generatedManifests, patchedManifests, patchFilePath, err := ca.getGeneratedAndPatchedManifests(ctx, cf, patchUpdated)
 	if err != nil {
-		relConfigFilePath, err := filepath.Rel(frs.baseDir, cf.Path)
+		relConfigFilePath, err := filepath.Rel(ca.rawfiles.baseDir, cf.Path)
```
2opremio: I don't like that we piggyback on `rawFiles`'s fields for functionality unrelated to `rawFiles`. I would keep a copy.

manifests/configaware.go
```diff
@@ -211,16 +188,16 @@ func (frs *fileResourceStore) updatePatchFile(ctx context.Context, cf *ConfigFil
 	if err != nil {
 		return err
 	}
-	newPatch, err := frs.manifests.CreateManifestPatch(generatedManifests, finalManifests,
+	newPatch, err := ca.rawfiles.manifests.CreateManifestPatch(generatedManifests, finalManifests,
```
2opremio: Same here (I will stop now :) )

manifests/manifests.go
```diff
@@ -29,19 +28,9 @@ type Manifests interface {
...
-		return fmt.Errorf("cannot write to internal buffer: %s", err)
-	}
-	if _, err := buffer.Write(manifest); err != nil {
-		return fmt.Errorf("cannot write to internal buffer: %s", err)
-	}
-	return nil
+	// AppendManifestToBuffer concatentates manifest bytes to a
+	// (possibly empty) buffer of manifest bytes; the resulting bytes
+	// should be parsable by `ParseManifest`.
+	// TODO(michael) should really be an interface rather than `*bytes.Buffer`.
```
2opremio: Agreed, can we make it a writer and rename it to `AppendManifest`?

(To be fair all the `[]byte`'s  should probably be readers and writers too)

manifests/configaware.go
2opremio: We can't call `SetWorkloadContainerImage()` since it operates on files. I would keep a copy of the manifests then.

sync/sync_test.go
```diff
@@ -81,7 +78,7 @@ func resourcesToStrings(resources map[string]resource.Resource) map[string]strin
 
 // Our invariant is that the model we can export from the cluster
 // should always reflect what's in git. So, let's check that.
-func checkClusterMatchesFiles(t *testing.T, rs resourcestore.ResourceStore, resources map[string]string, base string, dirs []string) {
+func checkClusterMatchesFiles(t *testing.T, rs manifests.Store, resources map[string]string, base string, dirs []string) {
```
2opremio:
```suggestion
func checkClusterMatchesFiles(t *testing.T, ms manifests.Store, resources map[string]string, base string, dirs []string) {
```

2opremio:
Good job, thanks! Minor comments (most important regarding shared ownership of fields between `rawFiles` and `configAware`).

Otherwise LGTM

manifests/configaware.go
squaremo: Yeah, `rawfiles` is doing double duty, by being the implementation for rawfiles; and (for configAware) being a holder for some stuff it needs, like the manifests. I'm not too nervous about this -- is the objection about maintaining encapsulation (rather that correctness)?

2opremio: Yes, encapsulation and clear resource ownership.

manifests/manifests.go
squaremo: I did give it a quick attempt before writing the comment, but there were complications -- I can't remember what they were :-(. Possibly just that the implementation uses methods of `*bytes.Buffer`.

manifests/configaware.go
squaremo: OK, it's a fair objection -- I'm trying to get away with too much :-) I'll fix it up.

PR #2142, opened 2019-06-10
Use openssh-client instead of openssh apk

willholley:
In the primary Flux container, `ssh` is only required as a client to connect to external SSH servers. Replacing the use of the `openssh` apk with `openssh-client` reduces the surface area of the container and removes a class of vulnerability/configuration warnings that get picked up by static analysis tools in relation to `sshd`.

### DIFF
--- a/docker/Dockerfile.flux
+++ b/docker/Dockerfile.flux
@@ -2,7 +2,7 @@ FROM alpine:3.9
 
 WORKDIR /home/flux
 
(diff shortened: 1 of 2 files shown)


stefanprodan:
The same applies to Helm Operator, can you please change it here also https://github.com/weaveworks/flux/blob/master/docker/Dockerfile.helm-operator

willholley:
thanks for the swift feedback @stefanprodan - done!

Issue #1676, opened 2019-01-20
Crypto support for git


davidkarlsen:
It would be great if https://www.agwa.name/projects/git-crypt/ or https://github.com/mozilla/sops could be supported OOTB.

While I am aware of solutions like sealed https://github.com/bitnami-labs/sealed-secrets (and also use it for my own charts for secrets) there are certain public charts which do not - hence there is a real need to support a secure git repo in some way.

Both options can be supported quite non-intrusively - simply add support for the binary and add config vars (at least for sops)

stephenmoloney:
@davidkarlsen 
> It would be great if https://www.agwa.name/projects/git-crypt/ or https://github.com/mozilla/sops could be supported OOTB.

These are useful projects and I've used git-crypt but plan on moving to `sops`. and I've recently come across another one that seems interesting in the k8s context: `https://github.com/Soluto/kamus` but I have given it a try yet.


With that said, however, I wonder if flux already has a workaround for your issue with the ValueFileSecrets feature?
Example chart where it might be useful:
`https://github.com/helm/charts/blob/d26f5d463e702b3ea83983904526c138dfff3bd9/stable/rabbitmq/templates/secrets.yaml#L1-L21`
In those instances, there is a rather nice workaround that flux provides and is documented here:
https://github.com/weaveworks/flux/blob/master/site/helm-integration.md#specvaluefilesecrets

This thread on slack goes into a bit of detail on it. https://weave-community.slack.com/archives/C4U5ATZ9S/p1546602423201000

whereismyjetpack:
I use `valueFileSecrets` with helm charts that don't support external secrets. Even if they do, I find that deployments roll nicer when helm manages all the bits and bobs. 

We use git-crypt for humans and sealed secrets for the cluster. This seems to work well

davidkarlsen:
valueFileSecrets amends this to a certain extent - but you'll have one manual flow (create the secret) - and one git-ops driven - using sops or git-crypt makes it fully git-driven.

kamus looks a bit like sealed-secrets

Multiply:
We've been using `helm secrets` (sops) for a while, and prefer it over SealedSecrets, as we can use AWS KMS to distribute keys to our team.

An easy way to integrate sops, would be to extend `valueFileSecrets` or add another one specifically for sops, that takes the encrypted file location, attempts to decrypt it using credentials you configured for flux globally, and profit.

michaelgeorgeattard:
[...]

PR #2159, opened 2019-06-13
Add git-secret support

arsiesys:
Hello,

To answer to some needs, I made a change to handle the git repository using git-secret. This allow us to encrypt our secrets.
https://git-secret.io/

If the parameter --git-secret is set to true in fluxd configuration and a .gitsecret directory exist in the root of the git, flux will execute a "git secret reveal -f" in the working git.
Indeed, to make it work, you need to have the key in the gpg keystore, that's why we are using the --git-gpg-key-import option for that. Happy it was already there.

I am not fully comfortable with GO, so I am open to any suggestion that could be easily applied on my changes. For any more complicated changes, please take the handover on this PR if you are really interested or give me good clues :p.

Related Issues:
#1676


git/operations.go
```diff
@@ -99,6 +99,14 @@ func checkPush(ctx context.Context, workingDir, upstream, branch string) error {
 	return execGitCmd(ctx, args, gitCmdConfig{dir: workingDir})
 }
 
+func secretUnseal(ctx context.Context, workingDir string) error {
+	args := []string{"secret", "reveal", "-f"}
+	if err := execGitCmd(ctx, args, gitCmdConfig{dir: workingDir}); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("git secret reveal -f"))
```
hiddeco:
```suggestion
		return errors.Wrap(err, "git secret reveal -f")
```

No reason to use `fmt.Sprintf` for a plain string.

git/working.go
```diff
@@ -89,6 +90,12 @@ func (r *Repo) Clone(ctx context.Context, conf Config) (*Checkout, error) {
 	}
 	r.mu.RUnlock()
 
+	if _, err := os.Stat(repoDir + "/.gitsecret"); err == nil && conf.GitSecret {
```
hiddeco: I would make the `.gitsecret` folder configurable, as by looking at the git-secret documentation, the user can control this by setting an `$SECRETS_DIR` environment variable.

This would require you to add `SECRETS_DIR` to (and maybe also `SECRETS_EXTENSION`) https://github.com/weaveworks/flux/blob/86eed4de40d539704862505a6d912726b323176c/git/operations.go#L28

In addition, I do not think I would check if the directory is there as a requirement to attempt an unseal, as the user has explicitly instructed us to run it by setting the flag. Having an additional 'hidden' check may cause confusion, instead, I would just bail out on the error `secretUnseal` returns.

arsiesys:
> Awesome to see how this benefits from the initial work for GPG signatures, and a sought after feature! medal_military
> 
> Please take a look at my (minor) comments.

Thanks for your comments, I applied your recommendations! :)

cmd/fluxd/main.go
```diff
@@ -114,6 +114,7 @@ func main() {
 		gitEmail     = fs.String("git-email", "support@weave.works", "email to use as git committer")
 		gitSetAuthor = fs.Bool("git-set-author", false, "if set, the author of git commits will reflect the user who initiated the commit and will differ from the git committer.")
 		gitLabel     = fs.String("git-label", "", "label to keep track of sync progress; overrides both --git-sync-tag and --git-notes-ref")
+		gitSecret    = fs.Bool("git-secret", false, `enable "git-secret reveal -f", can be completed with SECRETS_EXTENSION and SECRETS_DIR environment varialbles (see git-secret documentation)`)
```
hiddeco:
```suggestion
		gitSecret    = fs.Bool("git-secret", false, "if set, git-secret will be run on every git checkout")
```

aarnaud:
Nice, I hope we will have this feature in the next release

git/working.go
```diff
@@ -89,6 +90,12 @@ func (r *Repo) Clone(ctx context.Context, conf Config) (*Checkout, error) {
 	}
 	r.mu.RUnlock()
 
+	if conf.GitSecret {
```
stefanprodan: I think Flux should log an error at startup if the `git-secret` is enabled but `git-gpg-key-import` is empty. I would also add to the `git-secret` flag description that importing the GPG key is required.

squaremo: I think it's possible, though a major pain, to mount a GPG keyring rather than importing keys. So _not_ using the import does not necessarily mean there are no keys for `git secret` to use.

On the other hand: I anticipate mistakes here, so I do agree with Stefan's advice -- downgrade it to a warning, perhaps.

hiddeco:
@arsiesys couple of days have gone by since the last review, are you still planning to incorporate the feedback from Stefan and Michael?

arsiesys:
Hello, yes, I will. A bit lack of time to do that lately ! 😬

hiddeco:
@arsiesys no worries, if it does not fit your schedule right now I'd be happy to make the adjustments. Just let me know.

git/working.go
arsiesys: Hello, thanks for your comments, I tried to implement the warn message and completed the description, is it correct like this?

stefanprodan: Yes thanks

squaremo:
Nice work @arsiesys !

davidkarlsen:
@stefanprodan should the chart be updated too? What do you need for it - simply a flag to activate `--git-secret` ?

stefanprodan:
@davidkarlsen for experimental features there is an additional flag in the chart: https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L265

hardvintage:
Can i ask a question here...
**I enabled the options:**
`- --git-secret=true
- --git-gpg-key-import=/root/.gpg/`
**Imported successfully**
`ts=2020-09-07T12:21:22.416999372Z caller=main.go:347 info="imported GPG key(s) from /root/.gpg/" files=[xxx.gpg]`
**Get ERROR**
`ts=2020-09-07T12:17:56.882295512Z caller=images.go:17 component=sync-loop msg="polling for new images for automated workloads"
ts=2020-09-07T12:17:56.973595547Z caller=images.go:23 component=sync-loop error="getting unlocked automated resources: git secret reveal -f: gpg: decryption failed: No secret key\ngit-secret: abort: problem decrypting file with gpg: exit code 2: /tmp/flux-working161304987/xxx/xxx/xxx/config/local.json\n"`

When open a shell of the pod , i can import the gpg key and do git reveal, could anybody help?

PR #2171, opened 2019-06-19
Implement configurable timeout for RPC operations

hiddeco:
This assures that in a scenario where for example the Kubernetes API
is temporary rate limited, requests that are taking too long get
abandoned, rather than piling up.

As we have observed excessive amount of goroutines getting spawned
in a short amount of time, sometimes resulting in a shutdown of the
Flux daemon due to the spike in CPU (and RAM) that comes with it.

### DIFF
--- a/cluster/kubernetes/kubernetes.go
+++ b/cluster/kubernetes/kubernetes.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"sync"
@@ -127,7 +128,7 @@ func NewCluster(client ExtendedClient, applier Applier, sshKeyRing ssh.KeyRing, 
 // SomeWorkloads returns the workloads named, missing out any that don't
 // exist in the cluster or aren't in an allowed namespace.
 // They do not necessarily have to be returned in the order requested.
-func (c *Cluster) SomeWorkloads(ids []flux.ResourceID) (res []cluster.Workload, err error) {
+func (c *Cluster) SomeWorkloads(ctx context.Context, ids []flux.ResourceID) (res []cluster.Workload, err error) {
 	var workloads []cluster.Workload
 	for _, id := range ids {
 		if !c.IsAllowedResource(id) {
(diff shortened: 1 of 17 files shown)


squaremo:
Adding a context to the RPCs is :100:

I think this will need to go further though. The problem being addressed (in large part) is that Kubernetes API requests pile up, once they start needing to wait for rate limiting. To be able to abandon those when a request times out, the context must be propagated to those calls (i.e., to the cluster.Cluster methods, then to the Kubernetes client methods).

Sadly, the client-go API doesn't support contexts. What I suggest is, for the meantime, is that a context is passed through, and checked before each Kubernetes API call (or anything that occurs in a loop, at least).

cluster/kubernetes/kubernetes.go
```diff
@@ -281,7 +282,7 @@ func (c *Cluster) PublicSSHKey(regenerate bool) (ssh.PublicKey, error) {
 // the Flux instance is expected to have access to and can look for resources inside of.
 // It returns a list of all namespaces unless an explicit list of allowed namespaces
 // has been set on the Cluster instance.
-func (c *Cluster) getAllowedAndExistingNamespaces() ([]apiv1.Namespace, error) {
+func (c *Cluster) getAllowedAndExistingNamespaces(ctx context.Context) ([]apiv1.Namespace, error) {
```
squaremo: There's not much point in passing the context down to the leaves unless it's either given to the remote calls, or at least consulted. Since client-go doesn't support contexts, the only choice is the latter: in other words, to check that the context hasn't closed before making remote calls. It may time out or be cancelled _during_ the operation (including the time spent waiting for the rate limiter), but at least further operations won't be attempted, so overall, pressure will be lower.

The easiest way to check the context in this circumstance will be to see if `ctx.Err()` is `nil` or not:

```
if err := ctx.Err(); err != nil {
  return nil, err
}
ns, err := c.client.CoreV1().Namespaces().Get(name, meta_v1.GetOptions{})
// ...
```

cluster/kubernetes/kubernetes.go
```diff
@@ -240,7 +241,7 @@ func (c *Cluster) Export() ([]byte, error) {
 		}
 
 		for _, resourceKind := range resourceKinds {
-			workloads, err := resourceKind.getWorkloads(c, ns.Name)
+			workloads, err := resourceKind.getWorkloads(ctx, c, ns.Name)
```
squaremo: For this one I think you can get away with checking here, rather than in each `getWorkload` method. Similarly `getWorkloads`.

hiddeco: That is what I first had, the reason I choose to not do this is that in case Flux gets extended in the future (i.e. some other action calling `getWorkload`), it is already covered.

squaremo: Fair enough, yup :+1:

PR #2184, opened 2019-06-24
Force fetch tags on checkout local working clone

### LINKED ISSUE #2169: Working clones occasionally balk at fetching sync tag
A user that has fluxd 1.12.3 connected to weave cloud occasionally sees an error message in the UI, represented in the log like this:

```
ts=2019-05-29T06:51:26.34694615Z caller=logging.go:42 component=upstream method=ListServices error="git fetch --tags /tmp/flux-gitclone212733495 [refs/notes/flux-flux:refs/notes/flux]: From /tmp/flux-gitclone212733495\n * [new ref]         refs/notes/flux -> refs/notes/flux\n ! [rejected]        flux-sync -> flux-sync (would clobber existing tag)\n"
```

(NB I have anonymised the name of the ref)
I think the chain of events is something along the lines of:

 1. The UI calls the API ListServices
 2. fluxd clones its repo mirror into a temporary directory, at HEAD, so it can examine the files
 3. as part of the clone, it fetches the sync tag from the mirror to the temp dir
 4. for whatever reason, between the clone and the subsequent fetch, the tag has changed

hiddeco:
In Git 2.20 "git fetch" was taught to forbid updates to existing tags
without the "--force" option[1]. As Flux force-pushes a tag all the
time to keep track of the sync state, this would sometimes lead to
'would clobber existing tag' errors while making sure a fresh checkout
was up-to-date with its mirror, as described issue 2169[2].

To prevent this from happening: force fetch all tags from the mirror
while creating the local working clone, and _before_ fetching anything
else.

NB: We only have to do this while creating a working clone from a
mirror, as a mirror (bare clone) will accept non-fast-forward tag
changes from remote.

[1]: https://github.com/git/git/blob/master/Documentation/RelNotes/2.20.0.txt#L67-L71
[2]: #2169

Fixes #2169

### DIFF
--- a/git/working.go
+++ b/git/working.go
@@ -82,6 +82,18 @@ func (r *Repo) Clone(ctx context.Context, conf Config) (*Checkout, error) {
 	}
 
 	r.mu.RLock()
+	// Here is where we mimic `git fetch --tags --force`, but
+	// _without_ overwriting head refs. This is only required for a
+	// `Checkout` and _not_ for `Repo` as (bare) mirrors will happily
+	// accept any ref changes to tags.
+	//
+	// NB: do this before any other fetch actions, as otherwise we may
+	// get an 'existing tag clobber' error back.
+	if err := fetch(ctx, repoDir, r.dir, `'+refs/tags/*:refs/tags/*'`); err != nil {
+		os.RemoveAll(repoDir)
+		r.mu.RUnlock()
+		return nil, err
+	}
 	if err := fetch(ctx, repoDir, r.dir, realNotesRef+":"+realNotesRef); err != nil {
 		os.RemoveAll(repoDir)
 		r.mu.RUnlock()


squaremo:
This makes me nervous because it will apply to all refs fetched from the origin, not just the tag. That makes me think perhaps we should pull the specific tag, with a `+`, separately to everything else.

hiddeco:
> This makes me nervous because it will apply to all refs fetched from the origin, not just the tag.

This is already the case if I am reading the man page correctly.

> Unlike when pushing with git-push(1), any updates
outside of refs/{tags,heads}/* will be accepted
without + in the refspec (or --force), whether
that’s swapping e.g. a tree object for a blob, or a
commit for another commit that’s doesn’t have the
previous commit as an ancestor etc.

squaremo:
> any updates outside of refs/{tags,heads}/* will be accepted
> without + in the refspec

Since `refs/heads/*` are branches, that statement does not include branches; i.e., git will only fast-forward branches, when fetching.

hiddeco:
@squaremo this should be it.

PR #2186, opened 2019-06-24
Remove Weave Cloudness from fluxctl usage

### LINKED ISSUE #2182: Standardise authentication for fluxctl
fluxctl has an (optional) authentication scheme that is useless for anything but Weave Cloud. It'd be much better if it supported a standard authentication scheme, so it has a chance of working with third party kit.

Just supporting a [bearer token in a standard `Authorization` header](https://tools.ietf.org/html/rfc6750) would be enough to start with.

squaremo:
Removes Weave Cloud defaults from fluxctl.

 - Use an RFC6750 bearer token when connecting to an upstream service (Weave Cloud supports these; but it's also an easy, generic thing to support in any other such service)
 - Changes the behaviour of fluxctl so that you must supply both a URL and token to connect to an upstream service. Previously it was presumed you were connecting to Weave Cloud if you supplied a token at all.

The latter change is mildly backward _in_compatible, so we should be careful about rolling this out. In particular, any scripts or people that supply a token but not a URL to fluxctl will find that they get a warning, and fluxctl will probably try to using port-forwarding and fail.

Fixes #2182.

### DIFF
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -37,12 +37,12 @@ Connecting:
   # To a fluxd running in namespace "default" in your current kubectl context
   fluxctl list-workloads
 
-  # To a fluxd running in namespace "weave" in your current kubectl context
-  fluxctl --k8s-fwd-ns=weave list-workloads
+  # To a fluxd running in namespace "flux" in your current kubectl context
+  fluxctl --k8s-fwd-ns=flux list-workloads
 
-  # To a Weave Cloud instance, with your instance token in $TOKEN
-  fluxctl --token $TOKEN list-workloads
-
+  # To an upstream service, with a bearer token in $TOKEN
+  fluxctl --url https://example.com/upstream/flux --token $TOKEN list-workload
+s
 Workflow:
   fluxctl list-workloads                                                   # Which workloads are running?
   fluxctl list-images --workload=default:deployment/foo                    # Which images are running/available?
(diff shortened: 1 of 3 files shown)


jwenz723:
Seems like another thing that could be done to fluxctl to make it more of a native kubernetes style experience would be to change the `--k8s-fwd-ns` flag to just `--namespace` or `-n` so it matches kubectl.

hiddeco:
@jwenz723 this is not possible as there already is a `--namespace` (and `-n`) selector for e.g. listing workloads in a namespace with `fluxctl`.

https://github.com/fluxcd/flux/blob/master/cmd/fluxctl/list_workloads_cmd.go#L33

2opremio:
@squaremo Are you planning to keep working on this one?

squaremo:
> Are you planning to keep working on this one?

If there are more things to do to decouple from Weave Cloud, I can add to this yes. But it is waiting for an opportunity to break backward compatibility, rather than to be completed.

yebyen:
Can this be made backward compatible, and are there other (known, open?) systems that fluxctl is integrated with by now?

Seems like it would be unlikely for this change to get merged at this point, as flux v1 is in maintenance mode, unless it can be done without breaking anything. Can we leave the default URL intact and simply provide an opportunity to override, along with the warning? It would be very bad to break anything for Weave Cloud customers in particular, I think at least.

There can be an opportunity to help with migration to Flux v2 here, if this deprecation warning message is updated to point to the migration guide once that is in place.

kingdonb:
I have connected Weave Cloud to see how it's done with the latest version of Flux v1 (assuming you don't allow the weave-flux-agent to decide which version of Flux to install), and I think this change only affects `fluxctl`

I came to this conclusion because it looks like `fluxd` has already had some/most or all Weave Cloudness removed, eg. Flux Daemon will not automatically connect to Weave Cloud servers unless they are passed in at runtime with `--connect=`

eg:

```
- args:
  - --connect=https://cloud.weave.works./api/flux
  - --token=aaaabbbbccccddddeeeeffffgggghhhh
```

So, this change only affects `fluxctl`. I think it can be closed, without action. But as I haven't had a chance to read the source carefully yet, I won't close it. I'd just like to understand it at least well enough to say why it is backwards incompatible before I close it. I will get there soon. (Leaving open for now.)

kingdonb:
@squaremo Do you think I should rebase this and try to merge it, or close it out?

It is missing the DCO check and has conflicts, I haven't looked at them yet.

kingdonb:
Related to #3052

kingdonb:
I am closing this out as it is described as mildly backward incompatible, I understand why this is true better now.

No backward-incompatible changes in Flux v1 while maintenance period continues. (After that, it will be archived.)

This cannot be merged, at present though Flux v2 deftly removes all weave cloudiness 👍

PR #2201, opened 2019-06-27
Move ResourceID from root to resource package

stefanprodan:
- move ResourceID from root to resource package
- move Update type from policy pkg to resource pkg to avoid circular dependency
- rename resource.Update to resource.PolicyUpdate
- rename ResourceID and related functions to ID
- run go fmt on all files


squaremo:
ResourceID should really be `resource.ID`. This PR tidies a file away, but since it just introduces another package, it's a big diff for a modest tidy.

resource/policy.go
```diff
@@ -45,3 +45,11 @@ func ChangesForPolicyUpdate(workload Workload, update policy.Update) (map[string
 	}
 	return result, nil
 }
+
+type Updates map[ID]Update
+
+type Update struct {
```
squaremo: This is a weird place to put this type. At the least, I'd expect it to be called `PolicyUpdate`. But it probably belongs in another package, rather than here. As `update.Policy`, if that doesn't create more problems, for example.

stefanprodan:
@squaremo I've went with renaming `resource.Update` to `resource.PolicyUpdate` and I've fixed the import order.

squaremo:
> I've went with renaming `resource.Update` to `resource.PolicyUpdate`

Was it impossible to do the preferred change, moving it into `update`?

stefanprodan:
@squaremo 

```
import cycle not allowed
package github.com/weaveworks/flux/cmd/fluxctl
	imports github.com/weaveworks/flux/api
	imports github.com/weaveworks/flux/api/v11
	imports github.com/weaveworks/flux/api/v10
	imports github.com/weaveworks/flux/api/v6
	imports github.com/weaveworks/flux/cluster
	imports github.com/weaveworks/flux/resource
	imports github.com/weaveworks/flux/update
	imports github.com/weaveworks/flux/cluster
```

squaremo:
> import cycle not allowed

Ugh, nightmare. OK.

squaremo:
:+1:

Rebass and squash the suggestion commit, and the Update re-rename.
In the future, consider keeping code changes, and reformatting, in separate commits.

PR #2208, opened 2019-06-29
Update Flux CD diagram

stefanprodan:
- update Flux logo and add GitOps logo
- add memcached and describe the image metadata sync flow
- explain pull vs push with `git sync` vs `kubectl apply` 

flux-cd-diagram


hiddeco:
Is it an idea to incorporate the Helm operator in this diagram, or an additional one? It is often not clear to people how the relationship between Flux and the Helm operator works precisely.

stefanprodan:
I will make another diagram for the Helm Operator but that will probably not be on Flux main readme but in the helm-op docs.

PR #2211, opened 2019-07-01
Flux Adapter preparation

### LINKED ISSUE #2210: Add Upstream API to regular API
At present, there are a small number of methods in the Upstream API that are not in the flux API. These are methods that can be called via RPC from an upstream (i.e., Weave Cloud) connection, but not via HTTP (e.g., not by `fluxctl`).

In the course of separating the Upstream part of the daemon out into weaveworks/flux-adapter, these will have to move into the public API, since `flux-adapter` runs in a sidecar, and uses HTTP (via http/client) rather than having in-process access to the daemon.

I believe the upstream methods can just be added to the v11 Server interface, and the Upstream interface abandoned from now on.

squaremo:
- makes API changes so that all methods are available over HTTP, including NotifyChange (used for relaying webhooks, for example) /cc @alanjcastonguay 
 - adds some notes about what can be moved out of this repo and into `weaveworks/flux-adapter`

This (modulo passing tests and review) can be merged now, in principle, because it doesn't alter what is served over the Upstream connection -- it just adds to the HTTP API. But I am going to make it a draft until I've Actually Tried that.

Fixes #2210.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -8,10 +8,3 @@ import "github.com/weaveworks/flux/api/v11"
 type Server interface {
 	v11.Server
 }
-
-// UpstreamServer is the interface a Flux must satisfy in order to communicate with
-// Weave Cloud.
-type UpstreamServer interface {
-	v11.Server
-	v11.Upstream
-}
--- a/api/v11/api.go
+++ b/api/v11/api.go
@@ -18,8 +18,8 @@ type Server interface {
 	v10.Server
 
 	ListServicesWithOptions(ctx context.Context, opts ListServicesOptions) ([]v6.ControllerStatus, error)
-}
 
-type Upstream interface {
+	// NB Upstream methods move into the public API, since
+	// weaveworks/flux-adapter now relies on the public API
 	v10.Upstream
 }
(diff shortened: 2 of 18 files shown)


alanjcastonguay:
Looks like you missed some of the `api.UpstreamServer` references in tests during the great renaming.

squaremo:
This PR _also_ updates the Upstream connection so it connects to the v11 endpoint. This was introduced as wee while ago to support rollout progress with ListServicesWithOptions, but back-filled in the (versioned) RPC clients so that it would work with older fluxd too.

Usually there's a gap necessary between introducing a new version and rolling out a Flux release that uses it, so that the service part of it can be rolled out in Weave Cloud -- so probably what happened is that, since it worked with the back-filled RPC client, the latter part of the rollout was forgotten.

In any case, I've now Actually Tried It with

 - a flux daemon connecting to Weave Cloud (v11 endpoint)
 - a flux daemon pointed at the flux adapter, itself connected to Weave Cloud

and both of these work. Specifically, I was looking for all the UI elements to work OK, and for notifications to be propagated to the service.

PR #2212, opened 2019-07-02
Simple draft of a website ... not serious

dholbach:
This is not meant for real, just exploratory and fun.

- [x] using main README.md in a symlink (untouched)
- [ ] theme wants logo to be a SVG, ours is a PNG (I just renamed it, that's a hack)
- [x] logo brings a UI glitch with it(?), or are the lines intentional?
- [ ] some URLs need changing
- [x] there's a NSOmar link in the top we can remove
- [x] symlink images, resize them


dholbach:
Screenshot from 2019-07-02 16-36-33

In terms of what **I** can do this is as good as it gets. :joy: 

What I tried to explore in this PR is how we could (with re-using all of our current content) have a single-page website that doesn't require us to do maintain separate bits for Github's main README.md and a website.

PR #2219, opened 2019-07-03
Support both new fluxcd and original annotations

### LINKED ISSUE #2217: Support alternative fluxcd.io annotations
The annotations that are interpreted by flux will need to change, since they include the domain `weave.works`. But we don't want to just replace them, since that will break many installations (e.g., by not ignoring something that is supposed to be ignored). So: for some fairly large window, we'll need to support both sets of annotations.

squaremo:
The new prefix for annotations is `fluxcd.io`; but, many files will
use `flux.weave.works`, and we don't want to suddenly not understand
the policies in those.

This commit adds support for interpreting and updating policies as
annotations with either prefix. When updating, if an annotation uses
the old prefix, that will be retained; otherwise, the new prefix is
used.

This change is backward-compatible, but we may want to keep it back until we're releasing other "decoupling" changes.

Fixes #2217.

### DIFF
--- a/cluster/kubernetes/policies.go
+++ b/cluster/kubernetes/policies.go
@@ -3,8 +3,6 @@ package kubernetes
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
-
 	kresource "github.com/weaveworks/flux/cluster/kubernetes/resource"
 	"github.com/weaveworks/flux/resource"
 )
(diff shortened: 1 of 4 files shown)


hiddeco:
Getting this in to finish #2249 with the new annotation domain (and to support updates of existing annotations).

PR #2220, opened 2019-07-04
Document Helm operator rollback feature

hiddeco:
Per request (https://github.com/weaveworks/flux/pull/2006#issuecomment-507331706), and because well, it was missing...

### DIFF
--- a/site/helm-integration.md
+++ b/site/helm-integration.md
@@ -392,9 +392,10 @@ apiVersion: flux.weave.works/v1beta1
 kind: HelmRelease
 # metadata: ...
 spec:
+  # Listed values are the defaults.
   rollback:
     # If set, will perform rollbacks for this release.
-    enable: true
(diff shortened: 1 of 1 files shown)


site/helm-integration.md
```diff
@@ -367,6 +369,46 @@ undesirable because it makes it hard to specify exactly what you want,
...
+
+### Configuration
+
+```yaml
+apiVersion: flux.weave.works/v1beta1
+kind: HelmRelease
+# metadata: ...
+spec:
+  rollback:
+    # If set, will perform rollbacks for this release.
```
stefanprodan: Can you add a new comment line to each field and specify what's the default value if the field is omitted from the spec?

hiddeco: As the values I filled in are the defaults, I solved it in a different way.

PR #2223, opened 2019-07-04
Enable podsecuritypolicy

ArchiFleKs:
Enable podsecuritypolicy with a permissive default.

-->

### DIFF
--- a/chart/flux/templates/psp-clusterrole.yaml
+++ b/chart/flux/templates/psp-clusterrole.yaml
@@ -1,0 +1,17 @@
+{{- if .Values.rbac.pspEnabled }}
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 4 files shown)


stefanprodan:
@ArchiFleKs thanks for the PR. How was this tested? If you set `privileged: false` in the PSP and you add a privileged pod to Flux repo does Flux logs the PSP deny error?

stefanprodan:
Ok I've tested this on EKS. Flux will not capture the PSP deny error as it happens at replicaset level:

```
kubectl -n team1 describe replicasets podinfo-5d7d9fc9d5

Error creating: pods "podinfo-5d7d9fc9d5-" is forbidden: unable to validate against any pod security policy: [spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
```

PR #2226, opened 2019-07-05
Helm chart: Fix PSP binding and api group rbac

stefanprodan:
Tested on EKS 1.13


stefanprodan:
@hiddeco sorry for all this spam, I should've tested the previous PR better.

PR #2234, opened 2019-07-09
Chart:  re-implement registry.ecr.require

pinkavaj:
This parameter was not implemented.

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -190,6 +190,9 @@ spec:
           - --registry-rps={{ .Values.registry.rps }}
           - --registry-burst={{ .Values.registry.burst }}
           - --registry-trace={{ .Values.registry.trace }}
+          {{- if .Values.registry.ecr.require }}
+          - --registry-require=ecr
+          {{- end }}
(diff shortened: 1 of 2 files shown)


chart/flux/templates/deployment.yaml
```diff
@@ -190,6 +190,9 @@ spec:
           - --registry-rps={{ .Values.registry.rps }}
           - --registry-burst={{ .Values.registry.burst }}
           - --registry-trace={{ .Values.registry.trace }}
+          {{- if registry.ecr.require }}
```
hiddeco:
```suggestion
          {{- if .Values.registry.ecr.require }}
```

chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.13.1"
-version: 0.10.2
+version: 0.10.3
```
hiddeco:
```suggestion
version: 0.10.2
```

We bump versions before making the chart release, to ensure we only bump it once per release.

pinkavaj: oook

hiddeco:
Looks like this got removed on accident in #1918.

PR #2236, opened 2019-07-10
Chart: support Helm operator `--workers` flag

hiddeco:


chart/flux/templates/helm-operator-deployment.yaml
```diff
@@ -126,6 +126,7 @@ spec:
         - --charts-sync-interval={{ .Values.helmOperator.chartsSyncInterval }}
         - --update-chart-deps={{ .Values.helmOperator.updateChartDeps }}
         - --log-release-diffs={{ .Values.helmOperator.logReleaseDiffs }}
+        - --workers={{ .Values.helmOperator.workers }}
```
stefanprodan: Please use an `if` here to make the change backwards compatible, otherwise if someone uses this chart with an older Flux version it will crash at startup.

hiddeco: Turbo patched.

PR #2248, opened 2019-07-11
parse credentials without scheme

fnmeissner:
When the host within the docker config has format IP:PORT
url.Parse(host) fails. This commit will first try re-parsing
with prepended scheme before returning any error.


squaremo:
Thanks for this! I'm going to give it a quick go locally, then will review here.

Issue #1262, opened 2018-08-01
Support even more patterns of specifying images in FluxHelmRelease


squaremo:
From @justinbarrick's testing of #1258:

> A couple of third party charts do not follow any of these patterns:
> brigade uses controller.registry and controller.name to form the image name.
> stable/influxdb uses image.repo instead of image.repository.
> stable/external-dns uses image.name.

These are probably fairly commonly used; even though we (presently) require people to vendor the charts, and they could therefore alter them, it's friendlier if they don't have to.

(This may be a good time to consider implementing some scheme so that people can specify the containers using some scheme, in the FluxHelmRelease.)

stephenmoloney:
One pattern I see in quite a few places is 

```
image:
  image:
  imageTag:
```

I don't think this is currently supported by flux.

reference: https://github.com/helm/charts/search?q=imageTag%3A&unscoped_q=imageTag%3A

billimek:
It would be great if there was a way to tell flux what pattern to use instead of relying on a set of patterns/strings hard-coded into flux.

For example, keel provides a mechanism to define 'where' the image name and image tag are defined: https://keel.sh/v1/guide/documentation.html#Helm-example

squaremo:
Yar, having a little language to interpret would be modestly more flexible that a set of understood patterns. (I was thinking of a similar thing for dealing with custom resources).

hiddeco:
Before we can turn this issue into a PR we must agree on how a user should instruct the operator to look for images at certain paths in the `values` value.

@stefanprodan iirc you had annotations in mind; can you elaborate on this 'design'?

stefanprodan:
I propose the following annotations:

```
repository.flux.weave.works/<arbitrary-name>: "path.to.image.repository"
tag.flux.weave.works/<arbitrary-name>: "path.to.image.tag"
filter.flux.weave.works/<arbitrary-name>: "glob|semver|regexp"
``` 

Here is how Jenkins could look:

```yaml
kind: HelmRelease
  annotations:
    flux.weave.works/automated: "true"
    repository.flux.weave.works/master: "Master.Image"
    tag.flux.weave.works/master: "Master.ImageTag"
    filter.flux.weave.works/master: semver:~2.60
    repository.flux.weave.works/agent: "Agent.Image"
    tag.flux.weave.works/agent: "Agent.ImageTag"
    filter.flux.weave.works/agent: glob:3.27-*
spec:
  chart:
[...]

PR #2249, opened 2019-07-11
Support defining image paths for HelmReleases

hiddeco:
This commit adds support for supplying image paths for `HelmRelease`
resources through annotations, and introduces a new alias for image
tag filters.

For an image mapping to be valid, the image path _must_ be defined
and _may_ be supplemented by a tag and/or registry path. In case
the combination of resolved values does not result in a valid image
ref, the mapping will be ignored.

**Supported annotations:**

`registry.fluxcd.io/<alias>: <registry path>`
`repository.fluxcd.io/<alias>: <image path>`
`tag.fluxcd.io/<alias>: <tag path>`
`filter.fluxcd.io/<alias>: <tag filter>`

Image mappings are composed by `<alias>`.

**A valid annotation mapping is for example:**

```yaml
annotations:
  registry.fluxcd.io/container: some.custom.registry
  repository.fluxcd.io/container: some.custom.image
  tag.fluxcd.io/container: some.custom.tag
  filter.fluxcd.io/container: glob: master-*

  repository.fluxcd.io/anotherone: depth.knows.no.limits.image
```

which would map the image for the following `values` structure:

```yaml
values:
  some:
    custom:
      registry: docker.io
      image: fluxcd/flux-prerelease
      tag: master-abcdefg
  depth:
    knows:
      no:
        limits:
          image: fluxcd/helm-operator:0.10.0
```

This PR is in need of ~https://github.com/squaremo/kubeyaml/pull/11~ https://github.com/squaremo/kubeyaml/pull/14 to persist the actual changes to git.

Fixes: #1262


2opremio:
Where, `container` a parameter, right? (i.e. does the scheme support multiple containers?) 

The use of subdomains (as opposed to e.g. a path) to denote what you are specifying seems a bit funky, but we probably can't do better due to the restrictions of the annotation syntax imposed by k8s.

Maybe we could add another subdomain to make it more clear? e.g.
`tag.pathspec.flux.weave.works`?

hiddeco:
> Where, `container` a parameter, right? (i.e. does the scheme support multiple containers?)

`container` in this case is just an alias for your collection of mappings (it represents one image definition), it can be anything and multiple mappings can be defined.

> The use of subdomains (as opposed to e.g. a path) to denote what you are specifying seems a bit funky, but we probably can't do better due to the restrictions of the annotation syntax imposed by k8s.

Correct, this is the most Kubernetes native approach to defining the mapping and is conform their specifications for annotations.

> Maybe we could add another subdomain to make it more clear? e.g.
`tag.pathspec.flux.weave.works`?

Can not decide if the clarity improvement would weight up to the annoyance of a really long annotation key.

cluster/kubernetes/resource/fluxhelmrelease.go
```diff
@@ -220,18 +327,109 @@ func interpretAsImage(m mapper) (image.Ref, ImageSetter, bool) {
...
+	}
+
+	imageValue := v.Path(cim.Repository).Data()
+	if img, ok := imageValue.(string); ok {
+		if cim.Registry == "" && cim.Tag == "" {
+			if imgRef, err := image.ParseRef(img); err == nil {
+				return imgRef, func(ref image.Ref) {
+					v.SetP(ref.String(), cim.Repository)
+				}, true
+			}
```
squaremo: If this fails, I'm not sure it's worth continuing -- adding a registry and/or tag to an invalid image can't make it valid.

squaremo: (it does get checked again later, so it's safe as it is I believe)

cluster/kubernetes/resource/fluxhelmrelease.go
```diff
@@ -220,18 +327,109 @@ func interpretAsImage(m mapper) (image.Ref, ImageSetter, bool) {
...
+						v.SetP(ref.Tag, cim.Tag)
+					}, true
+				}
+			}
+		}
+	}
+
+	return image.Ref{}, nil, false
+}
+
```
squaremo: :heavy_check_mark: although I should mention it took me quite a lot of staring to get to this point :-)

cluster/kubernetes/resource/fluxhelmrelease.go
```diff
@@ -220,18 +327,109 @@ func interpretAsImage(m mapper) (image.Ref, ImageSetter, bool) {
...
 // Containers returns the containers that are defined in the
 // FluxHelmRelease.
 func (fhr FluxHelmRelease) Containers() []resource.Container {
 	var containers []resource.Container
-	// If there's an error in interpreting, return what we have.
-	_ = FindFluxHelmReleaseContainers(fhr.Spec.Values, func(container string, image image.Ref, _ ImageSetter) error {
+
+	containerSetter := func(container string, image image.Ref, _ ImageSetter) error {
```
squaremo: If you're giving this a name, `addContainer` might be better -- the implied analogy to `imageSetter` doesn't really work.

cluster/kubernetes/resource/resource.go
```diff
@@ -198,7 +205,11 @@ func unmarshalKind(base baseObject, bytes []byte) (KubeManifest, error) {
 		return &list, nil
 	case base.Kind == "FluxHelmRelease" || base.Kind == "HelmRelease":
 		var fhr = FluxHelmRelease{baseObject: base}
-		if err := yaml.Unmarshal(bytes, &fhr); err != nil {
+		// NB: workaround for go-yaml/yaml/issues/139
+		// By using github.com/ghodss/yaml to unmarshal HelmReleases.
+		// This effectively results in all keys of `Value`s being strings
+		// and not interface{}.
+		if err := jsonyaml.Unmarshal(bytes, &fhr); err != nil {
```
squaremo: Does this mean the `mapper` stuff above can be removed?

cluster/kubernetes/resource/fluxhelmrelease.go
```diff
@@ -220,18 +327,109 @@ func interpretAsImage(m mapper) (image.Ref, ImageSetter, bool) {
...
+
+func interpretMappedContainerImage(values map[string]interface{}, cim ContainerImageMap) (image.Ref, ImageSetter, bool) {
+	v, err := gabs.Consume(values)
+	if err != nil {
+		return image.Ref{}, nil, false
+	}
+
+	imageValue := v.Path(cim.Repository).Data()
+	if img, ok := imageValue.(string); ok {
+		if cim.Registry == "" && cim.Tag == "" {
```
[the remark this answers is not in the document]
hiddeco: That will definitely make it easier to understand. I added the utilities around `ContainerImageMap` much later and did not rethink if I could also improve this. Thanks for the suggestion.

cluster/kubernetes/manifests.go
```diff
@@ -133,7 +133,34 @@ func (m *manifests) ParseManifest(def []byte, source string) (map[string]resourc
...
+	// in to a valid image ref), fallback to the default behaviour.
+	//
+	// NB: we do this here and not in e.g. the `resource` package, to ensure
+	// everything _outside_ this package only knows about Kubernetes native
+	// containers, and not the dot notation YAML paths we invented for custom
+	// Helm value structures.
+	if hr, ok := res.(*kresource.FluxHelmRelease); ok {
+		if paths, err := hr.GetContainerImageMap(container); err == nil {
+			def, err := updateWorkloadImagePaths(def, id, paths, image)
+			if err == nil {
```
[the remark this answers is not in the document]
hiddeco: Agreed.

cluster/kubernetes/resource/resource.go
hiddeco: Above where? Quite a lot of 'mapper' things on this page...

PR #2253, opened 2019-07-12
Added a simple kustomize.yaml for deploying flux

jmtrusona:
Added a simple `kustomization.yaml` which defines the flux manifests as resources. This will allow others to use this repo as a remote base and use `kustomize` to apply their desired configuration instead of having to duplicate the manifests in their own repos.

For example:

```
DEMO=$(mktemp -d)

cat <<EOF >$DEMO/kustomization.yaml
bases:
- git@github.com:fluxcd/flux//deploy-kustomize

patchesStrategicMerge:
- args-patch.yaml
EOF

cat <<EOF >$DEMO/args-patch.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: flux
spec:
  template:
    spec:
      containers:
      - name: flux
        args:
        - --git-url=git@github.com:example/manifests
EOF
```

### DIFF
--- a/deploy/kustomization.yaml
+++ b/deploy/kustomization.yaml
@@ -1,0 +1,6 @@
+resources:
+- flux-account.yaml
+- flux-deployment.yaml
+- flux-secret.yaml
+- memcache-dep.yaml
+- memcache-svc.yaml
--- a/deploy-kustomize/kustomization.yaml
+++ b/deploy-kustomize/kustomization.yaml
@@ -1,0 +1,6 @@
+resources:
+- ../deploy/flux-account.yaml
+- ../deploy/flux-deployment.yaml
+- ../deploy/flux-secret.yaml
+- ../deploy/memcache-dep.yaml
+- ../deploy/memcache-svc.yaml
--- a/deploy-kustomize/kustomization.yaml
+++ b/deploy-kustomize/kustomization.yaml
@@ -1,6 +1,0 @@
-resources:
-- ../deploy/flux-account.yaml
-- ../deploy/flux-deployment.yaml
-- ../deploy/flux-secret.yaml
-- ../deploy/memcache-dep.yaml
-- ../deploy/memcache-svc.yaml


2opremio:
Would it make sense to have a similar directory for the helm operator?

jmtrusona:
> Would it make sense to have a similar directory for the helm operator?

Uh I haven't messed with the Helm stuff at all (don't really trust that tiller situation...) so I'll let others that may have more insight address that question. 

I'm open to any changes or additions to the PR that you'll feel is needed though.

stefanprodan:
That args patch will crash Flux since there are a couple of args that are required see https://github.com/stefanprodan/fluxcd-multi-tenancy

I would place the kustomization inside the deploy dir instead of creating a new one and document the args patch by adding all Flux arguments to it, similar to what we have in chart/flux/values.yaml

jmtrusona:
> That args patch will crash Flux since there are a couple of args that are required see https://github.com/stefanprodan/fluxcd-multi-tenancy
> 
> I would place the kustomization inside the deploy dir instead of creating a new one and document the args patch by adding all Flux arguments to it, similar to what we have in chart/flux/values.yaml

Thank you for the feedback. That `fluxcd-multi-tenancy` repo has some really nice documentation. 

A (sort of) quick side note that you may find interesting; our use case is similar but kind of the opposite of the `fluxcd-multi-tenancy` repo. We only have a single team but multiple clusters. Currently, we are not using `kustomize` and have a manifest repo per cluster which has made even the simplest changes like updating the `regexp` pattern on the flux annotation a large undertaking that touches many files across multiple repos. So the first step we're looking at is consolidating to a singular manifest repo with a different `--git-path` set on the fluxcd deployment for each cluster. The key to getting it to work was making sure that each deployment had a unique `--git-label` set as the deployments would fight over the `flux-sync` tag when a new image was pushed. Also, to make the transition to `kustomize` smoother, we're looking at running two fluxcd deployments per cluster (one for static manifests and one configured for `kustomize`) which now I'm thinking I will revisit my approach after reading your docs.  If I can find some time, I'd like to write up the approach we're taking in similar fashion as to what you've done.

Alright back to the review, I'll definitely move the `kustomization.yaml` to the `deploy` directory. I was just including that single `--git-repo` arg to keep the example in this pull request comment small and simple. Here's a more verbose version which is currently working:

```
DEMO=$(mktemp -d)

cat <<EOF >$DEMO/kustomization.yaml
bases:
- git@github.com:fluxcd/flux//deploy-kustomize

patchesStrategicMerge:
- args-patch.yaml
EOF

cat <<EOF >$DEMO/args-patch.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: flux
spec:
  template:
    spec:
      containers:
      - name: flux
        args:
        - --sync-interval=60s
        - --git-poll-interval=30s
        - --manifest-generation=true
        - --git-url=git@github.com:example/manifests
        - --git-branch=master
        - --git-path=kustomize/clusters/foo
        - --git-label=flux-sync-foo
        - --listen-metrics=:3031
EOF
```

I don't know that it makes sense to duplicate the documentation for all the fluxcd deployment arguments. There's a significant difference between the approach of documenting a `values.yaml` which is defining new variables (hence the need for documentation) for use within the Helm Chart templates and a `kustomize` overlay which should be exactly the same as the [deployment args documentation](https://github.com/fluxcd/flux/blob/master/deploy/flux-deployment.yaml#L107). I may need some clarification on that before knowing what steps to take next.

Along the same lines, were you thinking that `args-patch.yaml` should be something that's actually checked into the fluxcd repo? If so, do you have any ideas on an approach for that? 

I could add it as a long comment in the `deploy/kustomization.yaml` but it does involve multiple files so that may be a little confusing. I took a stab at it and didn't really like how it looked.

Just to reiterate, my intent for the scope of this PR was to keep to just add a simple `kustomization.yaml` that defines the resources so that it can be referenced as a remote base so that there isn't a need to duplicate the fluxcd deployment manifests in every manifest repo.

Thanks for your time!

jmtrusona:
Hey @stefanprodan,  just wanted to check back in on this review since I hadn't heard anything in a while and see if there were still any questions. I'm not a fan of letting a PR get stale so I'll just close it out this weekend if I don't hear back about it by then. 

Thanks

hiddeco:
Given @stefanprodan was on vacation last week, I am reopening this.

deploy/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+resources:
+- flux-account.yaml
+- flux-deployment.yaml
+- flux-secret.yaml
+- memcache-dep.yaml
+- memcache-svc.yaml
```
stefanprodan:
```suggestion
- memcache-svc.yaml

```

stefanprodan:
I think we should add a `Kustomize installer` section to Flux docs and have there a patch example, e.g. https://docs.flagger.app/install/flagger-install-on-kubernetes#install-flagger-with-kustomize

PR #2255, opened 2019-07-15
Make `fluxctl` respect the kube config context NS

### LINKED ISSUE #1877: Fluxctl should respect the namespace set in your current kube config context
Here are Under Armour we work with a number of kubernetes clusters and have 10-15 namespaces in each cluster so we heavily use `kubectx` and `kubens` (https://github.com/ahmetb/kubectx) to switch around from cluster to cluster and namespace to namespace.  It is really nice to know that kubectl will respect which cluster and namespace is currently configured in your kube context so you can easily do `kubectl get deployment` and only see the data for that specific cluster & namespace that is active in your kube context.

Fluxctl seems to respect the cluster portion of the context but not the namespace leaving you have to have constantly add an extra argument to your fluxctl commands like this `fluxctl list-controllers --namespace=ua-prod` or `fluxctl list-controllers --namespace=ua-stage`

I would like to request that fluxctl also respect the namespace configured in your kube context like it does for the cluster so you can use the following workflow
```
> kubectx us-red
> kubens ua-stage
> fluxctl list-controllers
... be show the controllers in the `us-red` cluster and `ua-stage` namespace
> kubens ua-prod
> fluxctl list-controllers
... be show the controllers in the `us-red` cluster and `ua-prod` namespace
```

Cheers

hiddeco:
By looking at the configured namespace in the active context of the
resolved config, if none is found we return the given `defaultNamespace`.

Fixes #1877

### DIFF
--- a/cmd/fluxctl/args.go
+++ b/cmd/fluxctl/args.go
@@ -7,6 +7,8 @@ import (
 	"os/exec"
 	"strings"
 
+	"k8s.io/client-go/tools/clientcmd"
+
 	"github.com/spf13/cobra"
 
 	"github.com/weaveworks/flux/update"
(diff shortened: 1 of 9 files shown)


squaremo:
https://github.com/fluxcd/flux/blob/master/cluster/kubernetes/namespacer.go#L40 has an example of doing this using k8s.io/client-go rather than shelling out, which may be a bit neater.

hiddeco:
> We should probably put this in a minor version release, wdyt?

Yep, as it will output completely different things than what it used to if a user has an active kube config with a namespace set.

PR #2256, opened 2019-07-15
Log correct `--registry-require` flag (remove D from end)

pinkavaj:


hiddeco:
I am reading your commit message as `--registry-require` _should be_ `--registry-required`, while it is the other way around. I would suggest to changing it to something like:

```
Log correct `--registry-require` flag (remove D from end)
```

PR #2257, opened 2019-07-15
Update contact info in README

squaremo:
This updates _most_ contact info and other links to point to the new repo or new Slack channel.

TODO

 - [ ] Get a replacement contact email for CoC


dholbach:
Cortex for example is very brief when it comes to the CoC, there's simply https://github.com/cortexproject/cortex/blob/master/code-of-conduct.md in the repository.

dholbach:
Not sure we'll need to keep listing Alexis. Maybe let's just point to the MAINTAINERS file instead? I'm sure all of you are available when it should come to any CoC violations.

squaremo:
Re CoC, I think we should do as cortex does

PR #2266, opened 2019-07-17
Let a named cluster role be used in chart

squaremo:
If you are stamping the Helm chart into a cluster a few times, it's
good to be able to refer to a single cluster role, rather than
creating one for each stamping.

This commit lets you supply a name for the cluster role to refer to;
to refer to an existing cluster role rather than creating one, use

    --set clusterRole.create=false --set clusterRole.name=pre-existing


squaremo:
Stefan, one for when you're back :-)

PR #2270, opened 2019-07-18
Fix file name in .flux.yaml docs

carlpett:
I'm assuming this is a typo, at least?


carlpett:
Build failure seems to be quite unrelated (cannot install packages in docker image) - how can I retrigger?

PR #2279, opened 2019-07-22
Enhance Git proxy support

### LINKED ISSUE #2179: Support git checkout over http proxy
**Describe the feature**
See https://gitolite.com/git-over-proxy.html
If socat could be included in the image then checkout against github would be supported for environments locked down behind a corporate proxy. This would also mean exposing an option to configure /root/.ssh/config via the helm-chart.

0. install flux
1. configure ssh-config to use socat

Feel free to add a diagram if that helps explaining things.

**Expected behavior**
Checkouts would go via http-proxy

hiddeco:
_For our beloved enterprise users._

By adding `socat` to the Flux and Helm operator image and whitelisting `GIT_PROXY_COMMAND` inheritance.

Ref: https://gitolite.com/git-over-proxy.html

Fixes #2179

### DIFF
--- a/docker/Dockerfile.flux
+++ b/docker/Dockerfile.flux
@@ -2,7 +2,7 @@ FROM alpine:3.9
 
 WORKDIR /home/flux
 
-RUN apk add --no-cache openssh-client ca-certificates tini 'git>=2.12.0' 'gnutls>=3.6.7' gnupg gawk
+RUN apk add --no-cache openssh-client ca-certificates tini 'git>=2.12.0' 'gnutls>=3.6.7' gnupg gawk socat
 RUN apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/testing git-secret
 
 # Add git hosts to known hosts file so we can use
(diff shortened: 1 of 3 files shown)


stefanprodan:
@hiddeco should we add this to helm-op?

hiddeco:
@stefanprodan I forgot for a moment that the Helm operator is of course suffering from the same issue. Will adapt the commit.

squaremo:
I'm not sure how it would work to use the "proxying the git protocol" technique from the reference; but the "proxying the ssh protocol" could work (requiring `socat` and an extra stanza in ssh_config).

I don't think the extra env entry is necessary for proxying SSH. But it doesn't hurt, and maybe it is possibly to use it some other way.

hiddeco:
> I'm not sure how it would work to use the "proxying the git protocol" technique from the reference; but the "proxying the ssh protocol" could work (requiring socat and an extra stanza in ssh_config).

Just mount is as a config map and set the environment variable to the mount path? Git will add additional parameters to the command it calls when `GIT_PROXY_COMMAND` is set (`$1` = hostname, `$2` = port).

PR #2287, opened 2019-07-23
Add `fluxctl install` command

2opremio:
Simlarly to [`linkerd install`](https://linkerd.io/2/reference/cli/install/), `fluxctl install` is meant to print out and tweak the Kubernetes manifests needed to install Flux in a cluster.

Its output is intended for piping into other commands such as:

```
# Install Flux and make it use Git repository git@github.com:<your username>/flux-get-started
fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' | kubectl -f -
```


cmd/fluxctl/setup_cmd.go
```diff
@@ -0,0 +1,101 @@
...
+		"Stop for amending the Flux manifests before pushing them to the Git repository")
+	cmd.Flags().StringSliceVarP(&opts.additionalFluxArgs, "extra-flux-args", "e", []string{},
+		"Additional arguments for Flux as CSVs, e.g. --extra-flux-arg='--manifest-generation=true,--sync-garbage-collection=true'")
+	return cmd
+}
+
+func (opts *setupOpts) RunE(cmd *cobra.Command, args []string) error {
+	//0. Read and patch embedded deploy/* manifests with what was passed to ops
+
+	// TODO. BTW, it's probably easier to embed chart/flux/* and use `helm template` as a library instead
```
2opremio: That's the same approach used by`linkerd install`: https://github.com/linkerd/linkerd2/blob/edge-19.7.4/cli/cmd/install.go#L745

hiddeco:
My vote goes to a `fluxctl setup` which outputs the YAMLs to `stdout`, based on a _released version_ of our Helm chart. Re-using the Helm chart means we do not have to maintain multiple templates, and a released version because:

1. The chart and the features it supports are on a different release schedule than `fluxd` and `fluxctl` releases. I think it is desirable to keep those on a separate release schedule, as this ensures complete freedom for both elements.
2. We have been talking about moving `chart/` to a dedicated repository, and to split the current chart into a `fluxd` and Helm operator chart, this would mean you can no longer embed from `chart/flux/*`.

2opremio:
Alright, will do. How about the command line options? Is that acceptable? I think linkerd does a more sophisticated thing, inspecting and exposing all the available chart options, but I think it's a bit overkill. I would rather keep it simple.

hiddeco:
@2opremio the `--extra-flux-args` is going to be tricky, as the Helm chart offers an `extraArgs` key, but also has some defaults / dedicated keys for arguments that are missing in this command.

The Helm operator setting the default and a user supplying the argument here will have the desired effect (as Flux takes the value of the last supplied argument), but will not return a pretty YAML, which is probably a requirement.

squaremo:
Yep, simple for now!

On command-line options: those you have are a good base. You can remove `--git-flux-subdir`, `--timeout`, `--amend` since they don't pertain to generating a config.

Since people can mess around with the output as much as they want, it's not that important to get all the bells and whistles in, or to have an escape for extra args. But looking ahead a bit, I would like to see

 - `--git-committer` (or `--git-user` and `--git-email`), since that will soon become mandatory
 - `--git-path` (as a string slice), since I expect it to be used a lot

as arguments.

squaremo:
Versioning is tricky isn't it.

If you build in whatever's in the files (either the chart or example deployment) at the time of release, that's probably going to have most of the bells and whistles from fluxd in that release -- but it won't correspond to a released chart, which always comes  after a fluxd release.

If you include the last released chart, in general it _won't_ have arguments for fluxd in the same release.

If you try to use the latest chart, fluxctl has to understand, for all supported versions -- which will have to include future versions, since _see above_ -- how to encode the arguments it's given into Helm values.

---

With those points in mind, here's an idea: make the deploy/ examples into very simple text templates, along with all the comments and commented-out bits, and bundle those into the fluxctl binary.

Pros:

 - will always (barring mistakes) be up-to-date with respect to the released fluxd
 - the output will retain all the useful comments and optional bits
 - easier to make it match the exact arguments supported by `fluxctl setup`

Cons:

 - a bit proprietary?

2opremio:
> With those points in mind, here's an idea: make the deploy/ examples into very simple text templates, along with all the comments and commented-out bits, and bundle those into the fluxctl binary.

How about using Kustomize as a library and use generated-patches to modify the manifests on deploy/* on the fly? This will allow us leave the the manifest files unchanged and save us from maintaining separate templates (or invent an annotation language to maintain the existing templates)

2opremio:
Or, we can create `deploy-templates/` , a Go-templated version of `deploy/` for the parameters used by `fluxctl setup`. We can embed those files in `fluxctl` and generate `deploy/` with the existing values during the build.

squaremo:
> How about using Kustomize as a library and use generated-patches to modify the manifests on deploy/* on the fly?

Pros:

 - not proprietary
 - serves secondary purpose as an example of how to configure flux using kustomize

Cons:

 - will wipe out the comments and commented-out sections, so the output won't be self-documenting

squaremo:
> Or, we can create `deploy-templates/` , a Go-templated version of `deploy/`

I reckon (either for this or for the kustomize version) just replace it. It doesn't especially serve a purpose other than as a starting point, now -- you can't just apply what's in there, for instance.

2opremio:
> serves secondary purpose as an example of how to configure flux using kustomize

My idea was to apply the patches on the fly and print the patches manifests.

2opremio:
> It doesn't especially serve a purpose other than as a starting point, now -- you can't just apply what's in there, for instance.

I did that all the time when developing (after tweaking the parameters). Also, the guides rely on it.

squaremo:
> I did that all the time when developing (after tweaking the parameters). Also, the guides rely on it.

Right, you had to tweak it, i.e., edit it. You would still be able to do that, especially if it's a kustomize base.

2opremio:
Ah, but you are saying there is no need for that after `fluxctl install`.

chart/flux/README.md
```diff
@@ -211,8 +211,8 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `git.url`                                         | `None`                                               | URL of git repo with Kubernetes manifests
 | `git.branch`                                      | `master`                                             | Branch of git repo to use for Kubernetes manifests
 | `git.path`                                        | `None`                                               | Path within git repo to locate Kubernetes manifests (relative path)
-| `git.user`                                        | `Weave Flux`                                         | Username to use as git committer
-| `git.email`                                       | `support@weave.works`                                | Email to use as git committer
+| `git.user`                                        | `Flux`                                               | Username to use as git committer
+| `git.email`                                       | `flux-dev@googlegroups.com`                          | Email to use as git committer
```
squaremo: This is a breaking change*, so I don't recommend including it in this PR.

*Why? Because if you installed the chart without supplying a value, installing the new chart will change the git committer.

cmd/fluxctl/install_cmd.go
```diff
@@ -28,74 +22,32 @@ func newInstall() *installOpts {
...
-	cmd.Flags().StringVarP(&opts.gitFluxPath, "git-flux-subdir", "p", "flux/",
-		"Directory within the Git repository where to commit the Flux manifests")
-	cmd.Flags().DurationVarP(&opts.timeout, "timeout", "t", 10*time.Second,
-		"Timeout duration for I/O operations")
-	cmd.Flags().StringVarP(&opts.namespace, "namespace", "n", "flux",
+	cmd.Flags().StringVarP(&opts.gitLabel, "git-label", "", "flux",
+		"Git label to keep track of Flux's sync progress; overrides both --git-sync-tag and --git-notes-ref")
+	cmd.Flags().StringVarP(&opts.gitUser, "git-user", "", "Flux",
+		"Username to use as git committer")
+	cmd.Flags().StringVarP(&opts.gitEmail, "git-email", "", "flux-dev@googlegroups.com",
```
squaremo: This should be mandatory, rather than defaulting.

cmd/fluxctl/install_cmd.go
```diff
@@ -28,74 +22,32 @@ func newInstall() *installOpts {
...
+		"Email to use as git committer")
+	cmd.Flags().StringVarP(&opts.namespace, "namespace", "", "flux",
 		"Cluster namespace where to install flux")
-	cmd.Flags().BoolVarP(&opts.amend, "amend", "a", false,
-		"Stop for amending the Flux manifests before pushing them to the Git repository")
-	cmd.Flags().StringSliceVarP(&opts.additionalFluxArgs, "extra-flux-args", "e", []string{},
+	cmd.Flags().StringSliceVarP(&opts.additionalFluxArgs, "extra-flux-args", "", []string{},
```
squaremo: Remove this. People can edit the output as well as they can go find the option they want and encode here.

chart/flux/README.md
squaremo: Similarly all the arguments etc below.

cmd/fluxctl/install_cmd.go
2opremio: Is it OK if we keep the functionality internally? (i.e. not exposed in a flag but accessible programmatically through a Go import). I am planning to use this in `eksctl`.

2opremio: To be frank I think I would find it super useful as a user. Flux arguments are the one thing which users will want to change all the time but it's not worth being exhaustive with all the options offered (i.e. I think that `--extra-flux-args` are a good compromise).

chart/flux/README.md
2opremio: Oh, right. So much for fixing things on the way :)

squaremo: Yeah I know, sorry about that!

cmd/fluxctl/install_cmd.go
squaremo: Yes to that first question.

On the second comment, I suppose you might want to script the generation of flux configs, and for that an escape would be more useful (because you'd figure out the value for it once, and run it many times).

For interactive use, it would surely be easier to generate the file, then edit it and figure out which things you need to add and update.

Anyway: the bit you need is to have the escape available in the code, so neither of us have to be right or wrong here :-)

cmd/fluxctl/install/install.go
```diff
@@ -0,0 +1,54 @@
+package install
```
squaremo: This should probably go under the top directory, if you want to use it from elsewhere.

2opremio: makes sense

cmd/fluxctl/install_cmd.go
2opremio: Would it make sense to default this to what's in the user'd the git configuration (if found)?

hiddeco: Don't think it would, most of the time this is set to the email address / account of some GitHub bot.

2opremio:
I think I have addressed all the comments. PTAL ASAP . Thanks!

site/annotations-tutorial.md
```diff
@@ -11,37 +11,12 @@ deployment](https://github.com/weaveworks/flux-get-started) and click on the
 
 ## Setup
 
-Get the source code of Flux:
+First, please [install fluxctl](fluxctl.md#installing-fluxctl).
```
hiddeco: Can you put these (doc related) changes in a separate commit / PR, so we are able to merge them _after_ the `fluxctl` release has been made?

2opremio: Sure

2opremio: Done!

deploy/flux-deployment.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   template:
     metadata:
       annotations:
-        prometheus.io/port: "3031" # tell prometheus to scrape /metrics endpoint's port.
+        prometheus.io.port: "3031" # tell prometheus to scrape /metrics endpoint's port.
```
2opremio: What the heck? The template has a slash.

2opremio: Ahhh, this happened during rebasing. I didn't incorporate the changes made on deploy/

squaremo:
> WIP: Add fluxctl setup

This could have done with a bit of a rebass before merging :-/ Never mind, done now.

PR #2292, opened 2019-07-24
Fix values merge in helm-operator

### LINKED ISSUE #1696: mergeValues doesn't merge values inside of maps
Running flux 1.9.0 and flux-helm-operator 0.5.3

I ran into this issue while trying to configure a HelmRelease that combines values from the `values` key with sensitive values from a `valuesFileSecret`. As an example, given the following values:
```
# values.yaml in a valuesFileSecret:
mychart:
  api_client_id: CLIENT_ID
  api_secret: SECRET_TOKEN
---
# spec.values in a HelmRelease:
mychart:
  image: my-registry/myapp:1.0.0
  ...
```
I was expecting that these would be combined like so before being applied to the helm release:
```
mychart:
  api_client_id: CLIENT_ID
  api_secret: SECRET_TOKEN
  image: my-registry/myapp:1.0.0
  ...
```
However, the resulting values reported by `helm get` _only_ include the keys in spec.values. This merging works for top-level keys in values YAMLs, so it only seems to affect keys inside of maps. Looking at the implementation of `mergeValues` in `flux/integrations/helm/release/release.go:274` I get the impression that keys inside of maps were intended to be combined this way. 

I'm able to work around this issue by avoiding these kinds of merges but it would be nice to be able to combine values this way. For instance, I could use a separate valueFileSecret for each set of API tokens.

obiesmans:
Fix values merge in helm-operator.

Values merges in `helm-operator` will overwrite instead of merge a dict that has it's keys
defined in more than two sources (e.g : Values and multiple ValuesFrom).

This happens because when merging, https://github.com/fluxcd/flux/blob/6077df2f010a107669a44244536f7fbf65eaa3d7/integrations/helm/release/release.go#L507-L514
first iteration makes `dest[k]` a `chartutil.Values` and then at the next iteration will see `isMap` evaluate to False, overwriting instead of merging the two maps.

```
type Values map[string]interface{}

func main() {
    var v Values
    _, ok := interface{}(v).(map[string]interface{})
    fmt.Print(ok) // ok is false
}
```

I wrote a unit test that feed Values from a HelmRelease, a Configmap and a
Secret to make sure it fails, then changed `mergValues` signature to make it pass.

```
go test -v ./integrations/helm/release/
=== RUN   TestValues
--- FAIL: TestValues (0.00s)
    release_test.go:75:
        	Error Trace:	release_test.go:75
        	Error:      	Expected value not to be nil.
        	Test:       	TestValues
    release_test.go:76:
        	Error Trace:	release_test.go:76
        	Error:      	Expected value not to be nil.
        	Test:       	TestValues
FAIL
FAIL	github.com/weaveworks/flux/integrations/helm/release	0.498s
```

After the signature change :
```
go test -v ./integrations/helm/release/
=== RUN   TestValues
--- PASS: TestValues (0.00s)
PASS
ok  	github.com/weaveworks/flux/integrations/helm/release	0.242s
```

Fixes #1696.

### DIFF
--- a/integrations/helm/release/release.go
+++ b/integrations/helm/release/release.go
@@ -488,9 +488,9 @@ func fhrResourceID(fhr flux_v1beta1.HelmRelease) resource.ID {
 	return resource.MakeID(fhr.Namespace, "HelmRelease", fhr.Name)
 }
 
-// Merges source and destination `chartutils.Values`, preferring values from the source Values
+// Merges source and destination map, preferring values from the source Values
 // This is slightly adapted from https://github.com/helm/helm/blob/2332b480c9cb70a0d8a85247992d6155fbe82416/cmd/helm/install.go#L359
-func mergeValues(dest, src chartutil.Values) chartutil.Values {
+func mergeValues(dest, src map[string]interface{}) map[string]interface{} {
 	for k, v := range src {
 		// If the key doesn't exist already, then just set the key to that value
 		if _, exists := dest[k]; !exists {
--- a/integrations/helm/release/release_test.go
+++ b/integrations/helm/release/release_test.go
@@ -1,0 +1,77 @@
+package release
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	flux_v1beta1 "github.com/weaveworks/flux/integrations/apis/flux.weave.works/v1beta1"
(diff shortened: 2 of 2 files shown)


hiddeco:
This PR fixes #1696.

PR #2152, opened 2019-06-12
Add read-the-docs support

dholbach:
After talking to @stefanprodan I realised that versioned docs won't work easily with Gitbook, we'd also just have one free repo.

read-the-docs might be the way to go. Here's a WIP PR to get this into shape.

To run this, do:
```sh
# in top-level directory
make serve-docs
```

### Content
- [x] fix index
- [x] use fonts we use in other places
- [x] break up docs into subdirectories
- [x] solve issues around over-flowing text boxes
- [x] find solution for [bla.md#anchor] links (https://github.com/rtfd/recommonmark/issues/8)
- [x] find solution for links in tables (https://github.com/ryanfox/sphinx-markdown-tables/issues/18)
- [x] Remove the "next read ..." mention in the docs - this is in the RTD UI already
- [ ] factor in [Hidde's feedback](https://github.com/weaveworks/flux/pull/2152#issuecomment-502773997)
- [ ] Helm Operator docs should be in a separate dir
- [ ] fix references from Weaveworks website

### Build
- [x] integrate https://hub.docker.com/r/ddidier/sphinx-doc/
- [x] add linkchecker to find broken links
- [x] hook up `make test-docs` with CI
- [ ] generally document how to use this
- [x] hook up with release/build process (what do we need here? document how to publish a new release branch in readthedocs?)

### Publish
- [x] decide on project-name to use with readthedocs.io
- [x] set up readthedocs.io account
- [ ] publish master
- [ ] pull in older releases
- [ ] link to docs from main README.md

PR #2306, opened 2019-07-29
Fix broken doc link

rytswd:
After #2152 has been merged, it looks like some of the doc had broken links to other docs due to the path changes. I just scanned all the `*.md` files that had some directory structure changes in that PR, and updated the entire repo with string replace.
I think I covered most use cases, but please shout if I missed / wrongly replaced any.


rytswd:
One correction, I did scan other files as well, and thus ended up touching `.txt`, `.tmpl` and `.go` one file each. They are purely for text usage, and I believe they are the correct fix.

As to the build failure, I'm uncertain how this PR can cause any build behaviour change from the previous run. I'm new to the flux repo, and would love if anyone can lend a hand looking into it.

docs/using/fluxctl.md
```diff
@@ -556,7 +556,7 @@ Commit customization
         --git-user
         --git-email
 
-        See [docs/daemon.md] for more information.
+        See [docs/features/daemon.md] for more information.
```
dholbach: Maybe let's make this a real clickable link?

PR #2312, opened 2019-07-30
Correct (my own) spelling of integrations

squaremo:
I fully expect this to lay waste to the build (on this branch), for a while.


squaremo:
I did not expect this to pass CI without further intervention*. Now, I am deeply confused.

*It does not complete `make check-generated` locally, because the code-generator got by go mod makes code that's incompatible with the apimachinery library also got by go mod.

squaremo:
BOOYAH! I have made it fail for the _right reason_:
https://circleci.com/gh/fluxcd/flux/7110

There are two problems with `make check-generated`:

 1. `integrations` was spelt incorrectly in the diff command, meaning any diffs in the actual package would not be detected
 2. k8s.io/code-generator puts its output in the `GOPATH`-mandated location -- it doesn't know about go mod. So all the generated files were off elsewhere in the filesystem, and no diffs were detected.

Having fixed those two problems, now we are at the thing that actually needs to be fixed:

 a. code-generator at the version indicated in go.mod generates code that is **not compatible** with apimachinery from the version indicated in go.mod.

squaremo:
I'm not sure it's possible to keep all of these things:

 - using client-go v11
 - using go modules
 - generating kubernetes client code

The version of k8s.io/code-generator that will produce code compatible with client-go v11 (roughly, that at the tag kubernetes-1.14.4) isn't kitted out for go mod. Attempting to run it from its pkg/mod location (as the current update_codegen.sh script does) won't work, because it wants to `go install` itself, and it can't write to its own directory (by design of go mod). Attempting to vendor it and run it from vendor/k8s.io/code-generator just doesn't work -- `go mod vendor` just doesn't put it in vendor, I don't know why.

squaremo:
Thanks @hiddeco and @dholbach 
Sorry to subject you to this necessary unpleasantness :-)

PR #2323, opened 2019-08-05
Update Slack information in MAINTAINERS file

hiddeco:


stefanprodan:
Can you also please change my handle, it's `stefanprodan` on the CNCF Slack.

Thanks

PR #2334, opened 2019-08-07
Support different targetNamespace for HelmRelease #2128

### LINKED ISSUE #2128: helm-operator: install resources to different namespace than HelmRelease itself
**Describe the feature**
When using plain helm to install resources, user can specify next:
`--tiller-namespace` - namespace where the tiller is located.
`--namespace` - namespace where resources should be installed.

so basically, when I want to use tiller in `kube-system` ns and install resources from chart to, let's say, `development` ns, I can do `helm install --tiller-namespace kube-system --namespace development`. 

however, if I'm using helm-operator and flux, I don't see where can I do that.
I can only specify `namespace` for `HelmRelease` resource, and all the things from my helm chart will be installed to the same ns where helmrelease is located.

**Expected behavior**
Add possibility to configure `helm --namespace ... ` feature, for example, user can specify separately namespace where tiller and helm-operator are located and separately namespace, where the resources should be installed:

```yaml
---
kind: HelmRelease
metadata:
  name: rabbit
  namespace: kube-system
spec:
  releaseName: rabbitmq
  namespace: some-other-namespace
  chart:
    repository: https://kubernetes-charts.storage.googleapis.com/
    name: rabbitmq
    version: 3.3.6
  values:
    replicas: 1
```

stealthybox:
Fixes #2128
I've tested this on a kind cluster with several cases.
Let me know if we should add some e2e tests.
We might also need to bump the CRD to `v1beta2`.

### DIFF
--- a/integrations/apis/flux.weave.works/v1beta1/types.go
+++ b/integrations/apis/flux.weave.works/v1beta1/types.go
@@ -5,7 +5,7 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/helm/pkg/chartutil"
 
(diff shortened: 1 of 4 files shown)


integrations/apis/flux.weave.works/v1beta1/types.go
```diff
@@ -33,17 +33,40 @@ func (fhr HelmRelease) ResourceID() resource.ID {
...
-	namespace := fhr.Namespace
-	if namespace == "" {
-		namespace = "default"
+	if fhr.Spec.ReleaseName == "" {
+		namespace := fhr.GetDefaultedNamespace()
+		targetNamespace := fhr.GetTargetNamespace()
+
+		if namespace != targetNamespace {
+			// prefix the releaseName with the administering HelmRelease namespace as well
+			return fmt.Sprintf("%s-%s-%s", namespace, targetNamespace, fhr.Name)
```
stefanprodan:
```suggestion
			return fmt.Sprintf("%s-%s", targetNamespace, fhr.Name)
```
I would keep the original name format to be in line with the Helm2 format.

squaremo:
> We might also need to bump the CRD to `v1beta2`.

"might" -- what would decide one way of the other?

stefanprodan:
The new field is opt-in and it doesn't change the default behaviour, so I would keep the version as it is since we'll be releasing v1 after the repo split.

integrations/apis/flux.weave.works/v1beta1/types.go
stealthybox: This behavior is only different in the case that there is a parent namespace.
When targetNamespace isn't specified to something different than the HelmRelease ns, the previous behavior is preserved.

This is here to prevent naming conflicts and show (in helm) where the HelmRelease is managed.

Ex:
team1-prod-cache
infra-clusterservices-externaldns
admin-test-privateingress

I'm fine to remove this, but I hope we agree these are helpful auto-generated names. :+1:

stefanprodan: The release name is used to compose the Kubernetes service names, here is an [example](https://github.com/helm/charts/blob/master/stable/drone/templates/service.yaml#L4). I can see how this would break setups that rely on the Helm v2 convention.

stealthybox: `fullname` is a conventional helper template :+1:
The drone chart you linked is using an older version.
It doesn't check if `Release.Name` contains `$name` before appending:
https://github.com/helm/charts/blob/b2425d3/stable/drone/templates/_helpers.tpl#L7-L15

If the release name were `team1-production-drone`,
that fullname would be `team1-production-drone-drone`.

If the release name were `production-drone` (not having the parent namespace prepended), 
it would be `production-drone-drone`.

helm-operator's name convention to prepend the target-namespace is unrelated, additional convention the user must already be aware of.

When composing charts, it's up to users to understand how release names affect resource names. There is no "Helm 2 format" we need to adhere to.
_______________________
Prepending the parent namespace to the release name is good for users because it references what namespace the HelmRelease is in when you do a `helm list`.
This allows you to see all of the releases together, despite them being in different target namespaces.

The one surprising case is modifying the `spec.targetNamespace` of an existing HelmRelease and having the generated name change. Users would have to set the `spec.releaseName` to keep their old release.

This is the only significant trade-off I see in prepending the additional namespace.
If we're not okay with this U/X, I can pull this part of the patch.

I really do see it better for users to have this though.
If you're managing a lot of HelmReleases with targetNamespaces from multiple parent namespaces, this will make it much easier to locate them all from Helm.

stefanprodan: Ok. Users can change the name in the custom resource if the conventions doesn't suits them.

stealthybox: Thanks for hearing me out :+1:

PR #2337, opened 2019-08-08
Publish Helm chart to charts.fluxcd.io

stefanprodan:
Push Flux chart to both fluxcd/flux and fluxcd/charts


stefanprodan:
@hiddeco I plan to delete the push to fluxcd/flux after we update the docs with charts.fluxcd.io that's way I've duplicated the commands.

PR #2345, opened 2019-08-09
Update logo

dholbach:
Update logo after work from @bia and @miklosp on this.

Renamed logo file as well and added it do docs site too. Let me know if I missed anything.


stefanprodan:
Renaming the logo breaks Helm Hub and probably other sites https://hub.helm.sh/charts/weave-flux/flux

hiddeco:
@stefanprodan would creating a  symlink work?

dholbach:
http://docs.fluxcd.io/en/latest/ shows the new logo too.

PR #2350, opened 2019-08-12
[snapcraft] stop using soon-to-be deprecated version-script

dholbach:
https://forum.snapcraft.io/t/version-script-deprecated-cannot-find-docs/12621


dholbach:
I'm sure there's a more elegant way to go about https://github.com/fluxcd/flux/pull/2350/files#diff-5fde7a6d86053f0e1d88c0a2a238941fR26-R28 - but it did the trick (don't want to exit during the build, but catch the return code).

PR #2351, opened 2019-08-12
update logos to be transparent

dholbach:


stefanprodan:
Can you also add the horizontal one and use it in the docs please?

Issue #2200, opened 2019-06-27
Create probes for flux standalone deployment


mpashka:
**Describe the feature**
Currently there are no probes in flux deployment manifest file [https://github.com/weaveworks/flux/blob/master/deploy/flux-deployment.yaml]
That means flux reports as ready before initialization and if flux stops doing things for some reason no actions will be taken.

**What would the new user story look like?**
1. User starts up standalone Flux as described here [https://github.com/weaveworks/flux/blob/master/site/get-started.md#deploying-flux-to-the-cluster]
2. During warmup flux reports flux pod is not ready.
3. If flux stops operating then flux pod is restarted.

**Expected behavior**
Flux must report to kubernetes that it is ready only when it is able to accept and process connections.
If flux stops operating by some reason then it must be restarted.

PR #2355, opened 2019-08-13
Add liveness and readiness probes

stefanprodan:
This PR adds liveness and readiness probes to Flux chart and deploy manifests.

Fix: #2200


chart/flux/templates/deployment.yaml
```diff
@@ -99,6 +99,18 @@ spec:
           - name: http
             containerPort: 3030
             protocol: TCP
+          livenessProbe:
+            httpGet:
+              port: 3030
+              path: /api/flux/v6/identity.pub
```
hiddeco: Given our quest earlier today, this is a smart intermediate solution :1st_place_medal:

PR #2369, opened 2019-08-16
Add automated image update examples to docs

stefanprodan:
The examples use `fluxcd.io` annotations instead of `flux.weave.works`. Do not merge this until Flux 1.14 is released.

### DIFF
--- a/docs/references/automated-image-update.md
+++ b/docs/references/automated-image-update.md
@@ -1,0 +1,87 @@
+# Automated deployment of new container images
+
+Flux can be used to automate container image updates in your cluster.
+Flux periodically scans the pods running in your cluster and builds a list of all container images.
+Using the image pull secrets, it connects to the container registries, pulls the images metadata
+and stores the image tag list in memcached.
+
+You can enable the automate image tag updates by annotating your deployments, statefulsets,
+daemonsets or cronjobs objects. You can also control what tags should be considered for an
+update by using glob, regex or semantic version expressions.
+
+> **Note:** that Flux only works with immutable image tags (`:latest` is not supported).
+Every image tag must be unique, for this you can use the Git commit SHA or semver when tagging images.
+
+## Examples
+
+What follows is a list of examples on how you can control the image update automation. If you're using Helm releases 
+please see the [Helm operator integration docs](helm-operator-integration.md).
+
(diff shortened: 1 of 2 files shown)


docs/references/automated-image-update.md
```diff
@@ -0,0 +1,89 @@
...
+The above configuration will make Flux update the `sidecar` when you push
+a tag for the `my-proxy` image that begins with `stg`.
+For the `app` container, Flux will update it when you push a tag for the
+`my-app` image that begins with `dev-`.
+
+To target a specific container the annotation format is `fluxcd.io/tag.<CONTAINER>: <TYPE>:<EXPRESSION>`.
+
+You can turn off the automation with `fluxcd.io/automated: "false"` or with `fluxcd.io/locked: "true"`.
+
+
```
hiddeco: Couple too many newlines here.

docs/references/automated-image-update.md
```diff
@@ -0,0 +1,89 @@
+# Automated deployment of new container images
```
hiddeco: I have also mentioned the `filter.fluxcd.io` annotation in #2298, as this annotation works for all workloads. May be worth to mention it somewhere here as 'existing'.

docs/references/automated-image-update.md
```diff
@@ -0,0 +1,89 @@
...
+Flux can be used to automate container image updates in your cluster.
+Flux periodically scans the pods running in your cluster and builds a list of all container images.
+Using the image pull secrets, it connects to the container registries, pulls the images metadata
+and stores the image tag list in memcached.
+
+You can enable the automate image tag updates by annotating your deployments, statefulsets,
+daemonsets or cronjobs objects. You can also control what tags should be considered for an
+update by using glob, regex or semantic version expressions.
+
+**Note** that Flux only works with immutable image tags (`:latest` is not supported).
```
hiddeco:
```suggestion
> **Note:** that Flux only works with immutable image tags (`:latest` is not supported).
```

to match how we style this in other places.

docs/references/automated-image-update.md
stefanprodan: Ah ok so that PR changes the annotations in the Helm docs. I'll remove it from here then.

PR #2384, opened 2019-08-21
Install docs fixes

stefanprodan:
Fix fluxctl install command, replace `--git-path` with `--git-paths`.
Fix podinfo example, replace `PODINFO_UI_COLOR`  with `PODINFO_UI_MESSAGE`.
Fix Kustomize install, add `namespace.yaml` to resources.

### DIFF
--- a/docs/tutorials/get-started-kustomize.md
+++ b/docs/tutorials/get-started-kustomize.md
@@ -42,6 +42,8 @@ Create a kustomization file and use the Flux deploy YAMLs as base:
 ```sh
 cat > fluxcd/kustomization.yaml <<EOF
 namespace: flux
+resources:
+  - namespace.yaml
 bases:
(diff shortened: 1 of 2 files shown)


stefanprodan:
@hiddeco can we pin the `stable` to a master commit instead of tag? The current fluxctl install command is broken 😢

hiddeco:
@stefanprodan not possible, `stable` is magic from RTD based on published semver (git) tags.

stefanprodan:
I would switch back to master then...

2opremio:
> Fix fluxctl install command, replace `--git-path` with `--git-paths`.

Oh dear. It sucks that it's inconsistent with `fluxd`, sorry about that. Since it's already released, should we simply accept both?

stefanprodan:
@2opremio yes I think accepting both would be best.

Issue #2386, opened 2019-08-21
Install documentation for fluxctl identity incorrect


neilshannon:
**Describe the bug**
After a fresh install via fluxctl 1.14.0, the command fluxctl identity does not work due to searching for a pod labeled [app=flux] instead of [name=flux] which the install creates.

**To Reproduce**
Steps to reproduce the behaviour:
0. Start with a fresh minikube cluster
1. Install fluxctl 1.14.0
2. Follow the instructions on the "getting started" page
```bash
export GHUSER="replace"
fluxctl install \
--git-user=${GHUSER} \
--git-email=${GHUSER}@users.noreply.github.com \
--git-url=git@github.com:${GHUSER}/flux-get-started \
--git-paths=namespaces,workloads \
--namespace=flux | kubectl apply -f -
```
3. Execute `fluxctl identity`
```bash
○ → fluxctl identity
Error: Could not create a dialer: Could not get pod name: Could not find running pod for selector: labels "app=flux"
Run 'fluxctl identity --help' for usage.
```

**Expected behavior**
The SSH key should be displayed as the output.

**Additional context**
Add any other context about the problem here, e.g 
- Flux version: 1.14.0
- Kubernetes version: 1.15.2
- Git provider: github

**Workaround**
`fluxctl --k8s-fwd-labels=name=flux --k8s-fwd-ns=flux identity`

hiddeco:
Resolved in #2387.

PR #2387, opened 2019-08-21
Fix fluxctl identity and sync docs

stefanprodan:
`fluxctl install` doesn't use the same label that `identity` wants.

Fix: #2386


neilshannon:
I took a crack at just fixing the template.  https://github.com/fluxcd/flux/pull/2388

PR #2388, opened 2019-08-21
modify the install pod label to app=flux to match the fluxctl defaults

neilshannon:
Updated the deployment template to match the default pod selector [app=flux] as configured in the fluxctl command https://github.com/fluxcd/flux/blob/master/cmd/fluxctl/root_cmd.go#L75.


neilshannon:
Fix: #2386

stefanprodan:
You have rerun the code gen with `make generate-deploy`

hiddeco:
This needs a seal of approval from @2opremio, as I do not know the underlying reason of the label change (and there may be one).

deploy/flux-deployment.yaml
```diff
@@ -2,12 +2,12 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: flux
+  app: flux
```
HighwayofLife: I don't think `app` is a valid metadata. See: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta
You might be looking to use:
```
metadata:
  name: flux
  labels:
    app: flux
```

neilshannon: Right you are -- good eye.  The label in the deployment template the only thing I needed to change.

hiddeco:
@neilshannon can you smash your commits together so we do not end up with 3 commits (and a merge commit) in `master`?

dholbach:
For me this does not work:

```console
[daniel@reef flux ]$ ./bin/fluxctl identity
Error: Could not create a dialer: Could not get pod name: Could not find running pod for selector: labels "app=flux"
Run 'fluxctl identity --help' for usage.
[daniel@reef flux ]$ ./bin/fluxctl identity --k8s-fwd-ns default
Error: Could not create a dialer: Could not get pod name: Could not find running pod for selector: labels "app=flux"
Run 'fluxctl identity --help' for usage.
[daniel@reef flux ]$ 
```

stefanprodan:
@dholbach have you followed the latest install docs? 
https://docs.fluxcd.io/en/latest/tutorials/get-started.html#set-up-flux

dholbach:
Ugh, no I haven't. This is a bit of a "breaking" change, no?

stefanprodan:
The braking change is in `fluxctl install` but I don't think it was intentional, @2opremio can say more.

stefanprodan:
If I remember correctly, `fluxctl` used to search for the `flux` namespace. If it doesn't do that anymore then the Helm chart instructions are also broken https://github.com/fluxcd/flux/blob/master/chart/flux/README.md#setup-git-deploy

2opremio:
> This needs a seal of approval from @2opremio, as I do not know the underlying reason of the label change (and there may be one).

There is none. Note that the label didn't change compared to what was under `deploy/` (see https://github.com/fluxcd/flux/pull/2287/files ) which is what I used as a reference.

squaremo:
NB the port forwarding _first_ tries what's given in the argument (which defaults to `app=flux`), _then_ tries `name in {flux, fluxd, weave-flux-agent}`. That hasn't changed. If it appears broken, it must be for some other reason.

2opremio:
OK, I have fixed the root problem at https://github.com/fluxcd/flux/pull/2394 . I am not closing this PR since we may anyways choose to change the label.

neilshannon:
So #2394  fixes the label selector fallback, but the documentation on fluxctl still states the default label selector for commands using `--k8s-forward-labels` is [app=flux].  Shouldn't we be consistent and update the messaging on fluxctl too?  

Example for identity:

```sh
○ → fluxctl identity -h
Display SSH public key

Usage:
  fluxctl identity [flags]

...

Global Flags:
      --k8s-fwd-labels stringToString   Labels used to select the fluxd pod a port forward should be created for. You can also set the environment variable FLUX_FORWARD_LABELS (default [app=flux])
```

Example for sync:

```sh
○ → fluxctl sync -h
synchronize the cluster with the git repository, now

Usage:
  fluxctl sync [flags]

...

Global Flags:
      --k8s-fwd-labels stringToString   Labels used to select the fluxd pod a port forward should be created for. You can also set the environment variable FLUX_FORWARD_LABELS (default [app=flux])
```

HighwayofLife:
This change will work, but you'd still have to use `--k8s-fwd-ns flux`, and then allows this to work because it's going to be searching for `app=flux` in that chosen namespace.
Using `--k8s-fwd-ns` is still problematic and also difficult to remember. It should be using `--namespace|-n`, but that's a PR for a different bit of code.

squaremo:
> It should be using `--namespace|-n`, but that's a PR for a different bit of code.

`-n` `--namespace` is used (optionally) to identify the target of subcommands, e.g., `fluxctl list-services -n demo`

squaremo:
> the documentation on fluxctl still states the default label selector for commands using --k8s-forward-labels is [app=flux]. Shouldn't we be consistent and update the messaging on fluxctl too?

@neilshannon What sort of update were you thinking of? (The default is reported correctly; the fallback is not mentioned though, and perhaps that's misleading?)

stefanprodan:
The Kubernetes deployment label selector is an [immutable field](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates), changing this would prevent people from upgrading Flux. Going to close this as I don't think such a breaking change is justifiable.

PR #2393, opened 2019-08-22
chart: add the ability to disable the memcached deployment

antonosmond:
In some cases it can be useful to deploy flux without memcached e.g. when not using the image scanning features or if you want to use an external memcached service deployed separately from the flux chart.
This change adds the ability to disable the memcached deployment in the helm chart and / or override the memcached hostname.

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -248,6 +248,8 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `registry.dockercfg.enabled`                      | `false`                                              | Mount `config.json` via Secret into the Flux Pod, enabling Flux to use a custom docker config file
 | `registry.dockercfg.secretName`                   | `None`                                               | Kubernetes secret with the docker config.json
 | `registry.dockercfg.configFileName`               | `/dockercfg/config.json`                             | Alternative path/name of the docker config.json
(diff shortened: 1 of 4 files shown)


stefanprodan:
@antonosmond have you tested Flux without memcached? If I remember right, it will not sync with git without it.

antonosmond:
@stefanprodan The change is more about using an external memcached.
In our use case, we have various "domains" each with multiple applications.
Each of those domains runs in its own namespace and we deploy a flux per domain/namespace.
This means we have like 10 flux deployments and we don't want a memcached deployment for each flux. A single instance of memcached should be sufficient.

chart/flux/README.md
```diff
@@ -248,6 +248,8 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `registry.dockercfg.enabled`                      | `false`                                              | Mount `config.json` via Secret into the Flux Pod, enabling Flux to use a custom docker config file
 | `registry.dockercfg.secretName`                   | `None`                                               | Kubernetes secret with the docker config.json
 | `registry.dockercfg.configFileName`               | `/dockercfg/config.json`                             | Alternative path/name of the docker config.json
+| `memcached.enabled`                               | `true`                                               | Create a memcached deployment and service. Set to `false` to use an external memcached service
```
stefanprodan: Please change this to `When set to false you must set an external memcached service`

PR #2394, opened 2019-08-22
Fix portforward error check

2opremio:
We were not checking portforwarding errors properly, causing fluxctl to only test the first label selector provided, and in turn causing to fail selecting on `name={"flux", "fluxd", "weave-flux-agent"}` as intended by https://github.com/fluxcd/flux/blob/d18267b6bfb1ca485f13d04efdb8d4b41dc66393/cmd/fluxctl/root_cmd.go#L129

### DIFF
--- a/cmd/fluxctl/portforward.go
+++ b/cmd/fluxctl/portforward.go
@@ -21,7 +21,7 @@ func tryPortforwards(ns string, selectors ...metav1.LabelSelector) (p *portforwa
 			return
 		}
 
-		if !strings.Contains(err.Error(), "Could not find pod for selector") {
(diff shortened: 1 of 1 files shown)


2opremio:
For traceability reasons, the problem was introduced when vendoring this change https://github.com/justinbarrick/go-k8s-portforward/commit/649dc88798f27b6f2165d30fd4b614fde71068a9#diff-fd0ea03735acb6add12bf1cca551cdd4 and not updating the check consistently.

However, the main cause is how shitty error selection is in Goland.

hiddeco:
> However, the main cause is how shitty error selection is in Golang.

~Maybe Go v3 will solve this...~

PR #2412, opened 2019-08-29
Fix the resolving of a single v1 HelmRelease

hiddeco:
Due to a mistake the name of the resource was used as a namespace
argument, as a result several functionalities relying on this method
did no longer work as it was unable to find the given workload (e.g.
automatic image updates and fluxctl workload selector flags). Passing
along the namespace (as intended) resolves all this.


kingdonb:
This resolved it for me, I can do HelmRelease automation now 👍TYVM

PR #2419, opened 2019-09-02
Release 1.14.2

squaremo:
In the interest of simplicity, I just fast-forwarded release/1.14.x instead of cherry-picking, which included a change or two relating the chart.


squaremo:
> The enhancement would require a minor semver bump following strict rules but with some creativity it falls into 'tweak of a shortcoming in an existing feature' category.

Right yeah -- I'm bending the rules a bit. I could cherry-pick it out, but there seems little reason to hold it back; and, I don't feel it warrants a whole minor version bump on its own.

PR #2427, opened 2019-09-04
Simplify the files included in kube-config in snapcraft.yaml

dholbach:
cc @stealthybox


dholbach:
I pinged Leigh, as this was relevant to some of his work, but can be reviewed by anyone really.

stealthybox:
Does this allow recursive access to any kubeconfig file in the `~/.kube` directory?
If so, +++++++100

snap/snapcraft.yaml
```diff
@@ -58,10 +58,8 @@ plugs:
   kube-config:
     interface: personal-files
     read:
-    - $HOME/.kube/config
-    - $HOME/.minikube/client.key
-    - $HOME/.minikube/client.crt
-    - $HOME/.minikube/ca.crt
+    - $HOME/.kube
+    - $HOME/.minikube
```
stealthybox: To support `k3d` we would need access to the host's `$XDG_CONFIG_HOME`, but that does not appear to be supported well by `snap`: [thread](https://forum.snapcraft.io/t/accessing-xdg-config-home-of-the-host/9780/10)

We can handle the common case by just hard-coding `~/.config`:
```suggestion
    - $HOME/.minikube
    - $HOME/.config/k3d
```

dholbach: Thanks a lot for digging into this!

dholbach:
Thanks for the edit @stealthybox - this is ready to go now. :heavy_check_mark:

dholbach:
This had to go through the store review process, but is live in `edge` now:

```
Track    Arch    Channel    Version          Revision
latest   amd64   stable     1.14.2           30
                 candidate  ^                ^
                 beta       ^                ^
                 edge       1.14.2+bb576fad  35
```

PR #2429, opened 2019-09-04
update gitSecret feature to handle git.Export

arsiesys:
Since the following commit, we do not call the Clone function anymore. As a result, the git-secret feature do not work anymore, the apply is done before doing the reveal of the secrets.

https://github.com/fluxcd/flux/commit/6eabb29d2501e1b8f3c6e4c18453cced20cdfdca#diff-ba9f11ecc3497d9993b933fdc2bd61e5, 

I am giving the gitSecret option value to the daemon object.. I did not find a better way to do it :(.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -613,7 +613,6 @@ func main() {
 		SigningKey:  *gitSigningKey,
 		SetAuthor:   *gitSetAuthor,
 		SkipMessage: *gitSkipMessage,
-		GitSecret:   *gitSecret,
 	}
 
 	repo := git.NewRepo(gitRemote, git.PollInterval(*gitPollInterval), git.Timeout(*gitTimeout), git.Branch(*gitBranch), git.IsReadOnly(*gitReadonly))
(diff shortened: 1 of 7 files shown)


stefanprodan:
@arsiesys the tests are failing:

```
git/export_test.go:32:28: not enough arguments in call to repo.Export
```

squaremo:
> I am giving the gitSecret option value to the daemon object.. I did not find a better way to do it :(.

Could it be an option for the Repo, and called as part of repo.go#workingClone, perhaps? If that's possible, I would much prefer it to passing a boolean argument through.

arsiesys:
> > I am giving the gitSecret option value to the daemon object.. I did not find a better way to do it :(.
> 
> Could it be an option for the Repo, and called as part of repo.go#workingClone, perhaps? If that's possible, I would much prefer it to passing a boolean argument through.

Hello! Sadly, I just made some tests and sadly, I need to do the git secret reveal after the checkout, it can't be done right after the workingClone cause this one is going to be with master (before the checkout).

I i think, I could eventually, move the call of the secretReveal to the Export function after the checkout call.

hiddeco:
@squaremo I performed some CPR on the PR (:recycle: :fish:) and gave `Export` a dedicated `SecretUnseal` method so the boolean doesn't need to be passed around everywhere. PTAL.

PR #2430, opened 2019-09-05
Update project in errors etc

squaremo:


squaremo:
> Did we finally get rid of all of them? Time will tell...

There's a few places left that refer to weaveworks, but are not visible to end users (or otherwise meaningful) -- tests, a couple of imports. I don't think there's a pressing case for removing them right now.

Issue #2431, opened 2019-09-05
[AKS] Mounting /etc/ubernetes/azure.json no longer possible; need option in chart to mount from secret instead


Korijn:
Related AKS bug report: https://github.com/Azure/AKS/issues/729
External-DNS has the same issue, they've rolled out support to mount azure.json through a secret to work around the issue. See [their solution here](https://github.com/helm/charts/blob/master/stable/external-dns/templates/deployment.yaml#L329-L337).

I think the same should be done here.

PR #2434, opened 2019-09-06
[Bugfix] Mount AKS service principal through secret instead of hostPath for ACR support

Korijn:
Add `registry.acr.secretName` chart parameter to work around permission errors on AKS when using ACR.

Fixes https://github.com/fluxcd/flux/issues/2431.

This [mimics the approach by external-dns](https://github.com/helm/charts/blob/5998aa6d6ac37813753285668a75287fcfc110ad/stable/external-dns/templates/deployment.yaml#L329-L337).


Korijn:
Can you give an indication of when the next chart release is planned? Don't need it urgently, but just so I can decide whether it is worth the effort of working with a local copy temporarily for my needs.

PR #2438, opened 2019-09-09
Using https protocol not require ssh config

vyckou:
Following changes would allow:
- native support for git remote using HTTPS transport, using username and password for git clone, by using git-apikey and git-user arguments.
- usage of git-apikey and git-user arguments together with https transport, will skip ssh key pair creation
- will hide potential secrets in logs
- by using git-apikey argument and HTTPS transport, forces to remove username/password from git-url argument. That is done intentionally, as git-apikey should be treated not as config map, but as secret

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	_ "net/http/pprof"
+	"net/url"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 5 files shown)


linuxbsdfreak:
@vyckou. 

At the moment i am testing the following with https for connecting to github by configuring the values.yaml for the flux helm chart 

```
git:
  url: "https://GITHUB_API_TOKEN@github.com/username/gitops-helm-release.git"
```
I also see that the API Key is displayed in the flux-helm-operator pod logs 

My Questions:
- What do i need to change in the future while configuring flux ?
- Does the logging of the API_TOKEN not happen anymore after the fix?

Appreciate your feedback?

hiddeco:
@linuxbsdfreak with a build from this PR, you would provide the HTTPS address to the `--git-url` flag and supply the token using the `--git-apikey` flag.

To make use of this in the chart, it would require changes to the chart to make it easy to use as it does not know about this flag at the moment. For the time being, it is however possible to supply it manually by utilizing the `additionalArgs` value.

vyckou:
@linuxbsdfreak, @hiddeco  is right, provide
```
--git-url=https://github.com/username/gitops-helm-release.git
--git-apikey=GITHUB_API_TOKEN
```
despite that, you **need** to provide `--git-user=GITHUB_USER_FOR_API`
otherwise, url parsers and SafeURL() function wont be able to identify GITHUB_API_TOKEN as secret, and it will treat as username. As a result you will continue to see the API_TOKEN in the logs

`"https://GITHUB_API_TOKEN@github.com/username/gitops-helm-release.git"` GITHUB_API_TOKEN by default will be treated as username and only in structure 
`"https://GITHUB_USERNAME:GITHUB_API_TOKEN@github.com/username/gitops-helm-release.git"` GITHUB_API_TOKEN is treated as password

I am not using helm charts, therefore simply forgot to tackle that part. Will try to align that too.

cmd/fluxd/main.go
```diff
@@ -596,12 +598,42 @@ func main() {
...
+			os.Exit(1)
+		}
+
+		if parsedGitURL.Scheme != "https" {
+			logger.Log("err", "git-apikey argument can be used only with HTTPS transport")
+			os.Exit(1)
+		}
+
+		if parsedGitURL.User.Username() != "" {
+			logger.Log("err", "Set git username not in git-url, but use git-user argument instead")
```
hiddeco: The `--git-user` is not used in the way you think it is used, or at least, this implementation does not take into account that it is used as the _full name_ of the user (e.g. `John Doe <john@doe.com>`). While I think GitHub does some behind-the-scene parsing to still get this information presented in their UI, the data in git will be off.

However, given that we want this feature in in a short timeframe, I will volunteer to make the required changes.

squaremo:
Is it possible to interpolate an env entry into the `--git-url` value? That might obviate the extra arguments -- you could just put the authkey into a secret and mount that to get the env entry. (This would be better than supplying it as an argument, too.)

hiddeco:
Closing this as it has been superseded by above PR.

I have added you in as a co-author in my version @vyckou, as it is a rework of your work, hope this eases the pain of me closing this. :tulip:

PR #2449, opened 2019-09-16
Update chart doc: add nodeSelector and tolerations for memcached

victorsalaun:
Update the chart README.md adding `nodeSelector` and `tolerations` for memcached


victorsalaun:
Hello. I don't understand why the build fails after updating a readme file.
Looking at the build logs, I understand the `modTime` is not correct. Do I have to update the values of modTime?

stefanprodan:
The timestamps changed because you've submitted the PR from your master branch instead of a feature branch. I've rebased your master, should be good now.

victorsalaun:
I see. I will know for the next time. Thanks @stefanprodan

PR #2454, opened 2019-09-19
Update namespace in clusterrolebinding for the example deployment yaml

ettiee:
The example deployment configs in `./deploy` has a small mismatch with the deployment configs applied through `fluxctl install`. 

- `flux` ClusterRoleBinding subject `flux` ServiceAccount specifies default namespace, through `fluxctl install` is `flux` namespace

### DIFF
--- a/deploy/flux-account.yaml
+++ b/deploy/flux-account.yaml
@@ -34,4 +34,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: flux
-    namespace: default
+    namespace: flux


hiddeco:
@ettiee thanks for your contribution. Changing only the namespace there would lead to issues as none of the other deployments had a namespace set (and would still be applied to the namespace configured in your `kubectl` context).

I have aligned all examples in `deploy/` to include a namespace in #2475, so that having a namespace set in your `kubectl` context will no longer lead to issues.

Many thanks though for submitting this, it is much appreciated :medal_sports:

Issue #1363, opened 2018-09-14
[windows] Publish fluxctl on chocolatey


dholbach:
As `fluxctl` builds on Windows (cf #1214), and chocolatey.org was [mentioned](https://weave-community.slack.com/archives/C4U5ATZ9S/p1535728473000100?thread_ts=1535728403.000100) [multiple](https://weave-community.slack.com/archives/C4U5ATZ9S/p1535731509000100) [times](https://github.com/weaveworks/flux/issues/1214#issuecomment-417964621), it'd be great if we could publish it there as well to make it more accessible.

@geofflamrock indicated interest in taking a look. Anyone else?

JimPruitt:
Not to step on anyone toes here, but I ran across this and decided to give it a go.  The package is currently pending a review.  https://chocolatey.org/packages/fluxctl/1.8.0

dholbach:
Nice work, @JimPruitt! Please let us know how things go!

I have a couple of questions. Do you know more about this? What's the process for 
- Uploading the package? Is it just a matter of taking the binary from the release page and uploading it to chocolatey.org? Can this be done my multiple people? Can this maybe even be automated?
- Downloading/installing it, so we can document it?

JimPruitt:
Thanks @dholbach!  Sorry if this is too verbose...

**Uploading Packages**
As far as uploading packages,  packages are submitted to Chocolatey.org and the guidelines and processes for submission are here: https://github.com/chocolatey/choco/wiki/CreatePackages .  One of the guidelines states that  the source code for the package should be publicly available.  However,  even with that in place, the package won't be trusted by the Chocolatey community if someone like me (an independent contributor) hosts and maintains the code.  If the package came from Weaveworks itself then it's more 'trustable'.  So, It would be better if Weaveworks assumed control of the code for building this package.  (Perhaps then you're also likely to see more contributions.)  

You're more than welcome to take what I've done here and host the code out of Weaveworks Github project if you wish.  (My intention here was just to contribute to this really cool project!)

**Downloading/Installing**
The code for building this package is kept here: https://github.com/JimPruitt/chocolatey-packages , and this particular package is encapsulated into a folder called 'fluxctl'.   Here are the steps for getting up and running:

1. Install chocolatey on your build machine  (see https://chocolatey.org/docs/installation on how to do this). You can build on Windows machines and (even better) Debian.  (I assume a preference)
3. clone the repo at  https://github.com/JimPruitt/chocolatey-packages
[...]

PR #2457, opened 2019-09-20
Update install doco to include chocolatey

chrisfowles:
#1363 Adds fluxctl install instructions to installation instructions for windows installations via chocolatey.

Cleans up code block format (indent vs fenced)


dholbach:
Hum... I had proposed https://github.com/chrisfowles/flux/pull/1 - I'll propose that for inclusion separately then. :-D

dholbach:
Moving to https://github.com/fluxcd/flux/pull/2459

PR #2458, opened 2019-09-20
Add HMCTS to who is using flux in production

timja:


hiddeco:
@timja one small note, would be able to both rebase the PR and put the name a bit higher on the list so we maintain alphabetical order?

timja:
> @timja one small note, would be able to both rebase the PR and put the name a bit higher on the list so we maintain alphabetical order?

done

PR #2459, opened 2019-09-20
reflow commit customization bits, fix link

dholbach:
Just a few markup changes and a fixed link to the daemon.md reference.


stefanprodan:
@dholbach instead of using the GitHub UI to update the branch, please use `git rebase` next time so we don't end up with two commits.

PR #2461, opened 2019-09-22
Update kubectl to v1.14 and kustomize to v3.2

stefanprodan:
- update kubectl to v1.14.7
- update kustomize to v3.2.0
- update e2e to Kind to v0.5.1 (Kubernetes v1.15.3)


stefanprodan:
With this and the imports moved to `fluxcd` I would say a minor release should be next. I would like to do the `pkg` move and after that do a Flux release, so that helm-operator will use the `/pkg` imports. Does this sound ok to you?

PR #2470, opened 2019-09-24
Support Git over HTTPS with creds from env vars

### LINKED ISSUE #2280: https protocol for git support
For now, flux does support HTTPS protocol for gitops repository only partially. There are lots of code and functionality, which would not be required to execute, if git-url would be https:// and not ssh://

As you are using git CLI client and invoking git commands to deal with gitops repo, passing url https://apikey@github.com/org/repo.git works, but everything which is linked about SSH keypairs are not required and adds additional overhead for the setup and runtime

If using HTTPS protocol instead of SSH would not require:
* creation of the SSH keypair on flux deployment in K8s
* usage of the ssh generator
* k8s secret with private key

On the git SCM side, that would bring pluses as:
* adding flux git user to the gitops repo(s) with r/w access, instead of putting public keys on the repos

this behaviour would enable to run flux in Openshift environments without granting flux a priv mode. - creation of the dockerfile with USER directive is not working, as often hardened Openshift setups expects userid with custom ranges.

hiddeco:
This is a rework of #2438, and fixes #2280.

This PR makes supplying HTTP(S) basic auth credentials in the
`--git-url` secure. Places where the full remote origin were logged
have been modified to only log the `Remote.SafeURL()` so that the
authentication key is never exposed in the logs.

Securely supplying the credentials without exposing them in the
Flux workload is possible by adding two environment variables (e.g.
`GIT_AUTHUSER` and `GIT_AUTHKEY`) to a Kubernetes secret, and using
Kubernetes mechanics to define them in the Flux pod by using an
`envFrom`. The variables can then be used in `--git-url` argument as
documented in the Kubernetes documentation[1]:
`--git-url=https://$(GIT_AUTHUSER):$(GIT_AUTHKEY)@github.com/an/example.git`

If the `--git-url` uses a HTTP(S) scheme; the generation of an SSH
private key and the setup of the SSH keyring is now disabled.

[1]: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

---

This includes changes to the documentation, (generated) example deployments and the Helm chart to offer full support.


squaremo:
The doc updates are welcome, but a lot of the changes aren't much to do with supporting HTTPS -- can you split them into a separate commit, at least?

chart/flux/values.yaml
```diff
@@ -275,6 +277,14 @@ kube:
 # additionalArgs:
 # - --connect=ws://fluxcloud
 
+# The contents of the secret will be defined as environment variables
+# in the Flux container (supported key values by Flux for now are
+# `GIT_AUTHUSER` and `GIT_AUTHKEY`). Once defined, you can use the
+# variables in your `git.url`:
+# `https://$(GIT_AUTHUSER):$(GIT_AUTHKEY)@github.com/fluxcd/flux-get-started.git`
```
squaremo: What happens if someone only wants to supply the authkey, as shown in https://github.com/fluxcd/flux/pull/2438#issuecomment-530265107?

cmd/fluxd/main.go
```diff
@@ -319,7 +320,13 @@ func main() {
 		}
 	}
 
-	if *sshKeygenDir == "" {
+	// Used to determine if we need to generate a SSH key and setup a keyring
```
squaremo: Is the keyring ever needed unless the git URL uses `ssh` (or the shorthand for it)? I.e., could you just check the scheme.

docs/guides/use-git-https.md
```diff
@@ -0,0 +1,60 @@
+# Using Git over HTTPS
+
+Instead of making use of Flux' capabilities to generate an SSH private
+key, or [supplying your own](provide-own-ssh-key.md). It is possible to
```
squaremo:
```suggestion
key, or [supplying your own](provide-own-ssh-key.md), it is possible to
```

docs/guides/use-git-https.md
```diff
@@ -0,0 +1,60 @@
...
+set the `GIT_AUTHUSER` and `GIT_AUTHKEY` environment variables and use
+these in your `--git-url` argument to provide your HTTPS basic auth
+credentials without having to expose them as a plain value in your
+workload.
+
+> **Note:** supplying the `GIT_AUTHKEY` in the `--git-url` will
+> disable the generation of a private key and prevent the setup
+> of the SSH keyring.
+
+> **Note:** the variables _must be escaped with `$()`_ for Flux
```
squaremo: It's not Flux that requires the parens, it's Kubernetes. So a better explanation here might be

> env variables used in Kubernetes manifests are escaped with `$()`

go.mod
```diff
@@ -20,7 +20,6 @@ require (
 	github.com/imdario/mergo v0.3.7
 	github.com/instrumenta/kubeval v0.0.0-20190804145309-805845b47dfc
 	github.com/justinbarrick/go-k8s-portforward v1.0.4-0.20190722134107-d79fe1b9d79d
-	github.com/modern-go/reflect2 v1.0.1 // indirect
```
squaremo: how is this ...

pkg/git/operations.go
```diff
@@ -35,6 +35,8 @@ var allowedEnvVars = []string{
 	// these are for Google Cloud SDK to find its files (which will
 	// have to be mounted, if running in a container)
 	"CLOUDSDK_CONFIG", "CLOUDSDK_PYTHON",
+	// these are used for supporting HTTPS basic auth
+	"GIT_AUTHUSER", "GIT_AUTHKEY",
```
squaremo: Are you certain these need to be exposed to `git` -- won't it just be given the URL with those values already interpolated?

cmd/fluxd/main.go
hiddeco: Using this approach is safer as the shorthand errors while parsing it as an URL due to the missing protocol. Which results in the same.

chart/flux/values.yaml
hiddeco: This would never work in any setup. Although git supports only passing along the username (and fetching the actual token from some keystore), passing along a _token_ as first parameter before the `@` will always result in it being parsed as an username.

In fact, passing along any credentials in an URL has been [deprecated with RFC3986](https://tools.ietf.org/html/rfc3986#section-3.2.1) but is backported / still supported by many.

go.mod
hiddeco: I have had this for any PR I have been working on in the past days, and decided to fix it here.

chart/flux/values.yaml
squaremo: > passing along a _token_ as first parameter before the `@` will always result in it being parsed as an username

I would have expected this as well. Can you explain the example given in that comment? Is it invalid?

go.mod
squaremo: Use `go mod`, they said, it'll be so much better ...

chart/flux/values.yaml
hiddeco: I think it comes from not following the GitHub documentation closely, or not being familiar with basic auth.

We could theoretically block things if we detect that only a `GIT_AUTHKEY` is present in the given `--git-url`; but I am not sure if that is worth the technical debt it would introduce.

pkg/git/operations.go
hiddeco: You are right, this is a left-over from the previous approach I attempted.

chart/flux/values.yaml
squaremo: I couldn't find an example _in the github docs_ that showed a key being used in username position. So I think it's a safe course of action to assume a username will be provided.

docs/guides/use-git-https.md
```diff
@@ -0,0 +1,60 @@
...
+     name: flux
+   ...
+   spec:
+     containers:
+     - name: flux
+       envFrom:
+       - secretRef:
+           name: flux-git-auth
+       args:
+       - --git-url=https://$(GIT_AUTHUSER):$(GIT_AUTHKEY)@<USER>/flux-get-started.git
```
squaremo: Where does the meta-variable `<USER>` come from?

pkg/daemon/daemon.go
```diff
@@ -542,7 +542,7 @@ func (d *Daemon) NotifyChange(ctx context.Context, change v9.Change) error {
 	switch change.Kind {
 	case v9.GitChange:
 		gitUpdate := change.Source.(v9.GitUpdate)
-		if gitUpdate.URL != d.Repo.Origin().URL && gitUpdate.Branch != d.GitConfig.Branch {
+		if gitUpdate.URL != d.Repo.Origin().SafeURL() && gitUpdate.Branch != d.GitConfig.Branch {
```
squaremo: I'm not completely sure this will work with an HTTPS URL, but no more sure that it worked before :-S

PR #2471, opened 2019-09-24
Tweak logged warning of unsupported resource kind

hiddeco:
As the previous warning message did not provide enough context to the
user to be able to figure out why it was being marked as 'unsupported'.

Context: https://weave-community.slack.com/archives/C4U5ATZ9S/p1569340270209200


hiddeco:
@stefanprodan it is included in the resource name `<namespace>:<kind>/<name>`.

PR #2473, opened 2019-09-24
Ensure generated template modtime is Unix epoch

hiddeco:
vsfgen which we use for maintaining an embedded copy of our templates
does also include the modtime of the files.

This poses a problem in combination with git, as in a project managed
by git the modtime of a file depends on the time it was changed
through a local git action (e.g. a checkout from a different state
which results in a file write on disk) instead of the time of the
actual change.

This commit adds a wrapper around `http.FileSystem` that modifies
the file mod times to always return Unix epoch. So that the only
things resulting in a new `generated_template.gogen.go` are folder
and/or file content changes.


2opremio:
Nice, thanks!!!

PR #2475, opened 2019-09-25
Use `flux` namespace in `deploy/` examples

hiddeco:
This is so that the configuration is not effected by e.g. a namespace
context set in the `kubectl` config of the user, as this caused
problems during the creation of the `ClusterRoleBinding` which was set
to the `default` namespace, while the resources themselves would be
created in the user their configured namespace.

It also simplifies the 'Get started with Kustomize' guide, as a user
does no longer need to create the namespace as a prerequisite but is
still able to overwrite it by adding a `namespace: <targetNamespace>`
to their `kustomization.yaml`.

Supersedes #2454

### DIFF
--- a/docs/tutorials/get-started-kustomize.md
+++ b/docs/tutorials/get-started-kustomize.md
@@ -24,28 +24,14 @@ First you'll need a git repository to store your cluster desired state.
 In our example we are going to use [`fluxcd/flux-get-started`](https://github.com/fluxcd/flux-get-started).
 If you want to use that too, be sure to create a fork of it on GitHub.
 
-Create a directory and add the `flux` namespace definition to it:
-
-```sh
-mkdir fluxcd
-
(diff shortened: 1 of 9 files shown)


docs/tutorials/get-started-kustomize.md
```diff
@@ -24,26 +24,11 @@ First you'll need a git repository to store your cluster desired state.
...
-EOF
-```
-
-Create a kustomization file and use the Flux deploy YAMLs as base:
+Create a directory, and add a `kustomization.yaml` file that uses the
+Flux deploy YAMLs as a base:
 
 ```sh
 cat > fluxcd/kustomization.yaml <<EOF
-namespace: flux
```
stefanprodan: I would keep the namespace declaration to make it obvious that Flux will be deployed in `flux` namespace.

Issue #2043, opened 2019-05-13
As an engineer, I want to specify my flux configuration in a single file


mjpitz:
**Describe the feature**
Right now, users can provide a large range of command line options to help configure flux. It would be great if this configuration could be pulled from a single config file instead of needing to hard code options on the container itself. By decoupling the container from the configuration, the two can be managed independently.

**What would the new user story look like?**
How would the new interaction with Flux look like? E.g.
0. The flux container should allow an option `--config` to be specified. This options should point to a file path location where the flux configuration is specified.
1. If specified, the configuration should be read from the user specified location.
2. Existing command line options (e.g. `--git.url`) should override their `--config` file counterparts.

**Expected behavior**
* Users can specify either a config file or the existing command line options.
* The command line options override any config file value.
* Configuration can be managed independently from the container.

squaremo:
This makes sense to me. It would lead the way for config being mounted from a ConfigMap, which might  make e.g., automatic configuration much easier.

tkeech1:
I'd like to work on adding support for ConfigMaps. It would be nice to use Viper because it supports hierarchical overrides and works with pflag. Please let me know if that's ok.

squaremo:
> I'd like to work on adding support for ConfigMaps.

Neat! Is there special support needed for ConfigMaps, over just getting config from a file? Or should I read this as "support for config from a file, which can be mounted from a configmap"?

stefanprodan:
I think Viper is the right choice for us. The config file can come from a ConfigMap or Kubernetes secret but that shouldn't matter since Flux/Viper will just look for a file on disk.

tkeech1:
My thought was to just add support for ConfigMaps. In a helm deployment, ConfigMap values would come from values.yaml. For a non-helm deployment, ConfigMap values would be specified in a new file, flux-configmap.yaml. In either case, the config will be loaded as a ConfigMap and the flux container will mount the ConfigMap values into environment variables. Existing command-line options will override environment variables.

Please let me know if you have something else in mind.

stefanprodan:
I think Flux should look for a config file on disk. The configmap would be mounted as a volume by the Helm chart and `fluxctl install`. If people want to use https git addresses, then they should be able to switch to a secret instead of a configmap since the git url will contain sensitive information.

vyckou:
[...]

PR #2479, opened 2019-09-26
Load fluxd arguments from config file

tkeech1:
This PR is addresses [2043](https://github.com/fluxcd/flux/issues/2043). It adds support for a fluxd configuration file that is created as a secret or configmap which is mounted in the fluxd container. This change only supports local config files (no support for Github URLs yet). Command-line arguments override values specified in the config file.

There are two new command line options:

**config.file** - a yaml config file to configure fluxd
**config.asConfigMap** - a boolean value that specifies if the file is loaded as a config map (if false or unspecified, the config file is loaded as a secret)

**Helm:**
For a Helm install, the flux-config.yaml file must be located in the chart/flux directory.

`helm install chart/flux --set helmOperator.create=true --set helmOperator.createCRD=false --set config.file=flux-config.yaml --set config.asConfigMap=true --namespace flux`

**Fluxctl:**
`fluxctl install --config-file=chart/flux/flux-config.yaml --config-as-configmap=false --namespace=flux | kubectl apply -f -`

Changes are not yet documented in the references or guides. Thanks for any feedback.


squaremo:
Ohhhh, this looks promising! What do you reckon is left to do @tkeech1 ?

stefanprodan:
I think this would be a good moment to introduce a config struct e.g.:

```go
type Config struct {
	GitUrl      string `mapstructure:"git-url"`
	GitBranch   string `mapstructure:"git-branch"`
}

func main() {
	// read file
	viper.ReadInConfig()

	// bind flags 
	viper.BindPFlags(fs)

	// bind environment variables
	viper.SetEnvPrefix("FLUXCD")
	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
	viper.AutomaticEnv()

	// unmarshal instead of calling viper.GetString for each config field
	var config Config
	viper.Unmarshal(&config)
}
```

tkeech1:
Thanks for the feedback. I added the config struct and updated the documentation. Please let me know if you'd like any other changes.

squaremo:
I started rebasing this, fixed a few things along the way, then realised I had diverged from the PR branch. I'm working here: https://github.com/fluxcd/flux/compare/issue/2043-config-file.

I've preserved the authorship of commits that are substantially similar; we can either consider a PR from that branch, or I can push to this PR (and we don't have to decide right now ..)

2opremio:
Closing in favor of #2544

Issue #2477, opened 2019-09-25
Context deadline exceeded error using manifest-generation and kustomize


davidpristovnik:
**Describe the bug**
All of our deployments use kustomize to render the manifests. Directory structure looks like this with up to a few hundreds of deployments: 
```
├── .flux.yaml
├── flux1
│   └── kustomization.yaml
├── flux2
│   └── kustomization.yaml
├── flux3
│   └── kustomization.yaml
├── flux4
│   └── kustomization.yaml
└── flux5
    └── kustomization.yaml
```
Each `kustomization.yaml` has a reference to a common base, with multiple workloads/deployments. 
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

bases:
- ../../../../overlays/flux

namePrefix: flux1-

commonLabels:
  app.kubernetes.io/instance: flux1
  hostname: flux1.somedomain.com

imageTags:
- name: flux
  newTag: 2019-08-07_15-11-24_master_87ed9db687
commonAnnotations:
  fluxcd.io/automated: "true"
  fluxcd.io/tag.nginx: glob:*_master_*
  fluxcd.io/tag.uwsgi: glob:*_master_*
```
We use `--manifest-generation` feature, with `.flux.yaml` config looking like this:
```
version: 1
commandUpdated:
  generators:
    - command: |
        for k in `find . -name kustomization.yaml`; do
          echo ---; kustomize build `dirname $k`
        done
  updaters:
    - containerImage:
        command: |
          instance=${FLUX_WL_NAME}
          cd $instance
          kustomize edit set image "$FLUX_IMG:$FLUX_TAG"
      policy:
        command: |
          instance=${FLUX_WL_NAME}
          cd $instance
          kustomize edit add annotation -f "fluxcd.io/$FLUX_POLICY:$FLUX_POLICY_VALUE"
```
This setup works fine for small number of deployments but when trying to enable automated release cycle or do a `fluxctl release --all --update-all-images' it fails with a timeout error. 

**Logs**
```
[...]

PR #2481, opened 2019-09-26
Make sync operations timeout configurable

stefanprodan:
Add `--sync-timeout` to fluxd command flags and `sync.timeout` to the Helm chart (defaults to `1m` same as before).

Workaround for: #2477 and #1857


hiddeco:
In #1857, where the timeouts were originally reported, we discussed ways of optimization instead of making it a configurable flag. However, shortly after this the manifest factorization feature was introduced, which gives people the power to run anything during sync. As optimizing the unknown is a hard thing to do, I think this change now is the obvious (and easiest) solution.

PR #2483, opened 2019-09-26
eliminate staticcheck warnings: S1005

beautytiger:


hiddeco:
Thank you for submitting this PR and thinking about codestyle. I am personally opposed to this "S1005" rule and will give you a short explanation why.

While Golang may not technically require us to put a blank identifier there, it is an indicator for me that something has multiple values without having to think much about it. If you remove it, I would need to look at the type of the value, classify it as being single or multi valued, and then reach the same conclusion as I would have had with the blank identifier being there.

Given having it there does not cost anything performance wise, and [the value will be discarded harmlessly](https://golang.org/doc/effective_go.html#blank). It will safe me a lot of time during refactors or reviews, at no cost.

beautytiger:
@hiddeco 
Thanks for your detailed explaination, I see the value of this pattern.

PR #2491, opened 2019-10-02
Build: release modifications

hiddeco:
- Stop publishing images to Weaveworks DockerHub
- Republish release tag with a v prefix, to make the release available to Go Mod


hiddeco:
> I think the fluxctl download URLs will stop working.

I think they will continue to work as long as we continue publishing our releases on GitHub _without_ the `v` added to the tag. The CI will take care of adding an additional Go Mod-compliant tag, and all other script / formula things that depend on the non-`v` tag won't notice a thing.

PR #2495, opened 2019-10-03
Update Helm operator dependency to `v1.0.0-rc2`

hiddeco:


stefanprodan:
Any idea why `github.com/weaveworks/flux` is still a dependency?

hiddeco:
@stefanprodan only thing I can think of is a circular nested dependency:

`rc2` -> `v1.15.0` -> `rc1` -> `weaveworks/flux@v1.14.x`

stefanprodan:
Ah right, I see it on the helm-op deps...

PR #2514, opened 2019-10-13
Rename sync marker key to fluxcd.io/sync-mark

stefanprodan:
Rename sync marker key from `flux.weave.works/sync-hwm` to `fluxcd.io/sync-mark`.

### DIFF
--- a/pkg/sync/secret.go
+++ b/pkg/sync/secret.go
@@ -11,7 +11,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-const syncMarkerKey = "flux.weave.works/sync-hwm"
+const syncMarkerKey = "state.fluxcd.io/high-water-mark"
 
 // NativeSyncProvider keeps information related to the native state of a sync marker stored in a "native" kubernetes resource.
 type NativeSyncProvider struct {


stefanprodan:
Shouldn't we drop the `flux.weave.works` naming if the annotations are not user-facing?

stefanprodan:
Because `hwm` doesn't mean anything to me

squaremo:
> Shouldn't we drop the `flux.weave.works` naming if the annotations are not user-facing?

AIUI annotations without a prefix are reserved for users. From https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set:

> If the prefix is omitted, the annotation Key is presumed to be private to the user. Automated system components (e.g. kube-scheduler, kube-controller-manager, kube-apiserver, kubectl, or other third-party automation) which add annotations to end-user objects must specify a prefix.

So it's pretty clear that in this case, the annotation should have a prefix. It is less clear from that (but reasonable to argue) that annotations added by users to control automation should have a prefix.

squaremo:
> Because `hwm` doesn't mean anything to me

I don't think "mark" has any more explanatory power. But neither do I think there's an especial need for it to be self-explanatory. How about `high-water-mark`?

stefanprodan:
What about `git.fluxcd.io/high-water-mark`?

squaremo:
> What about `git.fluxcd.io/high-water-mark`?

I like the idea of using the "subdomain"; maybe we can make it clear that it's effectively internal state that way. To tie it into the command-line flags, perhaps `state.fluxcd.io/high-water-mark`?

squaremo:
What's the backward-compatibility story here -- at a glance, I'd expect fluxd to fail to read an existing high ware mark, and send a bunch of spurious events, before it writes to the new annotation. (When updating it, it should remove the old one too)

stefanprodan:
Yes indeed, after upgrading Flux it will push a bunch of events 😢

2opremio:
@stefanprodan Did the spurious events made you reconsider this PR? Should we close it?

Issue #1128, opened 2018-06-08
support triggering via webhook


davidkarlsen:
Having to wait for flux to detect changes in the git repo is a bit of a drag / delay. It would be nice if flux could support being called via a webhook to trigger immediate refresh of the repo.

squaremo:
We have a prototype of this in weave cloud, and the hooks are there in the daemon (i.e., it's just engineering in weave cloud keeping it back).

We could also have these in the daemon. It's a bit trickier, because
 1. you don't usually want to expose the flux API to the internet (so you've have to have some kind of ingress for just some routes)
 2. it wouldn't necessarily deal with all the formats of payload, e.g., from github etc.

davidkarlsen:
Great news - I guess it can be a bit nicer in weave cloud (UI etc).
It would make sense to have in the daemon for users running flux only - no weave cloud. Since the git setup is really in the daemon the format should not matter much if it is just a signal to pull the configured repo (e.g. no need to parse payload).

Morriz:
+1
it is how most CICD workflows are triggered...check out how drone.io does it: (example: github.com) you create an oAuth app with creds for drone to be able to create a deploy hook automatically (generating a unique token per webhook) for the repo it wants to register for. So no need to manually create a deploy hook for every repo, just configure for your git host (github or whatever) once. 

It would be most elegant to create a dashboard service (with a nice UI showing all flux releases), which would be accessible to these app generated webhooks. That service would then publish those events to the flux instances, triggering them to do their thing. Maybe that is possible with hosted Flux, but I am not aware of the paid offering's options...

fikriauliya:
Has it already in available open source version?

warmfusion:
This feels like a real problem to support clear pipelines for CI/CD workflows. Our dev team wants to _know_ their deployment is activated as they would then execute secondary flows such as cache clears, routing behaviour or smoke/QA testing against the deployment.

Without having an explicit hook (and ideally a blocking wait) we can't use Flux for any deployment on our systems as the async delay reduces our options for automated hands-off continuous delivery and notification.

squaremo:
> Our dev team wants to _know_ their deployment is activated

This is less to do with webhooks and more with providing a layer of interpretation on the Kubernetes API. 

But I do grasp the requirement: to be able to spin up tests, you need to know when the system under test is ready.

> Without having an explicit hook (and ideally a blocking wait)

[...]

PR #2515, opened 2019-10-13
Add an API to trigger a Git sync

proskehy:
As per the discussion in #1128, there are currently no convenient API methods that could be used as a webhook interface for Git. For the simple need of letting Flux know that there are changes to be synced, `/api/flux/v11/sync-git` can be used. 
It requires no arguments, only triggers a `Repo.Notify()` call, same as `/api/flux/v11/notify` does if the supplied payload checks out.

### DIFF
--- a/pkg/api/v11/api.go
+++ b/pkg/api/v11/api.go
@@ -17,6 +17,8 @@ type ListServicesOptions struct {
 type Server interface {
 	v10.Server
 
+	SyncGit(context.Context) error
+
 	ListServicesWithOptions(ctx context.Context, opts ListServicesOptions) ([]v6.ControllerStatus, error)
 
 	// NB Upstream methods move into the public API, since
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -550,6 +550,11 @@ func (d *Daemon) NotifyChange(ctx context.Context, change v9.Change) error {
 	return nil
 }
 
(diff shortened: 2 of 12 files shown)


squaremo:
As discussed in #1128, I think webhook receivers belong in a separate listener/process/container (e.g., a sidecar). In particular, requiring the flux API to be available to an ingress -- for it to route requests to -- makes this a non-starter.

There is still a potential use case for the changes in this PR, and that is for _fluxctl_ to ask the daemon to refresh its repo (whatever it is). Using `../notify` to do this would require the repo URL and branch to be supplied, which doesn't really make sense for `fluxctl`. However: `fluxctl` does already have `fluxctl sync`, which operates by submitting a job that refreshes the git repo, then waits for the HEAD revision to be applied to the cluster. If this is the use case that you are about, then adding a `fluxctl sync --nowait` which submits the job, but doesn't wait for the outcome, might be a more direct way of achieving it.

squaremo:
I am going to close this PR, with thanks to @proskehy for their work putting it together.

For those interested in webhook receivers specifically, https://github.com/fluxcd/flux-recv is a webhook receiver you can run as a sidecar to flux. It processes GitHub, GitLab, and DockerHub webhook notifications (including verifying them).

For those interested in being able to tell `fluxctl` to trigger a sync, asynchronously -- well I'd like to know if anyone is interested in that, because it would be fairly easy to add a flag to `fluxctl sync`, as described above.

PR #2520, opened 2019-10-16
Improve performance of fetching cluster resources when deploying flux to sync in all namespaces

waseem-h:
### what
Improve logic for fetching resources from the cluster
### why
Flux fetches resources from all namespaces on a per namespace basis even if we tell flux to deploy in all namespaces. This becomes an issue when there are a lot of namespaces and a lot of resource kinds.
### use-case
We have 300+ namespaces and 25+ CRDs in our cluster and it takes about 13.5 minutes for only the resources to be fetched which is a problem for us since we want to sync changes every minute. This PR improves the performance significantly by fetching resources from all namespaces if that is enabled instead of a per namespace basis.


waseem-h:
I would also like to mention that this reduces the number of calls made to the API Server by a huge amount.
Previously lets say you had 30 namespaces and 20 different kinds of resources, then flux would have to make 600 requests to fetch the state. With this, flux only makes 20 requests to the API Server and retrieves the same state. This is assuming that you're allowing flux to watch in all namespaces. If flux is running on selected namespaces then this PR has no affect on the number of calls

waseem-h:
@hiddeco It would be great if you could review this today

waseem-h:
@stefanprodan squashed

stefanprodan:
@waseem-h you can deploy the fix on your cluster using this image `fluxcd/flux-prerelease:master-f1698542`

waseem-h:
Thanks @stefanprodan. Any idea when the new minor is going to be released?

stefanprodan:
Probably in two weeks from now.

waseem-h:
Awesome

PR #2525, opened 2019-10-17
Break generational cycle

squaremo:
Adapted from commit msg:

For `fluxctl install`, we embed a set of templates in the fluxctl binary by generating a fake filesystem as Go code (pkg/install/generated_templates.gogen.go). But the code that generates the fake filesystem also _depends_ on the fake filesystem, since it does double duty as a program for generating example manifests from those templates.

This leads to a problem: if anything has upset the generation process, it's not possible to regenerate the files, since the generation tool won't build.

A related problem, worked around elsewhere (#2473), is that it's easy to introduce spurious differences in the generated code -- which, since it's checked in -- means painful merges.

In general we don't care about the generated Go code, only about what it produces. This PR removes it from git and from the uncommitted change detection (`make check-generated`); and uses a new capability of `fluxctl install` to write to individual files to generate the example files under deploy/, removing the need for the code generating program to depend on the code it's generating.

### DIFF
--- a/cmd/fluxctl/install_cmd.go
+++ b/cmd/fluxctl/install_cmd.go
@@ -2,14 +2,19 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 
 	"github.com/fluxcd/flux/pkg/install"
 )
 
-type installOpts install.TemplateParameters
+type installOpts struct {
+	install.TemplateParameters
+	outputDir string
+}
 
 func newInstall() *installOpts {
 	return &installOpts{}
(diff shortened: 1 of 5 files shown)


squaremo:
NB: there is one change to the example deployments in `deploy/` as a result of this: the git label, email and user are set explicitly to example values rather than left to default, since we require them to be supplied to `fluxctl install`.

I think this is a _good thing_, but I have little idea of how `deploy/` gets used nowadays -- @stefanprodan, do you think it is OK?

cmd/fluxctl/install_cmd.go
```diff
@@ -23,22 +28,23 @@ func (opts *installOpts) Command() *cobra.Command {
 fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' | kubectl -f -`,
 		RunE: opts.RunE,
 	}
-	cmd.Flags().StringVarP(&opts.GitURL, "git-url", "", "",
+	cmd.Flags().StringVar(&opts.GitURL, "git-url", "",
```
squaremo: All of these were using `*VarP`, the variant for providing a shortcut single-letter flag, then giving an empty string as the shortcut :man_shrugging:

stefanprodan:
@squaremo I see no problem with setting git label, email and user, but I would use `flux-sync` as the label default or better yet use `sync-state=secret`.

squaremo:
@stefanprodan Thanks -- I'll change the label. I am reluctant to change too much about either fluxctl install or the example manifests here, so I'll leave the "better yet" to another PR.

deploy/flux-deployment.yaml
```diff
@@ -145,6 +145,9 @@ spec:
         # Include this if you want to restrict the manifests considered by flux
         # to those under the following relative paths in the git repository
         # - --git-path=subdir1,subdir2
+        - --git-label=flux
```
squaremo:
```suggestion
        - --git-label=flux-sync
```

hiddeco:
The change is a great improvement for us, both in complexity and code reduction, and improved maintainability.

I am however wondering if there is any impact on `eksctl` and how they currently make use of the `install` package, as iirc, they make use of the embedded file that is no longer there.

stefanprodan:
I this this breaks eksctl, wksctl, firekube and any other project that depends on the install package.

squaremo:
> any other project that depends on the install package

WHYYYYYYY

squaremo:
> > any other project that depends on the install package
> 
> WHYYYYYYY
> (did they do that)

Oh yes, it was the motivation for the `install` package in the first place.

2opremio:
I think it's OK to break the API (eksctl vendors it anyways) but fluxctl should still supply a way for eksctl to generate the files (ideally in memory and including the filenames)

hiddeco:
@2opremio would it be possible for them to generate their own embedded file based on the templates they receive from vendoring the package?

cmd/fluxctl/install_cmd.go
2opremio: That was my bad, didn't really think about the alternatives

squaremo:
I can commit the generated file, and restore the change detection, while still breaking the dependency that was causing problems. (It's just a shame go doesn't let us avoid it)

2opremio:
Yes, I guess that would be possible, but in practice that's pushing the problem to the consumer which I don't think it's ideal (particularly if it's not just eksctl which uses this functionality)

pkg/install/generated_templates.gogen.go
```diff
@@ -19,39 +19,39 @@ var templates = func() http.FileSystem {
 	fs := vfsgen۰FS{
 		"/": &vfsgen۰DirInfo{
 			name:    "/",
-			modTime: time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC),
+			modTime: time.Date(2019, 10, 17, 13, 27, 25, 101756322, time.UTC),
```
hiddeco: Is this on purpose or did you forget to restore the `modTimeFS` wrapper?

squaremo: Honestly I was hoping it wasn't needed any more. Since check-generated will only fail if the instantiated templates are different to those in deploy/, there's no need to commit meaningless changes to the generated code (i.e., just because a template got a fresh mtime).

Thus ugly conflicts will only arise if there are genuinely changes on both branches -- and those should be resolvable just by making sure the non-generated files are merged OK, then regenerating this file.

Does this follow, or have I misunderstood the motivation of modTimeFS?

squaremo: Actually, I can check, since you outlined how this happens in Slack:

> 1. work on some branch that is older than master
> 2. rebase this branch, on master, the timestamp of your files is now newer than master
> 3. switch back to master, because git does not see any changes (they are equal by content to what's there), git will not touch them
> 4. profit, your timestamp is now newer than master and check-generated will fail

squaremo: I can cause problems by doing this:

 1. from a `base` branch, create branch `test`
 2. commit a change to a template in `test`
 3. run `make check-generated`, observe that it fails (as it should)
 4. commit the regenerated Go source in pkg/install/, and the generated yamls in deploy/
 5. checkout `base`, make any commit, then checkout `test` again -- this will touch the template you changed, since it will be replaced each time you change branch
 6. at this point, `make check-generated` will succeed (it produces the same deploy/ yaml that is committed); but it will regenerate the Go code, with different timestamps.
 7. rebasing on `base`, or similar operations, will probably break because of the spuriously regenerated Go file.

So it's not a disastrous problem -- everything will still build, and you can get past it (just overwrite the Go file) if you understand what's going on -- but it is a bit of a trap, and as with the origin problem, will come up frequently.

hiddeco: > but it is a bit of a trap, and as with the origin problem, will come up frequently.

I think this is enough reason to put it back, as we became tired of the issue pretty soon last time.

squaremo: > I think this is enough reason to put it back

Me also too

PR #2530, opened 2019-10-18
Added a readonly flag to the install command

sebikul:
Allow the daemon to be installed with the readonly flag present
using the install command. Modified the cmd flags and the deployment
template.


sebikul:
This change modifies the `generated_templates.gogen.go` file present in the repo. As it is a generated file, I decided against including it in this PR. Please let me know if it should also be added for this feature to be complete.

squaremo:
> This change modifies the `generated_templates.gogen.go` file present in the repo. As it is a generated file, I decided against including it in this PR. Please let me know if it should also be added for this feature to be complete.

Whether or not to commit generated files is often a unwelcome choice isn't it. The scheme we use is to commit the file, but check that it produces the right results (by running `fluxctl install -o deploy ...` and comparing to what's committed there). The easiest way to do it is to run

    make check-generated

This will regenerate the particular file (as will `make`, if you've changed the templates). If it succeeds, then your change hasn't upset the expected output in deploy/.

You may find that it fails because you've introduced a whitespace change in the output -- use the whitespace-suppressing mustaches `{{-` (suppress preceding whitespace) and `-}}` (suppress following whitespace) in the template to avoid this, or if it's desired whitespace, commit the change to `deploy/`.

pkg/install/templates/flux-deployment.yaml.tmpl
```diff
@@ -159,6 +159,9 @@ spec:
 
         # Tell flux it has readonly access to the repo (default `false`)
         # - --git-readonly
+{{ if .GitReadOnly }}
```
squaremo: For neatness in the output, it would be nice to output _either_ the arg or it's commented-out analogue. E.g.,

```
{{ if .GitReadOnly }}
  - --git-readonly
{{ else }}
  # - --git-readonly
{{ end }}
```

squaremo:
I've rebased it as a prerequisite for merging .. waiting for CI to pass...

Issue #2199, opened 2019-06-27
Create metric for flux manifest errors


mpashka:
**Describe the feature**
Add metric to inform users about invalid manifest or flux inability to apply manifest to k8s by some reason.

Sometimes user can make a mistake and push wrong manifest into git. Or sometimes manifest is correct but flux can't apply it - it is common use case for big configmaps - flux uses "apply" to push configmap to kubernetes and that implies saving configmap content into annotations. And in kubernetes there is limitation on annotations size. So if configmap size is bigger than some limit (something like 255Kb) then `kubectl apply` reports error and manifest can't be applied. There are flux events that can be used to detect this but in standalone flux installation it's much more convenient to create metrics with information about manifest errors. Metrics that can be easily checked any time. Also that metrics can be used to rise prometheus alarm which is de facto standard for k8s installations.

**Expected behavior**
I expect new metric with a number of invalid manifest files in flux git repository. And metric with a number of errors that occured during applying manifests to kubernetes. And possible a metric with a total number of manifests, but that one is only for reference.

Also it can be good idea to create flux prometheus dashboard that can be used for checking flux health and prometheus alarms for flux that can be used to inform users about invalid manifests.

ksaritek:
@stefanprodan & @mpashka, is it a difficult ticket for a newbie? is anybody working on that?

2opremio:
I think we can close this, since it's addressed in https://github.com/fluxcd/flux/issues/2199

ArthurSens:
Is it possible to add `git-repository`, `git-path`, and `file` to this metric?
If I'm using one Prometheus server to monitor multiple cluster/repositories, I'd like to create an alert that tells me exactly where is the error that is causing the failed sync

PR #2535, opened 2019-10-21
Create metric for flux manifest errors

mpashka:
https://github.com/fluxcd/flux/issues/2199

### DIFF
--- a/docs/references/monitoring.md
+++ b/docs/references/monitoring.md
@@ -14,5 +14,15 @@ The following metrics are exposed:
 | `flux_daemon_queue_duration_seconds`     | Duration of time spent in the job queue before execution
 | `flux_daemon_queue_length_count`         | Count of jobs waiting in the queue to be run
 | `flux_daemon_sync_duration_seconds`      | Duration of git-to-cluster synchronisation
+| `flux_daemon_sync_manifests`             | Number of manifests being synced to cluster
 | `flux_registry_fetch_duration_seconds`   | Duration of image metadata requests (from cache)
 | `flux_fluxd_connection_duration_seconds` | Duration in seconds of the current connection to fluxsvc
+
+Flux sync state can be obtained by using the following PromQL expressions:
+* `delta(flux_daemon_sync_duration_seconds_count{success='true'}[6m]) < 1` - for general flux sync errors - usually if 
+that is true then there are some problems with infrastructure or there are manifests parse error or there are manifests 
+with duplicate ids.
+
+* `flux_daemon_sync_manifests{success='false'} > 0` - for git manifests errors - if true then there are either some 
+problems with applying git manifests to kubernetes - e.g. configmap size is too big to fit in annotations or 
+immutable field (like label selector) was changed. 
--- a/pkg/daemon/sync.go
+++ b/pkg/daemon/sync.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
+	"github.com/fluxcd/flux/pkg/metrics"
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
 	"path/filepath"
@@ -157,6 +158,7 @@ func doSync(ctx context.Context, manifestsStore manifests.Store, clus cluster.Cl
 		switch syncerr := err.(type) {
 		case cluster.SyncError:
 			logger.Log("err", err)
+			updateSyncManifestsMetric(len(resources)-len(syncerr), len(syncerr))
 			for _, e := range syncerr {
 				resourceErrors = append(resourceErrors, event.ResourceError{
 					ID:    e.ResourceID,
@@ -167,8 +169,15 @@ func doSync(ctx context.Context, manifestsStore manifests.Store, clus cluster.Cl
 		default:
 			return nil, nil, err
 		}
+	} else {
+		updateSyncManifestsMetric(len(resources), 0)
 	}
 	return resources, resourceErrors, nil
+}
+
+func updateSyncManifestsMetric(success, failure int) {
+	syncManifestsMetric.With(metrics.LabelSuccess, "true").Set(float64(success))
+	syncManifestsMetric.With(metrics.LabelSuccess, "false").Set(float64(failure))
 }
 
 // getChangedResources calculates what resources are modified during
--- a/pkg/daemon/sync_test.go
+++ b/pkg/daemon/sync_test.go
@@ -12,8 +12,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/go-kit/kit/log"
-
 	"github.com/fluxcd/flux/pkg/cluster"
 	"github.com/fluxcd/flux/pkg/cluster/kubernetes"
 	"github.com/fluxcd/flux/pkg/cluster/kubernetes/testfiles"
@@ -26,6 +24,9 @@ import (
 	registryMock "github.com/fluxcd/flux/pkg/registry/mock"
 	"github.com/fluxcd/flux/pkg/resource"
 	fluxsync "github.com/fluxcd/flux/pkg/sync"
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	promdto "github.com/prometheus/client_model/go"
 )
 
 const (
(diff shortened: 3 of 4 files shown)


pkg/daemon/sync_test.go
```diff
@@ -26,6 +24,9 @@ import (
 	registryMock "github.com/fluxcd/flux/pkg/registry/mock"
 	"github.com/fluxcd/flux/pkg/resource"
 	fluxsync "github.com/fluxcd/flux/pkg/sync"
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	io_prometheus_client "github.com/prometheus/client_model/go"
```
stefanprodan: Please rename this import to `promdto`

mpashka: Done

stefanprodan:
I think the metric should be named `sync_error` and it should act like a boolean, values can be 0 or 1. 

Labels:
- `fetch` (records git errors)
- `generate` (records .flux.yaml generators errors)
- `parse` (records manifestsStore errors)
- `apply` (records kubectl apply errors)

With this metric and labels one can compose alerts with `avg_over_time(flux_daemon_sync_error [5m]) > 0` and compose messages like `flux sync apply failed` or `flux sync parse falied`.

squaremo:
It seems like the main thrust of this PR is to be able to know the number of manifests that failed to apply, so you can alert on manifests failing to apply.

There's a metric which record how long each sync takes, and whether it was successful or not -- `flux_demon_sync_duration_seconds` -- but that might be too broad for your purposes, since it will label a sample as unsuccessful if any error came up while syncing, e.g., it failed to post a webhook.

So I think there's room another metric to record specifically the application step. A gauge counting how many things were applied is appropriate -- I would record how many were attempted, and label according to whether they were successful applied or not.

Tangentially for your purpose, a metric timing (and labelling the success) of each phase of the syncing might be useful too. E.g., generating the manifests, parsing the manifests, doing the application, doing the garbage collection.

squaremo:
> I think the metric should be named `sync_error` and it should act like a boolean

I'm not a fan of metrics that are there just so you can write a very specific alert as simply as possible, but provide no other information. What if you want to alert on the _proportion_ of apply errors to successes? Or an alert on whether the number of manifests being applied is not what you expected -- a mistake in configuration could mean you suddenly have no manifests, and no errors.

mpashka:
Main idea of initial request - https://github.com/fluxcd/flux/issues/2199 - is to provide possibility to get information about current flux state - is everything is in sync and works as expected or something can't be synced and manual intervention is needed. flux_demon_sync_duration_seconds doesn't give flux current state information.
I suppose metric is proper way to get flux _state_.

squaremo:
Sorry, this difference of opinion is going to hold up the PR. @stefanprodan Can we agree a minimal change that makes it mergeable?

I suggest: make the metric a gauge of attempted manifest applications, labelled by success or failure. In other words,

    flux_daemon_sync_manifests{success=true|false}

If you want to alert on _any_failures, you can use

    flux_daemon_sync_manifests{success=false} > 0

stefanprodan:
I'm ok with `flux_daemon_sync_manifests{success=true|false}`. It's easy to reason about and create alerts.

mpashka:
So if sync was successful we put flux_daemon_sync_manifests{success=true}=manifest_number, if not - flux_daemon_sync_manifests{success=false}=manifest_errors
If there was error reading manifests or parsing manifest we put flux_daemon_sync_manifests{success=true}=0 and flux_daemon_sync_manifests{success=false}=number_of_parse_errors? Or just flux_daemon_sync_manifests{success=false}=1?

squaremo:
I'm not sure of the most convenient way to get these numbers, but

```
flux_daemon_sync_manifests{success=false}=manifest_errors
flux_daemon_sync_manifests{success=true}=total_manifest_number - manifest_errors
```

manifest_errors is the number of resource errors received back from the sync, as you have it now (sync.go L163).

mpashka:
Done

mpashka:
Done

pkg/daemon/sync.go
```diff
@@ -53,6 +54,7 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 	if d.GitSecretEnabled {
 		ctxt, cancel := context.WithTimeout(ctx, d.GitTimeout)
 		if err := working.SecretUnseal(ctxt); err != nil {
+			updateSyncManifestsMetric(0, 1)
```
squaremo: Remove these ones -- the metric measures the number of manifests (/resources), so setting 0 or 1 is the wrong unit. You will be able to see the success or failure of syncs in the metric `flux_daemon_sync_duration`.

pkg/daemon/sync.go
```diff
@@ -61,13 +63,15 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 	// Retrieve change set of commits we need to sync
 	c, err := getChangeSet(ctx, ratchet, newRevision, d.Repo, d.GitTimeout, d.GitConfig.Paths)
 	if err != nil {
+		updateSyncManifestsMetric(0, 1)
```
squaremo: As above

pkg/daemon/sync.go
```diff
@@ -61,13 +63,15 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
...
 		return err
 	}
 
 	// Run actual sync of resources on cluster
 	syncSetName := makeGitConfigHash(d.Repo.Origin(), d.GitConfig)
 	resourceStore, err := d.getManifestStore(working)
 	if err != nil {
+		updateSyncManifestsMetric(0, 1)
```
squaremo: As above

pkg/daemon/sync.go
```diff
@@ -149,6 +153,7 @@ func doSync(ctx context.Context, manifestsStore manifests.Store, clus cluster.Cl
 	logger log.Logger) (map[string]resource.Resource, []event.ResourceError, error) {
 	resources, err := manifestsStore.GetAllResourcesByID(ctx)
 	if err != nil {
+		updateSyncManifestsMetric(0, 1)
```
squaremo: As above

pkg/daemon/sync.go
```diff
@@ -165,12 +171,20 @@ func doSync(ctx context.Context, manifestsStore manifests.Store, clus cluster.Cl
 				})
 			}
 		default:
+			updateSyncManifestsMetric(0, 1)
```
squaremo: As above

mpashka:
That's correct. Success or failure of the overall sync process can be obtained by running queries:
delta(flux_daemon_sync_duration_seconds_count{success='true'}[6m]) < 1
or
rate(flux_daemon_sync_duration_seconds_count{success='false'}[6m]) > 0

So we need only measure the case where manifests are correct but there were some problems applying manifests to kubernetes - e.g. configmap size is too big to fit in annotations or immutable field (like label selector) was changed.

Probably it is better to put glux alarming info into documentation as well. I suppose https://github.com/fluxcd/flux/blob/master/docs/references/monitoring.md is a good place for it.

pkg/daemon/sync.go
mpashka: Probably this one is a single place where using metrics still makes sense but it must be adjusted to include number of manifests.
This is because `flux_daemon_sync_duration_seconds_count` usually say something is wrong with infrastructure and thus devops intervention is required. `flux_daemon_sync_manifests` on the other hand is info about invalid manifests in git.
The problem is in either case we can't say how many manifest were succeeded here because effectively success number is 0.

So @squaremo is it ok to leave metric here and put number of manifest parsing errors into it? Or just remove the line?

squaremo: I get what you mean -- it _is_ a problem with individual manifests, we just don't have the numbers handy. How about this: remove it from here for now, and we can consider a further change after this is merged: something like expanding this metric to have a label for the particular stage at which something failed; or, having a more fine-grained metrics for the duration (and success/failure) of sync phases generation / parsing / validation / application / gc.

pkg/daemon/sync.go
mpashka: Fixed.

docs/references/monitoring.md
```diff
@@ -14,5 +14,15 @@ The following metrics are exposed:
 | `flux_daemon_queue_duration_seconds`     | Duration of time spent in the job queue before execution
 | `flux_daemon_queue_length_count`         | Count of jobs waiting in the queue to be run
 | `flux_daemon_sync_duration_seconds`      | Duration of git-to-cluster synchronisation
+| `flux_daemon_sync_manifests`             | Number of successfully synchronized manifests, manifest sync errors
```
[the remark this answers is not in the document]
mpashka: Done

PR #2543, opened 2019-10-23
Run e2e test with bats

2opremio:
This paves the way for adding more e2e tests

Also, stop testing the Helm Operator in the e2e test (since the Helm Operator now lives in its own repo and should be tested there)


squaremo:
Cool. Good job on factoring out the various setup tasks, that is the hard bit!

PR #2544, opened 2019-10-23
Alternate #2479 (config file)

squaremo:
This is my rebase and modest rewrite of #2479, as a PR to inform discussion on how to push it on. 

Main differences:

 - I've left the chart changes out. They can be made separately to the fluxd changes (and often are). In any case, there are different formulations of supporting config files in the chart, and examining that would hold things up
 - `fluxctl install` checks if the mandatory args are supplied in the config file, if they are not passed in flags
 - I took the opportunity to rearrange the flag handling in fluxd/main.go, which (modulo the config file change) had become quite messy
 - fluxd should exit if it finds a config file but can't load it


squaremo:
> * there are different formulations of supporting config files in the chart

Just so I don't leave this dangling:

There are essentially three mechanisms for getting config settings into a configmap or resource via a Helm chart:
 - read from a file that comes with the chart
 - construct the configmap/secret from values given to the chart (either the whole content, or individual settings)
 - refer to a configmap or secret constructed "out of band" (i.e., not by the Helm chart)

1. In the case of bootstrapping Flux using the Helm chart, I'd expect people to supply the various parameters (`git.url` etc.) to the invocation of `helm install`. Asking them to compile a config file first, then supply it, makes more work for them (and more opportunity for spelling mistakes). More problematically, the chart can't refer to a file that doesn't come _with_ the chart -- so this involves people fetching the chart, putting the file in the right place, then invoking `helm install` referring to the local chart directory.

So for this use case, I think it's most useful to be able to tell the chart to create a configmap/secret using the config given _as values_. E.g., to take `git.url` etc., and put them in another resource, rather than as arguments to the fluxd container. (A similar argument can be made for `fluxctl install`.)

2. For the purpose of updating an installation (that uses a config file), it is going to be easier to edit an existing file than to recapitulate all the arguments. For this it would be most convenient to supply the config to the chart using `--set-file`. This would be my second choice for bootstrapping -- it needs a bit more work initially, but if you are going to update using the chart, you'll have to do it anyway.

3. When _automating_ updates (e.g., using the Helm operator), one could either refer to a configmap/secret that's generated some other way, like using Kustomize; or, the values can be supplied and the values->configmap from 1) used.

cmd/fluxd/main.go
```diff
@@ -224,17 +239,61 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Configure viper to check for a environment variables or a config file;
```
2opremio: do we really want to enable env variables? I guess it's one env variable per flag, which seems overkill.

Also, the configuration file may end up evolving into something which isn't flag (e.g. when supporting multiple git repositories) point at which it wouldn't make a lot of sense to use env variables anymore.

2opremio:
Some initial comments (from a conversation with @squaremo ):
* I am missing versioning (since the configuration format may change in the future and may not be flat.
* I have my doubts about using env variables (for the same reason)
* I am missing an e2e test
* It would be good if Flux could pick up changes on the config file on the fly. However that's hard at this point because a lot of things are built initially based on the initial config values. @squaremo wisely point out that it would probably better to watch for config changes and orderly restart the internal logic of the daemon when that happens.

cmd/fluxd/main.go
squaremo: The original PR included that and I assumed it was part of the desired change; I don't myself see any strong argument for having everything set by environment variable. It would simplify things to remove it (and it can always be added later, now the original PR has shown us how)

squaremo: The original PR included that and I assumed it was part of the desired change; I don't myself see any strong argument for having everything set by environment variable. It would simplify things to remove it (and it can always be added later, now the original PR has shown us how)

pkg/config/config.go
```diff
@@ -0,0 +1,88 @@
...
+package config
+
+import (
+	"fmt"
+	"time"
+)
+
+const ConfigPath = "/etc/fluxd/conf"
+const ConfigName = "flux-config"
+const FluxConfigVersion = "v1"
```
2opremio: Can you put this in a single const block?

We should start doing some linting int e

pkg/config/config.go
```diff
@@ -0,0 +1,88 @@
...
+	K8sVerbosity             int           `mapstructure:"k8s-verbosity"`
+	SSHKeygenDir             string        `mapstructure:"ssh-keygen-dir"`
+	ManifestGeneration       bool          `mapstructure:"manifest-generation"`
+	Connect                  string        `mapstructure:"connect"`
+	Token                    string        `mapstructure:"token"`
+	RPCTimeout               time.Duration `mapstructure:"rpc-timeout"`
+	DockerConfig             string        `mapstructure:"docker-config"`
+}
+
+func (c Config) IsValid() error {
```
2opremio: Is this invoked implicitly by Viper?

chart/flux/flux-config.yaml
```diff
@@ -0,0 +1,5 @@
+git-url: "git@github.com:weaveworks/flux-get-started"
```
2opremio: This should contain a version, shouldn't it?

In fact ... shouldn't the helm installation e2e test be failing because of this?

squaremo: This is in the chart, and should probably be excised from this PR entirely.
I think it won't affect the end-to-end test because there's nothing in the chart that includes this file. Since it doesn't get mounted into the container, it doesn't get checked for validity.

pkg/config/config.go
squaremo: No, it's called in main.go: https://github.com/fluxcd/flux/pull/2544/files#diff-350e484490d6241fa81c13942d160480R263

It's a bit of a dance because viper wants to cleverly do everything itself.

chart/flux/flux-config.yaml
2opremio: Ah, I see

pkg/config/config.go
2opremio: Sorry, I missed it

chart/flux/flux-config.yaml
stefanprodan: Let's remove the chart changes from this PR and address it separately.

2opremio:
🤞 to tests working after rebasing

cmd/fluxctl/install_cmd.go
```diff
@@ -25,49 +29,87 @@ func (opts *installOpts) Command() *cobra.Command {
...
+	cmd.Flags().StringVar(&opts.params.GitEmail, "git-email", "", "Email to use as git committer")
+	cmd.Flags().BoolVar(&opts.params.ConfigAsConfigMap, "config-as-configmap", false,
+		"Create a ConfigMap to hold the Flux configuration given with --config-file. If false, a Secret will be used. Ignored if --config-file is not given.")
+	cmd.Flags().BoolVar(&opts.params.GitReadOnly, "git-readonly", false, "Tell flux it has readonly access to the repo")
+	cmd.Flags().BoolVar(&opts.params.ManifestGeneration, "manifest-generation", false, "Whether to enable manifest generation")
+	cmd.Flags().StringVar(&opts.params.Namespace, "namespace", "", "Cluster namespace in which to install Flux")
+	cmd.Flags().BoolVar(&opts.params.RegistryScanning, "registry-scanning", true, "Scan container image registries to fill in the registry cache")
+
+	// These are flags for control of the output, etc.
+	cmd.Flags().StringVar(&opts.configFile, "config-file", "", "Make this file into a secret or configmap for Flux to mount as config")
```
2opremio:
```suggestion
	cmd.Flags().StringVar(&opts.configFile, "config-file", "", "Make this file into a secret (or configmap if --config-as-configmap is supplied) for Flux to mount as config")
```

2opremio: We should also explain what formats are supported for the config file. I think we should be restrictive the format (toml or yaml, for instance) instead of allowing every format viper supports.

(The minimum requirement IMO is that it can support key-value pairs and multiple git repos)

cmd/fluxctl/install_cmd.go
```diff
@@ -25,49 +29,87 @@ func (opts *installOpts) Command() *cobra.Command {
...
+	// To make sure the mandatory flags are set, we mimic what fluxd
+	// will do, and load from both the config file (if supplied, it
+	// will be mounted into the fluxd container as a configmap or
+	// secret) and the flags (which will go into the template as fluxd
+	// args).
+
+	if opts.configFile != "" {
+		viper.SetConfigFile(opts.configFile)
+		if err := viper.ReadInConfig(); err != nil {
+			return fmt.Errorf("unable to read config at %s (possibly a missing file extension, try .yaml or .json): %s", opts.configFile, err)
```
2opremio: I would only support a format

cmd/fluxctl/install_cmd.go
2opremio: My reasoning to restricting it to a single format is that:
1. Right now we only need key-values but we may want to store something more complicated than that in the future (e.g. multiple git repos) and not all formats will support them in the same way.
2. I think it can make it confusing for the end user (and will also make support more complicated) if we support every format viper supports.

cmd/fluxd/main.go
```diff
@@ -97,7 +99,15 @@ func (set stringset) has(possible string) bool {
 }
 
 func main() {
-	// Flag domain.
+	// --- Flags ---
+
+	// Generally we want to allow configuration to be supplied in a
+	// config file OR in the environment OR via command-line
```
2opremio: I thought we agreed on not accepting env variables. Is this simply not up-to-date?

2opremio:
I think `main()` has gotten big enough to justify doing the flag parsing and configuration creation in a separate file.

cmd/fluxd/main.go
```diff
@@ -106,106 +116,119 @@ func main() {
...
-		// k8s-scope settings
-		k8sNamespaceWhitelist = fs.StringSlice("k8s-namespace-whitelist", []string{}, "restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set")
-		k8sAllowNamespace     = fs.StringSlice("k8s-allow-namespace", []string{}, "restrict all operations to the provided namespaces")
-		k8sDefaultNamespace   = fs.String("k8s-default-namespace", "", "the namespace to use for resources where a namespace is not specified")
-		k8sExcludeResource    = fs.StringSlice("k8s-unsafe-exclude-resource", []string{"*metrics.k8s.io/*", "webhook.certmanager.k8s.io/*", "v1/Event"}, "do not attempt to obtain cluster resources whose group/version/kind matches these glob expressions. Potentially unsafe, please read its documentation first")
-		k8sVerbosity          = fs.Int("k8s-verbosity", 0, "klog verbosity level")
+		versionFlag = fs.Bool("version", false, "get version number")
 
+		// not present in the config struct, since "extra for experts" or only relevant when running locally
+		kubernetesKubectl = fs.String("kubernetes-kubectl", "", "optional, explicit path to kubectl tool")
```
2opremio: I don't completely disagree, but it seems some users are running Flux in a separate cluster (CC @bricef )

cmd/fluxd/main.go
```diff
@@ -106,106 +116,119 @@ func main() {
...
-		manifestGeneration = fs.Bool("manifest-generation", false, "search for .flux.yaml files to generate manifests")
-
-		// upstream connection settings
-		upstreamURL = fs.String("connect", "", "connect to an upstream service e.g., Weave Cloud, at this base address")
-		token       = fs.String("token", "", "authentication token for upstream service")
-		rpcTimeout  = fs.Duration("rpc-timeout", 10*time.Second, "maximum time an operation requested by the upstream may take")
-
-		dockerConfig = fs.String("docker-config", "", "path to a docker config to use for image registry credentials")
+		sshKeyBits = optionalVar(fs, &ssh.KeyBitsValue{}, "ssh-keygen-bits", "-b argument to ssh-keygen (default unspecified)")
+		sshKeyType = optionalVar(fs, &ssh.KeyTypeValue{}, "ssh-keygen-type", "-t argument to ssh-keygen (default unspecified)")
```
2opremio: Why doesn't it make sense to put these two in the config file?

cmd/fluxd/main.go
2opremio: I don't see the harm in having this in the config file

cmd/fluxd/main.go
```diff
@@ -232,17 +253,57 @@ func main() {
 		os.Exit(0)
 	}
 
+	viper.SetConfigName(config.ConfigName)
+	viper.AddConfigPath(config.ConfigPath)
+
```
2opremio:
```suggestion
viper.SetConfigType("foo")
```

once/if we settle on a single format

cmd/fluxd/main.go
```diff
@@ -232,17 +253,57 @@ func main() {
...
+			os.Exit(2)
+		}
+	} else {
+		fmt.Fprintf(os.Stderr, "Error: using configuration at %s, with command-line overrides\n", viper.ConfigFileUsed())
+	}
+	// Bind Viper to the pflags defined above
+	viper.BindPFlags(fs)
+
+	var config config.Config
+	viper.Unmarshal(&config)
```
2opremio: Is the returned error intentionally ignored?

cmd/fluxd/main.go
```diff
@@ -232,17 +253,57 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "Info: config file not found: %s\n", err.Error())
+		} else {
+			fmt.Fprintf(os.Stderr, "Error: found config file at %s but failed to load it: %s\n", viper.ConfigFileUsed(), err.Error())
+			os.Exit(2)
+		}
+	} else {
+		fmt.Fprintf(os.Stderr, "Error: using configuration at %s, with command-line overrides\n", viper.ConfigFileUsed())
+	}
+	// Bind Viper to the pflags defined above
+	viper.BindPFlags(fs)
```
2opremio: Why is the error ignored?

cmd/fluxd/main.go
```diff
@@ -232,17 +253,57 @@ func main() {
...
+	if err = config.IsValid(); viper.ConfigFileUsed() != "" && err != nil {
+		fmt.Fprintf(os.Stderr, "found config file at %s but it is not valid: %s\n", viper.ConfigFileUsed(), err.Error())
+		os.Exit(2)
+	}
+
+	// --- From this point, we can just consult the config struct for values ---
+
+	// viper.IsSet, which we could otherwise use for determining
+	// whether a flag has been supplied, is broken:
+	// https://github.com/spf13/viper/pull/331. So we have to proceed
```
2opremio: The PR has been merged, does it make sense to bump viper remove this?

test/e2e/lib/install.bash
```diff
@@ -49,26 +49,40 @@ function uninstall_flux_with_helm() {
...
+
   kubectl apply -f "$HELMRELEASE_CRD_URL"
-  kubectl apply -k "${kustomtmp}/${kustomization_dir}" >&3
-  kubectl -n "${FLUX_NAMESPACE}" rollout status -w --timeout=30s deployment/flux
+  kubectl apply -k "${kustomtmp}/${kustomization_dir}/" >&3
+  kubectl -n "${FLUX_NAMESPACE}" rollout status -w --timeout=30s deployment/flux ||
+    (
+      kubectl -n "${FLUX_NAMESPACE}" describe deployment/flux
+      kubectl -n "${FLUX_NAMESPACE}" logs deployment/flux
+    )
```
2opremio: I think this should exit with a non-zero status, otherwise the test will continue

pkg/install/templates/flux-deployment.yaml.tmpl
```diff
@@ -142,20 +166,33 @@ spec:
...
+        - --git-path={{ StringsJoin .GitPaths "," }}
+        {{- else }}
         # Include this if you want to restrict the manifests considered by flux
         # to those under the following relative paths in the git repository
-        # - --git-path=subdir1,subdir2{{ end }}{{ if .GitLabel }}
-        - --git-label={{ .GitLabel }}{{ end }}{{ if .GitUser }}
-        - --git-user={{ .GitUser   }}{{ end }}{{ if .GitEmail }}
-        - --git-email={{ .GitEmail }}{{ end }}
+        # - --git-path=subdir1,subdir2
+        {{- end }}
```
2opremio: so much nicer, thanks

test/e2e/12_fluxctl_config.bats
```diff
@@ -0,0 +1,33 @@
...
+  install_git_srv
+  tmp_config=$(mktemp -d)
+  defer rm -rf "$tmp_config"
+  cat > "${tmp_config}/flux.yaml" << EOF
+flux-config-version: v1
+EOF
+  install_flux_with_fluxctl '12_flux_config' '' --config-file="${tmp_config}/flux.yaml"
+}
+
+@test "'fluxctl install --config-file=...' smoke test" {
```
2opremio: I would add a couple of scenarios:

1. Setting an actual configuration value as opposed to a an empty config file and making sure it reaches Flux.
2. Checking the precedence of flags over configuration file entries by setting a configuration option in both places.

A good candidate for this is the flux URL which can be retrieved with `fluxctl sync` (alternatively could maybe just `curl` into `/v9/git-repo-config` opening the door to the branch etc ... )

pkg/install/go.mod
```diff
@@ -3,9 +3,16 @@ module github.com/fluxcd/flux/pkg/install
...
+	github.com/ghodss/yaml v1.0.0 // indirect
+	github.com/gobwas/glob v0.2.3 // indirect
 	github.com/instrumenta/kubeval v0.0.0-20190918223246-8d013ec9fc56
+	github.com/pkg/errors v0.9.1 // indirect
 	github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749
 	github.com/shurcooL/vfsgen v0.0.0-20181202132449-6a9ea43bcacd
 	github.com/stretchr/testify v1.4.0
 	golang.org/x/tools v0.0.0-20200121210457-b3205ff6fffe // indirect
+	k8s.io/apimachinery v0.17.2 // indirect
+	k8s.io/helm v2.16.1+incompatible
```
2opremio: These are the sort of dependencies we were trying to avoid with a separate install module.

docs/tutorials/get-started.md
```diff
@@ -58,6 +58,25 @@ Wait for Flux to start:
...
+### Using a configuration file
+
+You can also use a configuration file to configure Flux instead of specifying command line arguments:
+
+```sh
+echo "git-url: git@github.com:${GHUSER}/flux-get-started" >> flux-config.yaml
+fluxctl install --config-file=flux-config.yaml --namespace=flux | kubectl apply -f -
+```
+
+The configuration file accepts any of the command-line
```
2opremio: Not really (as it is right now)

pkg/install/install.go
```diff
@@ -63,3 +72,18 @@ func FillInTemplates(params TemplateParameters) (map[string][]byte, error) {
 	}
 	return result, nil
 }
+
+func ConfigContent(configFile io.Reader, configAsConfigMap bool) (string, error) {
+	tmp, err := ioutil.ReadAll(configFile)
+	if err != nil {
+		return "", fmt.Errorf("unable to read config file: %s", err)
+	}
+	var f chartutil.Files
```
2opremio: Having helm as a dependency just for this seems overkill.

pkg/install/install.go
```diff
@@ -26,10 +27,18 @@ type TemplateParameters struct {
 	Namespace          string
 	ManifestGeneration bool
 	AdditionalFluxArgs []string
+	ConfigFileContent  string
```
2opremio: Given that this package is users externally. The value expected for this field is really awkward (particularly given its name) since it must be YAML with a the name of the file as a key and the body of the file (base64 encoded) as a value.

If we only support a config file format then there is no need to specify the filename (we can hardcode it), then we can decide to encode it as base64 or not based on whether a secret is used.

2opremio:
@squaremo sorry I caused more conflicts with #2800

2opremio:
I have a question. Is there a recommended way to deprecate configuration values with viper?

pkg/install/go.mod
stefanprodan: Oh no, please do not add Helm and Kubernetes back to the install package.

squaremo: To be fair, this is something that changed out from under me, and not something I *re*introduced.

test/e2e/12_fluxctl_config.bats
2opremio: I would add a third one:

3. Test that the default value of a configuration option is set appropriately (with flags and configuration files handling default values doesn't seem straightforward)

cmd/fluxctl/install_cmd.go
squaremo: Related thing: I'm tempted to make all the field names `camelCase`, rather than using hyphens, since the latter look pretty strange in a YAML file. Version 2 of the format could even group the arguments into substructs -- though at that point, I would start to worry about it being confused for the Helm chart values (which do that too, in a reasonably thought-through but different way).

cmd/fluxd/main.go
squaremo: Correct: not up to date.

cmd/fluxctl/install_cmd.go
squaremo: > I'm tempted to make all the field names `camelCase`

It is not .. easy .. to have the flags as `--git-user` and the config file entries as `gitUser`, understandably but annoyingly. Specifically: `viper.BindPFlags` will use all the hypehated flag names, while viper.Decode will use all the camelCase field names, and they won't be related to the flag names. Ugh.

cmd/fluxctl/install_cmd.go
squaremo: This is not really supported by viper, so far as I can tell. You can either support all of the formats as selected by the file extension, or specify an exact format and it will look for an extensionless file.

2opremio: I think we can either use a single format (my preference) or, alternatively, do some sort of content detection ourselves and pass the exact expected format to viper.

I don't really see a reason for supporting multiple formats at this point.

cmd/fluxd/main.go
squaremo: They've always been off to the side of the "normal" flags, perhaps since they pertain to bootstrapping Flux rather than how it operates once running, or because the values are sent straight to the `ssh` package (and not examined). I don't think either of those are a particularly strong rationale; on the other hand, it would take some reimplementation to be able to put them in the config file, and I don't think it's worth it.

2opremio: I see. IMO they should either be allowed in the configfile or deprecated. I think it's important to keep the invariant of "everything can be configured through both flags and the config file" for currently non-deprecated flags.

squaremo: > I think it's important to keep the invariant of "everything can be configured through both flags and the config file" for currently non-deprecated flags.

I don't. For example, `--kubernetes-kubectl` is a flag that is useful (well, kind of), but doesn't make sense in the config file, since it is about the environment in which fluxd runs rather than its behaviour when running.

Actually, a better example: `--k8s-secret-volume-mount-path` only exists to tell fluxd about where a volume has been mounted for it to use, i.e., you match its argument with a mount path elsewhere in the same Deployment definition. Putting it in the config file makes that into a trapeze act -- now, the values in two separate files, possibly from two completely different sources, must match, and there's no reason to vary them independently. (And I should really take that field out of the config file definition, now I have noticed it.)

squaremo: `--docker-config` arguably belongs in the same category; useful only when it matches something specified elsewhere in the Deployment manifest (or when running locally).

cmd/fluxd/main.go
squaremo: Nop

cmd/fluxd/main.go
squaremo: This method was intended to be for the situation in which the config file has no extension -- but, I have checked, and it turns out that you can set the config file name to e.g., "flux-config.yaml" and set the type, to get the effect desired here.

cmd/fluxd/main.go
squaremo: Nop .. I mean, it was like that when I got here :-P I'll correct it.

pkg/install/install.go
squaremo: When the original PR was made, Helm was already a dependency. The requirement is more modest now, though; this can just be done in a template, as you suggest elsewhere.

squaremo:
> I have a question. Is there a recommended way to deprecate configuration values with viper?

Doesn't look like it; I would think you have to do your own schema versioning.

pkg/install/install.go
squaremo: Yeah it's a bit of a burr on the install API that you have to know how to create the file content. It's a bit of a faff to get the templates to e.g., indent things reliably, but :crossed_fingers: possible.

2opremio:
Is this ready for another review round, or do you want to resolve the conflicts first? (I would wait for #2813 and maybe #2817 to land before resolving the conflicts though)

cmd/fluxctl/install_cmd.go
```diff
@@ -97,6 +153,14 @@ func (opts *installOpts) RunE(cmd *cobra.Command, args []string) error {
 	}
 
 	for fileName, content := range manifests {
+		// The way the templating works, it will run through each
+		// file; but some may elide the entire content if it's not
+		// needed. So, don't bother writing anything out if the
+		// template evaluated to just whitespace.
+		if len(bytes.TrimSpace(content)) == 0 {
```
2opremio: Is this necessary? It is already done in `FillInTemplates()` (except for the trimming)

https://github.com/fluxcd/flux/blob/324d5b2d162e0b76fcaab0f6140d5ca2951ae093/pkg/install/install.go#L59-L61

(it was included in the midst of this PR)

cmd/fluxd/config.go
```diff
@@ -0,0 +1,163 @@
...
+		fs.Int(flagName, def, desc)
+		bindOrBail(fieldName, flagName)
+	}
+
+	defineFloat64 := func(fieldName, flagName string, def float64, desc string) {
+		fs.Float64(flagName, def, desc)
+		bindOrBail(fieldName, flagName)
+	}
+
+	defineString("LogFormat", "log-format", "fmt", "change the log format.")
```
2opremio: I was hoping for the `define` functions to automate the transforming from flag format into config-file option format (e.g. from `log-format` to `LogFormat`)

2opremio: Ah, I see that acronyms make it a bit hard (e.g. `ci` to `CI`) . But `CI` to `ci` is straight-forward. So maybe we can do it the other way around.

test/e2e/12_fluxctl_config.bats
2opremio: This hasn't been addressed

cmd/fluxd/config.go
squaremo: The flags are not always straight-forward transliterations of the field names. For example: `k8s-unsafe-exclude-resource` vs `K8sExcludeResource`; `git-ci-skip` vs `GitCISkip`. Could it be made to line up? Probably, since it's the keys (that appear in the mapstructure tag) that we care about, less so the field names, which wouldn't pass linting as they are anyway.

cmd/fluxctl/install_cmd.go
squaremo: It is now, yes -- I created trouble for myself by fixing it while I was rebasing :-/

pkg/install/templates/flux-config.yaml.tmpl
```diff
@@ -0,0 +1,24 @@
+{{ if .ConfigFileContent -}}
```
2opremio: Sorry I am noticing this now. But we should probably handle the Memcached deployment in the same way with `.RegistryDisableScanning` (right now we exclude it in the installation code).

cmd/fluxd/config.go
2opremio: We don't need to fix it right now.

However, the acronyms case can be dealt with by counting consecutive capital letters.

As for mismatches like `Unsafe` there are alternatives (e.g. using `K8sUnsafeExcludeResource`, which frankly I think we should be using or only providing names for exceptions).

One neat way to solve this would be `reflect`ing on the `Config` struct (we could use annotations for the name exceptions). We could even include the flag metadata in the struct itself and make the `bind` calls automatically.

kingdonb:
Closing, as this appears to be a large change not suited for maintenance mode, and that it has already been moved to the "Off the table" column in the kanban board.

PR #2546, opened 2019-10-24
Add ignore IDEs file rules

mosesyou:
I think we can add some git ignore file rules to make developers more friendly when contributing to source code.


hiddeco:
Thanks for your contribution!

This has already been proposed once (#1346), and I agree with Stefan that this should be in your `$HOME/.gitignore` file. The reasoning behind this for me is:

- The `.gitignore` file in this project should ignore things that are the direct result of e.g. running the build locally as instructed in the `Makefile`, but that should never be committed to Git. The `bin`, `build`, and `cache` folders are great examples of this (and are included in the `.gitignore`).
- There are too many IDEs on this world to keep track of, and it is impossible to include them all, so maintaining this list will become a burden.
- Contributors will have this issue for any project they work on that does not include IDE `.gitignore` rules in the project. Including it in as much projects as possible is like patching the leaves of a tree, while it can be fixed at the roots (using a `$HOME/.gitignore` file).

PR #2549, opened 2019-10-24
Add `SafeURL` helper method to `GitRemoteConfig`

### LINKED ISSUE #2548: Fluxctl exposes `GIT_AUTHKEY` in console output
**Describe the bug**

When using `fluxctl sync` the used authentication key / token is prominently displayed as part of the console output.

**To Reproduce**
1. Upgrade fluxd and fluxctl to 1.15.0, and set the deployment up to [use Git over HTTPS](https://github.com/fluxcd/flux/blob/master/docs/guides/use-git-https.md)
2. Run `fluxctl sync`
3. See "Synchronizing with https://…:MYTOKEN@…" as the very first line of output

**Expected behavior**

The authentication token should be hidden like it's done in fluxd's log output.

**Logs**
N/A

**Additional context**
- Flux version: 0.15.0
- Helm Operator version: N/A
- Kubernetes version: 1.15.3
- Git provider: GitLab on premise
- Container registry provider: GitLab on premise

hiddeco:
During the development of the (secure) Git HTTPS credential feature,
I did not take the response of the `GitRepoConfig` API method into
account. As a direct result, the `fluxctl sync` command still exposes
the full Git URL in the logs.

This commit embeds the `git.Remote` in the `GitRemoteConfig` structure,
so that we can re-use the `SafeURL` helper of this structure in the
`fluxctl` code. To ensure the JSON representation stays the same, a
JSON struct tag has been added to the `git.Remote` URL field.

Fixes #2548

### DIFF
--- a/pkg/api/v6/api.go
+++ b/pkg/api/v6/api.go
@@ -49,7 +49,7 @@ type ControllerStatus struct {
 // --- config types
 
 type GitRemoteConfig struct {
-	URL    string `json:"url"`
+	git.Remote
 	Branch string `json:"branch"`
 	Path   string `json:"path"`
 }
--- a/cmd/fluxctl/sync_cmd.go
+++ b/cmd/fluxctl/sync_cmd.go
@@ -46,10 +46,10 @@ func (opts *syncOpts) RunE(cmd *cobra.Command, args []string) error {
 	case git.RepoReady:
 		break
 	default:
-		return fmt.Errorf("git repository %s is not ready to sync (status: %s)", gitConfig.Remote.URL, string(gitConfig.Status))
+		return fmt.Errorf("git repository %s is not ready to sync (status: %s)", gitConfig.Remote.SafeURL(), string(gitConfig.Status))
 	}
 
-	fmt.Fprintf(cmd.OutOrStderr(), "Synchronizing with %s\n", gitConfig.Remote.URL)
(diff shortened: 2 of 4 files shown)


squaremo:
Thought: would implementing `Remote.String()` (and perhaps making URL unexported) help with some of the messing around with SafeURL()?

hiddeco:
@squaremo I wasn't sure what modifications I could make to the structure without it effecting e.g. Weave Cloud or flux-web. If you think it is safe to do so, I am happy to adjust the commit.

squaremo:
> I wasn't sure what modifications I could make to the structure without it effecting e.g. Weave Cloud or flux-web

Hmm yeah true, it might affect serialisation. OK, we can leave that for now.

pkg/api/v6/api.go
```diff
@@ -54,6 +58,17 @@ type GitRemoteConfig struct {
 	Path   string `json:"path"`
 }
 
+func (c GitRemoteConfig) SafeURL() string {
```
squaremo: It's a shame to have to recapitulate this method here -- is there any way of changing the type def so it just uses git.Remote?

hiddeco: I embedded the `git.Remote` struct, and added a JSON tag to the `URL` field. This should ensure two things:

1. The JSON representation of the structure is the same
2. The old `(GitRemoteConfig{}).Remote.URL` should continue to work for whoever depends on it.

PR #2551, opened 2019-10-24
Update the automation interval flag in the chart

srueg:
Since the logs tell me to do so:
```
Flag --registry-poll-interval has been deprecated, changed to --automation-interval, use that instead
```

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -218,7 +218,7 @@ spec:
           {{- if .Values.manifestGeneration }}
           - --manifest-generation=true
           {{- end }}
-          - --registry-poll-interval={{ .Values.registry.pollInterval }}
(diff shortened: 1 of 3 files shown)


chart/flux/templates/deployment.yaml
```diff
@@ -220,7 +220,7 @@ spec:
           {{- if .Values.manifestGeneration }}
           - --manifest-generation=true
           {{- end }}
-          - --registry-poll-interval={{ .Values.registry.pollInterval }}
+          - --automation-interval={{ .Values.registry.pollInterval }}
```
hiddeco: Would you be able to change the `pollInterval` to `automationInterval` too, in a backwards compatible manner?

srueg: Like this?

PR #2552, opened 2019-10-24
Add 'fluxctl install' e2e smoke test

2opremio:
Also:

* Factor out polling machinery
* Reduce the polling interval during e2e tests so that they don't take as long

### DIFF
--- a/test/e2e/lib/install.bash
+++ b/test/e2e/lib/install.bash
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 
 function install_tiller() {
-if ! helm version > /dev/null 2>&1; then # only if helm isn't already installed
+  if ! helm version > /dev/null 2>&1; then # only if helm isn't already installed
     kubectl --namespace kube-system create sa tiller
     kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
     helm init --service-account tiller --upgrade --wait
-fi
+  fi
 }
 
 function uninstall_tiller() {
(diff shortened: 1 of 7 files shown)


test/e2e/lib/install.bash
```diff
@@ -46,11 +46,34 @@ function uninstall_flux_with_helm() {
...
+  $fluxctl_install_cmd | \
+    sed 's%docker\.io/fluxcd/flux:.*%fluxcd/flux:latest%' | \
+    sed "s%--git-email=foo%--git-email=foo\\$eol        - --git-poll-interval=10s%" | \
+    sed "s%--git-email=foo%--git-email=foo\\$eol        - --sync-interval=10s%" | \
+    sed "s%--git-email=foo%--git-email=foo\\$eol        - --registry-exclude-image=\*%" | \
+    kubectl apply -f -
+  kubectl -n "${FLUX_NAMESPACE}" rollout status deployment/flux
+  # Add the known hosts file manually (it's much easier than editing the manifests to add a volume)
+  local flux_podname=$(kubectl get pod -n flux-e2e -l name=flux -o jsonpath="{['items'][0].metadata.name}")
+  kubectl exec -n "${FLUX_NAMESPACE}" "${flux_podname}" -- sh -c "echo '$(cat ${FIXTURES_DIR}/known_hosts)' > /root/.ssh/known_hosts"
```
2opremio: This is pretty horrible. I'm open for suggestions.

2opremio: this comment is also aimed at the sed expressions above (I tried to make a comment on the full block of code but it seems I failed)

hiddeco: Doesn't the `install` package already support `additionalArguments`? What about instead of doing the `sed`, we make a small adjustment to `fluxctl install` to accept a string slice of additional arguments, and then utilize this here?

2opremio: I like your suggestions. I initially incorporated `--extra-flux-arguments`, but @squaremo thought it was better to edit the output.

@squaremo I still think there is a valid usecase here, could you reevaluate?

2opremio: @hiddeco regardless, I will _fix_ this in a separate PR. Are you OK with the rest?

test/e2e/lib/poll.bash
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+function poll_until_equals() {
+  local what="$1"
+  local expected="$2"
+  local check_cmd="$3"
+  poll_until_true "$what" "[ '$expected' = \"\$( $check_cmd )\" ]"
+}
+
+function poll_until_true() {
```
hiddeco: Can we either add a comment here that this waits for a max of `2m`, or make it more obvious by making it configurable?

I think otherwise we all individually end up making the `24*5` calculation multiple times during the debugging of a (time based) failing test.

2opremio: sure, that makes sense

2opremio: I was waiting for a test needed a different period/total wait time to add the parameters

test/e2e/lib/install.bash
squaremo: Is it possible to run the output of `fluxctl install` through kustomize to do the patching you need?

PR #2560, opened 2019-10-28
e2e: generate SSH secret per test case

hiddeco:
The issue with the setup before this commit surfaced while adding a
new test suite for GPG functionalities.

Due to `fluxctl install` creating a boilerplate secret with the same
name as the secret generated by `run.bash`, it was removed during the
teardown of the test suite which tests the command, as this simply
`kubectl delete -f -`s the output of `fluxctl install`.

This caused my own tests to never get past the point of booting a
new git server instance, as the secret it was trying to mount was
no longer present.

This commit adds two helper methods to `lib/install.bash`,
`generate_ssh_secret` and `delete_generated_ssh_secret` which should
be run during the setup and teardown of tests which require an SSH
key (and `known_hosts`) file to be present. Both functions accept
a parameter to control what the secret should be named, which falls
back to `flux-git-deploy`. This same behaviour has been introduced
to `install_git_srv`.

### DIFF
--- a/test/e2e/lib/install.bash
+++ b/test/e2e/lib/install.bash
@@ -69,8 +69,32 @@ function uninstall_flux_with_fluxctl() {
 
 }
 
+function generate_ssh_secret() {
+  local secret_name=${1:-flux-git-deploy}
+  local gen_dir
+  gen_dir=$(mktemp -d)
+
+  ssh-keygen -t rsa -N "" -f "$gen_dir/id_rsa"
+  kubectl create secret generic "$secret_name" \
+    --namespace="${FLUX_NAMESPACE}" \
+    --from-file="${FIXTURES_DIR}/known_hosts" \
+    --from-file="$gen_dir/id_rsa" \
+    --from-file=identity="$gen_dir/id_rsa" \
+    --from-file="$gen_dir/id_rsa.pub"
+  rm -rf "$gen_dir"
+  echo "$secret_name"
+}
+
+function delete_generated_ssh_secret() {
+  local secret_name=${1:-flux-git-deploy}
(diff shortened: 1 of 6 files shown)


test/e2e/lib/install.bash
```diff
@@ -69,8 +69,32 @@ function uninstall_flux_with_fluxctl() {
...
+function delete_generated_ssh_secret() {
+  local secret_name=${1:-flux-git-deploy}
+  kubectl delete -n "${FLUX_NAMESPACE}" secret "$secret_name"
+}
+
 function install_git_srv() {
-  kubectl apply -n "${FLUX_NAMESPACE}" -f "${E2E_DIR}/fixtures/gitsrv.yaml"
+  local secret_name=${1:-flux-git-deploy}
+
+  sed "s/\$GIT_SECRET_NAME/$secret_name/" <"${E2E_DIR}/fixtures/gitsrv.yaml" | kubectl apply -n "${FLUX_NAMESPACE}" -f -
```
2opremio: Maybe it's cleaner to use `envsubst`?

2opremio: Also, do we need it to make the name configurable?

hiddeco: For `envsubst` the variable needs to be set as an environment variable.

I made the name configurable so that we are able to bootstrap the secret for any kind of setup that we may need in a test case, and to be able to create e.g. different secrets for the Helm operator and the Flux daemon.

hiddeco:
@2opremio as you are sick and should rest, I am giving Stefan the honour of reviewing this as it is blocking for the next PR I have lined up to take care of the environment variables.

In case you see anything that should be fixed or aligned, please comment and I will fix it post-merge.

2opremio:
Thanks Stefan!!

Issue #2428, opened 2019-09-04
Kustomize and deleted resources


emas80:
**Describe the bug**
I have a working setup of Flux using Kustomize, everything has been done following the documentation. I am very happy of the final result.

The problem is, I have removed some resources (deployments + related configmap hpa etc) by deleting the yaml from Git, and during the first iteration Flux gets completely stuck and logs

`
ts=2019-09-04T20:04:21.001530336Z caller=loop.go:111 component=sync-loop event=refreshed url=ssh://git@github.com/XXXX.git branch=master HEAD=a300161964a00f8d26072f7bf0600e41cdba71ed
ts=2019-09-04T20:04:24.083994965Z caller=loop.go:85 component=sync-loop err="loading resources from repo: cannot patch generated resources: missing resource (team-blue:deployment/Deployment) for patch"
ts=2019-09-04T20:04:39.688918792Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2019-09-04T20:04:41.352761131Z caller=images.go:23 component=sync-loop error="getting unlocked automated resources: cannot patch generated resources: missing resource (team-blue:deployment/Deployment) for patch"
`

That is because, instead of first deleting the resources and then "fixing" the patch file, it first tries to apply the patch file and it complains because such resources do not exist anymore.
Nothing gets deleted from the cluster, please note that instead any change/new resource is correctly created and Git is synchonized in the case of any releases.

In order to unlock Flux, I have to remove the references from the patch file - patching the patch - and after that Flux picks the new file, and it is able to remove the resources from the cluster.

**To Reproduce**
Create some resources, apply a flux-patch.yaml - for example by pushing a new docker image that triggers the automatic release + git update.
Then delete the resource by removing the files from Git - everything except the content of the patch file.

**Expected behavior**
I would expect that the resources are deleted and the references removed from the patch file, because at the end of the day the patch file has been created by Flux automatically during a release.

**Additional context**
Add any other context about the problem here, e.g 
- Flux version: 1.14.2

jwenz723:
I experienced this issue today and had a difficult time figuring out what was happening because the error message is rather vague: `getting unlocked automated resources: cannot patch generated resources: missing resource (namespace:deployment/Deployment) for patch`

It would be helpful if the error message included the name of the deployment that it is trying to patch, or at least a reference to the file containing the patch that is failing.

squaremo:
[...]

Issue #2324, opened 2019-08-05
Can't release new version of app, not committing it to git


yellowmegaman:
**Describe the bug**
Can't perform `fluxctl release` for new release, not yet committed to git

**To Reproduce**
Steps to reproduce the behaviour:
0. GKE + flux
1. deploy any app via flux/git
2. build new version of your app
3. list images of your app via fluxctl
4. fluxctl release new tag

**Expected behavior**
Change to new tag, add tag to git by fluxd

**Logs**
```
$ fluxctl list-images --workload backend:statefulset/backend
E0805 18:46:19.076371    6706 portforward.go:385] error copying from local connection to remote stream: read tcp4 127.0.0.1:39745->127.0.0.1:38966: read: connection reset by peer
WORKLOAD                     CONTAINER                    IMAGE                                                                      CREATED
backend:statefulset/backend  backend                      gcr.io/titanium-messenger-001/mmts-backend                                 
                                                          |   k8s-master-7a7a4419fbca5e3c9dd307443b7fac0de1f5a6d3                    05 Aug 19 09:13 UTC
                                                          |   k8s-WEL-220-futures-to-tasks-76a18b9b071b5b80d2f7d4f45565ac53cd6c9422  05 Aug 19 07:16 UTC
                                                          |   k8s-WEL-220-futures-to-tasks-3a5d445a8c57b22832316def36e5ec611a261248  05 Aug 19 07:13 UTC
                                                          |   k8s-WEL-220-futures-to-tasks-22eb692819799d733c80efb575f89278c6a511bc  05 Aug 19 07:12 UTC
                                                          |   k8s-WEL-220-futures-to-tasks-0e1d936ae8dae40638475dd0c5155a204d64345a  05 Aug 19 07:03 UTC
                                                          |   k8s-latest                                                             02 Aug 19 11:54 UTC
                                                          |   k8s-master-40f8de44031ab06b2aec060cf603e38a65a44803                    02 Aug 19 11:54 UTC
                                                          '-> k8s-master-1d85e954134a0b576c8e3715a69d40a7cd362573                    02 Aug 19 11:27 UTC
                                                              k8s-fix-program-durations-26e19d3426245d17864f2a9487913f0796c876ea     01 Aug 19 10:42 UTC
                                                              k8s-master-41e13458a943122e70ccdeb28c510271bc30a1fd                    31 Jul 19 12:36 UTC
                             check-postgres-availability  sorintlab/stolon                                                           
[...]

PR #2565, opened 2019-10-30
Improve experience with .flux.yaml files

squaremo:
- include more information like filenames in errors
 - correct the implementation of resourceID so it uses the name of the
   resource

These two address a comment https://github.com/fluxcd/flux/issues/2428#issuecomment-540031938 that fluxd should at least provide information about which file/resource is failing, when syncing goes wrong.

 - return an error if asked to update a manifest, but there are no (relevant) update commands in the config file

This addresses #2324 by making it more obvious what has failed (in that issue, the complaint can be interpreted as "nothing happened and I don't know why").

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/pkg/term v0.0.0-20190109203006-aa71e9d9e942
 	github.com/prometheus/client_golang v1.1.0
+	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
 	github.com/ryanuber/go-glob v1.0.0
 	github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749
 	github.com/shurcooL/vfsgen v0.0.0-20181202132449-6a9ea43bcacd
--- a/pkg/manifests/configaware_test.go
+++ b/pkg/manifests/configaware_test.go
@@ -235,3 +235,44 @@ spec:
 	assert.NoError(t, err)
 	assert.Equal(t, expectedPatch, string(patch))
 }
+
+const mistakenConf = `
+version: 1
+commandUpdated: # <-- because this is commandUpdated, patchFile is ignored
+  generators:
+    - command: |
+       echo "apiVersion: extensions/v1beta1
+       kind: Deployment
+       metadata:
+         name: helloworld
+       spec:
+         template:
+           metadata:
+             labels:
+              name: helloworld
+           spec:
+             containers:
+             - name: greeter
+               image: quay.io/weaveworks/helloworld:master-a000001
+       ---
+       apiVersion: v1
+       kind: Namespace
+       metadata:
+         name: demo"
+  patchFile: patchfile.yaml
+`
+
+// This tests that when using a config with no update commands, and
+// update operation results in an error, rather than a silent failure
+// to make any changes.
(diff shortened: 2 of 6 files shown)


squaremo:
@2opremio I've moved this out of draft, and done more work to surface mistakes with config files; PTAnotherL when you have a sec. :apple:

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/pkg/term v0.0.0-20190109203006-aa71e9d9e942
 	github.com/prometheus/client_golang v1.1.0
+	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
```
2opremio: Where did this come from? (just wondering)

pkg/manifests/configfile.go
```diff
@@ -21,42 +23,60 @@ const (
...
-	Path       string
-	WorkingDir string
-	Version    int
+	Version int
 	// Only one of the following should be set simultaneously
 	CommandUpdated *CommandUpdated `yaml:"commandUpdated"`
 	PatchUpdated   *PatchUpdated   `yaml:"patchUpdated"`
+
+	path       string // the absolute path to the .flux.yaml
+	workingDir string // the absolute path to the dir in which to run commands or find a patch file
```
2opremio: Thanks! I should had thought about that!

pkg/manifests/configfile.go
```diff
@@ -66,20 +86,119 @@ func NewConfigFile(path, workingDir string) (*ConfigFile, error) {
...
+				updaters[len(result)-1].Policy.Command,
+				result[len(result)-1].Error,
+				r.Source(),
+				result[len(result)-1].Output,
+			)
+			return false, err
+		}
+	}
+	// We assume that the update changed the resource. Alternatively, we could generate the resources
+	// again and compare the output, but that's expensive.
```
2opremio: Uhmm, we could simply use Git for this, I think. I wonder why I didn't think about it before (I know, not part of the PR)

pkg/manifests/configaware_test.go
```diff
@@ -235,3 +235,44 @@ spec:
...
+           spec:
+             containers:
+             - name: greeter
+               image: quay.io/weaveworks/helloworld:master-a000001
+       ---
+       apiVersion: v1
+       kind: Namespace
+       metadata:
+         name: demo"
+  patchFile: patchfile.yaml
```
2opremio: Ideally, I think we should just reject this file and parse strictly (although returning an error when no updates are done is better and should have is certainly an improvement).

2opremio:
Very minor stuff, otherwise LGTM! 💯 (+ an extra 🥇 for the dispatching refactoring!)

go.mod
squaremo: I don't know, it keeps popping up whenever I run `make`; so whichever version of go I have, it thinks it's a dependency.

pkg/manifests/configfile.go
squaremo: Right; I think this was probably ported over from the more circumspect implementation.

pkg/manifests/configaware_test.go
squaremo: Yeah; validating against a schema was my next stop. While the particular error here would change, there would probably still be ways to have a config that doesn't quite work.

PR #2569, opened 2019-10-31
e2e: Add basic sync test

2opremio:
As part of it:

1. Implicitly set the environment when importing lib/env.bash
   (needed so that other library files can use the environment variables)
2. Allow accessing the git server (securely) through port-forwarding as
   part of install_git_srv
3. Merge installing gitsrv and creating the git ssh secret (so that we
   can reuse the underlying ssh keys for (2))

### DIFF
--- a/test/e2e/12_sync.bats
+++ b/test/e2e/12_sync.bats
@@ -36,7 +36,7 @@ function setup() {
   [ "$sync_tag_hash" = "$head_hash" ]
 
   # Add a change, wait for it to happen and check the sync tag again
-  sed -i '' 's%stefanprodan/podinfo:2.1.0%stefanprodan/podinfo:3.1.5%' "${clone_dir}/workloads/podinfo-dep.yaml"
+  sed -i'.bak' 's%stefanprodan/podinfo:2.1.0%stefanprodan/podinfo:3.1.5%' "${clone_dir}/workloads/podinfo-dep.yaml"
   git commit -am "Bump podinfo"
   head_hash=$(git rev-list -n 1 HEAD)
   git push
--- a/test/e2e/12_sync.bats
+++ b/test/e2e/12_sync.bats
@@ -1,0 +1,55 @@
+#!/usr/bin/env bats
+
+load lib/env
+load lib/install
+load lib/poll
+load lib/defer
+
+git_ssh_cmd=""
+git_port_forward_pid=""
+
+function setup() {
+  kubectl create namespace "$FLUX_NAMESPACE"
+  # Install flux and the git server, allowing external access
+  install_git_srv flux-git-deploy git_srv_result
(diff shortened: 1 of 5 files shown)


test/e2e/lib/install.bash
```diff
@@ -60,43 +63,49 @@ function install_flux_with_fluxctl() {
...
+    local_port=$(echo "$local_port" | sed 's%.*:\([0-9]*\).*%\1%')
+    local ssh_cmd="ssh -o UserKnownHostsFile=/dev/null  -o StrictHostKeyChecking=no -i $gen_dir/id_rsa -p $local_port"
+    # return the ssh command needed for git, and the PID of the port-forwarding PID into a variable of choice
+    eval "${external_access_result_var}=('$ssh_cmd' '$COPROC_PID')"
+  fi
 }
 
 function uninstall_git_srv() {
+  local secret_name=${1:-flux-git-deploy}
+  kubectl delete -n "${FLUX_NAMESPACE}" secret "$secret_name"
```
2opremio: This causes an error if we invoke `uninstall_git_srv()` after `uninstall_flux()` because the latter also removes the secret. So, I am tempted to remove it.

test/e2e/12_sync.bats
```diff
@@ -0,0 +1,55 @@
...
+  sync_tag_hash=$(git rev-list -n 1 flux)
+  [ "$sync_tag_hash" = "$head_hash" ]
+}
+
+function teardown() {
+  uninstall_flux_with_fluxctl
+  kill "$git_port_forward_pid"
+  uninstall_git_srv
+  kubectl delete namespace "$DEMO_NAMESPACE"
+  kubectl delete namespace "$FLUX_NAMESPACE"
```
hiddeco: I would remove `uninstall_flux_with_fluxctl` as `kubectl delete namespace` already takes care of removing Flux, and it will probably save some test run time.

test/e2e/lib/install.bash
hiddeco: Alternative would be to 
```suggestion
  kubectl delete -n "${FLUX_NAMESPACE}" secret "$secret_name" || true
```

2opremio: I was referring to the printed error (which makes debugging the output log confusing (placing a red herring).

It doesn't really matter that the command has a non-0 exit since its invoked in the cleanup teardown function.

test/e2e/12_sync.bats
2opremio: That's a fair point. I can also remove `uninstall_git_srv()`.

test/e2e/lib/install.bash
hiddeco: In that case, see updated suggestion.

PR #2570, opened 2019-10-31
fix: add --batch to gpg --import to fix secret key import in non-interactive shells

ekristen:
Recently a change was made in gpg 2.2.x IIRC that causes some weird issues with respect to importing secret keys in non-interactive shells.

The fix is to simply add `--batch` to the command. That is all this PR does.

### DIFF
--- a/pkg/gpg/gpg.go
+++ b/pkg/gpg/gpg.go
@@ -64,7 +64,7 @@ func ImportKeys(src string, trustImportedKeys bool) ([]string, error) {
 }
 
 func gpgImport(path string) error {
-	cmd := exec.Command("gpg", "--import", path)
+	cmd := exec.Command("gpg", "--import", "--batch", path)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("error importing key: %s", string(out))


hiddeco:
@ekristen do you have some background information about the `--batch`? I have been writing e2e tests for our GPG functionalities last week and was unable to detect any issues with our current approach.

This means that it either works as expected, or my end-to-end tests are insufficient. With both can be confirmed with some background information.

ekristen:
Unfortunately this is only something I've learned from experience.  Changing the flux container to have interactive & tty fixes the git-secret unlock portion of the code, but the key will never get imported properly unless `--batch` is used.

I just know that after trying to get gpg working in several CI's in several docker containers certain commands will simply not work without it. 

```--batch Use batch mode. Never ask, do not allow interactive commands.```

If I exec into the container and import the gpg key using the batch flag flux then starts to work properly so long as I have also redeployed the container ahead of time with interactive & tty.

Perhaps the end-to-end tests aren't being run in the same fashion the container runs on kubernetes clusters?

hiddeco:
@ekristen the end-to-end tests run in a [`kind`](https://github.com/kubernetes-sigs/kind) cluster, which is a stripped down Kubernetes. The only difference I can think of is a single vs multiple keys.

ekristen:
Ok, I was able to dig into this more. I'm sad to say this was user error :( -- sorta. The GPG key refused to give up its passphrase and had the illusion of working properly. Now that I have a GPG in place without a passphrase all is working properly. Sorry to have wasted your time.

hiddeco:
These PRs are (as a maintainer) honestly the best, as it means we did not have a bug for weeks, glad you got it resolved in your end. :beer:

PR #2575, opened 2019-11-04
Delete old integration tests

2opremio:


hiddeco:
@2opremio you need to add a `.gitkeep` in the `test/bin` directory, the tests are now failing because the target dir doesn't exist.

```
cp: cannot create regular file 'test/bin/helm': No such file or directory

Makefile:104: recipe for target 'test/bin/helm' failed

make: *** [test/bin/helm] Error 1
```

2opremio:
Ah, I will add an `mkdir -p` in the installation recipes

PR #2580, opened 2019-11-05
Sops support

PaulFarver:
This is a pull request that adds support for sops OOTB as described in #1676 

The point of this feature is to let flux decrypt files encrypted with [sops](https://github.com/mozilla/sops). Sops is ideal for encryption of yaml files, and has support for AWS KMS, GCP KMS, Azure Key Vault and PGP. The idea is to get flux to decrypt yaml manifests that have been encrypted with sops.

- The go.mozilla.org/sops library has been added as a dependency
- load.go has been updated. The Load func now takes a new boolean argument (sopsEnabled), used to specify whether to decrypt yaml files with sops before applying them to the cluster.
- The manifests struct has been updated with sopsEnabled, and the NewManifests function also takes this as an argument.
- A new option has been added to the cli:
  - `--sops` is a boolean option for specifying whether sops decryption should be enabled


carlpett:
This is nice! :tada: Did you have a cause to exec to the binary instead of linking sops as a library? Is something missing in `go.mozilla.org/sops/decrypt`?

PaulFarver:
> This is nice! 🎉 Did you have a cause to exec to the binary instead of linking sops as a library? Is something missing in `go.mozilla.org/sops/decrypt`?

@carlpett I completely missed that. Of course that makes more sense. Will try it.

2opremio:
@PaulFarver thanks a lot for the effort! 🌹 

I need to get familiar with `sops` and discuss it with the other maintainers to see what's the way forward. Please stay put.

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -170,3 +178,7 @@ func ParseMultidoc(multidoc []byte, source string) (map[string]KubeManifest, err
 	}
 	return objs, nil
 }
+
+func shouldDecrypt(path string) bool {
+	return strings.HasSuffix(path, ".enc.yaml") || strings.HasSuffix(path, ".enc.yml")
```
2opremio: Shouldn't this be using the suffix provided to the `--sops-suffix` flag?

PaulFarver:
The docker build seems to be failing when it has to run the known_hosts.sh. Seems gitlab is having some trouble

pkg/cluster/kubernetes/resource/load.go
PaulFarver: Yes. It has been corrected in 8d45ecda8c8b617cd6276f4f6c977541a66b06f7

carlpett:
Btw, it might make sense to wait for https://github.com/mozilla/sops/pull/566, to get proper versioned modules (and the new import path)

2opremio:
@PaulFarver it seems that the e2e error was transient (or due to a flaky test, I need to investigate).

Good news is it started working again after rebuilding.

PaulFarver:
@2opremio Thanks!

@carlpett I think it's a good idea to wait until they get that merged. Is there anything I could work on for this, in the mean time. I'm guessing there might be some documentation and tests to be updated.

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -39,7 +41,16 @@ func Load(base string, paths []string) (map[string]KubeManifest, error) {
 			}
 
 			if !info.IsDir() && filepath.Ext(path) == ".yaml" || filepath.Ext(path) == ".yml" {
-				bytes, err := ioutil.ReadFile(path)
+				var bytes []byte
+				decryptFile, err := shouldDecrypt(path, sopsPattern)
```
ajvb: Not familiar with this codebase so I may be misguided in mentioning this, but the way we usually do this at Mozilla is something like:
```golang
// Try to decrypt the conf using sops or load it as plaintext.
// If the configuration is not encrypted with sops, the error
// sops.MetadataNotFound will be returned, in which case we
// ignore it and continue loading the conf.
confData, err = decrypt.Data(data, "yaml")
if err != nil {
    if err == sops.MetadataNotFound {
        // not an encrypted file
        confData = data
    } else {
        return errors.Wrap(err, "failed to load sops encrypted configuration")
    }
}
```

examples: https://github.com/mozilla-services/foxsec-pipeline-contrib/blob/master/common/configuration.go#L71-L79 and https://github.com/mozilla-services/autograph/blob/master/main.go#L255-L268

2opremio: This has the advantage of not requiring a special filename, which is great!

However, we should confirm that the metadata check is really cheap, otherwise we cannot afford to do it with every file.

2opremio:
@PaulFarver I really like the shape this is taking! Some comments:

1. Let's definitely wait for https://github.com/mozilla/sops/pull/566 (which I guess will be merged today)
2. Let's stop requiring a special filename for `sops`-encrypted files and use the [trial-and-error approach](https://github.com/fluxcd/flux/pull/2580#discussion_r344954835) suggested by @ajvb 
3. Let's add a flag for enabling/disabling sops decryption: `--sops`. Let's make it opt-in unless the content-check is really cheap.

PaulFarver:
@2opremio Sounds good 👍 Working on it now

2opremio:
4. Can we add a warning like the following? https://github.com/fluxcd/flux/pull/2159/files#diff-350e484490d6241fa81c13942d160480R296 (tweaking the text, of course, since the user may be using something else, like KMS).

2opremio:
and ...
5. Can you add tests? A unit tests and, ideally, an end-to-end test as well

2opremio:
and ..
6. Can you add the new flag to `site/daemon.md` ?

elzapp:
How will this play along with Kustomize and other manifest generators?

2opremio:
> How will this play along with Kustomize and other manifest generators?

Good question! Since the detection is content-based, it will work transparently.

PaulFarver:
> How will this play along with Kustomize and other manifest generators?

As @2opremio has already stated, it should work transparently. I think you might run into some trouble, though. 

Sops verifies a file with a checksum in the metadata. The checksum depends on when the file was modified, the values, and key structure. Even the key position. Also that of unencrypted values.

As I understand it, manifest generation is run before the yaml files are loaded in flux, which is when the sops decryption takes place. That means you would have to generate the valid sops encrypted file with kustomize build. It might be possible, but I think it would be pretty tedious, and hard to maintain.

2opremio:
> That means you would have to generate the valid sops encrypted file with kustomize build. It might be possible, but I think it would be pretty tedious, and hard to maintain.

As a side note, `kustomize` is just one of the possibilities to generate manifests. Flux's approach to manifest generation ([`.flux.yaml` files](https://docs.fluxcd.io/en/stable/references/fluxyaml-config-files.html)) is generic and admits any command (as long as it's present in the Flux image, int he future we will support supplying images to run commands through [Ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/)).

elzapp:
> Sops verifies a file with a checksum in the metadata. The checksum depends on when the file was modified, the values, and key structure. Even the key position. Also that of unencrypted values.

Exactly this was my concern

PaulFarver:
@elzapp
> Exactly this was my concern. It might be feasable to have the secrets not included with kustomize, and just cating them into the generated manifest, if sops only is concerned with the individual documents inside the yaml.

Sops does have support for yaml multidocs, but you have to encrypt all documents in the file, so I don't think it aligns too well with the kustomize use case.

The poc of this feature worked in much the same way as the git-secret feature, where files are decrypted right after a checkout. This would support encryption of patches individually, which seems like a better workflow for people using manifest generation.

@2opremio We could pursue decrypt-first functionality, but with the trial-and-error [approach](https://github.com/fluxcd/flux/pull/2580#discussion_r344954835) suggested by @ajvb. What do you think?

elzapp:
Decrypt first would be the best, but it would require to resolve all the bases and resources from the kustomize files first, or decrypt any yaml the whole repository, not just those inside the git-path. (The latter wouldn't work if the kustomize files refers to resources outside the repository, but maybe that is an ok limitation.)

PaulFarver:
Right... I have another branch, that implements decrypt-first with the trial-and-error approach and the go.mozilla.org/sops library. If that is the way we want to move forward, I will switch the pull request to that branch.

2opremio:
> > How will this play along with Kustomize and other manifest generators?
> 
> Good question! Since the detection is content-based, it will work transparently (i.e. you should be able to generate sops-encrypted content)

After thinking about this further it turns out I wasn't completely right (sorry I didn't think it through right away).

On one hand, at the current state of this PR, sops decryption is attempted after reading the files from the filesystem and not when parsing them (which is used both when loading files from disk and by manifest generation).

There is an easy fix though. @PaulFarver can you move the `softSopsDecrypt()` call into `ParseMultidoc()` instead?

Depending on how `sops` encrypts yaml document streams, we may want to attempt `softSopsDecrypt()` on the full stream and/or on each individual document, after extracting it from the stream. The goal is making sure that manifest generation can output a hybrid YAML stream, with encrypted and unencrypted documents.

This will allow you to do manifest generation with sop-encrypted manifests, included in the generation output, like follows:

`.flux.yaml`
```
version: 1
patchUpdated:
  generators:
    - command: kustomize build .
    - command: cat *.enc.yaml
  patchFile: flux-patch.yaml
```

Where files with suffix `enc.yaml` are encrypted with sops. Alternatively you could use a `--git-path` not using manifest generation (i.e. without an associated `.flux.yaml`) , directly containing the sops-encrypted files.

Note that the `enc.yaml` files cannot be passed as an input [resource](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/fields.md#resources) to Kustomize. This has the disadvantage of not being able to process sops-encrypted with Kustomize (e.g. add namepace prefix to a decrypted secret). 

Now I realize this might be what @elzapp was referring to with _play along with Kustomize and other manifest generators_.

A decrypt-first (in-place) approach would indeed solve this problem but I think it's overkill since, as @elzapp mentioned:

> Decrypt first would be the best, but it would require to resolve all the bases and resources from the kustomize files first, or decrypt any yaml the whole repository, not just those inside the git-path.

Flux doesn't understand `kustomization.yaml` files (as I mentioned, Flux's approach to manifest generation is technology agnostic) which leaves walking every yaml file in the repo for decryption in every sync.

This wouldn't support external bases including sops-encrypted manifests and it would result in git repositories incompatible with standard Kustomize (which doesn't support sops manifests as input).

I think that implementing decrypt-first to solve this problem is just a workaround to compensate the lack of  interoperability between `sops` and Kustomize, which isn't Flux's  job to solve.

As an alternative, and in addition to the current PR, we could include the sops binary in the Flux container, allowing to explicitly decrypt the files before passing them to Kustomize, as follows

`.flux.yaml`
```
version: 1
patchUpdated:
  generators:
    - command: for I in *.enc.yaml; do sops -d "$I" -o "${I%*enc.yaml}yaml"; done
    - command: kustomize build .
  patchFile: flux-patch.yaml
```

`kustomization.yaml`
```
resources:
- foo.yaml
- bar.yaml
.....
```

File layout
```
├── .flux.yaml
├── kustomization.yaml
├── foo.enc.yaml
└── bar.enc.yaml
```

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -127,11 +130,18 @@ func looksLikeChart(dir string) bool {
...
 // constructs an object set from the resources represented therein.
-func ParseMultidoc(multidoc []byte, source string) (map[string]KubeManifest, error) {
+func ParseMultidoc(multidoc []byte, source string, sopsEnabled bool) (map[string]KubeManifest, error) {
 	objs := map[string]KubeManifest{}
-	decoder := yaml.NewDecoder(bytes.NewReader(multidoc))
 	var obj KubeManifest
 	var err error
+	cleartext := multidoc
+	if sopsEnabled {
+		cleartext, err = softSopsDecrypt(multidoc)
```
2opremio: Shouldn't we also call `softSopsDecrypt()` after https://github.com/fluxcd/flux/pull/2580/files#diff-c3f2eeebdc2b683408b96d1ed03621c5R156 in case we are decoding a document stream which isn't which contains individually encrypted documents?

In fact, how does sops encryption of document streams work? If it's one to one (i.e. when encrypting a stream of n documents leads to another stream of n encrypted documents) , then we should only call it there, since the stream itself is parseable in a normal fashion.

PaulFarver: According to the sops [documentation](https://github.com/mozilla/sops#33yaml-streams), the hash isn't computed for each individual document. It encrypts the entire file, and saves the hash of that. So tampering with one document would cause all of them to fail the validation step in the decryption.

In order to decrypt each individual document, we would have to unmarshal the yaml stream, and then marshal every document in order to try the softSopsDecrypt. This approach would work if a bunch of individually encrypted yaml files were joined in a stream and passed to the function, but in this case yaml documents that were encrypted in a yaml stream would not pass the hash validation step.

One interesting thing to note is that sops actually includes the same metadata in all the encrypted documents in a stream. So it would technically be possible to match the documents, but it wouldn't be pretty.

2opremio: I have been playing with `sops` and it indeed fails to decrypt isolated documents from an encrypted yaml stream, with a `MAC mismatch` error.

Also, `sops` doesn't transparently decrypt streams composed by encrypted and plain yaml manifests.

> So it would technically be possible to match the documents, but it wouldn't be pretty.

Yes, I think that's trying to be too smart.

With these restrictions, I can't think of a good way to make sops work well with manifest generation (at least with the current manifest generation scheme; in which generators print yaml streams to `stdout`).

I see two options:

1. Explicitly avoid automatic decryption for generated files. We decrypt sops files automatically for raw manifests (i.e. only in the Loadbut we also include the sops binary, to be used in `.flux.yaml` files.
2. Use the in-place decrypt-first scheme. However, I still think that trying to decrypt every file is overkill. Having a specific file suffix (like the PR originally proposed) may solve the problem.

I think we should go for (1), having raw sops manifest files being decrypted already adds a lot of value. Manifest generation can use the binary.

I would like to know @squaremo 's opinion on this one though

2opremio: BTW, @PaulFarver thanks a lot for enduring so many comments and design changes 🌷

PaulFarver: @2opremio No worries. I think it is awesome to see this level of interest 🥇

squaremo: I think it would be perfectly fine to support:

 1. SOPS decryption of plain YAMLs;
 2. calling `sops` as a command before running `kustomize build .`

I don't see a strong use case for decrypting the output of arbitrary commands. The conclusion I reach from the arguments in the thread above is that it's just too fiddly to organise for a correctly encrypted set of bytes, and you are much better off just doing 1) or 2).

I think this indicates that sops decryption should be done when reading a file in `cluster/kubernetes/resource.Load` (since that loads from files, and SOPS deals with files), but not attempted in ParseManifests.

If it can be written as a wrapper around `Manifests`, that might be a little neater than passing a boolean down through a lot of procedures.

2opremio: OK, it seems @squaremo and I agree (which doesn't always happen, so it's a good sign :) )

@PaulFarver can you please:
1) Move sops decryption back to `Load()`
2) Document (in the `--sops` flag help) that automatic `sops` decryption is not applied when using `.flux.yaml` files , and that the `sops` binary can be used for that.
3) Include the `sops` binary in the Flux image
4) Add test(s)

docker/Dockerfile.flux
```diff
@@ -37,6 +37,11 @@ ENTRYPOINT [ "/sbin/tini", "--", "fluxd" ]
 COPY --from=quay.io/squaremo/kubeyaml:0.7.0 /usr/lib/kubeyaml /usr/lib/kubeyaml/
 ENV PATH=/bin:/usr/bin:/usr/local/bin:/usr/lib/kubeyaml
 
+# Add the sops binary and allow it to be executed
+ARG SOPS_VERSION=3.4.0
+ADD https://github.com/mozilla/sops/releases/download/$SOPS_VERSION/sops-$SOPS_VERSION.linux /usr/local/bin/sops
```
2opremio: Can you install sops verifying its checksum and using the build cache system like, e.g. `kubectl`? 
https://github.com/fluxcd/flux/blob/518933bf8fc906e90e67c5a78439aa6039c1b6be/Makefile#L12
https://github.com/fluxcd/flux/blob/518933bf8fc906e90e67c5a78439aa6039c1b6be/Makefile#L94-L95
https://github.com/fluxcd/flux/blob/518933bf8fc906e90e67c5a78439aa6039c1b6be/Makefile#L109-L114

PaulFarver: It should be done correctly now, but please check the work. I can't say I fully understand this makefile, so I might have missed something.

2opremio: It is done correctly. Thanks!!

go.mod
```diff
@@ -35,6 +36,7 @@ require (
 	github.com/weaveworks/common v0.0.0-20190410110702-87611edc252e
 	github.com/weaveworks/go-checkpoint v0.0.0-20170503165305-ebbb8b0518ab
 	github.com/whilp/git-urls v0.0.0-20160530060445-31bac0d230fa
+	go.mozilla.org/sops v0.0.0-20190912205235-14a22d7a7060
```
2opremio: I think it would be clearer if we depended on an explicit version of `sops`

```suggestion
	go.mozilla.org/sops v3.4.0
```

(`14a22d7a7060` and `v3.4.0` point to the same commit)

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -8,14 +8,17 @@ import (
 	"os"
 	"path/filepath"
 
+	"go.mozilla.org/sops"
+	"go.mozilla.org/sops/decrypt"
+
```
2opremio:
```suggestion
```

pkg/cluster/kubernetes/resource/load.go
squaremo: > 3\. Include the `sops` binary in the Flux image

Why this?

cmd/fluxd/main.go
```diff
@@ -120,6 +120,7 @@ func main() {
 		gitSetAuthor = fs.Bool("git-set-author", false, "if set, the author of git commits will reflect the user who initiated the commit and will differ from the git committer.")
 		gitLabel     = fs.String("git-label", "", "label to keep track of sync progress; overrides both --git-sync-tag and --git-notes-ref")
 		gitSecret    = fs.Bool("git-secret", false, `if set, git-secret will be run on every git checkout. A gpg key must be imported using  --git-gpg-key-import or by mounting a keyring containing it directly`)
+		sopsEnabled  = fs.Bool("sops", false, `if set, decrypt sops encrypted manifest files with sops before applying them. Be aware that manifests generated with .flux.yaml are not decrypted, and should be handled with the included sops binary. Provide keys for sops in the same way as providing them for the binary`)
```
2opremio:
```suggestion
		sopsEnabled  = fs.Bool("sops", false, `if set, decrypt sops-encrypted manifest files with sops before applying them. Be aware that manifests generated with .flux.yaml are not automatically decrypted. Provide keys for sops in the same way as providing them for the binary`)
```

docs/references/daemon.md
```diff
@@ -93,6 +93,7 @@ Version controlling of cluster manifests provides reproducibility and a historic
 | --ssh-keygen-type                                |                                    | -t argument to ssh-keygen (default unspecified)
 | **manifest generation**
 | --manifest-generation                            | false                              | search for .flux.yaml files to generate manifests
+| --sops                                           | false                              | decrypt sops encrypted manifest files with sops before applying them. Be aware that manifests generated with .flux.yaml are not decrypted. This is due to the way sops handles yaml streams. It is recommended that generating manifests from encrypted files is handles with the included sops binary. Provide keys for sops in the same way as providing them for the binary, for example with --git-gpg-key-import. A full description of how to supply sops with a key can be found in the sops [Documentation](https://github.com/mozilla/sops#usage)
```
2opremio:
```suggestion
| --sops                                           | false                              | decrypt sops encrypted manifest files with sops before applying them. Be aware that manifests generated with .flux.yaml are not decrypted. This is due to the way sops handles yaml streams. It is recommended that generating manifests from encrypted files is handles with the included sops binary. Provide keys for sops in the same way as providing them for the binary, for example with --git-gpg-key-import. A full description of how to supply sops with a key can be found in the [sops documentation](https://github.com/mozilla/sops#usage)
```

squaremo:
Housekeeping to be aware of: when you are about to mark this "ready for review", please `git rebase -i master` or otherwise reduce this to a smaller number of commits, and elide the merges from master.

2opremio:
Apart from the (minor) comments I just added, the only things left in this PR are:
1. Adding tests (ideally a unit test and an e2e test)
2. As indicated by @squaremo at https://github.com/fluxcd/flux/pull/2580#discussion_r346240351 , use a `Manifests` variant instead of a boolean to dispatch whether sops-decryption should be attempted.
3. Squashing the commits

pkg/cluster/kubernetes/resource/load.go
2opremio: Because otherwise you cannot invoke `sops` from the `.flux.yaml` files without building your own image (or running an init container).

squaremo: Oh OK. I am a bit reluctant to keep adding binaries, but it is suggested by "support [...] calling `sops` as a command before running `kustomize build .`", so fair cop :-S

go.mod
PaulFarver: Yes, but that fails with, due to the semantic versioning requirement of go modules. This will be fixed in the future when https://github.com/mozilla/sops/pull/566 is included in a release.

pkg/cluster/kubernetes/resource/load_test.go
```diff
@@ -263,13 +264,57 @@ func debyte(r resource.Resource) resource.Resource {
...
+`
+	strat := func(arg []byte) ([]byte, error){
+		return nil, sops.MetadataNotFound
+	}
+	decrypted, err := softDecrypt([]byte(enc), strat)
+	if err != nil {
+		t.Error(err)
+	}
+	if string(decrypted) != enc {
+		t.Errorf("Expected: \n%#v\ngot:\n%#v", dec, decrypted)
```
2opremio: This error doesn't make sense (since `dec` wasn't returned). In fact, I would remove `enc` end only use `dec` (renaming it to `plaintextManifest`) or something in those lines.

pkg/cluster/kubernetes/resource/load_test.go
```diff
@@ -263,13 +264,57 @@ func debyte(r resource.Resource) resource.Resource {
 	return r
 }
 
+func TestSoftDecryption(t *testing.T){
```
2opremio: I don't think this test adds a lot of value. AFAICT it only tests the internal strategy mechanism.

PaulFarver:
Got some time to work on this again
- I've added some unit tests. It required me to change the design a little bit, in order to mock the decryption function, so a successful decryption could be tested.
- I bumped the version of sops to v3.5.0
- An e2e test would be really good for this feature, but I'm unsure what the best way to do it is. One approach is to generate a pgp key, encrypt a file and commit it to the gitsrv and run fluxctl sync. Another is to have the encrypted file and the key in https://github.com/fluxcd/flux-get-started, and then import the key before syncing. What do you think?

pkg/cluster/kubernetes/resource/load_test.go
```diff
@@ -319,9 +364,24 @@ func TestChartTracker(t *testing.T) {
...
 		fq := filepath.Join(dir, f)
-		if m, err := Load(dir, []string{fq}); err != nil || len(m) != 0 {
+		if m, err := Load(dir, []string{fq}, false); err != nil || len(m) != 0 {
 			t.Errorf("%q not ignored as a chart should be", f)
 		}
 	}
 
 }
+
+func TestLoadSomeWithSopsDecryption(t *testing.T) {
```
2opremio: This test adds value (since it makes sure that loading plain manifests works when sops is enabled). However, I am missing a test (by extending this one or adding an additional one) in which an encrypted manifests is decoded successfully.

The easiest would be using GPG. You may want to borrow ideas for scaffolding from the GPG package https://github.com/fluxcd/flux/tree/6a0e353061f68ba902161af3b39b4e6557c23646/pkg/gpg . Particularly https://github.com/fluxcd/flux/blob/6a0e353061f68ba902161af3b39b4e6557c23646/pkg/gpg/gpgtest/gpg.go#L18

pkg/cluster/kubernetes/resource/load_test.go
PaulFarver: Yeah. The idea was to avoid testing sops functionality in a unit test, since it should be covered by the sops test suite, but I get what you mean.

pkg/cluster/kubernetes/resource/load_test.go
PaulFarver: Thanks. This was my missing piece :)

I was trying to figure out how test an actual decryption

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -8,14 +8,29 @@ import (
...
 	"path/filepath"
 
+	sops "go.mozilla.org/sops/v3"
+	"go.mozilla.org/sops/v3/decrypt"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 )
 
+// decryptionStrategy is a type of function that reads data and return the decrypted data
+type decryptionStrategy func([]byte) ([]byte, error)
```
2opremio: Would you mind adding that this is needed for testing? (mocking the strategy)

2opremio:
> Got some time to work on this again

Great! Thanks for the continued effort!

> * I've added some unit tests. It required me to change the design a little bit, in order to mock the decryption function, so a successful decryption could be tested.

Makes sense! They are looking good! I added some comments

> * An e2e test would be really good for this feature, but I'm unsure what the best way to do it is. One approach is to generate a pgp key, encrypt a file and commit it to the gitsrv and run fluxctl sync. Another is to have the encrypted file and the key in https://github.com/fluxcd/flux-get-started, and then import the key before syncing. What do you think?

Yes, that's what I would do. The e2e commit [signing](https://github.com/fluxcd/flux/blob/6095242e5027c8f8ca933853d79c103b91a4eb0c/test/e2e/20_commit_signing.bats)/[verification](https://github.com/fluxcd/flux/blob/6095242e5027c8f8ca933853d79c103b91a4eb0c/test/e2e/20_commit_verification.bats) tests already does something similar, so I think you should be able to reuse their scaffolding.

pkg/cluster/kubernetes/resource/load_test.go
```diff
@@ -319,9 +321,44 @@ func TestChartTracker(t *testing.T) {
...
 		fq := filepath.Join(dir, f)
-		if m, err := Load(dir, []string{fq}); err != nil || len(m) != 0 {
+		if m, err := Load(dir, []string{fq}, false); err != nil || len(m) != 0 {
 			t.Errorf("%q not ignored as a chart should be", f)
 		}
 	}
 
 }
+
+func TestLoadSomeNoneEncrypted(t *testing.T) {
```
2opremio:
```suggestion
func TestLoadSomeWithSopsNoneEncrypted(t *testing.T) {
```

pkg/cluster/kubernetes/resource/load_test.go
```diff
@@ -319,9 +321,44 @@ func TestChartTracker(t *testing.T) {
...
+	objs, err := Load(dir, []string{dir}, true)
+	if err != nil {
+		t.Error(err)
+	}
+	if len(objs) != len(testfiles.ResourceMap) {
+		t.Errorf("expected %d objects from %d files, got result:\n%#v", len(testfiles.ResourceMap), len(testfiles.Files), objs)
+	}
+}
+
+func TestLoadSomeEncrypted(t *testing.T) {
```
2opremio:
```suggestion
func TestLoadSomeWithSopsAllEncrypted(t *testing.T) {
```

2opremio:
@PaulFarver are you done or are you still working on an end-to-end test?

pkg/cluster/kubernetes/resource/load.go
2opremio: This is not needed anymore, right?

PaulFarver: Correct. There is no need to mock up the decryption strategy any more

PaulFarver:
> @PaulFarver are you done or are you still working on an end-to-end test?

Since there is a case with full sops decryption in the unit test, it feels a bit overkill to write code that encrypts, commits, checks out and decrypts a file in an e2e test.

I want to do a bit of cleaning up, but otherwise I'm done.

2opremio:
> I want to do a bit of cleaning up, but otherwise I'm done.

Faaantastic. Please don't forget to rebase and squash your commits (it doesn't need to be just one, but not more than a handful of them)

tomjohnburton:
@PaulFarver GO PAUL. This is epic

PaulFarver:
I've cleaned up a bit. I've rebased and squashed, so it should be a little more neatly applied to master, than the other sixty-odd commits😄 

I got to thinking... It might be a little neater to split `NewManifests(namespacer, logger, sopsEnabled)` into `NewManifests(namespacer, logger)` and `NewSopsManifests(namespacer, logger, sopsEnabled)` in order to avoid polluting all other NewManifests calls. I've implemented that version, but I still have the other implementation on another branch, so I can switch to that, if it's preferable.

pkg/cluster/kubernetes/manifests.go
```diff
@@ -37,13 +37,24 @@ type manifests struct {
...
 func NewManifests(ns namespacer, logger log.Logger) *manifests {
 	return &manifests{
 		namespacer:       ns,
 		logger:           logger,
 		resourceWarnings: map[string]struct{}{},
+		sopsEnabled:      false,
+	}
+}
+
+func NewSopsManifests(ns namespacer, logger log.Logger, sopsEnabled bool) *manifests {
```
2opremio: Much nicer! But why still pass a Boolean? Shouldn't it always be enabled?

pkg/cluster/kubernetes/testfiles/data.go
```diff
@@ -28,7 +28,17 @@ func TempDir(t *testing.T) (string, func()) {
 
 // WriteTestFiles ... given a directory, create files in it, based on predetermined file content
 func WriteTestFiles(dir string) error {
-	for name, content := range Files {
+  return writeFiles(dir, Files)
+}
+
+// WriteEncryptedTestFiles ... given a directory, create files in it, based on predetermined file content.
+// These files are encrypted with sops using TestPrivateKey
+func WriteEncryptedTestFiles(dir string) error {
```
2opremio: Nit
```suggestion
func WriteSopsEncryptedTestFiles(dir string) error {
```

pkg/cluster/kubernetes/testfiles/data.go
```diff
@@ -347,3 +357,154 @@ spec:
         - containerPort: 8080
 `,
 }
+
+var EncryptedFiles = map[string]string{
```
2opremio: Nit
```suggestion
var SopsEncryptedFiles = map[string]string{
```

2opremio:
Really nice! This is ready to go apart from 3 minor comments.

cmd/fluxd/main.go
```diff
@@ -508,7 +513,11 @@ func main() {
 			logger.Log("err", err)
 			os.Exit(1)
 		}
-		k8sManifests = kubernetes.NewManifests(namespacer, logger)
+		if *sopsEnabled {
+			k8sManifests = kubernetes.NewManifests(namespacer, logger)
+		} else {
+			k8sManifests = kubernetes.NewSopsManifests(namespacer, logger)
```
2opremio: Shouldn't this be the other way around? (Sorry I missed this)

PaulFarver: Whoops, my bad. You didn't miss it. I just passed `*sopsEnabled` to the `NewSopsManifests` call before.

Good catch

PaulFarver: Fixed now

2opremio:
Nice, I think this is ready to go. Re-running the tests since they seem to be failing for unrelated reasons.

PaulFarver:
Cool! Thanks for the patience @2opremio 🥇 It's been a pleasure

2opremio:
@PaulFarver would you mind rebasing on top of master?

PaulFarver:
@2opremio Yep, done!

ryedin:
@PaulFarver sorry to activate an old thread, but I'm trying to figure out how this works when my file was encrypted with AWS KMS. Would I need to first mount (somehow) a volume into the flux container that contains the `~/.aws/credentials` for the IAM user that has permissions for the key? 

It looks like the code here is less about generically "enabling SOPS" and more specifically "enabling sops with GPG". Is that accurate?

2opremio:
KMS should work, you just need to configure the environment if the Flux container as if you were to invoke the sops binary (which is also included in the Flux image, so it's a good way to test that it works).

ryedin:
@2opremio that's essentially what I was asking how to do :) (`you just need to configure the environment of the Flux container...`)

sops uses the go aws sdk, which expects the credentials file to live at `~/.aws/credentials`. So in order to get KMS to work in flux we have a chicken and an egg problem of sorts. Or at least we have this one set of secrets that has to be pre-baked, as it were, into the environment...

I'm not sure the best way to solve that problem. Unless I'm really misunderstanding something here

tomjohnburton:
Don’t bother with credentials file. You can just export the variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY. Set them up as environment variables for your flux container

ryedin:
@tomjohnburton ... `face` / `palm` heh, thank you for that. 

that said, I use helm to install flux. I see `extraEnvs` is available here https://github.com/fluxcd/flux/blob/master/chart/flux/README.md#configuration

... is it considered secure to pass those secrets in like that? that's what I'm alluding to when I say chicken and egg. I'm just not sure if pushing those in directly as env vars is actually a smart/safe thing to be doing.

stefanprodan:
Instead of storing the AWS key in a Kubernetes secret or env vars you should consider using IAM roles for service accounts. With eksctl you could associate an IAM role to Flux service account so that Flux pod can access the KMS API see https://eksctl.io/usage/iamserviceaccounts/

ryedin:
ok, perfect! That's very helpful for when I get to the EKS stage. Currently proving everything out locally first. I think helm-secrets has what I need for this layer for now. First am going to push straight into env vars using a temp/throwaway IAM user just to prove out the e2e flow, then will work on improving with helm-secrets/volume mounts, then "harden" once in EKS and switching to the service account/role approach. 

Thank you for the help guys!

patrick-armitage:
Would anyone have a running example of eksctl-enabled-flux using sops + KMS key to decrypt files?

ryedin:
yes, this has been working very well. we use a simple script we named `setup-flux.sh`. Obviously change the env vars to be what you want to use, but we use a specific AWS IAM user for this. Ostensibly you could use roles instead, but we didn't get that far yet.

I've kept the stuff in there where we're allowing the script to control setting up multiple instances of flux that each watch a different branch of the repo and a specific path. You may or may not find that part useful, but it's how we are dealing with running multiple flux-enabled CI/CD projects in the same cluster. Obviously change paths and whatnot to match your needs.

The important bits are in the helm command `set` options near the bottom.
The sops related requirements are the `SOPS_KMS_ARN` env var and the access key/secret pair. This assumes you have setup a symmetric KMS key (or keys) and have the ARN(s) to plug in.


```
#!/bin/bash
set -e

if [[ ! -x "$(command -v kubectl)" ]]; then
    echo "kubectl not found"
    exit 1
fi

if [[ ! -x "$(command -v helm)" ]]; then
    echo "helm not found"
    exit 1
fi

if [[ ! -x "$(command -v fluxctl)" ]]; then
    echo "fluxctl not found"
    exit 1
fi

# requires the ENV calling this script to have flux aws user credentials
ACCESS_KEY_ID=$FLUX_AWS_ACCESS_KEY_ID
SECRET_ACCESS_KEY=$FLUX_AWS_SECRET_ACCESS_KEY

# default flux release name
FLUX_NAMESPACE=fluxcd

# default branch and path to sandbox
GITHUB_BRANCH=sandbox
GIT_PATH=kubernetes/flux/environments/sandbox/releases/sandbox

# allow overriding branch
if [[ $CLUSTER_ENV = "sandbox" ]]; then
  FLUX_NAMESPACE=flux-sandbox
  GITHUB_BRANCH=sandbox
  GIT_PATH=kubernetes/flux/environments/sandbox/releases/sandbox
  echo "using sandbox branch"
fi

if [[ $CLUSTER_ENV = "qa" ]]; then
  FLUX_NAMESPACE=flux-qa
  GITHUB_BRANCH=qa
  GIT_PATH=kubernetes/flux/environments/qa/releases/qa
  echo "using qa branch"
fi

if [[ $CLUSTER_ENV = "production" ]]; then
  FLUX_NAMESPACE=flux-production
  GITHUB_BRANCH=master
  GIT_PATH=kubernetes/flux/environments/production/releases/production
  echo "using master branch"
fi

# ensure namespace exists (each env uses it's own namespace)
kubectl create namespace $FLUX_NAMESPACE || echo "$FLUX_NAMESPACE namespace already exists"

# ensure helm repo added
helm repo add fluxcd https://charts.fluxcd.io

echo "Setting up flux controller"
helm upgrade -i $FLUX_NAMESPACE fluxcd/flux --wait \
  --namespace $FLUX_NAMESPACE \
  --set git.pollInterval=1m \
  --set git.ciSkip=true \
  --set git.url=git@github.com:<YOUR ORG>/<YOUR REPO> \
  --set git.path=$GIT_PATH \
  --set git.branch=$GITHUB_BRANCH \
  --set sops.enabled=true \
  --set extraEnvs[0].name=SOPS_KMS_ARN,extraEnvs[0].value=$SOPS_KMS_ARN \
  --set extraEnvs[1].name=AWS_ACCESS_KEY_ID,extraEnvs[1].value=$ACCESS_KEY_ID \
  --set extraEnvs[2].name=AWS_SECRET_ACCESS_KEY,extraEnvs[2].value=$SECRET_ACCESS_KEY

# output the ssh public key, which needs to be applied as a deploy key to the repo
echo "Flux has been installed. ssh public key:"
fluxctl identity --k8s-fwd-ns $FLUX_NAMESPACE

echo "Setting up helm-operator"
helm upgrade -i $FLUX_NAMESPACE-helm-operator fluxcd/helm-operator --wait \
  --namespace $FLUX_NAMESPACE \
  --set git.ssh.secretName=$FLUX_NAMESPACE-git-deploy \
  --set git.timeout=300s \
  --set helm.versions=v3 \
  && echo "helm-operator installed"

```

Once setup, this works great. Any SOPS encrypted manifests flux finds in the configured path in the repo, as long as they were encrypted with that same KMS ARN, will be automatically decrypted.

Make sure to remember for kubernetes `Secrets` manifests, the values need to be base64 encoded before you SOPS encrypt the doc. that bit me in the butt once or twice, but that's just basic k8s stuff.

ryedin:
And to be clear, there is nothing special about whether your cluster was setup using `eksctl` or not. Just make sure your current context is using the cluster you want to set flux up in. (highly recommend using `kubectx` if you're not already)

PR #2581, opened 2019-11-05
e2e: Add sync failure test

2opremio:
The test updates and duplicates a resource, making sure syncing catches the problem and no update is performed in the cluster.

### DIFF
--- a/test/e2e/12_sync.bats
+++ b/test/e2e/12_sync.bats
@@ -57,6 +57,7 @@ function setup() {
   sync_tag_hash=$(git rev-list -n 1 flux)
   head_hash=$(git rev-list -n 1 HEAD)
   [ "$sync_tag_hash" = "$head_hash" ]
+  podinfo_image=$(kubectl get pod -n demo -l app=podinfo -o"jsonpath={['items'][0]['spec']['containers'][0]['image']}")
 
   # Bump the image of podinfo, duplicate the resource definition (to cause a sync failure)
   # and make sure the sync doesn't go through
@@ -68,9 +69,9 @@ function setup() {
   # Wait until we find the duplicate failure in the logs
   poll_until_true "duplicate resource in Flux logs" "kubectl logs -n $FLUX_NAMESPACE -l name=flux | grep -q \"duplicate definition of 'demo:deployment/podinfo'\""
   # Make sure that the version of podinfo wasn't bumped
-  local podinfo_image
-  podinfo_image=$(kubectl get pod -n demo -l app=podinfo -o"jsonpath={['items'][0]['spec']['containers'][0]['image']}")
-  [ "$podinfo_image" = "stefanprodan/podinfo:2.1.0" ]
+  local podinfo_image_now
+  podinfo_image_now=$(kubectl get pod -n demo -l app=podinfo -o"jsonpath={['items'][0]['spec']['containers'][0]['image']}")
+  [ "$podinfo_image" = "$podinfo_image_now" ]
   # Make sure that the Flux sync tag remains untouched
   git pull -f --tags
   sync_tag_hash=$(git rev-list -n 1 flux)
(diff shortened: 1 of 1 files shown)


test/e2e/12_sync.bats
```diff
@@ -45,7 +47,38 @@ function setup() {
...
+  # Check the sync tag
+  git pull -f --tags
+  local sync_tag_hash
+  sync_tag_hash=$(git rev-list -n 1 flux)
+  head_hash=$(git rev-list -n 1 HEAD)
+  [ "$sync_tag_hash" = "$head_hash" ]
+
+  # Bump the image of podinfo, duplicate the resource definition (to cause a sync failure)
+  # and make sure the sync doesn't go through
+  sed -i'.bak' 's%stefanprodan/podinfo:2.1.0%stefanprodan/podinfo:3.1.5%' "${clone_dir}/workloads/podinfo-dep.yaml"
```
hiddeco:
```suggestion
  sed -i'.bak' 's%stefanprodan/podinfo:.*%stefanprodan/podinfo:3.1.5%' "${clone_dir}/workloads/podinfo-dep.yaml"
```

Will make it immune to upstream version changes (except if upstream changes it the tag to `3.1.5`).

hiddeco: (When changing this you probably also want to change the replace in the basic sync test).

test/e2e/12_sync.bats
```diff
@@ -45,7 +47,38 @@ function setup() {
...
+  cp "${clone_dir}/workloads/podinfo-dep.yaml" "${clone_dir}/workloads/podinfo-dep-2.yaml"
+  git add "${clone_dir}/workloads/podinfo-dep-2.yaml"
+  git -c 'user.email=foo@bar.com' -c 'user.name=Foo' commit -am "Bump podinfo and duplicate it to cause an error"
+  git push
+  # Wait until we find the duplicate failure in the logs
+  poll_until_true "duplicate resource in Flux logs" "kubectl logs -n $FLUX_NAMESPACE -l name=flux | grep -q \"duplicate definition of 'demo:deployment/podinfo'\""
+  # Make sure that the version of podinfo wasn't bumped
+  local podinfo_image
+  podinfo_image=$(kubectl get pod -n demo -l app=podinfo -o"jsonpath={['items'][0]['spec']['containers'][0]['image']}")
+  [ "$podinfo_image" = "stefanprodan/podinfo:2.1.0" ]
```
hiddeco: Same here, probably better to get the image after the first successful sync, and compare it to the version you now have in memory here.

2opremio:
@hiddeco PTAL

test/e2e/12_sync.bats
```diff
@@ -45,7 +47,39 @@ function setup() {
...
+  # Wait until we find the duplicate failure in the logs
+  poll_until_true "duplicate resource in Flux logs" "kubectl logs -n $FLUX_NAMESPACE -l name=flux | grep -q \"duplicate definition of 'demo:deployment/podinfo'\""
+  # Make sure that the version of podinfo wasn't bumped
+  local podinfo_image_now
+  podinfo_image_now=$(kubectl get pod -n demo -l app=podinfo -o"jsonpath={['items'][0]['spec']['containers'][0]['image']}")
+  [ "$podinfo_image" = "$podinfo_image_now" ]
+  # Make sure that the Flux sync tag remains untouched
+  git pull -f --tags
+  sync_tag_hash=$(git rev-list -n 1 flux)
+  [ "$sync_tag_hash" = "$head_hash" ]
```
hiddeco: I am used to having the changing or updated variable on the right side in comparisons. I also see a mixed usage of `local` and non-`local` variables, without a clear pattern in when they are (not) used.

2opremio: You are totally right. Fixed.

PR #2583, opened 2019-11-05
Added manifest generation flag

sebikul:
As with PR https://github.com/fluxcd/flux/pull/2530, this one adds the `manifest-generation` flag to the installation command.


2opremio:
Thanks! 🌷

PR #2587, opened 2019-11-06
e2e: Misc fixes and refactorings

2opremio:


test/e2e/20_commit_signing.bats
```diff
@@ -6,31 +6,32 @@ load lib/gpg
...
-  # shellcheck disable=SC2154
-  git_ssh_cmd="${git_srv_result[0]}"
-  export GIT_SSH_COMMAND="$git_ssh_cmd"
+  install_git_srv git_srv_result
   # shellcheck disable=SC2154
-  git_port_forward_pid="${git_srv_result[1]}"
+  export GIT_SSH_COMMAND="${git_srv_result[0]}"
   # Teardown the created port-forward to gitsrv.
-  defer kill "$git_port_forward_pid"
+  defer kill "${git_srv_result[1]}"
```
hiddeco: I _think_ this should be `defer "kill '${git_srv_result[1]}'"`?

2opremio: It's a number, so you don't really need to single-quote it. You can pass multiple arguments to defer since it expands them.

hiddeco: https://github.com/fluxcd/flux/pull/2587/files#diff-806c4694241259b6acd056c83308cb4fR40, should be like this then :-).

test/e2e/lib/install.bash
```diff
@@ -51,103 +53,47 @@ function uninstall_flux_with_helm() {
...
   # Everything goes into one directory, but not everything is
   # necessarily used by the kustomization
   cp -R "${E2E_DIR}"/fixtures/kustom/* "${kustomtmp}/"
-  local kustomization
-  kustomization="base"
-  if [ -n "$1" ]; then
-    # use the kustomization given instead; ../base will still be
-    # there to be used as a base
-    kustomization="$1"
+  if [ -n "$2" ]; then
```
2opremio:
```suggestion
  if [ -n "$key_values_varname" ]; then
```

test/e2e/20_commit_signing.bats
2opremio: Yes, I missed that

test/e2e/lib/template.bash
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+function fill_in_place_recursively() {
+  local -n key_values=$1 # pass an associate array as a nameref
+  local target_directory=${2:-.}
+  (# use a subshell to expose key-values as variables for envsubst to use
+    for key in "${!key_values[@]}"; do
+      export "$key"="${key_values[$key]}"
+    done
+    while IFS= read -r -d '' file; do
```
hiddeco: I think this (L10-17) could use a descriptive comment to not have to decipher the excellent bash wizardry you pull off here if you are just reading the code to get a general understanding.

PR #2588, opened 2019-11-07
add --namespace argument to override default namespace

### LINKED ISSUE #1260: Let the default namespace for resources be configured
If you have yamels that apply to more than one environment, where the namespace differs, you might want to have config that omits the namespace and let it default.

You can mount a `~/.kube/config` file into the container to change the behaviour of `kubectl apply` so that it defaults to a particular namespace; but fluxd is in general ignorant of this setting, and will for example report manifests that don't have a namespace as belonging to `"default"`.

moshloop:
Fixes #1260


2opremio:
Thanks a lot for the contribution 🏅  . I like this!

Is it urgent? (it may take a few days for me to get to review it)

2opremio:
My first few thoughts:
* I would rename `--namespace` to `--default-sync-namespace` (other suggestions are welcome, I just think that `namespace` is too unspecific, @hiddeco thoughts?)
* I am missing a test
* I need to think about the interaction of this with `.kube/config` and we should document which one takes precedence (I think it should be the flag which overrides whatever is in `.kube/config`)

moshloop:
> I would rename --namespace to --default-sync-namespace (other suggestions are welcome, I just think that namespace is too unspecific

The reason I used namespace as it mirrors the behaviour of `kubectl --namespace`

>  missing a test

Sure - will add one

pkg/cluster/kubernetes/namespacer.go
```diff
@@ -51,7 +47,7 @@ var getDefaultNamespace = func() (string, error) {
 		return c.Namespace, nil
 	}
 
-	return defaultFallbackNamespace, nil
+	return defaultNamespace, nil
```
2opremio: I think that the passed namespace should take precedence over the kubeconfig file mounted. So, I would:

1. Rename this function to `getKubeconfigDefaultNamespace()` and leave `defaultFallbackNamespace` as it was
2. Rename the `defaultNamespace` parameter of `NewNamespacer` to `defaultNamespaceOverride` and set it to the `fallback` without calling `getKubeconfigDefaultNamespace()` if `!= ""`

2opremio:
> The reason I used namespace as it mirrors the behaviour of `kubectl --namespace`

That makes sense, but we do other stuff with namespaces (like whitelisting them) and it can get confusing (e.g. the user may think that this will restrict flux to only deploy to the supplied namespace).

I would go for `--default-namespace-override`

2opremio:
Actually, for consistency, I would go for `--k8s-default-namespace` (the `override` suffix makes it a bit too long)

2opremio:
Also, can you add the new flag to `site/daemon.md` ?

moshloop:
@2opremio done

2opremio:
@moshloop I am confused, did you intentionally remove the commit and closed the PR?

PR #2590, opened 2019-11-07
remove replace directives

carnott-snap:
go.mod contains many replace entries that all consumers must include to
import logic. These are cancerous and tend to make consumers builds
difficult or outright fail. All have been removed, except for the one
required by helm-operator. This entry, while retained, has been
converted into a noop and should be removed when
fluxcd/helm-operator#97 is fixed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,12 +2,15 @@ module github.com/fluxcd/flux
 
 go 1.13
 
+// remove when https://github.com/docker/distribution/pull/2905 is released.
+replace github.com/docker/distribution => github.com/2opremio/distribution v0.0.0-20190419185413-6c9727e5e5de
+
 require (
 	github.com/Jeffail/gabs v1.4.0
 	github.com/Masterminds/semver v1.4.2
 	github.com/aws/aws-sdk-go v1.19.11
 	github.com/bradfitz/gomemcache v0.0.0-20190329173943-551aad21a668
-	github.com/docker/distribution v0.0.0-00010101000000-000000000000
+	github.com/docker/distribution v2.7.1+incompatible
 	github.com/evanphx/json-patch v4.1.0+incompatible
 	github.com/fluxcd/helm-operator v1.0.0-rc2
 	github.com/ghodss/yaml v1.0.0
@@ -35,26 +38,10 @@ require (
 	golang.org/x/sys v0.0.0-20190801041406-cbf593c0f2f3
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 	gopkg.in/yaml.v2 v2.2.2
-	k8s.io/api v0.0.0-20190313235455-40a48860b5ab
-	k8s.io/apiextensions-apiserver v0.0.0-20190315093550-53c4693659ed
-	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
+	k8s.io/api v0.0.0-20190708174958-539a33f6e817 // kubernetes-1.14.4
+	k8s.io/apiextensions-apiserver v0.0.0-20190315093550-53c4693659ed // kubernetes-1.14.4
+	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d // kubernetes-1.14.4
 	k8s.io/client-go v11.0.0+incompatible
 	k8s.io/helm v2.13.1+incompatible
 	k8s.io/klog v0.3.3
-)
-
-replace github.com/docker/distribution => github.com/2opremio/distribution v0.0.0-20190419185413-6c9727e5e5de
-
-// The following pin these libs to `kubernetes-1.14.4` (by initially
-// giving the version as that tag, and letting go mod fill in its idea of
-// the version).
-// The libs are thereby kept compatible with client-go v11, which is
-// itself compatible with Kubernetes 1.14.
-
-replace (
-	k8s.io/api => k8s.io/api v0.0.0-20190708174958-539a33f6e817
-	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
-	k8s.io/apiserver => k8s.io/apiserver v0.0.0-20190708180123-608cd7da68f7
-	k8s.io/client-go => k8s.io/client-go v11.0.0+incompatible
-	k8s.io/component-base => k8s.io/component-base v0.0.0-20190708175518-244289f83105
 )
--- a/go.sum
+++ b/go.sum
@@ -408,6 +408,7 @@ gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 honnef.co/go/tools v0.0.0-20180728063816-88497007e858/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 honnef.co/go/tools v0.0.0-20190106161140-3f1c8253044a/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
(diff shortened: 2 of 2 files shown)


2opremio:
Hi. Thanks for the contribution!

 However, I am afraid that those replacements are needed. What makes you think they are not?

It may be that flux even compiles without them but that doesn't mean they are not needed.

> These are cancerous and tend to make consumers builds difficult or outright fail

I would generally agree with that, but, they are needed. Also, could you provided an existing scenario in which the replacements are an issue?

carnott-snap:
> However, I am afraid that those replacements are needed. What makes you think they are not?
>
> It may be that flux even compiles without them but that doesn't mean they are not needed.

The repo builds and tests successfully without them, and I have a basic consumer package that compiles successfully. Replace lines do not affect dependent modules (per the spec), so I am unclear why they are necessary, can you explain this requirement?

> I would generally agree with that, but, they are needed. Also, could you provided an existing scenario in which the replacements are an issue?

I have interest in using transitive dependencies of flux, and do not want to maintain a list of eight replace directives for modules that I do not use (directly). There is no automated way to keep them in sync, and upgrading the flux version will cause breakage when these replace lines change.

2opremio:
The docker distribution one is needed to honor timeouts when accessing repositories and the Kubernetes ones to ensure Flux backwards compatible with Kubernetes. They are needed because different dependencies import different versions of the Kubernetes libraries, and unfortunately Golang doesn't support using multiple dependency versions.

> I have interest in using transitive dependencies of flux, and do not want to maintain a list of eight replace directives for modules that I do not use (directly). There is no automated way to keep them in sync, and upgrading the flux version will cause breakage when these replace lines change.

Sorry, but I don't see how that would cause things to break. As I see it, the impact of changing the replacements would not be worse than bumping the dependency versions without an explicit replacement rule.

Regardless, the replacements are there for a reason and we cannot simply remove them. We could consider bumping the replacement versions, but that's about it.

2opremio:
@carnott-snap to understand your usecase, can you elaborate on your usecase? Why are you importing flux?

carnott-snap:
> The docker distribution one is needed to honor timeouts when accessing repositories ...

I am sympathetic to this concern, however that change lives in a stale fork, and is over six months out of date. Is there work to upstream this fix? If so, we can leave the replace, but should link against the PR and warn about impending removal. If not, the logic should be internalised, so that consumers do not see the side effects and the maintenance burden is clear to the maintainers of this library.

> [T]he Kubernetes ones to ensure Flux backwards compatible with Kubernetes. They are needed because different dependencies import different versions of the Kubernetes libraries, and unfortunately Golang doesn't support using multiple dependency versions.

I do not see how `replace` solves this problem. As I understand things, if you require a given version of a library, you get that version. I am unclear how `replace` gives you any more backwards compatibility than `require`? If I have chosen too high a version for the tag, say you want `kubernetes-12.0.3`, I can fix this up.

> Sorry, but I don't see how that would cause things to break. As I see it, the impact of changing the replacements would not be worse than bumping the dependency versions without an explicit replacement rule.

I should be clear, while there the mentioned breakage relates to upgrades and will happen with `replace` or `require`. This is more about complexity and maintenance burden. If I have a `go.mod` file that works, it will continue to work. If I have two dependencies that require incompatible versions of kubernetes libraries, that will be a problem either way. My concern is that every dependency that is transitively linked to `flux` will need to manually keep in sync a `replace` block.

> @carnott-snap to understand your usecase, can you elaborate on your usecase? Why are you importing flux?

We are attempting to integrate with the `eksctl` repo, which depends on `flux`: https://github.com/weaveworks/eksctl/blob/master/go.mod#L25. Are you indicating that `flux` is not maintained as a library? If so, can we deprecate the external api surface, so that consumers will not be confused?

2opremio:
> I am sympathetic to this concern, however that change lives in a stale fork, and is over six months out of date. Is there work to upstream this fix?

https://github.com/docker/distribution/pull/2905 , which has been ignored by the maintainers so far.

> I do not see how `replace` solves this problem. As I understand things, if you require a given version of a library, you get that version

Not really, in my experience Go will override the version you require by the version required by your dependencies.

> My concern is that every dependency that is transitively linked to `flux` will need to manually keep in sync a `replace` block.

I don't think that's the case. You don't need to add a replace rule if you don't care about the specific version of your dependencies.

> Are you indicating that `flux` is not maintained as a library? If so, can we deprecate the external api surface, so that consumers will not be confused?

I wasn't trying to imply that. In fact, Flux intentionally exports the `install` package.

2opremio:
> If so, we can leave the replace, but should link against the PR and warn about impending removal

We should document it yeah.

2opremio:
> > I do not see how `replace` solves this problem. As I understand things, if you require a given version of a library, you get that version
> 
> Not really, in my experience Go will override the version you require by the version required by your dependencies.

Let's put it in a different way. If you manage to remove the replace directives without changing `go.sum` as a result, I will be more than happy to merge your PR. I doubt that's possible though.

carnott-snap:
> Let's put it in a different way. If you manage to remove the replace directives without changing `go.sum` as a result, I will be more than happy to merge your PR.

I believe 23325e13 meets these standards. There is one additional line in the `go.mod`, however it is only used during dependency verification and does not affect build or actual resolved versions:
```
k8s.io/api v0.0.0-20190313235455-40a48860b5ab/go.mod h1:iuAfoD4hCxJ8Onx9kaTIt30j7jUFS00AXQi6QMi99vA=
```
My initial change also included safe dependency upgrades, using `@latest`, this has been rolled back and I will submit these changes separately.

> docker/distribution#2905 , which has been ignored by the maintainers so far.

With the lack of traction on the PR, and the importance of this feature, it suggests we should make a real fork. I dislike imposing the `replace`, and think that a fork will allow you to maintain the desired logic in the long term.

This will involve publishing a valid module to `github.com/2opremio/distribution` (or `github.com/fluxcd/distrobution`, since this is the org that needs the logic) and migrate all the references from `github.com/docker/distrobution`.

> You are right, we should add a comment in `go.mod`.

Since I am unsure if you are willing to take the fork route, I have amened `go.mod` to include this comment and am happy to revise the wording to your liking.

> Not really, in my experience Go will override the version you require by the version required by your dependencies.

This is not true for the current state of the repo: a `replace` is required for `github.com/docker/distribution` because you use an non-existent pseudo-version, `v0.0.0-00010101000000-000000000000`. To correct this, and make `replace` opt in, I have replaced the pseudo-version a valid one: `v2.7.1+incompatible`. Unfortunately because of the dependency cycle between `flux` and `helm-operator`, this change alone is not enough. We will need to [break the history](https://github.com/fluxcd/helm-operator/pull/97#discussion_r343966184) before we can remove this requirement. It seems like this will benefit the `helm-operator` repo as well, so we can orchestrate that out of band.

go.mod
```diff
@@ -2,12 +2,15 @@ module github.com/fluxcd/flux
 
 go 1.13
 
+// TODO(carnott-snap): remove when docker/distribution#2905 is released.
```
2opremio:
```suggestion
// remove when https://github.com/docker/distribution/pull/2905 is released.
```

2opremio:
This is starting to look good, thanks.

> There is one additional line in the `go.mod`, however it is only used during dependency verification and does not affect build or actual resolved versions:

How do you know that?

carnott-snap:
> How do you know that?

There are two types of hashes in `go.sum`: module and `go.mod`. The former is the actual version that is used for compilation, thuse there is only one per module import path, where as the latter are required for every referenced version of a module in the dependency graph, thus there are many:
```
k8s.io/api v0.0.0-20190313235455-40a48860b5ab/go.mod h1:iuAfoD4hCxJ8Onx9kaTIt30j7jUFS00AXQi6QMi99vA=
k8s.io/api v0.0.0-20190708174958-539a33f6e817 h1:V6YPTc5fSnwv7EBjx6es9VyAki/6bqK4M3ECA6WwfBk=
k8s.io/api v0.0.0-20190708174958-539a33f6e817/go.mod h1:iuAfoD4hCxJ8Onx9kaTIt30j7jUFS00AXQi6QMi99vA=
```
As such, we can read these three lines as saying: you will find two module versions, `v0.0.0-20190313235455-40a48860b5ab` and `v0.0.0-20190708174958-539a33f6e817`, then use the more recent, `v0.0.0-20190708174958-539a33f6e817`, to compile with.

This can be confirmed by running `go list -m k8s.io/api`:
```
k8s.io/api v0.0.0-20190708174958-539a33f6e817
```

2opremio:
@carnott-snap makes sense, would you mind rebasing on master before I merge?

PR #2577, opened 2019-11-04
End-to-end test for garbage collection

squaremo:
This adds a test for syncing with garbage collection switched on. To make it easier to add arbitrary arguments to the flux deployment, it also adds a mechanism for turning the output of `fluxctl install` into a kustomization and patching it.

PR #2596, opened 2019-11-08
e2e: Test the local Flux version

2opremio:
https://github.com/fluxcd/flux/pull/2577 introduced a regression, ( at https://github.com/fluxcd/flux/pull/2577/files#diff-e76b06a8940cf68dbf66b91380d3dc28L58 )
causing the end-to-end tests to use the latest Flux release instead of the local version.


2opremio:
@alexmt please make sure to test using this change. Otherwise you won't be testing your local code.

PR #2599, opened 2019-11-11
Better error messages for generated manifests

### LINKED ISSUE #2595: Improve error messages involving resources from `.flux.yaml` files
In #2571 we see an error message which isn't very helpful:

```
loading resources from repo: duplicate resource from configuration/xxx/.flux.yaml and configuration/xxx/.flux.yaml.
```

The message should also include:
1. The ID of the resource (Namespace,Kind,Name)
2. The `--git-path` involved when loading the `.flux.yaml` file.

squaremo:
This PR improves the error messages for generated manifests, mainly by including the root path -- i.e., the --git-path value, or implicitly `.` -- as well as the path to the config file `.flux.yaml`, given relative to that.

For example, in the case of duplicate resources, here's the error message logged:

```
ts=2019-11-11T16:32:51.814302722Z caller=loop.go:101 component=sync-loop err="loading resources from repo: duplicate definition of resource <cluster>:namespace/demo generated by production/../.flux.yaml and stag
ing/../.flux.yaml"
```

NB this error message also names the resource in question, therefore fixes #2595.


pkg/manifests/configaware.go
```diff
@@ -208,18 +210,14 @@ func (ca *configAware) getResourcesByID(ctx context.Context) (map[string]resourc
 		if err != nil {
 			return nil, err
 		}
-		relConfigFilePath, err := cf.RelativeConfigPath()
-		if err != nil {
-			return nil, err
-		}
-		resources, err := ca.manifests.ParseManifest(resourceManifests, relConfigFilePath)
+		resources, err := ca.manifests.ParseManifest(resourceManifests, cf.Source())
```
2opremio: Passing the config file's source context instead of simply the `.flux.yaml` filepath will break [this check](https://github.com/fluxcd/flux/blob/859d7112c619a3f4c96fbc8d3e3bab29366ad687/pkg/daemon/sync.go#L224)

2opremio: I guess we can make the `Source()` method of resources a bit more sophisticated, separating context and filepath.

Or, we could add an extra `IsGenerated()` method for resources.

And/Or, we should probably reconsider always adding the generated resources to the commit log (which is the only place where I believe `Source()` is expected to follow certain structure)

squaremo: So it does. I have touched that code but I'd forgotten that it relied on `Source` for something other than presentation. (Possibly that is the case elsewhere too)

That is a shame; I either have to figure out how to reform that bit of code (in the smaller sense of not relying on the Source there, or in the larger sense of fixing how changed files are calculated); or, hack another thing into resources so that error messages have access to the extra bit of info.

squaremo: Of the alternatives (thanks for thinking those through), I prefer keeping Source for presentation only; which means either omitting generated resources from the event log, or calculating it some other way.

squaremo: Perhaps for the purpose of making error messages better I could find Another Way that is neutral with respect to `.Source`.

2opremio: Sounds good!

squaremo: Opened https://github.com/fluxcd/flux/issues/2601 for the bigger problem with that sync.go code.

squaremo: I've gone back to using the config file path as the source for generated manifests, so it won't affect the bit of code you spotted @2opremio .

pkg/manifests/configfile.go
```diff
@@ -33,8 +33,13 @@ type ConfigFile struct {
 	CommandUpdated *CommandUpdated `yaml:"commandUpdated"`
 	PatchUpdated   *PatchUpdated   `yaml:"patchUpdated"`
 
-	path       string // the absolute path to the .flux.yaml
-	workingDir string // the absolute path to the dir in which to run commands or find a patch file
+	// These are supplied, and can't be calculated from each other
+	originalPath string // the path relative to the repo root, for identifying generated manifests
```
2opremio: I would rename this to be more descriptive. `originalPath` by itself doesn't say much, it took me a while to figure. How about ....

```
filepathRelativeToRepoRoot
filepathAbsolute
workingDirAbsolute
```

In fact, it may be less confusing (in terms of the parameters passed to the constructor) to use the following (all absolute) paths:

```
repoRootDirPath
configPath
workingDir
```

pkg/manifests/configfile.go
```diff
@@ -97,24 +112,28 @@ func NewConfigFile(path, workingDir string) (*ConfigFile, error) {
...
 // identify resources generated from the config file.
-func (cf *ConfigFile) RelativeConfigPath() (string, error) {
-	return filepath.Rel(cf.workingDir, cf.path)
+func (cf *ConfigFile) RelativeConfigPath() string {
+	return cf.relativeConfigPath
+}
+
+// ConfigVsOriginalPath shows the config file as it relates to the
+// repo-relative path, e.g., `staging/../.flux.yaml`
+func (cf *ConfigFile) ConfigVsOriginalPath() string {
```
2opremio: I would call this `ConfigFileSearchPath()` or something in those lines, because it's the path used to find the configfile from the base path.

pkg/manifests/configfile.go
squaremo: > In fact, I think we can leave it as it was
> 
> ```
> path
> workingDir
> ```

Prior to this branch:
 - `workingDir`, as supplied, is the absolute path to the directory.
 - `path` is the absolute path to the config file.

What is lacking there is the path as given by `--git-path` (or implied by its absence), which can't be calculated from those two.

> Isn't what you currently call originalPath the relative path from the workingDir to path?

No, `originalPath` is the path from the root of the git repo to the directory under consideration; i.e., it's `filepath.Rel(root, workingDir)`.

In general you need three bits of information, to be able to report the things described in #2595 as well as process the config file: the absolute base for all the paths, so they can be found in the filesystem; the path under consideration or "working directory"; the path to the config file to use.

But, finally:

> I would rename this to something more descriptive.

Sure, I agree `originalPath` is obscure; I can rearrange or rename them so it's clearer what's going on.

pkg/manifests/configfile.go
squaremo: I'll have another think about it. It doesn't need to be a path, I just though that was a compact representation of the two pieces of information (the path under consideration, and where the config file is in relation to it).

2opremio:
It's looking much better now, thanks! Could you add a test?

pkg/manifests/configaware.go
```diff
@@ -208,20 +210,32 @@ func (ca *configAware) getResourcesByID(ctx context.Context) (map[string]resourc
...
-				return nil, fmt.Errorf("duplicate resource from %s and %s",
-					r.Source(), resourcesByID[id].resource.Source())
+
+		for id, generated := range resources {
+			if duplicate, ok := resourcesByID[id]; ok {
+				var duplicateErr error
+				switch {
+				case duplicate.configFile == cf:
+					// the duplicate is _also_ generated by this config file
+					duplicateErr = fmt.Errorf("duplicate definitions of resource %s generated by %s",
```
2opremio: Can you make the error messages consistent with https://github.com/fluxcd/flux/blob/6a0e353061f68ba902161af3b39b4e6557c23646/pkg/cluster/kubernetes/resource/load.go#L56 ?

squaremo: Yup

squaremo:
> Could you add a test?

Yus. And in fact, in doing so I discovered that duplicates among generated manifests are not detected, because cluster/kubernetes/resource/load.go doesn't attempt to detect them. What do you reckon, should I --

 - add duplicate detection in there
 - leave that case alone (or similarly, leave the later detection in as a backstop)

2opremio:
> because cluster/kubernetes/resource/load.go doesn't attempt to detect them

Can you elaborate? That file does detect duplicates at https://github.com/fluxcd/flux/blob/6a0e353061f68ba902161af3b39b4e6557c23646/pkg/cluster/kubernetes/resource/load.go#L56

squaremo:
> > because cluster/kubernetes/resource/load.go doesn't attempt to detect them
> 
> Can you elaborate? That file does detect duplicates at

In `Load` it detects when a manifest loaded from a file duplicates a manifest loaded from another file.
But in `ParseMultidoc`, which is used by `Load` as well as config generation, it doesn't attempt to detect duplicates _within_ a chunk of bytes. So duplicate definitions within a file or a generated chunk of bytes aren't detected.

2opremio:
> In `Load` it detects when a manifest loaded from a file duplicates a manifest loaded from another file.
> But in `ParseMultidoc`, which is used by `Load` as well as config generation, it doesn't attempt to detect duplicates _within_ a chunk of bytes. So duplicate definitions within a file or a generated chunk of bytes aren't detected.

Ah, now I understand, thanks. I was assuming that adding the resources one by one after parsing was enough. But it isn't, since `ParseMultidoc()` already returns a map, silently and implicitly coalescing resources with identical ID.

Thanks! This ended up being more work that one would expect.

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -158,10 +158,18 @@ func ParseMultidoc(multidoc []byte, source string) (map[string]KubeManifest, err
 		// contained resources we are after.
 		if list, ok := obj.(*List); ok {
 			for _, item := range list.Items {
-				objs[item.ResourceID().String()] = item
+				id := item.ResourceID().String()
+				if _, ok := objs[id]; ok {
+					return nil, fmt.Errorf(`duplicate definition of '%s' (in %s)`, id, source)
+				}
```
2opremio: Can we somehow factor out the duplicate-check and message? maybe in a local function?

squaremo: It occurred to me too, but I didn't bother because it would increase the line count, and the sites are close together. Have I drunk too much Go-aid? Possibly.

Issue #2605, opened 2019-11-12
unable to fetch node CPU metrics


texasbobs:
We are using Flux 1.13.2 and in some clusters, it no longer clones repos to be synced.  The error seems to indicate it is unable to execute a query on the Prometheus api.  
```
ts=2019-11-12T17:00:47.74181943Z caller=images.go:17 component=sync-loop msg="polling images"
ts=2019-11-12T17:00:54.041654513Z caller=images.go:27 component=sync-loop msg="no automated workloads"
ts=2019-11-12T17:04:33.083254717Z caller=loop.go:111 component=sync-loop event=refreshed url=git@github.myrepo.git branch=master HEAD=3e9
ts=2019-11-12T17:06:05.738649349Z caller=loop.go:85 component=sync-loop err="collating resources in cluster for sync: Error while fetching node metrics for selector : unable to fetch node CPU metrics: unable to execute query: Get http://prometheus-k8s.monitoring.svc:9090/api/v1/query?query=sum%281+-+rate%28node_cpu_seconds_total%7Bmode%3D%22idle%22%7D%5B1m%5D%29+%2A+on%28namespace%2C+pod%29+group_left%28node%29+node_namespace_pod%3Akube_pod_info%3A%7Bnode%3D~%22url%7C url%7C url%22%7D%29+by+%28node%29&time=1573578364.541: dial tcp 10.233.9.37:9090: 
connect: connection refused"
```

How can we prevent this from stopping the sync?

Why does Flux care about CPU metrics at all?

It looked like there was a PR to ignore these types of errors back in 1.12.2. https://github.com/fluxcd/flux/pull/2009

squaremo:
> Why does Flux care about CPU metrics at all?

It doesn't. Possibly this is an error reported by the Kubernetes API client, either from something it is attempting, or something attempted by an intermediary, or the API server itself.

stefanprodan:
@squaremo it looks like our use of the discovery API triggers the metrics queries, this is indeed very troubling since a busy cluster with many metrics will add a huge delay to Flux sync.

squaremo:
> it looks like our use of the discovery API triggers the metrics queries

How .. even ... I don't ...

stefanprodan:
I think we should be using `ServerGroups()` and call `ServerResourcesForGroupVersion` for each group excluding the metrics one to avoid querying the metrics providers.

stefanprodan:
[...]

Issue #2192, opened 2019-06-25
 Error while fetching pod metrics for selector  in namespace \"kube-public\": no pods to fetch metrics for"


tomaszkiewicz:
**Describe the bug**

I installed flux on one of my clusters, got initial sync with blank directory in git repo, then added and commited a file with namespace definition created with `kubectl create ns testing --dry-run -o yaml > testing-ns.yaml`.

After I made a commit and pushed it I ran `fluxctl sync` and got the following results and no resources created:

For `fluxctl sync`:
```
Synchronizing with git@github.com:tomaszkiewicz/xxxxxx-flux.git
Revision of master to apply is e40a0d1
Waiting for e40a0d1 to be applied ...
Error: timeout
Run 'fluxctl sync --help' for usage.
```

Logs from flux container:
```
flux-7894f6bdd8-gdjnd flux ts=2019-06-25T18:58:20.581107249Z caller=loop.go:119 component=sync-loop jobID=192e9d45-920b-a138-9eb1-2e4b36be63b7 state=in-progress
flux-7894f6bdd8-gdjnd flux ts=2019-06-25T18:58:21.817843071Z caller=loop.go:131 component=sync-loop jobID=192e9d45-920b-a138-9eb1-2e4b36be63b7 state=done success=true
flux-7894f6bdd8-gdjnd flux ts=2019-06-25T18:58:23.004831912Z caller=loop.go:111 component=sync-loop event=refreshed url=git@github.com:tomaszkiewicz/xxxxxx-flux.git branch=master HEAD=e40a0d15ad0f30af11c7261e42001d5dd4786d63
flux-7894f6bdd8-gdjnd flux ts=2019-06-25T18:58:30.803863871Z caller=loop.go:85 component=sync-loop err="collating resources in cluster for sync: Error while fetching pod metrics for selector  in namespace \"kube-public\": no pods to fetch metrics for"
```

**To Reproduce**
Described above, I've also tested that on other cluster but there were no visible errors in flux container logs, but the result was the same - timeout and no resources created.

**Expected behavior**
Flux should create a namespace according to definition in yaml file.

**Logs**
Provided above.

**Additional context**
Add any other context about the problem here, e.g 
- Flux version: 1.13.0
- Helm Operator version: N/A
[...]

Issue #2120, opened 2019-06-03
Vague "server could not find the requested resource" message


jpds:
I temporarily removed the `metrics-server` deployment from our Kubernetes cluster and saw this appear in our flux logs:

```
flux/flux-7fcb7dd6c4-d9dpd[flux]: ts=2019-06-03T08:05:00.54547656Z caller=loop.go:90 component=sync-loop err="collating resources in cluster for sync: the server could not find the requested resource"
```

It'd be nice if the log message included what it couldn't find. :)

**Additional context**
Add any other context about the problem here, e.g 
- Flux version: 1.12.3
- Kubernetes version: 1.11
- Git provider: Github

2opremio:
That error is not very helpful. However I am not sure it's directly related to the `metrics-server`.

Did the error only appear once?

jpds:
Yep, only saw it after yanking `metrics-server` and not since.

ksaritek:
if no one is looking at that issue, I am a volunteer for that.

hiddeco:
@ksaritek given the size / impact of this bug, this one is all yours, as we (core maintainers) have some bigger issues to tackle first.

ksaritek:
Hi @jpds, is that metrics-server located in your config repo? 
I tested against minikube version: v1.2.0 and followed the steps at get-started-developing.md

At first:
* k8s: v1.15
* flux: 1.13.3 (master branch)

I enabled metrics-server first (minikube addons enable metrics-server). Then deploy flux and remove metrics-server. After that, I checked the logs but I did not come across the error. 

Then :
* k8s: v1.11.0
* flux: 1.12.3
I enabled metrics-server first (minikube addons enable metrics-server). Then deploy flux and remove metrics-server. After that, I checked the logs but I did not come across the error. 

Also that "the server could not find the requested resource" seems like an error message that returns from k8s API. So I checked to be sure that there are any place that modifying k8s error and seems like no. 

@hiddeco, any idea to do next?

PR #2606, opened 2019-11-12
Exclude the metrics APIs from resources discovery

stefanprodan:
This PR refactors the Kubernetes resources discovery to exclude the `*.metrics.k8s.io` API groups and thus avoiding querying the cluster metrics and failing to sync when the metrics providers are unavailable.

Tested on EKS with metrics server and by @texasbobs with Prometheus adapter.

Fix: #2605 
Fix: #2120
Fix: #2192

### DIFF
--- a/pkg/cluster/kubernetes/sync.go
+++ b/pkg/cluster/kubernetes/sync.go
@@ -22,7 +22,6 @@ import (
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 
 	"github.com/fluxcd/flux/pkg/cluster"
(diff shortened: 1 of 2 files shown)


2opremio:
Thanks @stefanprodan !

It looks good, however I am not sure whether the specific functions you are calling are cached (I vaguely recall that the memory-cache implementation of the discovery interface only applied caching to certain functions), can you check?

stefanprodan:
It uses the same client as before https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go 

When enabling the go client logging the caller is:
```
kubernetes_caller=k8s.io/client-go@v11.0.0+incompatible/discovery/cached/memory/memcache.go
```

pkg/cluster/kubernetes/sync.go
```diff
@@ -239,6 +240,9 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
 	}
 
 	for _, resource := range resources {
+		if resource == nil {
```
squaremo: Why is this needed?

pkg/cluster/kubernetes/sync.go
```diff
@@ -208,21 +207,23 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
...
+	sgs, err := c.client.discoveryClient.ServerGroups()
+	if sgs == nil {
+		return nil, err
+	}
+
+	resources := []*meta_v1.APIResourceList{}
+	for i := range sgs.Groups {
+		gv := sgs.Groups[i].PreferredVersion.GroupVersion
+		// exclude the *.metrics.k8s.io resources to avoid querying the cluster metrics
+		if !strings.Contains(sgs.Groups[i].Name, "metrics.k8s.io") {
```
squaremo: `Contains` seems a little scattershot, can it be more specific? For example, will the name _start_ with metrics.k8s.io; or, if it's a "subdomain", will it _end_ with `.metrics.k8s.io`.

pkg/cluster/kubernetes/sync.go
stefanprodan: To avoid panics, a resource can be nil. I've hit this while testing it, I can remove it if you think it's not worth it.

squaremo: The client can return nil entries? Eugh. OK, that's sensible then -- maybe signpost it with a brief comment.

pkg/cluster/kubernetes/sync.go
stefanprodan: Doing two comparisons (starts with and ends with) is faster than contains? The `metics.k8s.io` domain is reserved to Kubernetes aggregator API, I don't see how is a scattershot.

squaremo: > Doing two comparisons (starts with and ends with) is faster than contains?

I'm not suggesting doing two comparisons; I'm suggesting being more precise with the one comparison.

> The `metrics.k8s.io` domain is reserved to Kubernetes aggregator API

If that's the exact name, then test for equality. If the items you want to avoid will be `<something>.metrics.k8s.io`, test that they end with `.metrics.k8s.io`. Otherwise it could exclude some future group "whatever-metrics.k8s.io" inadvertently. Sure, it is not that likely such a group would be introduced, but: better to succeed precisely than to fail accidentally.

pkg/cluster/kubernetes/sync.go
```diff
@@ -208,21 +207,23 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
...
+	sgs, err := c.client.discoveryClient.ServerGroups()
+	if sgs == nil {
+		return nil, err
+	}
+
+	resources := []*meta_v1.APIResourceList{}
+	for i := range sgs.Groups {
+		gv := sgs.Groups[i].PreferredVersion.GroupVersion
+		// exclude the *.metrics.k8s.io resources to avoid querying the cluster metrics
+		if sgs.Groups[i].Name != "metrics.k8s.io" && !strings.HasSuffix(sgs.Groups[i].Name, ".metrics.k8s.io") {
```
squaremo: Oh, it can be _either_ of exactly metrics.k8s.io or a subdomain. Eesh. Well this is still better, I reckon -- thanks for following my suggestion up.

stefanprodan:
@squaremo anything else blocking this? I've rebase with master.

pkg/cluster/kubernetes/sync.go
```diff
@@ -208,21 +207,23 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
...
+			if r, err := c.client.discoveryClient.ServerResourcesForGroupVersion(gv); err == nil {
+				resources = append(resources, r)
```
squaremo: Could you just not append `r` if it's `nil`, so you don't need to guard that later?

stefanprodan: done

PR #2607, opened 2019-11-13
fix indents in the .flux.yaml reference

translucens:
indents of the environment variable list are broken and fixed them


2opremio:
Thanks a lot!!!

PR #2621, opened 2019-11-20
Honor KUBE_CONFIG env. variable, if present

denysvitali:
Whilst debugging I realised that there is no proper way to specify a custom kubernetes config location. This PR adds such a functionality so that interested users can use the `KUBE_CONFIG` environment variable as it already happens with `kubectl`.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -197,7 +197,11 @@ func main() {
 	fs.MarkDeprecated("registry-poll-interval", "changed to --automation-interval, use that instead")
 
 	var kubeConfig *string
-	{
+	kubeConfigEnv, hasKubeConfigEnv := os.LookupEnv("KUBE_CONFIG")
+	
+	if hasKubeConfigEnv {
+		kubeConfig = &kubeConfigEnv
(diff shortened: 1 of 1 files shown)


denysvitali:
Currently I'm full of tasks and I can't handle this change. As soon as I'll have more free time I'll make the requested changes 👍

chancez:
This should use `KUBECONFIG`, not `KUBE_CONFIG`, as that's the standard variable for all kube tooling, and honestly, if you just used `KUBECONFIG`, it probably would work since flux is just using the upstream libraries which likely already look at `KUBECONFIG`.

2opremio:
> This should use `KUBECONFIG`, not `KUBE_CONFIG`, as that's the standard variable for all kube tooling, and honestly, if you just used `KUBECONFIG`, it probably would work since flux is just using the upstream libraries which likely already look at `KUBECONFIG`.

Ah, good point about `KUBECONFIG`, however Flux doesn't look at `KUBECONFIG` due to how the libraries are invoked (see my comment above).

2opremio:
Closing in favor of #2741

PR #2625, opened 2019-11-21
add --k8s-default-namespace argument to override default namespace

moshloop:

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -18,7 +18,6 @@ import (
 	"syscall"
 	"time"
 
-	helmopclient "github.com/fluxcd/helm-operator/pkg/client/clientset/versioned"
 	"github.com/go-kit/kit/log"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
@@ -51,6 +50,7 @@ import (
 	"github.com/fluxcd/flux/pkg/remote"
 	"github.com/fluxcd/flux/pkg/ssh"
 	fluxsync "github.com/fluxcd/flux/pkg/sync"
+	helmopclient "github.com/fluxcd/helm-operator/pkg/client/clientset/versioned"
 )
 
 var version = "unversioned"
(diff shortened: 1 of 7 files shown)


2opremio:
For traceability, this overrides #2588 which was mysteriously closed by GitHub

cmd/fluxd/main.go
```diff
@@ -170,10 +170,10 @@ func main() {
...
 
 		// k8s-scope settings
-		k8sNamespaceWhitelist    = fs.StringSlice("k8s-namespace-whitelist", []string{}, "restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set")
-		k8sAllowNamespace        = fs.StringSlice("k8s-allow-namespace", []string{}, "restrict all operations to the provided namespaces")
-
-		k8sVerbosity             = fs.Int("k8s-verbosity", 0, "klog verbosity level")
+		k8sNamespaceWhitelist = fs.StringSlice("k8s-namespace-whitelist", []string{}, "restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set")
+		k8sAllowNamespace     = fs.StringSlice("k8s-allow-namespace", []string{}, "restrict all operations to the provided namespaces")
+		k8sVerbosity          = fs.Int("k8s-verbosity", 0, "klog verbosity level")
+		k8sDefaultNamespace   = fs.String("k8s-default-namespace", "default", "the namespace to use for resources where a namespace is not specified")
```
2opremio: `k8sVerbosity` is not a scope setting, can you move it to a separate block, like it was before?

2opremio: Also, I wouldn't set a default value for `k8s-default-namespace`. Using `default` would require setting it to empty when the user wants to supply a default namespace through a kubeconfig file.

pkg/cluster/kubernetes/namespacer.go
```diff
@@ -23,21 +23,24 @@ type namespaceViaDiscovery struct {
...
+func NewNamespacer(d discovery.DiscoveryInterface, defaultNamespaceOverride string) (*namespaceViaDiscovery, error) {
+	if defaultNamespaceOverride != "" {
+		return &namespaceViaDiscovery{fallbackNamespace: defaultNamespaceOverride, disco: d}, nil
+	}
+	fallback, err := getKubeconfigDefaultNamespace()
 	if err != nil {
 		return nil, err
 	}
 	return &namespaceViaDiscovery{fallbackNamespace: fallback, disco: d}, nil
 }
```
2opremio: Nitpick (sorry, I should had thought about this before). Can you make `getKubeconfigDefaultNamespace()` return `""` if a default namespace isn't found? (i.e.` ""` instead of `defaultFallbackNamespace`). A function called `getKubeconfigDefaultNamespace()` shouldn't do anything else.

Then, in `NewNamespacer()`, you can use `defaultFallbackNamespace` if nothing else worked.

```golang
func NewNamespacer(d discovery.DiscoveryInterface, defaultNamespaceOverride string) (*namespaceViaDiscovery, error) {
	if defaultNamespaceOverride != "" {
		return &namespaceViaDiscovery{fallbackNamespace: defaultNamespaceOverride, disco: d}, nil
	}
	kubeconfigDefaultNamespace, err := getKubeconfigDefaultNamespace()
	if err != nil {
		return nil, err
	}
	if kubeconfigDefaultNamespace != "" {
		return &namespaceViaDiscovery{fallbackNamespace: kubeconfigDefaultNamespace, disco: d}, nil
	}	
	return &namespaceViaDiscovery{fallbackNamespace: defaultFallBackNamespace, disco: d}, nil
}
```

moshloop:
@2opremio done

docs/references/daemon.md
```diff
@@ -85,6 +85,7 @@ Version controlling of cluster manifests provides reproducibility and a historic
 | --k8s-secret-data-key                            | `identity`                         | data key holding the private SSH key within the k8s secret
 | **k8s configuration**
 | --k8s-allow-namespace                            |                                    | restrict all operations to the provided namespaces
+| --k8s-default-namespace                          | default                           | the namespace to use for resources where a namespace is not specified
```
2opremio:
```suggestion
| --k8s-default-namespace                          |                                    | the namespace to use for resources where a namespace is not specified
```

2opremio: @moshloop I tried committing this change to save you hassle (and merge the PR), but I can't. Would you mind doing the honors?

PR #2626, opened 2019-11-22
Add program to generate changelog release entries

2opremio:
This Go program uses the GitHub API to automate the arduous task of creating a changelog entry for an upcoming release.

It:

* Automatically obtains the tag of the latest release (to know from what commit to start looking for Pull Requests).
* Identifies pull requests
* Attempts to categorize pull requests in subsections using tags
* Generates the list of contributors

~I still need to clean the code up before it's ready to merge (e.g. there's a huge `main()` function)~


2opremio:
@hiddeco I guess we could reuse this for the Helm Operator as well? Should I put it in a separate repo?

hiddeco:
@2opremio yes, this can definitely be reused by the Helm operator. May require a slight rework of the `CHANGELOG.md` so that the categories match, but doing this will probably also make it easier for people familiar with both projects to work through / read the changes.

2opremio:
I decided against moving this to a new repo (it didn't feel right to create a repo just for this).
 
I did add flags to make it runnable to for the helm operator by just providing `--gh-repo=helm-operator`

internal_docs/generate_release_changelog/main.go
```diff
@@ -0,0 +1,255 @@
...
+	return comparison.Commits, nil
+}
+
+func findPullRequestsInCommits(ctx context.Context, ghClient *github.Client, gitHubOrg string, gitHubRepo string,
+	commits []github.RepositoryCommit, commitProcessHook func()) ([]*github.PullRequest, error) {
+	// Store pull requests in visiting order, preventing duplicates
+	prs := []*github.PullRequest{}
+	visitedPRs := map[int]struct{}{}
+	for _, commit := range commits {
+		commitProcessHook()
```
2opremio: I am not super happy about this. It seems rather artificial. The alternative I considered is returning a channel and running asynchronously, but I think that's a bit overkill.

internal_docs/generate_release_changelog/main.go
```diff
@@ -0,0 +1,255 @@
...
+		pullRequestListOptions := &github.PullRequestListOptions{}
+		// Note: Unfortunately this will return all the PRs whose head branch contains the commit
+		// (i.e. not just the PR which incorporates the commit to the base repo).
+		// We deal with this by discarding duplicates.
+		prsForCommit, _, err := ghClient.PullRequests.ListPullRequestsWithCommit(ctx,
```
2opremio: Calling this on every commit is very slow (that's why I added the progress bar). As an alternative we could parse the commit messages looking for `Merge` messages (and messages ending with `#number` for rebased PRs, which don't have a Merge message). However, I deemed that to be a bit too fragile.

2opremio:
> Can we move it to a less ugly path? :-)

Sure! What do you suggest?

internal_docs/generate_release_changelog/main.go
squaremo: You can look for commits with more than one parent. That will catch all PR merges except those that are fast-forwards, which should be none.

squaremo:
> > Can we move it to a less ugly path? :-)
> 
> Sure! What do you suggest?

Maybe we could have:
 - internal/docs/releasing.md
 - internal/cmd/changelog/main.go

(of which, only the latter matters here)

internal_docs/generate_release_changelog/main.go
2opremio: > That will catch all PR merges except those that are fast-forwards, which should be none.

We explicitly disable that?

squaremo: Yes I believe so. (But of course, that may change, you never know)

2opremio: Yes, I would rather not make assumptions.

PR #2629, opened 2019-11-22
fix: added support for annotation for memcached

emas80:
Added support for annotation on the memcached pod

It would be nice to have it because since when we've added Istio then the sidecar is automatically injected into any pod that does not specify a given annotation.
So we are currently unable to remove the sidecar from the memcached pod.

### DIFF
--- a/chart/flux/templates/memcached.yaml
+++ b/chart/flux/templates/memcached.yaml
@@ -18,6 +18,10 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
+      annotations:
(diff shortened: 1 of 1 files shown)


chart/flux/templates/memcached.yaml
```diff
@@ -18,6 +18,10 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
+      annotations:
```
2opremio: Please include the annotations inside the `if` block

Issue #2550, opened 2019-10-24
"Act normal" .flux.yaml mode


squaremo:
In https://cloud-native.slack.com/archives/CLAJ40HV3/p1571794316152600, Eric describes a problem with .flux.yaml processing: if you have a .flux.yaml anywhere, but you want plain YAMLs in some places, there are constraints over how you can organise things.

Since the manifest generation will look in parent directories for a `.flux.yaml`, once you have one, no directory underneath can use just plain YAMLs -- unless you write a script to cat them all.

A nice solution to this might be to have a variety of `.flux.yaml` config that reverts to YAML-as-usual processing. Possibly this could be composed with other forms (in the same way you can supply several commands) -- it needs a bit of thinking about.

squaremo:
@2opremio I'd appreciate your thoughts on this -- have I misdiagnosed what's going on?

2opremio:
The diagnosis is correct, but I am not sure whether that's the best solution though.

Generally, I would recommend using two different `--git-path` parameters (one for the raw repos and another one for the .flux.yaml-driven ones). This may not work in all cases, but I would like to understand why first.

EricHorst:
@2opremio are you saying that two `--git-path` parameters are allowed on the same fluxd? Or are you saying to run two different flux instances with different `--git-path`? The latter seems like a non-starter. We already have administrative fluxd and tenant fluxd and doubling that number seems less than ideal.

Two paths can be emulated in one repo without the `--git-path` parameter with a structure like:
```
/kustom/.flux.yaml
/kustom/kustomize.yaml
/raw/plain.yaml
```
But in either case, using this structure, or specifying two paths with the fluxd parameter, the first level of directories is completely artificial and forces one to make unnatural organizational decisions. If we have a deployment that has both generated and raw manifests, the must be split them into two, possibly distant (depending on the depth of each tree), locations. Further, if the vast majority of the repo is generated, and only one or two exceptions are needed it makes the raw location an oddball and over-lookable. That is to say, the structure needs to make sense and be maintainable by my team without laying traps which lead to mistakes. Lastly our team provides infrastructure to dev team tenants and we have to explain/convince and support  engineers organizing their own flux repos so having a solution which doesn't impose a repo structure will fly with greater ease.

[...]

PR #2634, opened 2019-11-26
Simplify fluxyaml reference

squaremo:
This rewrites the first half of the document describing manifest
generation, in pursuit of these goals:

 - be clear, up front, about how --git-path and .flux.yaml interact,
   since people tend to misunderstand that;

 - write more in terms of how to operate the feature, rather than how
   it works;

 - explain patchUpdated, which should be preferred in most situations
   as less complicated, first.

I set out to include the same information as the original. Although it's a reference, I wrote fairly conversationally. This might mean it's more appropriate to put this elsewhere, and restore the material here as a reference. I'm happy to take advice.

This is intended to address, in part, #2550

### DIFF
--- a/docs/references/fluxyaml-config-files.md
+++ b/docs/references/fluxyaml-config-files.md
@@ -5,33 +5,47 @@ instead of having to include them in your git repo as YAML files. For
 example, you can use `kustomize` to patch a common set of resources to
 suit a particular environment.
 
-## Setting manifest generation up
-
 Manifest generation is controlled by the flags given to `fluxd`, and
 `.flux.yaml` files in your git repo.
 
-To enable it, pass the command-line flag `--manifest-generation=true`
-to `fluxd`.
+To enable it, you will need to
 
-For each path supplied with the flag `--git-path` (or `.` if none are
-supplied),
+ 1. pass the command-line flag `--manifest-generation=true`
+to `fluxd`.
+ 2. put at least one `.flux.yaml` file in the git repository.
+
+Where to put `.flux.yaml`, and what should be in it, are described in
+the sections following.
+
+## Setting manifest generation up
+
+The command-line flag `--git-path` (which can be given multiple
+values) marks a "target path" within the git repository in which to
+find manifests. If `--git-path` is not supplied, the top of the git
+repository is assumed to be the sole target path.
+
+Without manifest generation, fluxd will look for YAML files under each
+target path.
+
+With manifest generation **enabled**, fluxd will look for processing
+instructions in a file `.flux.yaml`, which can be located _at_ the
+target path, or in a directory _above_ it in the git repository.
 
  - if a `.flux.yaml` file is found, it is used **instead** of looking
-   for YAML files;
+   for YAML files, and no other files are examined for that target
+   path;
 
  - if no `.flux.yaml` file is found, the usual behaviour of looking
-   for YAML files is used.
+   for YAML files is adopted for that target path.
 
-The manifests from all the `--git-path` values -- read from YAML files
-or generated -- are combined before applying to the cluster. If
-duplicates are detected, an error is logged and the sync is abandoned.
+The manifests from all the target paths -- read from YAML files or
+generated -- are combined before applying to the cluster. If
+duplicates are detected, an error is logged and fluxd will abandon the
+attempt to apply manifests to the cluster.
 
-A `.flux.yaml` file can be _at_ the path mentioned, or in a directory
-_above_ it in the git repo. Here are some examples:
+Here are some examples:
 
-```sh
-# show normal files _and_ .flux.yaml
-flux-kustomization-example$ tree -a -P .flux.yaml -P '*' -I .git
+```
 .
 ├── base
 │   ├── demo-ns.yaml
@@ -40,12 +54,10 @@ flux-kustomization-example$ tree -a -P .flux.yaml -P '*' -I .git
 │   ├── podinfo-hpa.yaml
 │   └── podinfo-svc.yaml
 ├── .flux.yaml
-├── LICENSE
 ├── production
 │   ├── flux-patch.yaml
 │   ├── kustomization.yaml
 │   └── replicas-patch.yaml
-├── README.md
 └── staging
     ├── flux-patch.yaml
     └── kustomization.yaml
@@ -53,14 +65,16 @@ flux-kustomization-example$ tree -a -P .flux.yaml -P '*' -I .git
 
 In this case, say you started `fluxd` with `--git-path=staging`, it
 would find `.flux.yaml` in the top directory and use that to generate
-manifests.  The files and directories (if there were any) in
-`staging/` are ignored, in favour of the instructions given in the
-`.flux.yaml` file.
+manifests.  The other files and directories (if there were any) in
+`staging/` are not examined by fluxd, in favour of following the
+instructions given in the `.flux.yaml` file.
 
-In this modified example,
+This layout could also be used with `--git-path=production`.
 
-```sh
-flux-kustomization-example$ tree -a -P .flux.yaml -P '*' -I .git
+In this modified example, the `.flux.yaml` file has been moved under
+`staging/`:
+
+```
 .
 ├── base
 │   ├── demo-ns.yaml
@@ -68,25 +82,23 @@ flux-kustomization-example$ tree -a -P .flux.yaml -P '*' -I .git
 │   ├── podinfo-dep.yaml
 │   ├── podinfo-hpa.yaml
 │   └── podinfo-svc.yaml
-├── LICENSE
 ├── production
 │   ├── flux-patch.yaml
-│   ├── .flux.yaml
 │   ├── kustomization.yaml
 │   └── replicas-patch.yaml
-├── README.md
 └── staging
     ├── flux-patch.yaml
+    ├── .flux.yaml
     └── kustomization.yaml
 ```
 
-… the `.flux.yaml` file is under `production/`, so it will _only_
-take effect for `--git-path=production`.
+… since the `.flux.yaml` file is now under `staging/`, it will still
+take effect for `--git-path=staging`. However:
 
-Using `--git-path=staging` would **not produce a usable
+Using `--git-path=production` would **not produce a usable
 configuration**, because without an applicable `.flux.yaml`, the files
-under `staging/` would be treated as plain Kubernetes manifests, which
-they are plainly not.
+under `production/` would be treated as plain Kubernetes manifests,
+which they are plainly not.
 
 Note also that the configuration file would **not** take effect for
 `--git-path=.` (i.e., the top directory), because manifest generation
@@ -95,8 +107,9 @@ will not look in subdirectories for a `.flux.yaml` file.
 ## How to construct a .flux.yaml file
 
 `.flux.yaml` files come in two varieties: "patch-updated", and
-"command-updated". These refer to the way in which automated updates
-are applied to files in the repo:
+"command-updated". These refer to the way in which [automated
+updates](./automated-image-update.md) are applied to files in the
+repo:
 
  - when patch-updated, fluxd will keep updates in its own patch file,
    which it applies to the generated manifests before applying to the
@@ -136,7 +149,7 @@ stream, before being applied.
 Much of the time, it will only be necessary to supply one command to
 be run.
 
-The commands will be run with the `--git-path` being processed as a
+The commands will be run with the target path being processed as a
 working directory -- which is not necessarily the same directory in
 which the `.flux.yaml` file was found. [See below](#execution-context)
 for more details on the execution context in which commands are run.
@@ -144,7 +157,8 @@ for more details on the execution context in which commands are run.
 ### Using patch-updated configuration
 
 A patch-updated configuration generates manifests using commands, and
-records updates in a patch file.
+records updates as a set of [strategic merge][strategic-merge] patches
+in a file.
 
 For example, when an automated image upgrade is run, fluxd will do
 this:
@@ -168,9 +182,12 @@ patchUpdated:
 ```
 
 The `generators` field is explained just above. The `patchFile` field
-gives a path, relative to the `--git-path` value, in which to record
-patches. `fluxd` will create the file if necessary, and commit any
-changes it makes to git.
+gives a path, relative to the target path, in which to record
+patches. `fluxd` will create or update the file when needed, and
+commit any changes it makes to git.
+
+> Note: at present, it is necessary to manually remove patches that
+> refer to deleted manifests. See [issue #2428][#2428]
 
 ### Using command-updated configuration
 
@@ -210,8 +227,7 @@ several times.
 
 `generators` and `updaters` are run in a POSIX shell inside the fluxd
 container. This means that the executables mentioned in commands must
-be available in the [Flux container
-image](https://github.com/fluxcd/flux/blob/master/docker/Dockerfile.flux).
+be available in the [Flux container image][flux-dockerfile].
 
 Flux currently includes `kustomize` and basic Unix shell tools. If the
 tools in the Flux image are not sufficient for your use case, you can
@@ -235,10 +251,11 @@ repository with this structure:
 └──── [...]
 ```
 
-The commands in `.flux.yaml` will be executed with their working
+… the commands in `.flux.yaml` will be executed with their working
 directory set to `staging`.
 
-In addition, `updaters` are given arguments via environment variables:
+In addition, commands from `updaters` are given arguments via
+environment variables, when executed:
 
  * `FLUX_WORKLOAD`: the workload to be updated. Its format is
   `<namespace>:<kind>/<name>` (e.g. `default:deployment/foo`). For
@@ -266,3 +283,7 @@ minute. If you run into errors like `error executing generator
 command: context deadline exceeded`, you can increase the timeout with
 the `--sync-timeout` fluxd command flag or the `sync.timeout` Helm
 chart option.
+
+[#2428]: https://github.com/fluxcd/flux/issues/2428
+[flux-dockerfile]: https://github.com/fluxcd/flux/blob/master/docker/Dockerfile.flux
+[strategic-merge]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md
--- a/docs/references/fluxyaml-config-files.md
+++ b/docs/references/fluxyaml-config-files.md
@@ -227,14 +227,21 @@ several times.
 
 `generators` and `updaters` are run in a POSIX shell inside the fluxd
 container. This means that the executables mentioned in commands must
-be available in the [Flux container image][flux-dockerfile].
+be available in the running fluxd container.
 
 Flux currently includes `kustomize` and basic Unix shell tools. If the
-tools in the Flux image are not sufficient for your use case, you can
-include new tools in your own Flux-based image or, if the tools are
-popular enough, Flux maintainers can add them to the Flux image
-(please create an issue). In the future it may be possibly to specify
-an container image for each command.
+tools in the Flux image are not sufficient for your use case, you have
+some options:
+
+ - build your own custom image based on the [Flux
+   image][flux-dockerfile] that includes the tooling you need, and run
+   that image instead of `fluxcd.io/flux`;
+ - copy files from an `initContainer` into a volume shared by the flux
+   container, within the deployment.
+
+In the future it may be possibly to specify an container image for
+each command, rather than relying on the tooling being in the
+filesystem.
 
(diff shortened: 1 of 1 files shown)


2opremio:
Really nice! 

By now it has become clear that this document was really confusing for users. Part of the problem is that it was [extracted from a design document](https://docs.google.com/document/d/1ebAjaZF84G3RINYvw8ii6dO4-_zO645j8Hb7mvrPKPY/edit?ts=5db9b7f0), meant to be precise and not necessarily user-friendly.

I will take a deep look tomorrow

2opremio:
Really nice! 

By now it has become clear that this document was really confusing for users. Part of the problem is that it was [extracted from a design document](https://docs.google.com/document/d/1ebAjaZF84G3RINYvw8ii6dO4-_zO645j8Hb7mvrPKPY/edit?ts=5db9b7f0), meant to be precise and not necessarily user-friendly.

I will take a deep look tomorrow

2opremio:
> This might mean it's more appropriate to put this elsewhere, and restore the material here as a reference. I'm happy to take advice.

I think it's fine to write it in a friendlier way. We already have the [design document](https://docs.google.com/document/d/1ebAjaZF84G3RINYvw8ii6dO4-_zO645j8Hb7mvrPKPY) if we ever want to look at (which I doubt).

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
-# Manifest factorization through `.flux.yaml` configuration files
+# Manifest generation through `.flux.yaml` configuration files
 
-## Enabling search of `.flux.yaml` files
+This feature lets you generate Kubernetes manifests with a program,
+instead of having to include them in your git repo as YAML files. For
```
2opremio:
```suggestion
instead of having to include them in your Git repository as YAML files. For
```

I am not sure we are consistent in terminology across the docs. If we use _git_ and _repo_ everywhere else I am fine with it. `Consistency > formal pickiness`

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+example, you can use `kustomize` to patch a common set of resources to
+suit a particular environment.
 
-To enable it supply the command-line flag `--manifest-generation=true` to `fluxd`.
+## Setting manifest generation up
 
-## Goal
+Manifest generation is controlled by the flags given to `fluxd`, and
+`.flux.yaml` files in your git repo.
```
2opremio: Same

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
-required, the two main ones being:
+To enable it, pass the command-line flag `--manifest-generation=true`
+to `fluxd`.
 
-* Having a staging/canary cluster and production cluster. The resources from the staging cluster are regularly promoted
-  to the production cluster. In addition, there are long-term differences between the cluster resources (e.g. different
-  security keys, different database endpoints etc ...).
-* Federation. Different clusters run in separate regions, usually with very similar resources but different
-  configurations.
+For each path supplied with the flag `--git-path` (or `.` if none are
```
2opremio: Maybe we should introduce what `--git-path` is beforehand.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
-To enable it supply the command-line flag `--manifest-generation=true` to `fluxd`.
+## Setting manifest generation up
 
-## Goal
+Manifest generation is controlled by the flags given to `fluxd`, and
+`.flux.yaml` files in your git repo.
 
-It is a common pattern to run very similar resources in separate clusters. There are various scenarios in which this is
-required, the two main ones being:
+To enable it, pass the command-line flag `--manifest-generation=true`
```
2opremio: Maybe we should highlight this a bit more, since it's essential for the feature. Do we use a highlight box or something like that in other parts of the docs?

docs/references/fluxyaml-config-files.md
2opremio: Also, it may be worth clarifying that `.` means the root directory of the repository.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+For each path supplied with the flag `--git-path` (or `.` if none are
+supplied),
 
-The main goal of `.flux.yaml` configuration files is to help deploying similar resources/clusters:
+ - if a `.flux.yaml` file is found, it is used **instead** of looking
+   for YAML files;
 
-* with minimal replication of resource definitions
-* while keeping Flux neutral about the factorization technology used
+ - if no `.flux.yaml` file is found, the usual behaviour of looking
```
2opremio: It may be worth briefly describing the usual behavior

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+   for YAML files is used.
 
-## File-access behaviour in Flux
+The manifests from all the `--git-path` values -- read from YAML files
+or generated -- are combined before applying to the cluster. If
+duplicates are detected, an error is logged and the sync is abandoned.
```
2opremio:
```suggestion
duplicates are detected, an error is logged and Flux will abort the synchronization with the cluster.
```

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
-   in the manifest files, adjusting it if necessary)
-2. Update the manifest files of [workload](fluxctl.html#workloads).
-   Specifically, flux can update:
-    * container images, when releasing a new image version. A release can happen manually or automatically, when a new
-      container image is pushed to a repository.
-    * annotations, which establish the release policy of a workload (e.g. whether it should be automatically released,
-      whether it should be locked from releasing, what image tags should be considered for automated releases …)
+```sh
+# show normal files _and_ .flux.yaml
+flux-kustomization-example$ tree -a -P .flux.yaml -P '*' -I .git
```
2opremio: Is the _sophisticated_ tree command and `flux-kustomization-example` name necessary? I think it doesn't add much value as is and can add confusion.

I think we should either be explicit about using a real-world example (even adding a prior `git clone` command on https://github.com/weaveworks/flux-kustomize-example ) or just show the tree.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+└── staging
+    ├── flux-patch.yaml
+    └── kustomization.yaml
+```
 
-Flux can be configured to confine the scope of (1) and (2):
+In this case, say you started `fluxd` with `--git-path=staging`, it
+would find `.flux.yaml` in the top directory and use that to generate
+manifests.  The files and directories (if there were any) in
+`staging/` are ignored, in favour of the instructions given in the
```
2opremio: Nitpick (the files are actually used)
```suggestion
`staging/` are not traversed by Flux, in favour of the instructions given in the
```

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+`.flux.yaml` file.
 
-* To specific (sub)directories (flag `--git-path`)
-* To a Git branch other than `master` (flag `--git-branch`)
+In this modified example,
```
2opremio: I think it's worth indicating what is being changed (I had to look at it a couple of times).

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+│   └── replicas-patch.yaml
+├── README.md
+└── staging
+    ├── flux-patch.yaml
+    └── kustomization.yaml
+```
 
-Flux allows you to declare configuration files to override file operations (1) and (2), by declaring commands which
-perform equivalent actions.
+… the `.flux.yaml` file is under `production/`, so it will _only_
```
2opremio: Super nitpick: I would probably change the example to moving the `.flux.yaml` file inside `staging` and saying that it's equivalent to the prior configuration, after setting `--git-path=staging`.

Also, it may be worth mentioning that (before making the change) you can specify the production environment by specifying `--git-path=production` without moving the `.flux.yaml` file.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
-        command: containerImage_updater_command2 ciu2arg1 ciu2arg2 ...
-      policy:
-        command: policy_updater_command2 pu2arg1 pu2arg2 ...
-```
+## How to construct a .flux.yaml file
 
-> **Note:** For a simpler approach to updates, Flux provides a `patchUpdated` configuration file variant.
+`.flux.yaml` files come in two varieties: "patch-updated", and
+"command-updated". These refer to the way in which automated updates
+are applied to files in the repo:
```
2opremio: Worth adding a link to the relevant documentation for automated updates?

docs/references/fluxyaml-config-files.md
2opremio: And, I am surely biased but, wouldn't it make sense to (at least briefly) present a `.flux.yaml` file before getting into their interaction with `--git-path`?

As a reader I am not sure I would be able to understand what is going on without having a rough high-level idea of what the configuration will do on that `--git-path`

2opremio: > And, I am surely biased but, wouldn't it make sense to (at least briefly) present a `.flux.yaml` file before getting into their interaction with `--git-path`?

I think it would make sense to show it and indicate it will be explained later.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+patchUpdated:
+  generators:
+  - command: jk generate ./index.js
+  patchFile: flux-patch.yaml
 ```
 
-The commands in `.flux.yaml` will be executed with their working directory set to `staging.`
+The `generators` field is explained just above. The `patchFile` field
+gives a path, relative to the `--git-path` value, in which to record
+patches. `fluxd` will create the file if necessary, and commit any
```
2opremio: Nitpick:
```suggestion
patches. `fluxd` will create the file when needed, and commit any
```

2opremio: Also, it may be worth mentioning that it's the user's responsibility to keep the patch file consistent with the generated manifests (e.g. if a manifests is removed, their patch counterpart needs to be removed too).

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+for more details on the execution context in which commands are run.
 
-3. `fluxd` will ignore any other yaml files under that path (e.g. resource manifests).
+### Using patch-updated configuration
 
-Generators and updaters are intentionally independent, in case a matching updater cannot be provided. It is hard to
-create updaters for some factorization technologies (particularly Configuration-As-Code). To improve the situation,
-a separate configuration file variant (`patchedUpdated`) is provided, which will be described later on.
+A patch-updated configuration generates manifests using commands, and
+records updates in a patch file.
```
2opremio: Is it worth mentioning it's a Strategic Merge patch? We could hide that fact to avoid confusion, but it's their responsibility consistent with the manifests they generate.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
-```yaml
-version: 1
-patchUpdated:
-  generators:
-    - command: helm template ../charts/mychart -f overrides.yaml
-  patchFile: flux-patch.yaml
-```
+Flux currently includes `kustomize` and basic Unix shell tools. If the
+tools in the Flux image are not sufficient for your use case, you can
+include new tools in your own Flux-based image or, if the tools are
```
2opremio: Is it worth mentioning that an `init` container can be used, without needing to create a custom image?

e.g. https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-init-containers-plugin-downloads.html

Maybe we can include an example adding an extra binary.

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+directory set to `staging`.
 
-```yaml
-version: 1
-patchUpdated:
-  generators:
-    - command: kustomize build .
-  patchFile: flux-patch.yaml
-```
+In addition, `updaters` are given arguments via environment variables:
```
2opremio: Nitpick:

```suggestion
In addition, the `updaters` of the `commandUpdated` configuration are given arguments via environment variables:
```

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,268 @@
...
+image](https://github.com/fluxcd/flux/blob/master/docker/Dockerfile.flux).
 
-```yaml
-version: 1
-patchUpdated:
-  generators:
-    - command: helm template ../charts/mychart -f overrides.yaml
-  patchFile: flux-patch.yaml
-```
+Flux currently includes `kustomize` and basic Unix shell tools. If the
```
2opremio:
```suggestion
Flux currently includes `kustomize`, `helm` and basic Unix shell tools. If the
```

2opremio: Ignore this, we don't include the binary in the image (we only use it for testing).

docs/references/fluxyaml-config-files.md
stefanprodan: I agree that explaining what `--git-path` does would make the docs better.

docs/references/fluxyaml-config-files.md
stefanprodan: I would go with a `git clone && cd flux-kustomization-example` command and remove `flux-kustomization-example$`.

docs/references/fluxyaml-config-files.md
stefanprodan: Many users were confused about GC and patches. We should mention that even if Flux GC removes a deployment the patch will error out if not removed manually

docs/references/fluxyaml-config-files.md
stefanprodan: I would include an example of `kustomize edit set image`

docs/references/fluxyaml-config-files.md
2opremio: BTW, I should move the repo to the `fluxcd` org.

@squaremo Do you want me to do it before merging this PR?

stefanprodan: GitHub deals with redirects, I would just move it now.

docs/references/fluxyaml-config-files.md
squaremo: We use "git repo" and "git repository" here and there. Of those I think I prefer "git repository". Using a capital gives it a bit too much emphasis I reckon.

docs/references/fluxyaml-config-files.md
squaremo: Cool, OK. I was self-consciously not piling up too much text before getting to the good stuff, but you are both right that `--git-path` deserves a bit of explaining.

docs/references/fluxyaml-config-files.md
squaremo: I wanted to show a directory structure, but it doesn't really matter how I get it, so I'll just remove the command from the examples.

docs/references/fluxyaml-config-files.md
squaremo: I did not attempt to give a working example of `commandUpdated` updaters, because the thought of trying to formulate one terrified me. `kustomize edit set imagetag` sets the image for _all things that use it_, last time I looked.

docs/references/fluxyaml-config-files.md
squaremo: I've used "git repository" consistently, now.

docs/references/fluxyaml-config-files.md
squaremo: I put a bit of explanation up-front, and substituted the term "target path" for `--git-path` afterwards. I got the term from the earlier text, and it seems OK to me, but I'm happy to adopt another.

docs/references/fluxyaml-config-files.md
squaremo: I moved it to the top and made the absolute requirements clearer.

docs/references/fluxyaml-config-files.md
squaremo: "looking for YAML files" pretty much is a brief explanation of the usual behaviour, isn't it?

docs/references/fluxyaml-config-files.md
squaremo: Rephrased, though not exactly as suggested

docs/references/fluxyaml-config-files.md
squaremo: Rephrased, with "examined" rather than "traversed"

docs/references/fluxyaml-config-files.md
squaremo: Yes good point that. Also adopted your suggestion below of sticking with `staging` in the example.

docs/references/fluxyaml-config-files.md
squaremo: Umm yes it probably is worth giving an example of that, fiddly though it is. Maybe I'll do so in another commit.

docs/references/fluxyaml-config-files.md
EricHorst: How about "the usual behaviour of recursively reading all YAML files under the target path."

docs/references/fluxyaml-config-files.md
squaremo: I've mentioned the need to manually remove patches when manifests go.

docs/references/fluxyaml-config-files.md
squaremo: I've mentioned it, along with building a custom image, as one of the possibilities here. I stopped short of giving a worked example -- maybe later.

squaremo:
> As a formerly confused user I heartily endorse the new document and submit unsolicited Approval.

:pray: Thank you Eric, as a formerly confused user, your opinion is very welcome here!

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,296 @@
...
+the sections following.
 
-The main goal of `.flux.yaml` configuration files is to help deploying similar resources/clusters:
+## Setting manifest generation up
 
-* with minimal replication of resource definitions
-* while keeping Flux neutral about the factorization technology used
+The command-line flag `--git-path` (which can be given multiple
+values) marks a "target path" within the git repository in which to
+find manifests. If `--git-path` is not supplied, the top of the git
```
2opremio: Maybe it's worth indicating that multiple `--git-paths` can be provided?

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,296 @@
...
+    ├── flux-patch.yaml
+    ├── .flux.yaml
+    └── kustomization.yaml
+```
+
+… since the `.flux.yaml` file is now under `staging/`, it will still
+take effect for `--git-path=staging`. However:
+
+Using `--git-path=production` would **not produce a usable
+configuration**, because without an applicable `.flux.yaml`, the files
```
2opremio: Even if redundant, I think it's worth indicating that a `.flux.yaml` file cannot be found in that case because there is non in `/production` or its parent directories (`/`)

docs/references/fluxyaml-config-files.md
```diff
@@ -1,229 +1,296 @@
...
-├── .flux.yaml
-├── staging/
-├──── [...]
-├── production/
-└──── [...]
+```yaml
+version: 1
+patchUpdated:
+  generators:
+  - command: jk generate ./index.js
```
2opremio: It may be worth indicating that `jk` is not in the container, or indicate how to install it, or simply use `kustomize`. Otherwise I think it may be confusing.

docs/references/fluxyaml-config-files.md
squaremo: You mean like saying "which can be given multiple values"?

docs/references/fluxyaml-config-files.md
squaremo: I suppooooooose

docs/references/fluxyaml-config-files.md
2opremio: Yes. Note that it's also valid to provide multiple `--git-path` arguments (pflag will coalesce them)

squaremo: Yup, that's what I mean by "provide multiple values" -- use the flag to provide multiple values. I don't want to get into how the flag parsing works in this doc though.

docs/references/fluxyaml-config-files.md
squaremo: Changed it to be `generator_command`, so as not to distract from the shape of the YAML, which is the point here.

PR #2637, opened 2019-11-27
Upgrading to latest memcached image which uses debian buster

adusumillipraveen:
Moving to latest image as debian stretch has vulnerabilities(  which our CSP keeps warning us about ) and also because it has other advantages.

### DIFF
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -234,7 +234,7 @@ memcached:
   enabled: true
   hostnameOverride:
   repository: memcached
-  tag: 1.5.15
+  tag: 1.5.20
   pullSecret:
   createClusterIP: true
   verbose: false
--- a/deploy/memcache-dep.yaml
+++ b/deploy/memcache-dep.yaml
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
       - name: memcached
-        image: memcached:1.5.15
+        image: memcached:1.5.20
         imagePullPolicy: IfNotPresent
         args:
         - -m 512   # Maximum memory to use, in megabytes
--- a/pkg/install/templates/memcache-dep.yaml.tmpl
+++ b/pkg/install/templates/memcache-dep.yaml.tmpl
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
       - name: memcached
-        image: memcached:1.5.15
+        image: memcached:1.5.20
         imagePullPolicy: IfNotPresent
         args:
         - -m 512   # Maximum memory to use, in megabytes


chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.15.0"
-version: 0.15.0
+version: 0.15.1
```
2opremio: Chart releasing has its own pace

```suggestion
version: 0.15.0
```

2opremio:
I think it may be better to drop the patch version and simply use `1.5`. Mutable tags are an anti-pattern, but ensuring implicit patch upgrades is probably worth it (we already do it, intentionally or not in the Flux-based alpine image).

@squaremo thoughts?

chart/flux/Chart.yaml
timja: @2opremio are you able to apply the suggestion, @adusumillipraveen is on my team and is on leave now till next year

squaremo:
> I think it may be better to drop the patch version number and simply use `1.5`

You can certainly do that for a particular installation; that's an easier decision to endorse than making it so for everyone.

2opremio:
Actually, after thinking about it further, it's not like we do when building Flux. The base image used by Flux is immutable but it's always the same when deploying (so, you cannot end up with different images when re-deploying, for instance).

chart/flux/Chart.yaml
2opremio: Unfortunately I don't have permission to push to @adusumillipraveen 's repository

2opremio: Actually, I was able to push from the command line, but GitHub didn't let me commit the suggestion.

timja: I had an issue with that last week, applying a suggestion for someone else, and getting access denied. They then showed a screenshot showing the allow edits from maintainers was ticked =/

2opremio: Yeah, that's what seems to be happening here. I think that the suggestion application doesn't work for PRs from forks.

2opremio:
I had to regenerate `pkg/install/generated_templates.gogen.go` and push it to master directly (see d16d92e3d03b8e8a2bb8bc3368f14481ea4e1423) because this missed it.

It should had been automatically detected by CI, but `make check-generated` didn't work in the latest build of this PR (https://circleci.com/gh/fluxcd/flux/8813 ) nor in the merge build in master ( https://circleci.com/gh/fluxcd/flux/8819 )

I noticed when working on a later, unrelated PR (#2626 ) for which `make check-generated` did work ( https://circleci.com/gh/fluxcd/flux/8826 ). Why that happen is beyond me at this point.

PR #2638, opened 2019-11-27
Reset switch in .flux.yaml

squaremo:
Briefly: let people reset the processing rules in `.flux.yaml`, so you can have a `.flux.yaml` at the top of the git repo, but switch that off for a particular `--git-path`.

(At present this is branched from another PR, because I will need to write a bit of documentation about this new thing, and better to do so in revised docs)

### DIFF
--- a/docs/references/fluxyaml-config-files.md
+++ b/docs/references/fluxyaml-config-files.md
@@ -41,6 +41,10 @@ target path, or in a directory _above_ it in the git repository.
  - if no `.flux.yaml` file is found, the usual behaviour of looking
    for YAML files is adopted for that target path.
 
+ - a `.flux.yaml` file containing the `files` directive resets the
+   behaviour to looking for YAML files. This is explained further,
+   below.
+
 The manifests from all the target paths -- read from YAML files or
 generated -- are combined before applying to the cluster. If
 duplicates are detected, an error is logged and fluxd will abandon the
@@ -107,10 +111,59 @@ Note also that the configuration file would **not** take effect for
 `--git-path=.` (i.e., the top directory), because manifest generation
 will not look in subdirectories for a `.flux.yaml` file.
 
+### The `files` directive
+
+The `files` directive indicates that the target path should be treated
+as though it had _no_ `.flux.yaml` in effect. In other words, fluxd
+will look for YAML files under the directory, and update manifests
+directly by rewriting the YAML files.
+
+Here's an example `.flux.yaml` with the `files` directive:
+
+```
+version: 1
+files: {}
+```
+
+(The `{}` is an empty map, which acts as a placeholder value).
+
+This is to account for the case in which you have a `.flux.yaml`
+higher in the directory tree, applying to several target paths beneath
+it, but want to have a directory wth regular YAMLs as well.
+
+In the following example, the top-level `.flux.yaml` would take effect
+for `--git-path=staging` or `--git-path=production`.
+
+But if you wanted `yamls/permissions.yaml` to be applied (as it is),
+you could put a `.flux.yaml` containing `files` in that directory, and
+specify `--git-path=staging,yamls`.
+
+```
+.
+├── .flux.yaml
+├── base
+│   ├── demo-ns.yaml
+│   ├── kustomization.yaml
+│   ├── podinfo-dep.yaml
+│   ├── podinfo-hpa.yaml
+│   └── podinfo-svc.yaml
+├── production
+│   ├── flux-patch.yaml
+│   ├── kustomization.yaml
+│   └── replicas-patch.yaml
+├── yamls
+│   ├── .flux.yaml # (with "files" directive)
+│   └── permissions.yaml
+└── staging
+    ├── flux-patch.yaml
+    └── kustomization.yaml
+```
+
 ## How to construct a .flux.yaml file
 
-`.flux.yaml` files come in two varieties: "patch-updated", and
-"command-updated". These refer to the way in which [automated
+Aside from the special case of the `files` directive, `.flux.yaml`
+files come in two varieties: "patch-updated", "command-updated". These
+refer to the way in which [automated
 updates](./automated-image-update.md) are applied to files in the
 repo:
 
--- a/docs/references/fluxyaml-config-files.md
+++ b/docs/references/fluxyaml-config-files.md
@@ -41,9 +41,8 @@ target path, or in a directory _above_ it in the git repository.
  - if no `.flux.yaml` file is found, the usual behaviour of looking
    for YAML files is adopted for that target path.
 
- - a `.flux.yaml` file containing the `files` directive resets the
-   behaviour to looking for YAML files. This is explained further,
-   below.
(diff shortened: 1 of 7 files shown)


pkg/manifests/configfile.go
```diff
@@ -23,15 +24,79 @@ const (
...
 )
 
+// This is easier to read as YAML, trust me.
+const configSchemaYAML = `
+"$schema": http://json-schema.org/draft-07/schema#
+definitions:
+  command:
+    type: object
+    required: ['command']
+  version: { const: 1 }
```
2opremio: Shouldn't we bump the version?

2opremio:
Can you extend the `.flux.yaml` documentation with the additional syntax?

pkg/manifests/configfile.go
```diff
@@ -23,15 +24,79 @@ const (
...
+  properties:
+    version: { '$ref': '#/definitions/version' }
+    files:
+      additionalProperties: false
+  additionalProperties: false
+`
+
+var ConfigSchema *jsonschema.Schema
+
+func init() {
```
2opremio: Instead of using `init()` could you rename this to `mustCompileSchema(string)` (or the likes) and do:

```
var ConfigSchema *jsonschema.Schema = mustCompileSchema(configSchemaYAML)
```

pkg/manifests/configfile.go
squaremo: Why? No-one has used this yet.

2opremio: But we are changing the file format

squaremo: Versioning here is to prevent misunderstandings about what the file means; in particular, to make sure that you aren't giving fluxd an instruction it doesn't know how to carry out.

Can you think of a potential mistake that would be avoided by bumping the version here?

squaremo: I am not absolutely against it, but it does mean writing code that would discriminate between the two versions (assuming we'd want to not invalidate everyone's config immediately.)

pkg/manifests/configfile.go
squaremo: Sure. I presume it would compile to the same thing, is there a reason to prefer this formulation?

stefanprodan: `init()` functions play a special role in go, to avoid confusion I would rename it.

squaremo: @stefanprodan Yes, I know, I am using it for the special role.

2opremio: > is there a reason to prefer this formulation?

It is more explicit and readable

pkg/manifests/configfile.go
2opremio: > Can you think of a potential mistake that would be avoided by bumping the version here?

No, not really.

> I am not absolutely against it, but it does mean writing code that would discriminate between the two versions (assuming we'd want to not invalidate everyone's config immediately.)

Yeah, that's a pain.

I think that, technically, we should bump the version number but I am not sure its worth the pain.

Maybe we could use minor version numbers for backward-compatible syntax changes (without needing to handle it differently in the code). Or maybe we can just leave it at version 1 .

squaremo:
Now I'm writing some docs, it occurs to me that this:

```
version: 1
files: {}
```

looks like it says "include the following files (none)". The map value `{}` is just to have a value for the field, and it's a map because maybe we want to add more stuff there later, and otherwise any value would do.

Thinking thinking thinking ...

pkg/manifests/configfile.go
squaremo: okdokes

squaremo:
> Now I'm writing some docs, it occurs to me that this:
> 
> ```
> version: 1
> files: {}
> ```
> 
> looks like it says "include the following files (none)".

Use another word:

```
version: 1
plainFilesMode: {}
```

 * doesn't look like it's supposed to be including or excluding specific files

Use another, constant value:

```
version: 1
files: true
```

 * it's obvious that it's a toggle (although what is the meaning of `files: false`?)
 * adding more configs means a breaking schema change

Of these I think I prefer the first.

2opremio:
how about something a bit more descriptive

```
version: 1
scanThisDirForManifestFiles: true
```

stefanprodan:
I like `files: {}`, it gives us the the possibility to add glob filters later on, something like:

```
version: 1
files:
 include: "*prod.yaml"
 exclude: "crd*"
```

squaremo:
> I like `files: {}`, it gives us the the possibility to add glob filters later on

Yeah I like it for that reason too; it's just that the empty value (for now, the only possible value) gives the wrong impression.

squaremo:
Replaced `files: {}` with `scanForFiles: {}`, which I think is descriptive enough, and avoids the ambiguity. Possibly.

docs/references/fluxyaml-config-files.md
```diff
@@ -107,9 +110,58 @@ Note also that the configuration file would **not** take effect for
...
+
+This is to account for the case in which you have a `.flux.yaml`
+higher in the directory tree, applying to several target paths beneath
+it, but want to have a directory wth regular YAMLs as well.
+
+In the following example, the top-level `.flux.yaml` would take effect
+for `--git-path=staging` or `--git-path=production`.
+
+But if you wanted `yamls/permissions.yaml` to be applied (as it is),
+you could put a `.flux.yaml` containing `files` in that directory, and
```
2opremio:
```suggestion
you could put a `.flux.yaml` containing `scanForFiles` in that directory, and
```

pkg/manifests/configaware_test.go
```diff
@@ -385,3 +384,30 @@ func TestDuplicateInGenerators(t *testing.T) {
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "duplicate")
 }
+
+func TestJustFiles(t *testing.T) {
+	// +-- config
+	//   +-- .flux.yaml (patchUpdated)
+	//   +-- rawfiles
+	//     +-- .flux.yaml (files)
```
2opremio:
```suggestion
	//     +-- .flux.yaml (scanForFiles)
```

pkg/manifests/configaware_test.go
```diff
@@ -385,3 +384,30 @@ func TestDuplicateInGenerators(t *testing.T) {
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "duplicate")
 }
+
+func TestJustFiles(t *testing.T) {
```
2opremio:
```suggestion
func TestSccanForFiles(t *testing.T) {
```

pkg/manifests/configfile_test.go
```diff
@@ -5,11 +5,96 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"gopkg.in/yaml.v2"
+	//	jsonschema "github.com/xeipuuv/gojsonschema"
```
2opremio:
```suggestion
```

squaremo:
Merge it while it's hot!

Thanks @2opremio and @stefanprodan for reviews and comments.

Issue #2651, opened 2019-12-02
Make `fluxctl install` work with Open Shift


richardcase:
**Describe the bug**
When i run the following against open shift:

```bash
fluxctl install \
--git-user=${GHUSER} \
--git-email=${GHUSER}@users.noreply.github.com \
--git-url=git@github.com:${GHUSER}/flux-get-started \
--git-path=namespaces,workloads \
--namespace=flux | oc apply -f -
```
(same issue if you use `kubectl` instead of `oc`)

The flux pod starts ok but the memcache pod fails to start. Looking at why i see the following:

```yaml
pods "memcached-78f7c9df46-" is forbidden: unable to validate against
any security context constraint:
[spec.containers[0].securityContext.securityContext.runAsUser: Invalid
value: 11211: must be in the ranges: [1000510000, 1000519999]]
```

**To Reproduce**
Run the command above against an OpenShift cluster.

**Expected behavior**
I would expect flux and memcached to start and deploy podinfo (from the sample repo)

**Additional context**
I have got this working by manually changing the yaml to comment out/remove the security context entry for the memcached deployment:

```yaml
        ports:
        - name: clients
          containerPort: 11211
        #securityContext:
        #  runAsUser: 11211
        #  runAsGroup: 11211
        #  allowPrivilegeEscalation: false
```

If you now run the yaml in memcached starts. With this change OpenShift will assign the memcached pods to the `restricted` [SCC](https://docs.openshift.com/enterprise/3.0/architecture/additional_concepts/authorization.html#security-context-constraints).

We could make this more explicit by creating a service account for memcached and explicitly assign that to an SCC (either `restricted` or `nonroot`) when targetting openshift.

This also relates to #1125

richardcase:
As i have this working i am happy to make the fix

2opremio:
> As i have this working i am happy to make the fix

That would be fantastic!

> and explicitly assign that to an SCC (either `restricted` or `nonroot`) when targetting openshift.

The problem is that `fluxctl install` doesn't know what you are using the yamls for

richardcase:
> The problem is that fluxctl install doesn't know what you are using the yamls for

I've added a `--oc` flag in my changes so that we can:
```bash
fluxctl install \
--git-user=${GHUSER} \
--git-email=${GHUSER}@users.noreply.github.com \
--git-url=git@github.com:${GHUSER}/flux-get-started \
--git-path=namespaces,workloads \
[...]

Issue #1125, opened 2018-06-07
Document extra setup/permissions when using OpenShift


squaremo:
When using

    fluxd --ssh-keygen-dir=/var/fluxd/keygen --git-url=whatever

This shows in the logs:

```
ts=2018-06-06T13:43:35.453724903Z caller=main.go:132 version=1.4.0
ts=2018-06-06T13:43:35.58808823Z caller=main.go:208 err=“exit status 255”
```

The error in the log is from attempting to create the SSH keyring. After a bit of fiddling,

```
ts=2018-06-06T14:22:11.737167092Z caller=main.go:208 err=“failed to chmod identity file: chmod /etc/fluxd/ssh/identity: read-only file system”
```

It's supposed to be a read-only filesystem, so that's OK; but we should only try to chmod it if it's not already the correct permissions, so something went wrong with the mount, or the original config.

Apparently, doing

```
oc adm policy add-scc-to-user anyuid system:serviceaccount:flux:flux
```

(`oc` is the OpenShift command-line tool) and using `readOnly: false` for the secret volume fixes it, but I think this is a fudge at best -- it should be entirely happy with a read-only key.

ZeeFiix:
You wrote "After a bit of fiddling...".
I have also the problem witch exit status 255.

@squaremo : How can i increase logging so i can see the reason of the exit status 255 ?

squaremo:
I expect I meant that I put in extra logging. You could exec into the container and check the mode bits of the file -- that would at least tell you whether it's the exact problem described above.

I don't know how to fix it through; it seems that OpenShift has its own idea of how to mount volumes.

severi:
I got flux up and running on Openshift (v3.9) by giving the service account permissions to run as privileged 

```
oc adm policy add-scc-to-user privileged -z flux
```

and adding the following the the deployment.yaml

```
      containers:
      - name: flux
        securityContext:
          runAsUser: 0
```
This is most likely needed because Openshift normally runs the containers with a randomly generated user, which does not have permissions to edit the mounted secret.

Also I had to create the empty secret ( `flux-secret.yaml` ) before creating the deployment.

Just a side note regarding other things I changed from the default installation: 
- I ran flux in a single namespace and didn't give the service account permissions to other namespaces.
- added custom `known_hosts` as I'm using an enterprise github installation rather than github.com
- changed the default namespace  that resources are deployed to

severi:
[...]

PR #2652, opened 2019-12-02
Changes for openshift as a install target

richardcase:
Added the `--oc` flag to indicate that the target of the install is
an OpenShift cluster. The effect of this flag is to not apply the security context part of the memcached deployment. It also then allows you to use the following flags:

* `--oc-add-memcached-scc` - this will create a memcached specific service account,  SCC and also updates the deployment to use the service account
* `--oc-add-flux-scc` - this will create a flux specific SCC.

If you don't specify the `--oc-add-*` flags then the default SCC will be applied to the flux and memcached pods. 

Still a work in progress, adding additional changes and tests.......

Relates to: #2651


2opremio:
@richardcase is this ready to review?

richardcase:
@2opremio - yes i think it is probably ready for review

pkg/install/templates/memcache-account-oc.tmpl
```diff
@@ -0,0 +1,47 @@
...
+users:
+    - memcached
+volumes:
+- configMap
+- downwardAPI
+- emptyDir
+- persistentVolumeClaim
+- projected
+- secret
+{{ end }}
```
2opremio:
```suggestion
{{ end }}

```

pkg/install/templates/memcache-dep.yaml.tmpl
```diff
@@ -27,8 +28,8 @@ spec:
         # - -vv    # Uncomment to get logs of each request and response.
         ports:
         - name: clients
-          containerPort: 11211
+          containerPort: 11211 {{ if not .OpenShiftTarget}}
```
2opremio:
```suggestion
          containerPort: 11211{{ if not .OpenShiftTarget}}
```

pkg/install/templates/memcache-dep.yaml.tmpl
```diff
@@ -27,8 +28,8 @@ spec:
...
         securityContext:
           runAsUser: 11211
           runAsGroup: 11211
-          allowPrivilegeEscalation: false
+          allowPrivilegeEscalation: false {{ end }}
```
2opremio:
```suggestion
          allowPrivilegeEscalation: false{{ end }}
```

2opremio:
Apologies for taking so long to review this (Christmas vacation got on the way). What is the purpose of `oc-add-memcached-scc`and `oc-add-flux-scc`. Shouldn't they always be added when using `--oc`?

I am not familiar with SCCs so sorry if that's obvious. At the very least, I think that the Memcache service account should always be added, to make it consistent with this code

```
    spec:{{ if .OpenShiftTarget}}
      serviceAccountName: memcached{{ end }}
```

richardcase:
> I am not familiar with SCCs so sorry if that's obvious. At the very least, I think that the Memcache service account should always be added, to make it consistent with this code

@2opremio  - they are a OpenShift specific way of giving pods specific permissions. A bit like PodSecurityPolices.  If you don't specify an explicit SCC then OpenShift will assign one of the builtin in SCCs to the pod. I kept them as optional flags for this reason but we would live without the explicit SCCs which would be easier.

richardcase:
And maybe we can change the `--oc` flag to `--no-security-contect`?

2opremio:
> And maybe we can change the `--oc` flag to `--no-security-contect`?

I like that, since then we wouldn't have target-specific flags.

richardcase:
> I like that, since then we wouldn't have target-specific flags.

Great, i'll make that change now

richardcase:
@2opremio - i made the change for `--no-security-context`

pkg/install/templates/memcache-account.yaml.tmpl
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: ServiceAccount
```
2opremio: Why do we need a specific service account for memcached? Why not use the flux service account?

pkg/install/install_test.go
```diff
@@ -39,12 +61,22 @@ func TestFillInTemplatesAllParameters(t *testing.T) {
...
+		NoSecurityContext:  false,
+	}
+
+	manifests := testFillInTemplates(t, 6, params)
+
+	memDeploy := manifests["memcache-dep.yaml"]
+	deployment := testExtractDeployment(t, memDeploy)
 
+	container := deployment.Spec.Template.Spec.Containers[0]
+	if container.SecurityContext == nil {
```
2opremio: Please create a separate test for this `TestFillInTemplatesAllParameters` is supposed to only do that

cmd/fluxctl/install_cmd.go
```diff
@@ -51,6 +51,7 @@ fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' | ku
 	cmd.Flags().BoolVar(&opts.RegistryScanning, "registry-scanning", true,
 		"scan container image registries to fill in the registry cache")
 	cmd.Flags().StringVarP(&opts.outputDir, "output-dir", "o", "", "a directory in which to write individual manifests, rather than printing to stdout")
+	cmd.Flags().BoolVar(&opts.NoSecurityContext, "no-security-context", false, "Ensure that no security context information is added to the pod specs. Defaults to 'false'")
```
2opremio: I think negatives are confusing. How about `--add-security-context` and defaulting it to `true`?

2opremio:
BTW, we should probably do the same changes on the Helm chart

2opremio:
Also, we should update the documentation with this. I don't know if we already include how to install Flux for OpeShift but we should probably add/update that.

pkg/install/templates/memcache-account.yaml.tmpl
richardcase: We probably don't. I will remove

pkg/install/install_test.go
richardcase: Good point, i have created this test

cmd/fluxctl/install_cmd.go
richardcase: Yes it is confusing, very good point. I have changed this.

richardcase:
> BTW, we should probably do the same changes on the Helm chart

Ah yes, i hadn't thought of that. I have updated it now.

> Also, we should update the documentation with this. I don't know if we already include how to install Flux for OpeShift but we should probably add/update that.

The docs do need to be updated. There is #1125 about updating the docs for openshift. Would you like me to update the docs as pary of this change? Or do a separate doc PR at a later date?

2opremio:
No preference

deploy/memcache-account.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
2opremio: I think you forgot to remove this

richardcase: Yep missed that after removing the memcached service account.. Removed the file now.

richardcase:
> No preference

I'll do a separate PR and link to #1125.

pkg/install/templates/memcache-dep.yaml.tmpl
```diff
@@ -18,6 +18,7 @@ spec:
     spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
+      serviceAccountName: flux
```
2opremio: Do you need an explicit account to make it work on openshift? Flux needs permissions Memecached doesn't, so, the default account is a better choice.

Sorry for not writing this before.

richardcase: Previously I had always tested it with an explicit service account. But testing it just now it doesn't need it. I will remove it and push a change now.

pkg/install/install_test.go
```diff
@@ -1,16 +1,22 @@
...
 
 	"github.com/instrumenta/kubeval/kubeval"
 	"github.com/stretchr/testify/assert"
+
+	v1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
-func testFillInTemplates(t *testing.T, expectedManifestCount int, params TemplateParameters) {
+func testFillInTemplates(t *testing.T, expectedNoManifestst int, params TemplateParameters) map[string][]byte {
```
2opremio:
```suggestion
func testFillInTemplates(t *testing.T, expectedNoManifests int, params TemplateParameters) map[string][]byte {
```

pkg/install/install_test.go
```diff
@@ -24,6 +30,22 @@ func testFillInTemplates(t *testing.T, expectedManifestCount int, params Templat
 			}
 		}
 	}
+
+	return manifests
+}
+
+func testExtractDeployment(t *testing.T, data []byte) *v1.Deployment {
+	manifest := string(data)
+	manifest = strings.Replace(manifest, "---\n", "", -1)
```
2opremio: Why is this needed?

2opremio: (I think it should work without it)

pkg/install/install_test.go
```diff
@@ -24,6 +30,22 @@ func testFillInTemplates(t *testing.T, expectedManifestCount int, params Templat
 			}
 		}
 	}
+
+	return manifests
+}
+
+func testExtractDeployment(t *testing.T, data []byte) *v1.Deployment {
```
2opremio: I would simply inline this function

pkg/install/install_test.go
```diff
@@ -24,6 +30,22 @@ func testFillInTemplates(t *testing.T, expectedManifestCount int, params Templat
...
+	manifest := string(data)
+	manifest = strings.Replace(manifest, "---\n", "", -1)
+
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewBuffer([]byte(manifest)), 4096)
+	deployment := new(v1.Deployment)
```
2opremio:
```suggestion
	var deployment v1.Deployment
```

2opremio:
Minor nits, otherwise LGTM

pkg/install/install_test.go
richardcase: Changed.

pkg/install/install_test.go
richardcase: Its not needed. I was having another issue at the time. Removed.

pkg/install/install_test.go
richardcase: I haven't inlined this as its used in 2 different test functions.

pkg/install/install_test.go
richardcase: Made a change to this

richardcase:
@2opremio - thanks for the review. I've made some minor changes based on the nits.

pkg/install/install_test.go
2opremio: Ah, true. Can you rename it to `unmarshalDeployment`?

2opremio:
One additional nit. Could you rebase on master?

2opremio:
BTW, thanks a lot for coping with all my comments. This PR took way longer to merge than it should had.

richardcase:
> One additional nit. Could you rebase on master?

All done.

2opremio:
Note to self: this contains both helm chart changes and `fluxctl onstall` changes so it should go both into the Flux and Chart CHANGELOGs

PR #2658, opened 2019-12-04
Fix link in troubleshooting docs

Crevil:
The linked issue #1422 is not rendered in Read the docs as a link.

https://docs.fluxcd.io/en/stable/troubleshooting.html#flux-is-taking-a-long-time-to-apply-manifests-when-it-syncs

This change fixes the rendering to follow the `[fluxcd/flux#number]` format used
else where in the docs.


2opremio:
Thanks!

PR #2659, opened 2019-12-04
`fluxctl build` command for debugging configs locally

squaremo:
This adds a (slightly rough and ready, so far) subcommand to fluxctl which will generate manifests as fluxd would in the cluster. The idea is to give people a way to check that their local changes will give the results they expect.

```
$ fluxctl build -C ~/scratch/flux-kustomize-example/ -p staging --manifest-generation
building in /home/mikeb/scratch/flux-kustomize-example/
connecting to Kubernetes cluster https://192.168.39.67:8443...connected
generating manifests...done
<cluster>:namespace/demo
demo:service/podinfo
demo:deployment/poddy
demo:horizontalpodautoscaler/podinfo
```

(the output is just the names for now, to demonstrate that it's doing something)

Since fluxd needs to connect to the Kubernetes API for this task, so does `fluxctl build` -- it will use the usual means of figuring out how to connect a client. It will also need any tools that are mentioned in a `.flux.yaml` to be in the path, of course.

TODO

 - [ ] output the actual manifests as a YAML stream

### DIFF
--- a/cmd/fluxctl/build.go
+++ b/cmd/fluxctl/build.go
@@ -1,13 +1,26 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"path/filepath"
+	"time"
 
+	"github.com/go-kit/kit/log"
 	"github.com/spf13/cobra"
+	disco "k8s.io/client-go/discovery"
+	"k8s.io/client-go/tools/clientcmd"
+
+	kube "github.com/fluxcd/flux/pkg/cluster/kubernetes"
+	"github.com/fluxcd/flux/pkg/manifests"
 )
 
 type buildOpts struct {
+	base       string
+	paths      []string
+	generation bool   // use manifest generation
+	defaultNS  string // override the default namespace
 }
 
 func newBuild() *buildOpts {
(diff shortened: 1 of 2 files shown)


cmd/fluxctl/build.go
```diff
@@ -0,0 +1,129 @@
...
+	cmd := &cobra.Command{
+		Use:   "build",
+		Short: "Generate manifests from a local directory as fluxd would",
+		RunE:  opts.RunE,
+	}
+
+	cmd.Flags().StringVarP(&opts.base, "base", "C", ".", "run as if invoked in the directory given; the working directory is treated as the root, for the purpose of interpreting config")
+	cmd.Flags().StringSliceVarP(&opts.paths, "path", "p", nil, "use the path(s) given as target paths to be examined; if not supplied, the root directory is assumed as the sole target path")
+	cmd.Flags().BoolVarP(&opts.generation, "manifest-generation", "g", false, "act the same as fluxd with --manifest-generation; i.e., look for .flux.yaml files in the target path(s) and use the instructions there")
+	cmd.Flags().StringVar(&opts.defaultNS, "default-namespace", "", "give the default namespace explicitly, rather than taking it from KUBE_CONFIG")
```
2opremio: we should make this `--k8s-default-namespace` to make it consistent with `fluxd`

2opremio:
Thanks @squaremo ! I am really excited about this.

I am not sure about the `build` name. 

A natural addition to printing the manifests is doing manifest updates locally (i.e. a local version of `lock`/`unlock`, `automate`, `release` etc ...). The idea I had in mind was to do so through a flag e.g. `--local=<basedir>` flag. So, instead of `fluxctl build`, we could do `fluxctl sync --local=<basedir>`. This opens the door to simulating the full sync from a local directory (we probably want to do a `dry-run` in the cluster when doing `sync --local`).

2opremio:
Also, I am not sure we should be printing the manifests by default. It's of little use when things work as expected (maybe add a `--debug` flag?)

squaremo:
> I am not sure about the `build` name.
> 
> A natural addition to printing the manifests is doing manifest updates locally (i.e. a local version of `lock`/`unlock`, `automate`, `release` etc ...). What I had in mind was to do so through a flag e.g. `fluxctl lock --local=<basedir>`.
> 
> So, instead of `fluxctl build`, we could do `fluxctl sync --local=<basedir>`. This opens the door to simulating the full sync from a local directory (we probably want to do a `dry-run` in the cluster when doing so, to avoid conflicts with the real repo used by Flux).

This is not intended to open the door to syncing from a local repo, it's intended to give people a way to check that their (probably locally changed) repo will generate a correct configuration. That's the reason it's _not_ nominally linked to `sync`, and the reason it will (once I implement it) print the manifests out.

Ideally you would be able to run it without access to the cluster, so it could be used in CI, but for the use case of just checking things by hand this is good enough I think.

I'm not against implementing local operations, I just don't think `sync --local` is a good way to give people access to this.

> Also, I am not sure we should be printing the full manifests by default. It's of little use when things work as expected

It is exactly the thing that's of use! If the point is to be able to see the result of your local change (and perhaps do further operations on the generated manifests), you have to see the result. And that _is_ the intended point here.

I'm amenable to having a `-q` that suppresses the full manifests and just prints the names, since yes, sometimes you will just want to know if succeeded; and/or a flag to write each manifest into its own file somewhere, for later examination (like `fluxctl install -o` does).

2opremio:
> This is not intended to open the door to syncing from a local repo, it's intended to give people a way to check that their (probably locally changed) repo will generate a correct configuration.

I understand, but making sure that the manifests would be able to go through an actual sync dry-run would be valuable.

> I'm amenable to having a `-q` that suppresses the full manifests and just prints the names, since yes, sometimes you will just want to know if succeeded; and/or a flag to write each manifest into its own file somewhere, for later examination (like `fluxctl install -o` does).

I can agree to that :)

squaremo:
> > This is not intended to open the door to syncing from a local repo, it's intended to give people a way to check that their (probably locally changed) repo will generate a correct configuration.
> 
> I understand, but making sure that the manifests would be able to go through an actual sync dry-run would be valuable.

Yes I see why that would be useful

    fluxctl build . | kubectl apply -f- --dry-run

... though that would not tell you what would be garbage collected. I do feel like it's a distinct use case.

Making this one accessible via sync would mean more modality there (i.e., more disjoint ways a single command can behave depending on its flags):

    fluxctl sync # ask fluxd to refresh git and apply to the cluster
    fluxctl sync --local # generate manifests locally and apply them to the cluster
    fluxctl sync --local --dry-run # generate manifests and pretend to apply them to the cluster
    fluxctl sync --local --print # generate manifests and print them out rather than applying them

and other combinations no doubt. The point being that these aren't really all the same operation any more, or even that closely related, so giving them the same name isn't really justified.

2opremio:
Another thing is that, since everything is running client side (which is great), there can be a disconnect with that's running in the server (e.g an old version of fluxctl) processing manifests differently.

I think we should try to connect to the cluster to confirm the server runs the same version and if not (or the cluster is not reachable) show a warning .

2opremio:
Actually, the connection the the cluster is already happening in the root command, I think. So, we will need to massage that.

squaremo:
> Actually, the connection the the cluster is already happening in the root command, I think

https://github.com/fluxcd/flux/pull/2659/files#diff-e066ae5c7a64b2277cb1b2dfe4503fc8R106

2opremio:
> > Actually, the connection the the cluster is already happening in the root command, I think
> 
> https://github.com/fluxcd/flux/pull/2659/files#diff-e066ae5c7a64b2277cb1b2dfe4503fc8R106

Ah, true, sorry I missed that

squaremo:
> I think we should try to connect to the cluster to confirm the server runs the same version and if not (or the cluster is not reachable) show a warning .

This is possible (there's a `v6/version` in the API), but I don't know if it would be much of a guide. The version has changed a fair bit without the manifest generation itself changing (so, false positives), and the version alone doesn't encompass all the things that might affect the generation, like the version of bundled tooling (false negatives too).

On the other hand, I do think it's an important consideration! This command is way less useful if you can't be fairly sure you'll get the same result in the cluster. How about if you can run the command with the exact container image you're running? (or even, possibly, in the same cluster -- without having to send the repo contents to it, ideally).

squaremo:
> How about if you can run the command with the exact container image you're running

The most direct way to accomplish this would be to make the flux container image work this way -- either by including another binary (fluxctl is not currently included), or adding it as a subcommand of fluxd.

kingdonb:
Closing. ("off the table")

PR #2667, opened 2019-12-09
Add Canva to production users

groodt:


hiddeco:
Thank you @groodt :100:

PR #2670, opened 2019-12-09
Update docs on annotations in HelmReleases

### LINKED ISSUE #2663: Invalid image ID on HelmRelease
### Describe the bug
Flux incorrectly updates `HelmRelease`s when the tag is specified separately from the image, resulting in an invalid image ID error. I think also there's been a regression of the fix for #1290, or it wasn't fixed as I expected from reading that issue description. 

### To Reproduce
1. Configure FluxCD and the Helm Operator.
2. Define a `HelmRelease` for a chart with an `imageTag` value, e.g. `stable/traefik` or `elastic/filebeat`. A sample `HelmRelease` is included below
3. Set Flux automation annotations, including to point Flux at the `imageTag` value
4. Watch Flux update the image reference to the latest tag (1.7.19 at the time of writing)
5. Watch further Helm deployments fail

#### Sample HelmRelease
```yaml
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: traefik
  namespace: dev
  annotations:
    fluxcd.io/automated: "true"
    fluxcd.io/tag.chart-image: semver:~1.7
    # Tell Flux where to find the tag value
    tag.fluxcd.io/traefik: imageTag
spec:
  releaseName: belligerent-dog
  chart:
    repository: https://kubernetes-charts.storage.googleapis.com
    name: traefik
    version: 1.79.0
  rollback:
    enable: true
  values:
    ## Default values for Traefik
    image: traefik
    imageTag: v1.7
```

**Expected behavior**
Flux overwrites the `imageTag` value with the new tag version. Instead, however, it overwrites the `image` value, appending the tag. This results in Helm attempting to deploy an image like `traefik:v1.7.19:v1.7`. Flux logs a message stating `expected image name as either <image>:<tag> or just <image>: invalid image ID`.

**Additional context**
- Flux version: 1.16
- Helm Operator version: 1.0.0-rc4
- Kubernetes version: 1.13 (GKE)
- Git provider: GitHub
- Container registry provider: Docker Hub

alastairs:
Clearly indicate that the `repository.fluxcd.io/<alias>` annotation is required, and mark the Filter section as deprecated: @2opremio [indicated](https://github.com/fluxcd/flux/issues/2663#issuecomment-562667083) that the `fluxcd.io/tag.chart-image` annotation is "a legacy and alternative way to supply a filter".

Fixes #2663.

### DIFF
--- a/docs/references/helm-operator-integration.md
+++ b/docs/references/helm-operator-integration.md
@@ -91,14 +91,14 @@ alias overrules a detected image.
 The following annotations are available, and `repository.fluxcd.io`
 is required for any of these to take effect.
 
-| Annotation                         |                  | Required? |
+| Annotation                         | Value            | Required? |
 |------------------------------------|------------------|   :---:   |
 | **`repository.fluxcd.io/<alias>`** | `sub.repo`       |     ✅    |
 | `registry.fluxcd.io/<alias>`       | `sub.reg`        |           |
 | `tag.fluxcd.io/<alias>`            | `sub.tag`        |           |
 | `filter.fluxcd.io/<alias>`         | `glob: master-*` |           |
 
-Two images specified in a `HelmRelease` as an example:
+The following example `HelmRelease` specifies two images:
 
 ```yaml
 metadata:
@@ -106,7 +106,7 @@ metadata:
     # image and tag
     repository.fluxcd.io/app: appImage
     tag.fluxcd.io/app: appTag
-    filter.tag/app: 'glob: *'
+    filter.fluxcd.io/app: 'glob: *'
     # nested image with registry and tag
     registry.fluxcd.io/submarine: sub.marinesystem.reg
     repository.fluxcd.io/submarine: sub.marinesystem.img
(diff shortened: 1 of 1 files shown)


docs/references/helm-operator-integration.md
```diff
@@ -125,7 +125,7 @@ spec:
         tag: version
 ```
 
-#### Filter
+#### Filters (deprecated)
```
hiddeco: Filters are not deprecated, the preferred annotation for it just changed from `fluxcd.io/tag.<alias>` to `filter.fluxcd.io/<alias>`.

alastairs: Does `chart-image` still have the same effect if supplied as the alias for `filter.fluxcd.io/<alias>`?

hiddeco: Yes, `chart-image` is a static alias for an auto detected top-level image but functions like any other alias.

alastairs: Ok, thanks @hiddeco. Updated :+1:

PR #2679, opened 2019-12-11
Be more generous with the unit test timeout

### LINKED ISSUE #2672: Unit and end-to-end tests are flakey
They fail from time to time. e.g. https://circleci.com/gh/fluxcd/flux/8957

@hiddeco @squaremo @stefanprodan please add more cases here (if you see them) so that I can fix them.

2opremio:
Since the timeout is per package and not per test, I set it to 5m instead of 1m

Addresses part of #2672


2opremio:
e2e tests failed, re-running

PR #2681, opened 2019-12-11
Run manifest image update unit tests in parallel

2opremio:
I previously disabled parallel tests in 7d9e02dac62b331c04915d540e6fd42e96e2c308 thinking it would fix the flaky unit tests in CircleCI.
    
Parallelization wasn't the culprit. On the way I also found out that tests were not really running in parallel due to `t.Parallel()` being called after the test is run.

Follow up from #2674 , #2672 and #1507


2opremio:
Unit tests now take ~1 minute ( https://circleci.com/gh/fluxcd/flux/9076 ) vs 1:20 in master ( https://circleci.com/gh/fluxcd/flux/9079 )

PR #2682, opened 2019-12-11
Run manifest policy update unit tests in parallel

2opremio:
This follows the same scheme as https://github.com/fluxcd/flux/pull/2681

### DIFF
--- a/pkg/cluster/kubernetes/policies_test.go
+++ b/pkg/cluster/kubernetes/policies_test.go
@@ -186,7 +186,7 @@ func TestUpdatePolicies(t *testing.T) {
 		},
 	} {
 		t.Run(c.name, func(t *testing.T) {
-			cLocal := c // Use copy to avoid races between parallel tests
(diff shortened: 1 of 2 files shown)


pkg/cluster/kubernetes/policies_test.go
```diff
@@ -186,13 +186,15 @@ func TestUpdatePolicies(t *testing.T) {
 		},
 	} {
 		t.Run(c.name, func(t *testing.T) {
-			caseIn := templToString(t, annotationsTemplate, c.in)
-			caseOut := templToString(t, annotationsTemplate, c.out)
+			cLocal := c // Use copy to avoid races between parallel tests
```
squaremo: It's not a race between tests you are worried about here, it's between each test and the thread in which the loop is run.

2opremio: Indeed, it's the loop which modifies the variable. That's what I actually meant, but I phrased it wrongly. Will fix (for #2681 as well).

PR #2684, opened 2019-12-11
Pseudo randomize push check tag

### LINKED ISSUE #2683: Multiple flux instances pointing to the same repo can fail writability check due to shared flux-write-check tag
**Describe the bug**

**To Reproduce**
0. Configure multiple flux instances pointing to the same github repo at different branches
1. When there is latency talking to github, errors are seeing in the logs and the changes to the helm charts cannot be applied.

**Expected behavior**
Multiple flux instances pointing to the same github repo at different branches shouldn't use the same flux-write-check tag to check writability since the github commands can fail when the tag unexpectedly points to a different commit id.

**Logs**
flux-write-check -> flux-write-check (cannot lock ref 'refs/tags/flux-write-check': is at 4dbd5cc088259f4e21c7847ca0ea3f2adf62d476 but expected 9043c404310bf76b963b8fb8ddc786e92f9f1c61)\nerror: failed to push some refs to 'git@github.ec.va.gov:EPMO/bip-pacman-config.git'\n"

**Additional context**
- Flux version: 1.15.0

hiddeco:
When multiple Flux instances are synchronizing with the same repository
but different branches, and the upstream has some latency, they get
confused about the write check they are performing due to the tag
being overwritten with a commit hash that does not belong to the branch
they are synchronzing with.

This commit pseudo randomizes the tag used to check if we can write to
the upstream repository and removes the `--force` options from the tag
and push commands so that the tag is never accidentally overwritten and
multiple instances no longer get in each other's way.

Fixes #2683

### DIFF
--- a/pkg/git/operations.go
+++ b/pkg/git/operations.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


2opremio:
Is the [e2e failure](https://circleci.com/gh/fluxcd/flux/9093) legitimate? It's hard to tell from the logs :S

pkg/git/operations.go
```diff
@@ -98,19 +99,26 @@ func add(ctx context.Context, workingDir, path string) error {
...
-	// --force just in case we fetched the tag from upstream when cloning
-	args := []string{"tag", "--force", CheckPushTag}
+	// we need to pseudo randomize the tag we use for the write check
+	// as multiple Flux instances can perform the check simultaneously
+	// for different branches, causing commit reference conflicts
+	b := make([]byte, 5)
+	if _, err := rand.Read(b); err != nil {
+		return err
+	}
+	pseudoRandPushTag := fmt.Sprintf("%s-%x", CheckPushTag, b)
```
2opremio: I think we should rename `CheckPushTag` to `CheckPushTagPrefix`

PR #2686, opened 2019-12-11
e2e: poll for sync tag update

hiddeco:
This ensures multiple attempts are made before bailing, as it can
take some time for Flux to push the tag after applying the change
to the cluster.

Addresses part of #2672


2opremio:
You beat me to it! I think #2687 is more robust, since it pulls the tags in every attempt.

hiddeco:
Yeah,

PR #2693, opened 2019-12-16
Release 1.17.0

2opremio:
Also, rephrase the `--sops` flag help.


hiddeco:
I think the versions defined in the manifests should also be bumped.

2opremio:
Thanks, I always miss that. Done!

2opremio:
> The origin branch (`2opremio:release/1.16.0`) is slightly confusing, but the contents of the PR seem okay to me and release ready.

Yeah, I used a wrong name for the branch but it shouldn't impact the release.

PR #2698, opened 2019-12-16
Remove Helm Operator from Flux chart

stefanprodan:
- update Flux to 1.17.0
- remove helm-op manifests and link to helm-op dedicated chart
- update helm-op install in get started docs (unblocks https://github.com/fluxcd/flux-get-started/pull/25)


stefanprodan:
After this is merged we should update the get started guides to use the `flux.io` CRD

stefanprodan:
We have to update the docs as well ref: https://github.com/fluxcd/flux-get-started/pull/25

test/e2e/lib/install.bash
```diff
@@ -22,12 +22,8 @@ function uninstall_tiller() {
...
 
 function install_flux_with_helm() {
-  local create_crds='true'
-  if kubectl get crd fluxhelmreleases.helm.integrations.flux.weave.works helmreleases.flux.weave.works > /dev/null 2>&1; then
-    # CRDs existed, don't try to create them
-    echo 'CRDs existed, setting helmOperator.createCRD=false'
-    create_crds='false'
-  fi
+
+  kubectl apply -f https://raw.githubusercontent.com/fluxcd/helm-operator/master/deploy/flux-helm-release-crd.yaml
```
hiddeco: I would point this to a tagged release as otherwise the tests can become flakey due to upstream changes.

Given that within Flux the CRD is only used to automate things, I think the maintenance is low (or can be automated if proven tiresome by e.g. fetching the latest published semver version).

stefanprodan: Done

2opremio:
Thanks @stefanprodan ! I will cut a helm-chart release. Would you mind updating the guides?

Also, maybe we should create a new documentation release (or point stable to master) since otherwise the `stable` documentation will show the old guides.

hiddeco:
@2opremio `stable` can not be pointed to master as far as I am aware, it is a feature from RTD which automatically picks the latest semver release.

stefanprodan:
@2opremio no need to cut a release, the chart gets published with a git tag and I did that.

PR #2701, opened 2019-12-17
Add Mintel to production users

nabadger:


nabadger:
Went production today with flux :)

PR #2715, opened 2019-12-28
Add context option to fluxctl

roffe:
I'm building some shell scripts for our developers that includes fluxctl and since we have multiple contexts in our kubectl config this option is a must for some of our usage cases where id like to run fluxctl against different clusters

### DIFF
--- a/cmd/fluxctl/args.go
+++ b/cmd/fluxctl/args.go
@@ -52,7 +52,18 @@ func getUserGitConfigValue(arg string) string {
 	return strings.Trim(res, "\x00")
 }
 
-var getKubeConfigContextNamespace = func(defaultNamespace string) string {
+// Wrapper for getKubeConfigContextNamespace where the return order is
+// 1. Given namespace
+// 2. Namespace from current kubeconfig context if 1. is empty
+// 3. Default namespace specified if 1 is empty and 2 fails
+func getKubeConfigContextNamespaceOrDefault(namespace string, defaultNamespace string, kubeConfigContext string) string {
+	if namespace == "" {
+		return getKubeConfigContextNamespace(defaultNamespace, kubeConfigContext)
+	}
+	return namespace
+}
+
+var getKubeConfigContextNamespace = func(defaultNamespace string, kubeConfigContext string) string {
 	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		clientcmd.NewDefaultClientConfigLoadingRules(),
 		&clientcmd.ConfigOverrides{},
@@ -61,7 +72,13 @@ var getKubeConfigContextNamespace = func(defaultNamespace string) string {
 		return defaultNamespace
 	}
 
-	cc := config.CurrentContext
+	var cc string
+	if kubeConfigContext == "" {
+		cc = config.CurrentContext
+	} else {
+		cc = kubeConfigContext
+	}
+
 	if c, ok := config.Contexts[cc]; ok && c.Namespace != "" {
 		return c.Namespace
 	}
--- a/cmd/fluxctl/root_cmd.go
+++ b/cmd/fluxctl/root_cmd.go
@@ -19,6 +19,7 @@ import (
 )
 
 type rootOpts struct {
+	Context   string
 	URL       string
 	Token     string
 	Namespace string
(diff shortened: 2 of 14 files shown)


cmd/fluxctl/root_cmd.go
```diff
@@ -70,6 +71,8 @@ func (opts *rootOpts) Command() *cobra.Command {
 		PersistentPreRunE: opts.PersistentPreRunE,
 	}
 
+	cmd.PersistentFlags().StringVar(&opts.Context, "context", "",
```
2opremio:
```suggestion
	cmd.PersistentFlags().StringVar(&opts.Context, "k8s-context", "",
```

cmd/fluxctl/root_cmd.go
```diff
@@ -70,6 +71,8 @@ func (opts *rootOpts) Command() *cobra.Command {
...
+		fmt.Sprint("The kubectl context to use, will use your current selected if not specified"))
```
2opremio:
```suggestion
		fmt.Sprint("The kubeconfig context to use, will use your current selected if not specified"))
```

docs/references/fluxctl.md
```diff
@@ -224,6 +224,7 @@ Available Commands:
   version        Output the version of fluxctl
 
 Flags:
+      --context string                  The kubectl context to use
```
2opremio:
```suggestion
      --context string                  The kubeconfig context to use
```

cmd/fluxctl/root_cmd.go
roffe: to follow kubectl semantics i suggest not adding `k8s-`

docs/references/fluxctl.md
roffe: to follow kubectl semantics i suggest not adding k8s-

cmd/fluxctl/root_cmd.go
2opremio: I think it's more important to make the flag consistent with the rest of the kubernetes flags (which are `k8s`-prefixed)

2opremio: Uhm, although we already have `--kube-config`. @stefanprodan / @hiddeco what do you think?

stefanprodan: I would keep the `k8s` prefix only for the port forward component, we used that prefix to avoid collision with the existing `namespace` flag but I see no reason to use that prefix for context.

hiddeco: I am on the fence because `--context` is very generic.

`--k8s-context` makes more sense from my perspective as an engineer and maintainer, as the flag controls not _Flux' context_ but the _Kubernetes context_ that should be used to perform an action with `fluxctl`, and while `--kube-config` may not be prefixed with `k8s`, the `kube` still classifies what it controls (and that it is not directly related to Flux).

From a user's perspective however, it would make more sense to just go for `--context` so that I would not have to think or get frustrated.

hiddeco: Helm offers the same functionality and uses `--kube-context`. 

```console
hidde@baldr:~$ helm get -h

...

Flags:
  -h, --help   help for get

...
      --kube-context string              name of the kubeconfig context to use
      --kubeconfig string                path to the kubeconfig file

```

This would be in line with our `--kube-config` flag.

roffe:
just get the ball rolling. potatoe or potata, the options are already cluttered

2opremio:
Just leave at `--context`

2opremio:
Can you squash your commits and rebase on master? Otherwise it LGTM

roffe:
@2opremio should be ok now

cmd/fluxctl/args.go
```diff
@@ -52,7 +52,18 @@ func getUserGitConfigValue(arg string) string {
 	return strings.Trim(res, "\x00")
 }
 
-var getKubeConfigContextNamespace = func(defaultNamespace string) string {
+// Wrapper for getKubeConfigContextNamespace where the return order is
+// 1. Given namespace
+// 2. Namespace from current kubectl context if 1. is empty
+// 3. Default namespace specified if 1 is empty and 2 fails
+func getKubeConfigContextNamespaceOrDefault(namespace string, defaultNamespace string, kubectlContext string) string {
```
2opremio:
```suggestion
func getKubeConfigContextNamespaceOrDefault(namespace string, defaultNamespace string, kubeConfigContext string) string {
```

cmd/fluxctl/args.go
```diff
@@ -52,7 +52,18 @@ func getUserGitConfigValue(arg string) string {
...
+	if namespace == "" {
+		return getKubeConfigContextNamespace(defaultNamespace, kubectlContext)
+	}
+	return namespace
+}
+
+var getKubeConfigContextNamespace = func(defaultNamespace string, kubectlContext string) string {
```
2opremio:
```suggestion
var getKubeConfigContextNamespace = func(defaultNamespace string, kubeConfigContext string) string {
```

2opremio:
I missed a couple of `kubectl` references. Happy to merge after that

PR #2719, opened 2020-01-02
Namespace whitelisting in helm chart without clusterRole

PaulFarver:
When not using a clusterRole, flux gets restricted to the namespace it is deployed in.

This pull request adds an option to let flux monitor and manipulate resources in other namespaces, without being deployed with a clusterrole

This is done by creating roles and rolebindings in all namespaces listed in a new value `allowedNamespaces`

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -202,10 +202,11 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `env.secretName`                                  | ``                                                   | Name of the secret that contains environment variables which should be defined in the Flux container (using `envFrom`)
 | `rbac.create`                                     | `true`                                               | If `true`, create and use RBAC resources
 | `rbac.pspEnabled`                                 | `false`                                              | If `true`, create and use a restricted pod security policy for Flux pod(s)
+| `allowedNamespaces`                               | `[]`                                                 | Allow flux to manage resources in the specified namespaces. The namespace flux is deployed in will always be included
 | `serviceAccount.create`                           | `true`                                               | If `true`, create a new service account
 | `serviceAccount.name`                             | `flux`                                               | Service account to be used
 | `serviceAccount.annotations`                      | ``                                                   | Additional Service Account annotations
-| `clusterRole.create`                              | `true`                                               | If `false`, Flux and the Helm Operator will be restricted to the namespace where they are deployed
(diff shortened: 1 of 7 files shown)


chart/flux/README.md
```diff
@@ -202,10 +202,11 @@ The following tables lists the configurable parameters of the Flux chart and the
...
+| `allowedNamespaces`                               | `[]`                                                 | Allow flux to manage resources in the specified namespaces. The namespace flux is deployed in will always be included
 | `serviceAccount.create`                           | `true`                                               | If `true`, create a new service account
 | `serviceAccount.name`                             | `flux`                                               | Service account to be used
 | `serviceAccount.annotations`                      | ``                                                   | Additional Service Account annotations
-| `clusterRole.create`                              | `true`                                               | If `false`, Flux and the Helm Operator will be restricted to the namespace where they are deployed
+| `clusterRole.create`                              | `true`                                               | If `false`, Flux and the Helm Operator will be restricted to the namespaces given in `allowedNamespaces` and the namespace where they are deployed
```
stefanprodan: Helm Operator is no longer deployed with this chart, please remove it from the readme and values.

Thanks

chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.17.0"
-version: 1.0.0
+version: 1.1.0
```
stefanprodan: Please undo this change, the version bump happens at release time.

PaulFarver: Right. Fixed now

chart/flux/README.md
PaulFarver: I updated the documentation as well as the comment in the values file

chart/flux/templates/deployment.yaml
```diff
@@ -175,7 +175,7 @@ spec:
           {{- end }}
           args:
           {{- if not .Values.clusterRole.create }}
-          - --k8s-allow-namespace={{ .Release.Namespace }}
+          - --k8s-allow-namespace={{ join "," (append .Values.allowedNamespaces .Release.Namespace) }}
```
2opremio: Uhm, this may result in `.Release.Namespace` being duplicated which may cause problems (or inefficiencies). Would you mind changing the type of `NewCluster()`'s `allowedNamespaces` argument and `Cluster.allowedNamespaces` from `[]string` to a set (`map[string]struct{}`)?

PaulFarver: Good catch. I didn't think of that 👍

Issue #2676, opened 2019-12-10
Add --sync-interval to Values.yaml


texasbobs:
Currently, the sync-interval is set to the value for --git-poll-interval.  We would like to be able to set the git-poll-interval to 1m and set the sync-interval to 5m.  This would allow us to force a sync and overlay any manual editing that might have been performed.

This would allow the output of the helm template command to be used without any manual editing.

### values.yaml
```
  # Period at which to poll git repo for new commits
  pollInterval: "1m"
  # Period at which it will apply what’s in git to the cluster absent new commits
  syncInterval: "5m"
```
### deployment.yaml
- --sync-interval={{ .Values.git.syncInterval }}

kingdonb:
I believe this has been closed through #2721. Thanks for using Flux!

PR #2721, opened 2020-01-03
Separate Git Poll Interval from Sync Interval

jessebye:
Allows for specifying different values for the git poll interval and sync interval. Previously, it used the `git.pollInterval` value for both. This behavior was fine if using a longer poll interval, but caused problems if setting the poll interval to a much shorter value (i.e., less than one minute).

By separating these values, the chart user can now use a short git poll interval (to quickly catch changes committed to git) while keeping the sync interval set to a longer interval, allowing Flux to operate normally without being overwhelmed by frequent syncs.

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -210,7 +210,8 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `service.type`                                    | `ClusterIP`                                          | Service type to be used (exposing the Flux API outside of the cluster is not advised)
 | `service.port`                                    | `3030`                                               | Service port to be used
 | `sync.state`                                      | `git`                                                | Where to keep sync state; either a tag in the upstream repo (`git`), or as an annotation on the SSH secret (`secret`)
-| `sync.timeout`                                    | `None`                                               |  Duration after which sync operations time out (defaults to `1m`)
+| `sync.timeout`                                    | `None`                                               | Duration after which sync operations time out (defaults to `1m`)
+| `sync.interval`                                   | `<git.pollInterval>`                                 | Controls how often Flux will apply what’s in git, to the cluster, absent new commits (defaults to `git.pollInterval`)
 | `git.url`                                         | `None`                                               | URL of git repo with Kubernetes manifests
 | `git.readonly`                                    | `false`                                              | If `true`, the git repo will be considered read-only, and Flux will not attempt to write to it
 | `git.branch`                                      | `master`                                             | Branch of git repo to use for Kubernetes manifests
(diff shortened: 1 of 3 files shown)


chart/flux/templates/deployment.yaml
```diff
@@ -210,7 +210,7 @@ spec:
           - --git-set-author={{ .Values.git.setAuthor }}
           - --git-poll-interval={{ .Values.git.pollInterval }}
           - --git-timeout={{ .Values.git.timeout }}
-          - --sync-interval={{ .Values.git.pollInterval }}
+          - --sync-interval={{ .Values.sync.interval }}
```
2opremio: I favor this change, but it breaks backwards compatibility. A safer solution would be to keep setting `--sync-interval` to `.Values.git.pollInterval` **unless** `.Values.sync.interval` is provided.

jessebye: @stefanprodan thoughts on this? I can make this change if you prefer that approach to the one I took. I don't have a strong preference either way.

2opremio:
Fixes #2676

chart/flux/templates/deployment.yaml
jessebye: @2opremio I've reworked this so it will default to `.Values.git.pollInterval` if `.Values.sync.interval` isn't provided.

jessebye: If all is good, what does the release cadence for helm charts look like?

chart/flux/README.md
```diff
@@ -209,7 +209,8 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `service.type`                                    | `ClusterIP`                                          | Service type to be used (exposing the Flux API outside of the cluster is not advised)
 | `service.port`                                    | `3030`                                               | Service port to be used
 | `sync.state`                                      | `git`                                                | Where to keep sync state; either a tag in the upstream repo (`git`), or as an annotation on the SSH secret (`secret`)
-| `sync.timeout`                                    | `None`                                               |  Duration after which sync operations time out (defaults to `1m`)
+| `sync.timeout`                                    | `None`                                               | Duration after which sync operations time out (defaults to `1m`)
+| `sync.interval`                                   | `<git.pollInterval>`                                 | Controls how often Flux will apply what’s in git, to the cluster, absent new commits (defaults to `5m`)
```
2opremio:
```suggestion
| `sync.interval`                                   | `<git.pollInterval>`                                 | Controls how often Flux will apply what’s in git, to the cluster, absent new commits (defaults to `git.pollInterval`)
```

2opremio:
Only a minor comment. Can you also rebase on top of master instead of merging master in?

Thanks

jessebye:
@2opremio not being familiar with rebasing, I think I probably screwed up the process. Should I open a new PR with just my changes on the latest from master to clean things up?

jessebye:
@2opremio or @stefanprodan I could really use this feature for something we are working on, and I think this PR is 99% there. Can I get some guidance how to get this across the finish line? 🙏

2opremio:
@jessebye sure, if you give me write access to your branch I will try to rebase It for you tomorrow

hiddeco:
@2opremio was already working on it.

@jessebye I think I have successfully cleaned up the commit history, thanks a lot for your contribution :tulip:

chart/flux/templates/deployment.yaml
2opremio: There will be a release soon (CC @stefanprodan )

PR #2664, opened 2019-12-06
Upgrade aws-sdk-go to support IRSA (IAM Roles for Service Accounts)

aackerman:
…eatures of AWS EKS

This [blog post](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) outlines the IAM Roles for Service Accounts feature for AWS EKS. The post also states the minimum version of `aws-sdk-go` that supports these features is `1.23.13`. These changes upgrade `aws-sdk-go` to the latest `1.25.48` to support those features.

To make the changes I did the following and everything reported fine on macOS.

```
go get -u github.com/aws/aws-sdk-go
make
make test
```

Please bear with me if this isn't the expected approach, I don't have a lot of experience with go.

PR #2722, opened 2020-01-06
Update aws-sdk-go to v1.27.0

stefanprodan:
Update to the latest SDK version as it seems that #2664 broke the ECR auth.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,10 @@ go 1.13
 // remove when https://github.com/docker/distribution/pull/2905 is released.
 replace github.com/docker/distribution => github.com/2opremio/distribution v0.0.0-20190419185413-6c9727e5e5de
 
+// fix go-autorest ambiguous import caused by sops
+// sops needs to update their deps ref: https://github.com/kubernetes/client-go/issues/628
+replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
+
 require (
 	github.com/Jeffail/gabs v1.4.0
 	github.com/Masterminds/semver v1.4.2
@@ -49,6 +53,3 @@ require (
 	k8s.io/helm v2.13.1+incompatible
 	k8s.io/klog v0.3.3
 )
-
-// fix sops/v3/azkv ambiguous import
-replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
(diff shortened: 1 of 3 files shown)


stefanprodan:
The update didn't fixed it, after adding logging this is the error we get (kiam):

```
2020-01-06T14:00:50.375362066Z ts=2020-01-06T14:00:50.374622922Z caller=aws.go:124 component=aws error="fetching region for AWS" err="EC2MetadataRequestError: failed to get EC2 instance identity document\ncaused by: EC2MetadataError: failed to make EC2Metadata request\n\tstatus code: 404, request id: \ncaused by: request blocked by whitelist-route-regexp \"/latest/meta-data/placement/availability-zone\": /latest/dynamic/instance-identity/document\n"
```

2opremio:
Then I think we should roll back #2664

2opremio:
@stefanprodan can you give that a try?

stefanprodan:
Simon confirmed that using the old SDK, Flux works with our kiam setup. I think we should find how to use the SDK not rollback to a deprecated version.

stefanprodan:
@aackerman are you using IAM Roles for Service Accounts with Flux 1.17 can you confirm that it works?

aackerman:
@stefanprodan yes the IAM Roles for Service Accounts features are working for me using the `docker.io/fluxcd/flux:1.17.0` image.

stefanprodan:
@2opremio others are reporting that Flux 1.17 works ok with Kiam, my guess is that we're using an older version of Kiam in Weave Cloud that's not compatible with the newer SDK. I think we should merge this PR and deal with Kiam in our cluster.

stefanprodan:
@2opremio we have a fix, in order to allow Flux to discover the AWS region, Kiam has to be configured with:
```
--whitelist-route-regexp=/latest/meta-data/placement/availability-zone
```

2opremio:
Awesome!!

go.mod
```diff
@@ -50,3 +49,6 @@ require (
 	k8s.io/helm v2.13.1+incompatible
 	k8s.io/klog v0.3.3
 )
+
+// fix sops/v3/azkv ambiguous import
+replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
```
2opremio: Can you move this with the other replace directive?

PR #2731, opened 2020-01-09
Update packages to Kubernetes 1.16

stefanprodan:
- update Kubernetes client-go to 1.16 and auxiliary packages
- update old clientsets for Kubernetes 1.16
- update Masterminds semver to v3.0.3
- update Helm Operator to RC.6
- pin packages to break free of `weaveworks/flux` circular dependency


go.mod
```diff
@@ -9,47 +9,63 @@ replace github.com/docker/distribution => github.com/2opremio/distribution v0.0.
 // sops needs to update their deps ref: https://github.com/kubernetes/client-go/issues/628
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
 
+// transitive requirement from Helm Operator (to be removed on helm-op 1.0.0 GA)
+replace (
+	github.com/docker/docker => github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0
```
hiddeco: As long as Helm does not drop this replace, we will never be able to drop it ourselves, as far as I am aware.

stefanprodan: I can change it to `to be re-evaluated on helm-op 1.0.0 GA`?

go.mod
```diff
@@ -9,47 +9,63 @@ replace github.com/docker/distribution => github.com/2opremio/distribution v0.0.
 // sops needs to update their deps ref: https://github.com/kubernetes/client-go/issues/628
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
 
+// transitive requirement from Helm Operator (to be re-evaluated on helm-op 1.0.0 GA)
+replace (
+	github.com/docker/docker => github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0
+	github.com/fluxcd/helm-operator => github.com/fluxcd/helm-operator v1.0.0-rc6
+	k8s.io/client-go => k8s.io/client-go v0.0.0-20191016111102-bec269661e48 // kubernetes-1.16.2
```
2opremio: Do we really need this?

2opremio: (why not simply use that version directly?)

stefanprodan: Is it a rhetorical question? :))

2opremio: hahaha, no, see https://github.com/fluxcd/flux/pull/2590

stefanprodan: Using the version directly doesn't remove the the RC.1 dep that brings Flux 1.15 that brings weaveworks/flux. This is just a first step, 2nd is https://github.com/fluxcd/helm-operator/pull/192

2opremio: Ah, I see

PR #2736, opened 2020-01-11
Add connect option to connect to your own upstream service

chainlink:
This adds the ability to set your own upstream service (ie. `Fluxcloud` or other) when installing the helm chart

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -252,6 +252,9 @@ spec:
           {{- if .Values.token }}
           - --connect=wss://cloud.weave.works/api/flux
           - --token={{ .Values.token }}
+          {{- if .Values.connect }}
+          - --connect={{ .Values.connect }}
+          {{- end }}
           {{- end }}
           {{- if and .Values.syncGarbageCollection.enabled (not .Values.syncGarbageCollection.dry) }}
           - --sync-garbage-collection={{ .Values.syncGarbageCollection.enabled }}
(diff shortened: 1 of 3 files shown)


edshelton:
This is great - I'm currently trying to set up Fluxcloud with our Flux agent which is deployed via Helm, but without this there's no nice way of setting the `--connect` flag for Fluxcloud to work.

🙏

stefanprodan:
@edshelton we provide the additional args param so you can use Fluxcloud, here it is https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L216-L217

edshelton:
> @edshelton we provide the additional args param so you can use Fluxcloud, here it is https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L216-L217

Oh, didn't see that. Thanks. I was looking for `extraArgs` which is what a lot of other charts use. It's also not in the README 🙂

chainlink:
I also missed that and agree `additionalArgs` should be surfaced in the readme. 

In general, I think `connect` is a fairly common option and should potentially be surfaced on it's own.

chart/flux/values.yaml
```diff
@@ -3,6 +3,8 @@
 # Weave Cloud service token
 token: ""
 
+connect: ""
```
hiddeco: Can you add a comment about what this can be used for.

chart/flux/templates/deployment.yaml
```diff
@@ -252,6 +252,9 @@ spec:
           {{- if .Values.token }}
```
hiddeco: I do not think this should be nested, can you think of a way that both values can be used (`token` / `connect`), while still making it possible to connect to the `cloud.weave.works` endpoints when only a token is given for backwards compatibility?

kingdonb:

PR #2741, opened 2020-01-13
Honor KUBECONFIG env variable in fluxd

2opremio:
Also, delegate the creation of incluster configuration to the Kubernetes library.

As a result, we are not making use of the `--in-cluster` argument anymore. Both the `--in-cluster` and `--kube-config` arguments have been deprecated (you can use the `KUBECONFIG` env variable instead).

Overrides #2621

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -71,6 +71,8 @@ const (
 	defaultGitSkipMessage = "\n\n[ci skip]"
 
 	RequireECR = "ecr"
+
+	k8sInClusterSecretsBaseDir = "/var/run/secrets/kubernetes.io"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cmd/fluxd/main.go
```diff
@@ -447,8 +450,10 @@ func main() {
 		}
 		clusterVersion = "kubernetes-" + serverVersion.GitVersion
 
-		if *k8sInCluster && !httpGitURL {
-			namespace, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		fileinfo, err := os.Stat(k8sInClusterSecretsBaseDir)
+		isInCluster := err != nil && fileinfo.IsDir()
```
tonyffrench: I think you meant `isInCluster := err == nil && fileinfo.IsDir()` ... curious how a change so broken could make it to master ?

2opremio: Because you didn't take part in the review :)

Now seriously, my bad. This did t get caught by the end to end tests because we generate our own ssh key there.. we should do that.

That's for reporting it and sorry for the inconvenience.

PR #2745, opened 2020-01-15
Disable Image Scanning with a flag [take 2]

2opremio:
Overrides #2524 

I made sure that the memcached client is not initialized when scanning is disabled. Also, I added proper errors when querying the registry if scanning is disabled.


pkg/update/release_image.go
```diff
@@ -196,6 +196,9 @@ func (s ReleaseImageSpec) calculateImageUpdates(rc ReleaseContext, candidates []
 		ref, err = s.ImageSpec.AsRef()
 		if err == nil {
 			singleRepo = ref.CanonicalName()
+			// FIXME(fons): we probably want to allow this operation even if image
+			//              scanning is disabled. We could either avoid the validation
+			//              or use an uncached registry.
```
hiddeco: The uncached registry is not an option, Michael shared insights about this in another issue (Flux had this in the beginning but it lead to timeouts).

stefanprodan:
If read-only is enabled then I think it should also disable image scanning as for a read-only Flux it doesn't make sense to maintain a a cache of images or depend on memcached. Furthermore I would make the read-only flag in the fluxctl install command omit the memcached deployment from the manifests output.

pkg/update/release_image.go
2opremio: That code only validates that the images exist, I think we can use an uncached registry for that. Otherwise releasing a specific image won't work.

2opremio:
@stefanprodan note that ListImagss doesn't work when disabling image scanning but it works in read-only. There may be other things on top of that that I haven't thought of.

2opremio:
I changed `--disable-image-scan` to `--registry-scanning`. Also, `--registry-scanning=false` implies `--read-only=false`

docs/references/daemon.md
```diff
@@ -57,7 +57,7 @@ Version controlling of cluster manifests provides reproducibility and a historic
 | --git-notes-ref                                  | `flux`                   | ref to use for keeping commit annotations in git notes
 | --git-poll-interval                              | `5m`                     | period at which to fetch any new commits from the git repo
 | --git-timeout                                    | `20s`                    | duration after which git operations time out
-| --git-readonly                                   | `false`                  | If `true`, the git repo will be considered read-only, and Flux will not attempt to write to it. Implies --sync-state=secret
+| --git-readonly                                   | `false`                  | If `true`, the git repo will be considered read-only, and Flux will not attempt to write to it. Implies --sync-state=secret and and --registry-scanning=false
```
stefanprodan:
```suggestion
| --git-readonly                                   | `false`                  | If `true`, the git repo will be considered read-only, and Flux will not attempt to write to it. Implies --sync-state=secret and --registry-scanning=false
```

2opremio:
PTAL, I added `--registry-scanning` to `fluxctl install` and made sure that it doesn't output the memcached manifests when using `--registry-scanning=false` or `--git-readonly=true`

pkg/install/templates/flux-deployment.yaml.tmpl
```diff
@@ -184,6 +184,9 @@ spec:
         # Serve /metrics endpoint at different port;
         # make sure to set prometheus' annotation to scrape the port value.
         - --listen-metrics=:3031
+{{- if not .RegistryScanning }}
+        - --registry-scanning=fale
```
stefanprodan:
```suggestion
        - --registry-scanning=false
```

2opremio: Oh dear. Good catch

squaremo:
> If read-only is enabled then I think it should also disable image scanning as for a read-only Flux it doesn't make sense to maintain a a cache of images or depend on memcached

Those two things are related, but not coupled. You may very well want to query the image metadata while not wanting to write to the git repo.

> Also, `--registry-scanning=false` implies `--read-only=false`

What, turning registry scanning _off_ forces it to treat the repo as _writable_? That doesn't make any sense.

2opremio:
> > Also, `--registry-scanning=false` implies `--read-only=false`
> 
> What, turning registry scanning _off_ forces it to treat the repo as _writable_? That doesn't make any sense.

I meant `--read-only=true` implies `--registry-scanning=false` (that's what the code does)

> Those two things are related, but not coupled. You may very well want to query the image metadata while not wanting to write to the git repo.

I was on the fence about this one, @stefanprodan convinced me about not making much sense to have the scanning just for `list-images`. Maybe we are wrong ....

2opremio:
We can still revert that, nothing was released yet.

cmd/fluxctl/install_cmd.go
```diff
@@ -33,21 +33,23 @@ fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' | ku
 	cmd.Flags().StringVar(&opts.GitBranch, "git-branch", "master",
 		"Git branch to be used by Flux")
 	cmd.Flags().StringSliceVar(&opts.GitPaths, "git-paths", []string{},
-		"Relative paths within the Git repo for Flux to locate Kubernetes manifests")
+		"relative paths within the Git repo for Flux to locate Kubernetes manifests")
```
squaremo: How come these all got lowercased, _en passant_ .. looking at the other subcommands, the usage messages tend to start with capitals

2opremio: Ops, I did it after `fluxd`'s flags (wrongly assuming that the rest of the `fluxctl` commands followed the as pattern)

2opremio: I am happy to revert that, but I think `fluxd` and `fluxctl` should be consistent

squaremo: I noticed it because it caused a merge conflict (easily solved, I just need to know which direction to solve it in ..)

Good point on being consistent with fluxd; I'll resolve the conflict with the least change, then do them all at once in another PR. (i.e., don't worry about reverting!)

2opremio: Great, thanks!

texasbobs:
When you mention `--read-only=false` are you referring to`--git-readonly` or is this a new parm?

We would have `--git-readonly-false` but would not want any access attempts to memcached.

2opremio:
> We would have `--git-readonly-false` but would not want any access attempts to memcached.

That's what the flag added in this PR does.

> When you mention `--read-only=false` are you referring to`--git-readonly` or is this a new parm?

We were referring to `--git-readonly|

texasbobs:
> > We would have `--git-readonly-false` but would not want any access attempts to memcached.
> 
> That's what the flag added in this PR does.

Ok. Thanks.  I guess I wasn't reading it that way.  We would like the following:

`--registry-scanning=false` and `--git-readonly=false`

But the comments seem to me to indicate if `--registry-scanning=false` then `--git-readonly=true` should be used.

```
registryScanning     = fs.Bool("registry-scanning", true, "scan container image registries to fill in the registry cache; --registry-scanning=false implies --read-only=true")
```



`

2opremio:
That's really interesting. Thre is a while discussion about that at #2793 

Would you mind explaining there why would you need  `--registry-scanning=false` and `--git-readonly=false`?

`--registry-scanning=false` disables a lot of functionality also disabled implicitly by `--git-readonly=true`

texasbobs:
> That's really interesting. Thre is a while discussion about that at #2793
> 
> Would you mind explaining there why would you need `--registry-scanning=false` and `--git-readonly=false`?
> 
> `--registry-scanning=false` disables a lot of functionality also disabled implicitly by `--git-readonly=true`

Perhaps we do not understand the functions but here is what we are thinking.  We want to use `--git-readonly=false` so that Flux can keep track of what has changed in the repo and what it has already applied.  We want to use `--registry-scanning=false`  because we are not using the feature pertaining to image versions.

So perhaps we don't understand what the `--git-readonly` is for. Is the only time Flux writes back to the git repo when it needs to update an image tag?

Based on the following, we thought that Flux needed to write back to the repo to keep track of what it had applied.

  # Set to `true` if you intend for Flux to not be able to push changes to git.
  # Also configure state.mode to `secret` since storing state in a git tag will no longer be possible.
  readonly: false
  # use `.sync.state: secret` to store flux's state as an annotation on the secret (instead of a git tag)
  state: git

2opremio:
> So perhaps we don't understand what the `--git-readonly` is for. Is the only time Flux writes back to the git repo when it needs to update an image tag?

`--git-readonly` prevents Flux from writing back to the Git repository. This happens when updating images, when updating policies (annotation metadata) and when pushing a tag (specified through `--git-label`).

> We want to use --git-readonly=false so that Flux can keep track of what has changed in the repo and what it has already applied. 

So, what you mean is that you what `--git-readonly=false` in order to keep letting Flux use an git tag to keep track of what it synced?

texasbobs:
> So, what you mean is that you what `--git-readonly=false` in order to keep letting Flux use an git tag to keep track of what it synced?

That is correct.  From what I can tell in the doc, that is the purpose of the flag.  I see that it would also control the updates for new images, but it seems you could have readonly=false and not apply new images.
```
Flux keeps track of the last commit that it’s applied to the cluster, by pushing a tag (controlled by the command-line flags --git-sync-tag and --git-label) to the git repository. This gives it a persistent high water mark, so even if it is restarted from scratch, it will be able to tell where it got to.
```

texasbobs:
Looking at #2790, I would think that `git-readonly=true implies registry-scanning=false` is right. Why scan registry entries if you can't update the repo. However 
`registry-scanning=false does not imply git-readonly=true` because you still might want to use the sync tag option.

squaremo:
> Why scan registry entries if you can't update the repo.

Because there are API methods for querying workloads and the images they use, and those depend on having scanned registry entries, and have nothing to do with whether you can write to git or not.

texasbobs:
> > Why scan registry entries if you can't update the repo.
> 
> Because there are API methods for querying workloads and the images they use, and those depend on having scanned registry entries, and have nothing to do with whether you can write to git or not.

So in that case, the two parms are totally unrelated.

2opremio:
> I would think that `git-readonly=true implies registry-scanning=false` is right. Why scan registry entries if you can't update the repo. However
> `registry-scanning=false does not imply git-readonly=true` because you still might want to use the sync tag option.

@texasbobs note that (regardless of what the help messages say, they are wrong right now) as it is in `master` right now `--git-readonly=true` implies `--registry-scanning=false` and not the other way around. So, you could still have `--registry-scanning=false` and `--git-readonly=false` and your usecase would be covered.

PR #2748, opened 2020-01-15
Make docker/image-tag work with multiple version tags

2opremio:
It turns out that the release process for, say 1.19.1
adds an extra tag v1.19.1, that made docker/image-tag
simply print both tags.

Also, I removed code related to the Helm Operator.

### DIFF
--- a/docker/image-tag
+++ b/docker/image-tag
@@ -10,15 +10,28 @@ if [ "${1:-}" = '--show-diff' ]; then
 fi
 
 # If a tagged version, just print that tag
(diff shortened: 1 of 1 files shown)


stefanprodan:
@2opremio have you consider this https://github.com/fluxcd/helm-operator/blob/master/docker/image-tag#L16

2opremio:
That code silently succeeds if two tags point to HEAD, I think the code in this PR is more robust (although it doesn't validate that the tags are semvers)

2opremio:
Is it OK to merge then? CC @hiddeco

Issue #2642, opened 2019-11-29
Make cluster resource exclusion configurable


2opremio:
**Describe the feature**

We already exclude metrics resources (see https://github.com/fluxcd/flux/pull/2606 , however, that is hardcoded, and (as seen in #2633 ) it may be useful to exclude additional resources.

**What would the new user story look like?**

Users are provided with a new `fluxd` flag, say `--k8s-exclude-resource=apigroup/version/kind`.

Users can supply:
* Fixed values like `--k8s-exclude-resource=coordination.k8s.io/v1/Lease` 
* Multiple values like `--k8s-exclude-resource=coordination.k8s.io/v1beta1/Lease,coordination.k8s.io/v1/Lease` 
* Glob experssions like `--k8s-exclude-resource=coordination.k8s.io/*/Lease` , `--k8s-exclude-resource=authrequests.dex.coreos.com/*`

**Expected behavior**

Flux won't attempt to obtain resources of that kind from the cluster.

2opremio:
I am now thinking that this could probably achieved thorough simple RBAC rules.

primeroz:
The RBAC way make sense, I will give it a try as soon as i can but as long as it does not error out in the logs that is as good as a filter in the args to me.

Might be a bit harder to handle in terms of new users and documentation though ? 

It would be easier to document something 
* `RBAC allows pulling everything, filter args obviously remove stuff from flux downloading ( and applying ? ) resources`
rather than 
* `This is a base RBAC, extend if you need more resources to be handled by flux`
   assuming that the default example rbac would be somewhat limited and operators would have to extend it on their own .. but as long as we get 403 in the logs and errors in the metrics it should be doable.

alanjcastonguay:
Nice; some resources that Flux isn’t expected to be manipulating (eg Pods and ReplicaSets) can be excluded for the common-case. That might make for a good example in the docs.

Issue #2554, opened 2019-10-25
couldn't get resource list for webhook.certmanager.k8s.io/v1beta1


aungzy:
**Describe the bug**
After installing flux in OpenShift 4.2 cluster, the operator was able to refresh git repository, but not able to sync resources.
I am seeing the following 2 error messages:

type="internal kubernetes error" 
kubernetes_caller=k8s.io/client-go@v11.0.0+incompatible/discovery/cached/memory/memcache.go:111 
err="couldn't get resource list for webhook.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request"

component=sync-loop 
err="collating resources in cluster for sync: unable to retrieve the complete list of server APIs: webhook.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request"

**To Reproduce**
Steps to reproduce the behaviour:
0. What's your setup? As per https://docs.fluxcd.io/en/latest/tutorials/get-started.html

**Expected behavior**
Flux should be able to successfully run `kubectl apply` command to sync resources.

**Additional context**
- Flux version: 1.15.0
- Kubernetes version: 1.14 (with OpenShift 4.2)
- Git provider: Gitlab CE

tomacMartin:
I had the same issue. Upgrading cert-manager solved the issue for me. 

Additional context

Flux version: 1.15.0
Kubernetes version: 1.16 (Rancher)
Git provider: Gitlab CE

ambis:
Same problem here. Started out of the blue as per the logs (2019-11-15 midnight). No changes in flux source repo, none of the cluster components other than my app itself is automated for updates.

Running k3s v0.10.2, cert manager 0.11.0 and flux 1.15.0.

I deleted cert-manager completely with delete -f and flux reapplied it (now that it started working), and boom, here comes the error again.

cmanzi:
Can confirm I'm seeing this with Flux 1.17.0 as well. Both cert-manager and metrics-server API services. Restarting the Flux pod seems to get fix it.

2opremio:
I will make sure we exclude `webhook.certmanager.k8s.io` resources by default, and make the excluded resources configurable as per https://github.com/fluxcd/flux/issues/2642

2opremio:
Note that this is a problem with Kubernetes or the configuration of Kubernetes itself, which fails to answer List requests for `webhook.certmanager.k8s.io/v1beta1`

aungzy:
Thanks @2opremio !

[...]

PR #2749, opened 2020-01-15
Make Kubernetes resources exclusion configurable

stefanprodan:
This PR adds the `--k8s-exclude-resource` to fluxd command args to make the Kubernetes resources exclusion configurable. 

By default the `*metrics.k8s.io/*`, `webhook.certmanager.k8s.io/*` and `v1/Event` resources are excluded.

Also objects with owner references are excluded from Flux discovery to avoid conflicts between Flux GC and Kubernetes GC.

Fix: #2554
Fix: #2642

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -178,8 +178,8 @@ func main() {
 		k8sNamespaceWhitelist = fs.StringSlice("k8s-namespace-whitelist", []string{}, "restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set")
 		k8sAllowNamespace     = fs.StringSlice("k8s-allow-namespace", []string{}, "restrict all operations to the provided namespaces")
 		k8sDefaultNamespace   = fs.String("k8s-default-namespace", "", "the namespace to use for resources where a namespace is not specified")
-
-		k8sVerbosity = fs.Int("k8s-verbosity", 0, "klog verbosity level")
+		k8sExcludeResource    = fs.StringSlice("k8s-exclude-resource", []string{"*metrics.k8s.io/*", "webhook.certmanager.k8s.io/*", "v1/Event"}, "do not attempt to obtain cluster resources whose group/version/kind matches these glob expressions")
+		k8sVerbosity          = fs.Int("k8s-verbosity", 0, "klog verbosity level")
 
 		// SSH key generation
 		sshKeyBits   = optionalVar(fs, &ssh.KeyBitsValue{}, "ssh-keygen-bits", "-b argument to ssh-keygen (default unspecified)")
@@ -505,7 +505,7 @@ func main() {
 		for _, n := range append(*k8sNamespaceWhitelist, *k8sAllowNamespace...) {
 			allowedNamespaces[n] = struct{}{}
 		}
-		k8sInst := kubernetes.NewCluster(client, kubectlApplier, sshKeyRing, logger, allowedNamespaces, *registryExcludeImage)
+		k8sInst := kubernetes.NewCluster(client, kubectlApplier, sshKeyRing, logger, allowedNamespaces, *registryExcludeImage, *k8sExcludeResource)
 		k8sInst.GC = *syncGC
 		k8sInst.DryGC = *dryGC
 
--- a/pkg/cluster/kubernetes/sync.go
+++ b/pkg/cluster/kubernetes/sync.go
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
+	"github.com/ryanuber/go-glob"
 	"io"
 	"os/exec"
 	"sort"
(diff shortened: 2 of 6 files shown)


pkg/cluster/kubernetes/sync.go
```diff
@@ -215,8 +215,16 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
 	resources := []*meta_v1.APIResourceList{}
 	for i := range sgs.Groups {
 		gv := sgs.Groups[i].PreferredVersion.GroupVersion
-		// exclude the *.metrics.k8s.io resources to avoid querying the cluster metrics
-		if sgs.Groups[i].Name != "metrics.k8s.io" && !strings.HasSuffix(sgs.Groups[i].Name, ".metrics.k8s.io") {
+
+		excluded := false
+		for _, exg := range c.groupExcludeList {
+			if strings.HasSuffix(sgs.Groups[i].Name, exg) {
```
hiddeco: Wonder if suffix matches are sufficient for all use-cases, or that the next enhancement request is going to be glob matches.

stefanprodan: For groups I think suffix matches are covering all use-cases but if we want to exclude Kubernetes Kinds then glob should be used.

2opremio: I had made a more flexible design a while ago, see https://github.com/fluxcd/flux/issues/2642

Yeah, I think we should use globing instead.

stefanprodan: That issue refers to kinds not groups, this is a different feature from what you are describing in the issue. I see no reason why would you need blogling when groups follow the DNS format, if you want to exclude all certmanager's groups you would add `certmanager.k8s.io`.

2opremio: Yes, it contains all groups versions and kinds, it's a superset of this PR. I think we should implement that since it's more flexible and the approach in this PR is not extensible (making it hard to add kinds later on).

I am happy to implement it myself if you don't feel like updating the PR.

stefanprodan: Ok I've changed to traget resources with glob expressions.

cmd/fluxd/main.go
```diff
@@ -178,8 +178,8 @@ func main() {
 		k8sNamespaceWhitelist = fs.StringSlice("k8s-namespace-whitelist", []string{}, "restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set")
 		k8sAllowNamespace     = fs.StringSlice("k8s-allow-namespace", []string{}, "restrict all operations to the provided namespaces")
 		k8sDefaultNamespace   = fs.String("k8s-default-namespace", "", "the namespace to use for resources where a namespace is not specified")
-
-		k8sVerbosity = fs.Int("k8s-verbosity", 0, "klog verbosity level")
+		k8sExcludeResource    = fs.StringSlice("k8s-exclude-resource", []string{"*metrics.k8s.io*", "*webhook.certmanager.k8s.io*"}, "do not include in Flux discovery the Kubernetes resources that match these glob expressions")
```
2opremio: Can you elaborate on the format to use?

pkg/cluster/kubernetes/sync.go
2opremio: Only GroupVersions are targeted. So we should either include Kinds (and not only use it for discovery) or rename the argument.

cmd/fluxd/main.go
2opremio: Also, the description may be confusing for the end user (since she doesn't know how discovery is used internally).

2opremio:
```suggestion
		k8sExcludeResources    = fs.StringSlice("k8s-exclude-resource", []string{"*metrics.k8s.io*", "*webhook.certmanager.k8s.io*"}, "do not attempt to obtain cluster resources whose group/version that matches these glob expressions")
```

pkg/cluster/kubernetes/sync.go
2opremio: Actually, I am on the fence about this. `--k8s-exclude-resources` is probably fine.

stefanprodan:
@2opremio please take a look again, the filtering happens at group level first (to avoid API calls errors when fetching the resources) and a 2nd time for the Kind matching. Also if you have a better description for the flag please write it here. 

PS. I've also excluded the Event kind as the events could fill Flux memory.

2opremio:
> please take a look again, the filtering happens at group level first (to avoid API calls errors when fetching the resources) and a 2nd time for the Kind matching. Also if you have a better description for the flag please write it here.

Ah, sorry, I had missed that (I was checking the PR on the phone)

2opremio:
How about

```
do not attempt to obtain cluster resources whose group/version/kind matches these glob expressions, e.g. coordination.k8s.io/v1beta1/Lease, coordination.k8s.io/*/Lease or coordination.k8s.io/*
```

pkg/cluster/kubernetes/sync.go
```diff
@@ -215,11 +239,19 @@ func (c *Cluster) getAllowedResourcesBySelector(selector string) (map[string]*ku
 	resources := []*meta_v1.APIResourceList{}
 	for i := range sgs.Groups {
 		gv := sgs.Groups[i].PreferredVersion.GroupVersion
-		// exclude the *.metrics.k8s.io resources to avoid querying the cluster metrics
-		if sgs.Groups[i].Name != "metrics.k8s.io" && !strings.HasSuffix(sgs.Groups[i].Name, ".metrics.k8s.io") {
+
+		excluded := false
+		for _, exp := range c.resourceExcludeList {
+			if glob.Glob(exp, gv) {
```
2opremio: Since the expression targets group/version/kind we should be globbing against `group/version/` otherwise `group/version/*` will be missed

stefanprodan: Done

cmd/fluxd/main.go
```diff
@@ -178,8 +178,8 @@ func main() {
 		k8sNamespaceWhitelist = fs.StringSlice("k8s-namespace-whitelist", []string{}, "restrict the view of the cluster to the namespaces listed. All namespaces are included if this is not set")
 		k8sAllowNamespace     = fs.StringSlice("k8s-allow-namespace", []string{}, "restrict all operations to the provided namespaces")
 		k8sDefaultNamespace   = fs.String("k8s-default-namespace", "", "the namespace to use for resources where a namespace is not specified")
-
-		k8sVerbosity = fs.Int("k8s-verbosity", 0, "klog verbosity level")
+		k8sExcludeResource    = fs.StringSlice("k8s-exclude-resource", []string{"*metrics.k8s.io/*", "webhook.certmanager.k8s.io/*", "v1/Event"}, "do not attempt to obtain cluster resources whose group/version/kind matches these glob expressions")
```
2opremio: This should be added to the daemon docs

stefanprodan: Done

PR #2750, opened 2020-01-16
build: Stop publishing Flux chart to gh-pages

stefanprodan:
The Flux chart repo hosted under Flux GitHub repo (gh-pages branch) has been deprecated, removing it from the CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,19 +29,6 @@ jobs:
             git config --global user.email fluxcdbot@users.noreply.github.com
             git config --global user.name fluxcdbot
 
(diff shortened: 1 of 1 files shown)


2opremio:
I arrived late, but https://github.com/fluxcd/flux/blob/master/chart/README.md  should had been updated as well to reflect this.

stefanprodan:
@2opremio the chart readme has been updated long time ago, we use `https://charts.fluxcd.io` as our chart repo since last year.

2opremio:
Well, it reads _We use GitHub pages to publish the chart as a versioned package_ , we should probably specify the page instead

Issue #2699, opened 2019-12-17
Created Timestamp issue with gcr.io repo not detected


chainlink:
**Describe the bug**
Flux will not detect Created timestamp for gcr.io repo.

Seeing this error in the logs:
```
ts=2019-12-17T03:05:02.232839857Z caller=repocachemanager.go:260 component=warmer canonical_name=gcr.io/my/registry auth={map[]} err="failed to parse 1 timestamp label(s) as RFC3339 (org.label-schema.build-date); ask the repository administrator to correct this as it conflicts with the spec" ref=gcr.io/my-registry:a6fb309227fe9f2354cf56eb72e10f60271ceab8
```

But if I run docker inspect on that image I get:
```
...
"Created": "2019-12-17T02:43:52.417086026Z"
...
```

If I run `list-images` the `CREATED` column is blank on those images and I see `15 images omitted`

Flux version: 1.16.0

I'm currently not using any labels in my docker containers (having trouble adding in build-date label via cloudbuild) I have two other kube installs running flux pulling from gcr.io with no issue (on older flux versions)

hiddeco:
The error is a warning, and may also be logged while it falls back to the `Created` timestamp. From what you are reporting, and to be precise 

> the `CREATED` column is blank on those images

this fallback seems however not to happen, or the wrong thing is parsed.

Can you share the `Config.Labels` output from `docker inspect <image>` with me? Plus the `fluxctl` version you are using (`fluxctl version`)?

chainlink:
Found it! Thanks for the prompt (I should have checked the `labels` of the docker image before...)

```
            "Labels": {
                "org.label-schema.build-date": "20191001",
                "org.label-schema.license": "GPLv2",
                "org.label-schema.name": "CentOS Base Image",
                "org.label-schema.schema-version": "1.0",
                "org.label-schema.vendor": "CentOS"
            }
```

Turns out the base image it setting the label, and in the wrong format to boot. Ugh

hiddeco:
@chainlink I am glad you consider this issue to be solved, but Flux should still have fallen back to the `Created` date you originally mentioned.

My suspicion is however that you are using an older `fluxctl` version, to be precise one before #2594 (`<=1.15.0`). Can you confirm everything works as expected using `fluxctl` `1.16.0` or `1.17.0`?

chainlink:
I'm using `1.16.0` on both the client and server. Will try upgrading to `1.17.0`

chainlink:
[...]

PR #2752, opened 2020-01-16
e2e: Add image release test

2opremio:
All tests until now run without using the registry scanner, so I thought it was about time to have an end-to-end test for image releases (which is a fundamental Flux feature)

~We probably shouldn't depend on external registries (particularly on `stefanprodan/podinfo` since it is often updated) for tests. I plan to add an internal registry to the tests, but that will come later. If this test becomes too flakey I am happy to disable it until then.~ (I have added a local registry for Flux)

Addresses part of #2593 (adds a test covering `fluxctl release`).

TODO:
- [x] Test automatic image releases
- [x] Test `HelmRelease` image updates


2opremio:
The bitnami images (e.g. bitnami/ghost ) from the flux-get-started repo have too many tags and we get rate-limited. This makes the test timeout.

I am considering to create and use a branch in flux-get-started guide, pointing to images from a registry local to the e2e tests. However, we would still need to load the redis/ghost/mongodb images in the local registry, which will take time.

Another option is to create a new example altogether, with tiny images. I wouldn't want to maintain that though.

@stefanprodan @hiddeco any other ideas?

hiddeco:
> Another option is to create a new example altogether, with tiny images. I wouldn't want to maintain that though.

We can build empty `FROM scratch` (or almost empty, given our label support) images in a loop in combination with a [`registry`](https://hub.docker.com/_/registry) container and set the labels using `--label 'date=$(date)`, building ten images would take about a second and it requires no maintenance.

2opremio:
> I think I can l live with that.

It's not ideal though, because I don't want to use Kind when working on the tests locally

2opremio:
Here is an intermediate solution. Only make Flux use a local registry without changing the k8s configuration.

Kubernetes will pull images in whatever way it used to but Flux will query the local registry.

When spawning Kind we can also use a mirror.

In fact, if we don't change the get-started example we also need to modify Flux's configuration.

2opremio:
I bit the bullet and I added a local registry to the test. Note that the registry is only used by Flux to scan images (and not by the cluster).

Adding an `/etc/hosts` entry for the registry allowed me to keep using https://github.com/fluxcd/flux-get-started without any changes to the images.

2opremio:
We may also want to change `Kind`'s  registry to a local one by leveraging https://github.com/fluxcd/flux/pull/2752#issuecomment-575423184 , but that's for later.

test/e2e/fixtures/crane_empty_img_tmpl/config.json
```diff
@@ -0,0 +1,65 @@
...
+      "-c",
+      "#(nop) ",
+      "ENV foo="
+    ],
+    "Image": "",
+    "Volumes": null,
+    "WorkingDir": "",
+    "Entrypoint": null,
+    "OnBuild": null,
+    "Labels": null
```
hiddeco: I think it would be smart to include our supported timestamp labels here, this would ensure our whitelist for this works (and make it easier to confirm #2699, but this is self-interest).

2opremio: I _think_ I just added what meant, although I am not sure about the value of adding labels with the same time value.

hiddeco: You did, I will make some adjustments while triaging the above issue so that it is both covered in end-to-end tests, and the dates will have an offset. Thanks!

PR #2760, opened 2020-01-17
Fix conditional statement determining in-cluster setup

hiddeco:
Follow up on #2741

### DIFF
--- a/test/e2e/21_ssh_key_generation.bats
+++ b/test/e2e/21_ssh_key_generation.bats
@@ -1,0 +1,32 @@
+#!/usr/bin/env bats
+
+function setup() {
+  load lib/env
+  load lib/install
+  load lib/defer
+
+  kubectl create namespace "$FLUX_NAMESPACE"
+
+  # Installing Flux with the defaults _without_ installing the git
+  # server (which generates an SSH key used by the server and Flux)
+  # will cause Flux to generate an SSH key.
+  install_flux_with_fluxctl
+}
+
+@test "SSH key is generated" {
+  run fluxctl identity --k8s-fwd-ns "$FLUX_NAMESPACE"
+  [ "$status" -eq 0 ]
+  [ "$output" != "" ]
+
+  fingerprint=$(echo "$output" | ssh-keygen -E md5 -lf - | awk '{ print $2 }')
+  [ "$fingerprint" = "MD5:$(fluxctl identity -l --k8s-fwd-ns "$FLUX_NAMESPACE")" ]
+}
+
+function teardown() {
+  run_deferred
+  # Although the namespace delete below takes care of removing most Flux
+  # elements, the global resources will not be removed without this.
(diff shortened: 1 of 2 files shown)


test/e2e/21_ssh_key_generation.bats
```diff
@@ -0,0 +1,27 @@
...
+
+  kubectl create namespace "$FLUX_NAMESPACE"
+
+  install_flux_with_fluxctl
+}
+
+@test "SSH key is generated" {
+  run fluxctl identity --k8s-fwd-ns "$FLUX_NAMESPACE"
+  [ "$status" -eq 0 ]
+  [ "$output" != "" ]
```
2opremio: I think we can do a bit better than this. Could parse the output and check it's a valid public key?

2opremio: Also, we may want to do polling, since the key may not be ready right after flux is deployed.

test/e2e/fixtures/kustom/21_ssh_key_generation/generation_patch.yaml
```diff
@@ -0,0 +1,3 @@
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: --k8s-secret-name=flux-git-deploy-generated
```
2opremio: Why is this necessary? I thought the key was generated out of the  out of the box with what's supplied by `fluxctl install`

2opremio: Ah, you are doing this to override the key we load in the tests.

Isn't it cleaner to avoid installing that secret?

2opremio:
Addresses part of #2593

test/e2e/fixtures/kustom/21_ssh_key_generation/generation_patch.yaml
hiddeco: They were left-overs from a previous version of this, I removed the arguments selecting the folder but forgot to remove the files.

2opremio: But, without doing letting flux generate the we won't be testing the big being fixed, I think.

hiddeco: Without deploying the git server an empty secret will be applied and Flux will generate the key.

2opremio: Ah, good point. Can you add a comment?

test/e2e/21_ssh_key_generation.bats
hiddeco: The key is now validated in two ways:

1. it is parsed by `ssh-keygen` to compute the fingerprint, which will fail for an invalid key
2. the fingerprint is compared to the output of `fluxctl identity -l`

PR #2762, opened 2020-01-18
Helm chart sops support

### LINKED ISSUE #2761: Helm chart sops support
Helm chart does not provide option to enable sops within values file. This would provide the expected user experience.

**Describe the feature**
Add sops.enabled in values.yaml and update deployment.yaml template.

**Expected behavior**
Deploy flux with sops enabled via helm.

GlorifiedTypist:
Helm chart does not provide option to enable sops within values file. This would provide the expected user experience.

Fixes #2761


GlorifiedTypist:
@stefanprodan Great thanks, commits have been squashed.

PR #2765, opened 2020-01-19
Add detailed error message in fluxctl sync

### LINKED ISSUE #2657: Improve `fluxctl sync` error when Flux cannot clone the git repository
From a user in Slack:

> Hi, could someone point me how can I troubleshoot `Error: git repository ssh://git@gitlab.mydomain.com/myrepo/flux.git is not ready to sync (status: new)`

The user got that error because Flux couldn't clone the repository (the user later fixed the problem by mounting a `known_hosts` file which was missing).

That error is not very helpful because:
* It leaks internal information of how we manage repos (`new` doesn't mean anything to users)
* It's not actionable

A better error would be:

```
Error: git repository ssh://git@gitlab.mydomain.com/myrepo/flux.git cannot be cloned: actual cloning error
```

dinosk:
* Extends v6.GitConfig to include `Error` field.

* In case of `fluxctl sync` error, prints the `Error` field
  in the output.

Closes #2657 

Signed-off-by: Dinos Kousidis <dinkousidis@gmail.com>

### DIFF
--- a/pkg/api/v6/api.go
+++ b/pkg/api/v6/api.go
@@ -58,6 +58,7 @@ type GitConfig struct {
 	Remote       GitRemoteConfig   `json:"remote"`
 	PublicSSHKey ssh.PublicKey     `json:"publicSSHKey"`
 	Status       git.GitRepoStatus `json:"status"`
+	Error        string            `json:"errors"`
 }
 
 type Deprecated interface {
--- a/cmd/fluxctl/sync_cmd.go
+++ b/cmd/fluxctl/sync_cmd.go
@@ -46,7 +46,10 @@ func (opts *syncOpts) RunE(cmd *cobra.Command, args []string) error {
 	case git.RepoReady:
 		break
 	default:
-		return fmt.Errorf("git repository %s is not ready to sync (status: %s)", gitConfig.Remote.URL, string(gitConfig.Status))
+		if gitConfig.Error != "" {
+			return fmt.Errorf("git repository %s is not ready to sync\n\nFull error message: %v", gitConfig.Remote.URL, gitConfig.Error)
+		}
+		return fmt.Errorf("git repository %s is not ready to sync", gitConfig.Remote.URL)
 	}
 
 	fmt.Fprintf(cmd.OutOrStderr(), "Synchronizing with %s\n", gitConfig.Remote.URL)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -624,7 +624,12 @@ func (d *Daemon) GitRepoConfig(ctx context.Context, regenerate bool) (v6.GitConf
 	origin := d.Repo.Origin()
 	// Sanitize the URL before sharing it
 	origin.URL = origin.SafeURL()
-	status, _ := d.Repo.Status()
+	status, err := d.Repo.Status()
+	gitConfigError := ""
+	if err != nil {
+		gitConfigError = err.Error()
+	}
+
 	path := ""
 	if len(d.GitConfig.Paths) > 0 {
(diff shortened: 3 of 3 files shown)


pkg/api/v6/api.go
```diff
@@ -58,6 +58,7 @@ type GitConfig struct {
 	Remote       GitRemoteConfig   `json:"remote"`
 	PublicSSHKey ssh.PublicKey     `json:"publicSSHKey"`
 	Status       git.GitRepoStatus `json:"status"`
+	Error        string            `json:"errors"`
```
2opremio: This will break backwards compatibility with clients using the v6 endpoint.

You would need to jump the API version to extend it. But, even then, I am not sure that's the best place for the error (to be fair I don't think that the status should be part of the config either, bit that's how it is).

dinosk: Ah right, thank you for reviewing.
I can think of making `git.GitRepoStatus` a map containing the `Error`, or adding another endpoint for the repo status?

2opremio: After taking a look at it, adding the error is probably OK but you should add a new version of the data structure and a new version of the endpoint (which will, in turn, require bumping the latest version of the API)

dinosk: Ok, I will update the PR

2opremio:
Thanks! I would like @squaremo to also have a look

In the meantime, can you fix the failing test? I thought it was a transient CI error, but it fails systematically

dinosk:
Tests are fixed, let me know if I should add sth else!

pkg/api/v6/api.go
squaremo: Since the Error field is optional, I don't think just adding it to the v6 API response breaks backward-compatibility. A way to reason about it is to do the matrix of old version / new version and see if mismatches will cause a problem.

In the following table, "Old" means "does not know about the Error field" and "New" means "expects to use the Error field".

|                    | Old fluxd | New fluxd |
|-----------------|------|--------|
| **Old fluxctl**| Operates as now | Will not deserialise Error field, and thereby operate as now |
| **New fluxctl** | Can act conditionally on presence of Error field | Operates as desired |

2opremio: Well, I was thinking about other clients other than `fluxctl`. I know there is people using the API, and their JSON parsers may be stricts or not do well with new fields.

If we don't care about those usecases then I guess it would be fine.

squaremo: Hmm yes true, my thinking is limited to the era before people started using the API directly. OK, new API version it is, ulp.

squaremo: On third thought -- it is a fair amount of trouble to percolate a new version through intermediating code (e.g., Weave Cloud, but no doubt other things). I would be prepared to accept the potential outcry over breaking strict parsing, as a trade for a smaller change overall. wdyt, when it's put in those terms, @2opremio ?

2opremio: OK, I think that's fine :) , @dinosk is (rightfully) going to hate me for it though.

squaremo: Sorry @dinosk! Blame me, I'm the trouble-maker.

So: to be clear, let's go back to how you initially had it, with an Error field added to `v6.GitConfig`, and that field being used if present in `cmd/fluxctl/sync_cmd.go`.

dinosk: 😅 it is no problem! I see your point, I’ll revert it to the initial changes

dinosk: _makes backup of new api version_
Ok, this is back to the initial changes

2opremio: Thanks for being so understanding!

PR #2777, opened 2020-01-21
Maintain `pkg/install` as a separate module with own deps

hiddeco:


go.mod
```diff
@@ -19,6 +19,9 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.0.0-20191016111102-bec269661e48 // kubernetes-1.16.2
 )
 
+// ugh
```
2opremio: I actually think it's pretty nifty, it saves you from bumping the dependency on every release.

hiddeco: It is, this was my brain still being traumatized by the 

```console
github.com/fluxcd/flux/cmd/fluxctl imports
        github.com/fluxcd/flux/pkg/install: no matching versions for query "latest"
```

message that popped up.

I am wondering however what the impact is when someone (weirdly) decides they want to depend on different `pkg/install` and `flux` module versions, or are local path overwrites ignored while depending on something?

hiddeco:
#2778 also removes the import in `tools.go`, and has a solid comment, so lets merge that.

PR #2778, opened 2020-01-21
Make pkg/install a Go module to reduce its dependencies

2opremio:
External projects (like github.com/weaveworks/eksctl, which in turn are used as a library) shouldn't need to pull the full Flux dependencies by only using `pkg/install`.

This will cause a million conflicts with #2776 but I wanted to try it ASAP.

Fix: #2779


2opremio:
And ... it turns out that @hiddeco beat me to it in #2777

PR #2785, opened 2020-01-24
Unmarshal Docker image labels separately

### LINKED ISSUE #2699: Created Timestamp issue with gcr.io repo not detected

hiddeco:
To prevent the created field from never getting set when `LabelTimestampFormatError` is returned, which is used as a fallback in case the labels have an invalid (non RFC3339 date-time) format.

Fixes #2699

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -128,28 +128,39 @@ interpret:
 	switch deserialised := manifest.(type) {
 	case *schema1.SignedManifest:
 		var man schema1.Manifest = deserialised.Manifest
-		// for decoding the v1-compatibility entry in schema1 manifests
+		// For decoding the v1-compatibility entry in schema1 manifests
+		// Ref: https://docs.docker.com/registry/spec/manifest-v2-1/
+		// Ref (spec): https://github.com/moby/moby/blob/master/image/spec/v1.md#image-json-field-descriptions
 		var v1 struct {
 			ID      string    `json:"id"`
 			Created time.Time `json:"created"`
 			OS      string    `json:"os"`
 			Arch    string    `json:"architecture"`
+		}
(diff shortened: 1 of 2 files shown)


pkg/registry/client.go
```diff
@@ -134,22 +134,28 @@ interpret:
 			Created time.Time `json:"created"`
 			OS      string    `json:"os"`
 			Arch    string    `json:"architecture"`
+		}
+		if err = json.Unmarshal([]byte(man.History[0].V1Compatibility), &v1); err != nil {
+			return ImageEntry{}, err
+		}
+
+		var config struct {
```
2opremio: Please add a comment on why the config is unmarshaled separately. Also, it would be a good idea to add comment to each structure definition, linking to their authoritative source (e.g. OCI image standard )

hiddeco: The authoritative sources of the image labels are already recorded in `image/image.go`. I have in-lined comments for the others.

PR #2788, opened 2020-01-27
Make clone cleanups more robust

### LINKED ISSUE #2713: Flux may be leaking files in /tmp
**Describe the bug**
Flux fills up /tmp

**To Reproduce**
Large monorepo. Flux runs fine for a few days, then slowly leaks disk and then ultimately runs out of space. I've mounted a PVC at /tmp so that it doesn't consume disk on the node itself.

**Expected behavior**
Fluxcd should probably clean-up after itself?

**Logs**
```
ts=2019-12-27T00:58:18.491235642Z caller=images.go:23 component=sync-loop error="getting unlocked automated resources: mkdir /tmp/flux-working546801009: no space left on device"
```

**Additional context**
- Flux version: 1.17.0

2opremio:
* Fixes a systematic clone leak (introduced in #1807 , released in 1.14 ) in every bump of the sync tag. This probably justifies a new patch release and it probably fixes #2713 . I still don't understand how this has escaped our radar until #2713  
* Removes the temporary cloning directory if `git clone` fails for whatever reason
* It stops hiding underlying errors in`Export.Clean()` and prints them wherever a logger is available (not perfect, but better).

### DIFF
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -488,7 +488,7 @@ func (r *Repo) workingClone(ctx context.Context, ref string) (string, error) {
 	}
 	path, err := clone(ctx, working, r.dir, ref)
 	if err != nil {
-		os.RemoveAll(path)
+		os.RemoveAll(working)
 	}
 	return path, err
 }
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -486,5 +486,9 @@ func (r *Repo) workingClone(ctx context.Context, ref string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return clone(ctx, working, r.dir, ref)
+	path, err := clone(ctx, working, r.dir, ref)
+	if err != nil {
+		os.RemoveAll(path)
+	}
+	return path, err
 }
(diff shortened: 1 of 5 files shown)


2opremio:
As a future optimization, we don't really need a working directory for bumping and pushing the tag, which would save quite a bit of space in big repos (e.g. in #2713  ~1GB of data would be saved)

squaremo:
> we don't really need a working directory for bumping and pushing the tag

You _do_ need to have the changed tag visible locally only when it has been pushed to the upstream repo, though.

pkg/git/repo.go
```diff
@@ -486,5 +486,9 @@ func (r *Repo) workingClone(ctx context.Context, ref string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return clone(ctx, working, r.dir, ref)
+	path, err := clone(ctx, working, r.dir, ref)
+	if err != nil {
+		os.RemoveAll(path)
```
squaremo: It's a quibble, but this ought to be

    os.RemoveAll(workingDir)

since that is the directory created here. At one point `clone` created a directory _under_ the working dir it was given, now it returns the same directory -- if it did something like that again, this would start leaking directories.

It's probably better to remove the possibility of it being a different directory and change `clone` to return only an error, so there's no mistaking which path is used. But for now: removing the `workingDir` guards against that regression.

2opremio: Ah, good catch! Thanks

Issue #2779, opened 2020-01-22
Make pkg/install a sub-module


stefanprodan:
Right now using the `install` package as a library in other projects like [eksctl](https://github.com/weaveworks/eksctl/blob/master/go.mod) means adding Flux as a dependency and that pulls Kubernetes packages, Docker, Helm etc. It would be great if `install` would be a separate go module so that it can be used without having to pull all Flux dependencies.

PR #2789, opened 2020-01-27
Fix `make generate-deploy`

2opremio:
It was failing with

```
$ make generate-deploy
go generate ./pkg/install
go: directory pkg/install is outside main module
make: *** [pkg/install/generated_templates.gogen.go] Error 1
```

Also, make sure that the dependencies of `pkg/install/generate.go` are included in `go.mod` (I didn't know that the `tools` build tag was treated specially).

Followup from #2779

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,7 +174,7 @@ generate-deploy: $(GOBIN)/fluxctl
 		--namespace flux
 
 $(GENERATED_TEMPLATES_FILE): pkg/install/templates/*.tmpl pkg/install/generate.go
-	go generate ./pkg/install
+	cd ./pkg/install &&  go generate .
 
 check-generated: generate-deploy
 	git diff --exit-code -- deploy/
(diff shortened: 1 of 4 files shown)


2opremio:
> It was failing with
> 
> ```
> $ make generate-deploy
> go generate ./pkg/install
> go: directory pkg/install is outside main module
> make: *** [pkg/install/generated_templates.gogen.go] Error 1
> ```
> 
> Also, make sure that the dependencies of `pkg/install/generate.go` are included in `go.mod` (I didn't know that the `tools` build tag was treated specially.
> 
> Followup from #2779

Maybe we should fix this instead by moving `pkg/install/generate.go` to `pkg/cmd/something`. I don't think that external users of `pkg/install` should need to pull `vfsgen` since it's only a compile-time dependency. 

Now, maybe this is just a micro-optimization?

Even better than that it would be for Go to figure out which of the dependencies from a given package are used and not enforce building and getting everything in their `go.mod` (that would have saved used from needing #2779 in the first place. Maybe this isn't possible for some fundamental reason I am missing?

Issue #2797, opened 2020-01-29
Default namespace option (--k8s-default-namespace) has no effect


michaelbeaumont:
**Describe the bug**

In spite of setting the `--k8s-default-namespace` option, the default namespace isn't applied to namespaced manifests without a specific `namespace`.

Seems to be [present in the `gitops-engine` branch](https://github.com/fluxcd/flux/issues/2614) as well

**To Reproduce**

Steps to reproduce the behaviour:

1. Setup `flux` including the `--k8s-default-namespace` option. E.g.:
```
$ fluxctl install \
--git-user=test \
--git-email=test@users.noreply.github.com \
--git-url=https://github.com/michaelbeaumont/flux_manifests \
--git-path=/ \
--git-readonly=true \
--namespace=flux > out.yaml
$ patch -p0 << EOF
> --- out.yaml    2020-01-29 14:21:09.447424403 +0100
+++ out_p.yaml   2020-01-29 14:21:45.957538447 +0100
@@ -190,6 +190,8 @@

         # Include this to enable git signature verification
         # - --git-verify-signatures
+
+        - --k8s-default-namespace=test

         # Tell flux it has readonly access to the repo (default `false`)
         - --git-readonly
> EOF
```

2. A simple pod manifest:
https://github.com/michaelbeaumont/flux_manifests/tree/master

**Expected behavior**

The pod should be deployed in the `test` namespace, instead:
```
$ kubectl get pod -n default -o yaml
apiVersion: v1
kind: Pod
metadata:
  name: sleep
spec:
  containers:
  - name: sleep
    image: busybox
    command: ['sh', '-c', 'echo Started && sleep 3600']
```

**Additional context**

- Flux version: 1.17.1
- Kubernetes version: 1.16

2opremio:
:man_facepalming: after reviewing https://github.com/fluxcd/flux/pull/2625 , I can see that the option is not honoured when invoking `kubectl`.

I will fix that.

michaelbeaumont:
I was actually hoping to fix it myself!

2opremio:
Even better, please do!

It should be as simple as passing `--namespace` to `kubectl` (assuming that respects the namespace in resources that provide it explicitly)

michaelbeaumont:
`kubectl` seems to throw an error if the namespaces disagree so that won't work. I guess the intended usage is to limit operations to a single NS, so it makes sense to throw an error.

2opremio:
[...]

PR #2799, opened 2020-01-29
Fix syncing with --k8s-default-namespace

michaelbeaumont:
Fixes https://github.com/fluxcd/flux/issues/2797

The problem was that the namespace was set on the `KubeManifest` next to the raw `Resource` but then when the `KubeManifest` is synced as a `Resource`, the raw bytes are taken.
Since we're reserializing in `applyMetadata` anyway, I just update the namespace there.

### DIFF
--- a/pkg/cluster/kubernetes/sync.go
+++ b/pkg/cluster/kubernetes/sync.go
@@ -377,6 +377,15 @@ func applyMetadata(res resource.Resource, syncSetName, checksum string) ([]byte,
 		mixin["labels"] = mixinLabels
 	}
 
+	// After loading the manifest the namespace of the resource can change
+	// (e.g. a default namespace is applied)
+	// The `ResourceID` should give us the up-to-date value
+	// (see `KubeManifest.SetNamespace`)
+	namespace, _, _ := res.ResourceID().Components()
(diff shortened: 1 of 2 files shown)


pkg/cluster/kubernetes/sync.go
```diff
@@ -377,6 +377,11 @@ func applyMetadata(res resource.Resource, syncSetName, checksum string) ([]byte,
 		mixin["labels"] = mixinLabels
 	}
 
+	namespace, _, _ := res.ResourceID().Components()
+	if namespace != kresource.ClusterScope {
+		mixin["namespace"] = namespace
```
2opremio: Nice and simple. Much better than my suggestion at https://github.com/fluxcd/flux/issues/2797#issuecomment-579900826

Can you please add a detailed comment explaining why we are doing this?

michaelbeaumont: done

2opremio: Fantastic

2opremio:
Can you squash the commits?

2opremio:
And rebase on top of the current master?

michaelbeaumont:
done

Issue #2704, opened 2019-12-20
Support signature verification of signed merges from unsigned branches


jstevans:
**Describe the feature**
`fluxd` when configured with `--git-verify-signatures` accepts a signed commit `A` as long as there is a completely-signed merge-base between that commit and flux's current revision.

**What would the new user story look like?**
1. We would like to have a workflow where our `master` branch is unsigned.
2. There is some human-initiated automation that creates a signed (`no-fast-forward`) merge commit onto our `release` branch.

**Expected behavior**
`fluxd` would allow that merge commit by reducing the current constraint from "All commits between `HEAD..(fluxd_current_revision)`" to...
* **cheap:** `--first-parent` when calling [`git log`][0] << [`repo.CommitsBetween()`][1] << [`latestValidRevision`][2] contains only signed commits
* **better:** Allow any commit that has at least one set containing only signed commits that connects it to `fluxd`'s current revision. This could be done by reconstructing the git tree in code, and doing the walk from `HEAD` to `fluxd_current_revision` while treating unsigned commits as dead-ends.

[0]:https://github.com/fluxcd/flux/blob/3b5c66d617b1323027d56e3ec690302959df0167/pkg/git/operations.go#L254
[1]:https://github.com/fluxcd/flux/blob/3b5c66d617b1323027d56e3ec690302959df0167/pkg/git/repo.go#L242-L248
[2]:https://github.com/fluxcd/flux/blob/master/pkg/daemon/daemon.go#L888

hiddeco:
Short comment as I will be off from today on until after new year, but wanted to give this some attention so you know it has been looked at (by the person who wrote the initial signature verification implementation).

First: thanks a lot for diving into the code and getting familiar with it before making the request. I think your enhancement request is valid, and can see why you want to have this.

As this feature goes hand-in-hand with security, I need to give your proposed options some thought and map out what the impact of the difference between both of them is. To be continued...

hiddeco:
Finally had some time to dive into this @jstevans, thanks for waiting.

[I read a good long read about this topic](https://mikegerwitz.com/2012/05/a-git-horror-story-repository-integrity-with-signed-commits), where your *cheap*  proposal is [also described](https://mikegerwitz.com/2012/05/a-git-horror-story-repository-integrity-with-signed-commits#script-merge).

[...]

PR #2803, opened 2020-01-30
Support multiple signature verification strategies with  `--gitVerifySignaturesMode`

jstevans:
Add a new flag that puts the `--gitVerifySignatures` behavior into one
of two modes:

    * "all" (default) uses the existing `--gitVerifySignatures` behavior

    * "first-parent" uses the behavior described in issue #2704, wherein
        only each first-parent of each commit from the tip of the flux
        branch are considered when verifying GPG signatures

Fixes issue #2704

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -134,9 +134,10 @@ func main() {
 		gitTimeout      = fs.Duration("git-timeout", 20*time.Second, "duration after which git operations time out")
 
 		// GPG commit signing
-		gitImportGPG        = fs.StringSlice("git-gpg-key-import", []string{}, "keys at the paths given will be imported for use of signing and verifying commits")
-		gitSigningKey       = fs.String("git-signing-key", "", "if set, commits Flux makes will be signed with this GPG key")
-		gitVerifySignatures = fs.Bool("git-verify-signatures", false, "if set, the signature of commits will be verified before Flux applies them")
+		gitImportGPG               = fs.StringSlice("git-gpg-key-import", []string{}, "keys at the paths given will be imported for use of signing and verifying commits")
+		gitSigningKey              = fs.String("git-signing-key", "", "if set, commits Flux makes will be signed with this GPG key")
+		gitVerifySignatures        = fs.Bool("git-verify-signatures", false, "(deprecated) sets --git-verify-signatures-mode=all when set")
+		gitVerifySignaturesModeStr = fs.String("git-verify-signatures-mode", fluxsync.VerifySignaturesModeDefault, fmt.Sprintf("if git-verify-signatures is set, which strategy to use for signature verification (one of %s)", strings.Join([]string{fluxsync.VerifySignaturesModeNone, fluxsync.VerifySignaturesModeAll, fluxsync.VerifySignaturesModeFirstParent}, ",")))
 
 		// syncing
 		syncInterval = fs.Duration("sync-interval", 5*time.Minute, "apply config in git to cluster at least this often, even if there are no new commits")
@@ -203,6 +204,7 @@ func main() {
 	fs.MarkDeprecated("k8s-namespace-whitelist", "changed to --k8s-allow-namespace, use that instead")
 	fs.MarkDeprecated("registry-poll-interval", "changed to --automation-interval, use that instead")
 	fs.MarkDeprecated("k8s-in-cluster", "no longer used")
+	fs.MarkDeprecated("git-verify-signatures", "changed to --git-verify-signatures-mode, use that instead")
 
 	var kubeConfig *string
 	{
@@ -309,6 +311,19 @@ func main() {
 		*automationInterval = *registryPollInterval
 	}
 
+	// Maintain backwards compatibility with the --git-verify-signatures
+	// flag, but only if the --git-verify-signature-mode is not set to a
+	// custom (non default) value.
+	var gitVerifySignaturesMode fluxsync.VerifySignaturesMode
+	if *gitVerifySignatures && !fs.Changed("git-verify-signatures-mode") {
+		gitVerifySignaturesMode = fluxsync.VerifySignaturesModeAll
+	} else {
+		if gitVerifySignaturesMode, err = fluxsync.ToVerifySignaturesMode(*gitVerifySignaturesModeStr); err != nil {
+			logger.Log("error", err.Error())
+			os.Exit(1)
+		}
+	}
+
 	// Sort out values for the git tag and notes ref. There are
 	// running deployments that assume the defaults as given, so don't
 	// mess with those unless explicitly told.
@@ -346,7 +361,7 @@ func main() {
 
 	// Import GPG keys, if we've been told where to look for them
 	for _, p := range *gitImportGPG {
-		keyfiles, err := gpg.ImportKeys(p, *gitVerifySignatures)
+		keyfiles, err := gpg.ImportKeys(p, gitVerifySignaturesMode != fluxsync.VerifySignaturesModeNone)
 		if err != nil {
 			logger.Log("error", fmt.Sprintf("failed to import GPG key(s) from %s", p), "err", err.Error())
 		}
@@ -658,7 +673,7 @@ func main() {
 		"user", *gitUser,
 		"email", *gitEmail,
 		"signing-key", *gitSigningKey,
-		"verify-signatures", *gitVerifySignatures,
+		"verify-signatures-mode", gitVerifySignaturesMode,
 		"sync-tag", *gitSyncTag,
 		"state", *syncState,
 		"readonly", *gitReadonly,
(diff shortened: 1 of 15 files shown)


docs/references/git-gpg.md
```diff
@@ -240,3 +240,26 @@ b. Sign the sync tag by yourself, with a key that is imported, to point
    $ git tag --force --local-user=<key id> -a -m "Sync pointer" <tag name> <revision>
    $ git push --force origin <tag name>
    ```
+
+### Choosing a `--gitVerifySignaturesMode`
```
hiddeco: We (try to) maintain a document width of `<=80` for our documentation files.

pkg/install/generated_templates.gogen.go
```diff
@@ -1,216 +1,216 @@
-// Code generated by vfsgen; DO NOT EDIT.
```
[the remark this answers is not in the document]
jstevans: Hrm, no, but I'll revert. I'm running on Windows (really, a WSL2-backed instance of Ubuntu), so perhaps that had some effect on the whitespace 😄

cmd/fluxd/main.go
```diff
@@ -134,9 +134,10 @@ func main() {
...
 
 		// GPG commit signing
-		gitImportGPG        = fs.StringSlice("git-gpg-key-import", []string{}, "keys at the paths given will be imported for use of signing and verifying commits")
-		gitSigningKey       = fs.String("git-signing-key", "", "if set, commits Flux makes will be signed with this GPG key")
-		gitVerifySignatures = fs.Bool("git-verify-signatures", false, "if set, the signature of commits will be verified before Flux applies them")
+		gitImportGPG            = fs.StringSlice("git-gpg-key-import", []string{}, "keys at the paths given will be imported for use of signing and verifying commits")
+		gitSigningKey           = fs.String("git-signing-key", "", "if set, commits Flux makes will be signed with this GPG key")
+		gitVerifySignatures     = fs.Bool("git-verify-signatures", false, "if set, the signature of commits will be verified before Flux applies them")
+		gitVerifySignaturesMode = fs.String("git-verify-signatures-mode", "all", fmt.Sprintf("if git-verify-signatures is set, which strategy to use for signature verification (one of {$s}", strings.Join([]string{"all", "first-parent"}, ",")))
```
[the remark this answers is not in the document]
jstevans: Thanks for the comments! I've updated -- please let me know what you think of the update!

cmd/fluxd/main.go
```diff
@@ -345,8 +346,14 @@ func main() {
...
 
 	// Import GPG keys, if we've been told where to look for them
+	gitVerifySignaturesMode, err := fluxsync.ToVerifySignaturesMode(*gitVerifySignaturesModeStr)
+	if err != nil {
+		logger.Log("error", err.Error())
+	}
+
 	for _, p := range *gitImportGPG {
-		keyfiles, err := gpg.ImportKeys(p, *gitVerifySignatures)
+		gitVerifySignatures := gitVerifySignaturesMode != fluxsync.VerifySignaturesModeNone
```
hiddeco: I think this can be placed _outside_ the `for` loop.

cmd/fluxd/main.go
```diff
@@ -202,6 +202,7 @@ func main() {
 	fs.MarkDeprecated("k8s-namespace-whitelist", "changed to --k8s-allow-namespace, use that instead")
 	fs.MarkDeprecated("registry-poll-interval", "changed to --automation-interval, use that instead")
 	fs.MarkDeprecated("k8s-in-cluster", "no longer used")
+	fs.MarkDeprecated("git-verify-signatures", "changed to --git-verify-signatures-mode, use that instead")
```
hiddeco: In addition to marking the flag as deprecated, setting the deprecated `--git-verify-signatures` flag should still result in `--git-verify-signatures-mode=all` below the surface; _but only_ if `--git-verify-signatures-mode` has not been set too.

pkg/sync/provider.go
```diff
@@ -12,6 +13,37 @@ const (
...
+	case "":
+		return VerifySignaturesModeNone, nil
+	case VerifySignaturesModeNone:
+		return VerifySignaturesModeNone, nil
+	case VerifySignaturesModeAll:
+		return VerifySignaturesModeAll, nil
+	case VerifySignaturesModeFirstParent:
+		return VerifySignaturesModeFirstParent, nil
+	default:
+		return VerifySignaturesModeNone, fmt.Errorf("%s is not a valid git-verify-signatures-mode", s)
```
hiddeco: Nitpick for observability in logs

```suggestion
		return VerifySignaturesModeNone, fmt.Errorf("'%s' is not a valid git-verify-signatures-mode", s)
```

jstevans:
@hiddeco I'm having trouble what went wrong in the CI build -- I've read over the contributing docs, but I don't think I understand how to investigate this issue. Can you give some guidance?

cmd/fluxd/main.go
```diff
@@ -345,8 +347,27 @@ func main() {
...
+		logger.Log("error", err.Error())
+	}
+
+	// default to no signature verification
+	gitVerifySignaturesMode, err := fluxsync.ToVerifySignaturesMode(fluxsync.VerifySignaturesModeNone)
+
+	if err != nil {
+		logger.Log("error", err.Error())
+	}
+	if *gitVerifySignaturesModeStr != fluxsync.VerifySignaturesModeDefault {
```
hiddeco: `pflag` offers a `Changed` method, using this method you can check if a flag has been modified from its default value. Using this flag this could be reduced to:

```go
if fs.Changed("git-verify-signatures") && !fs.Changed("git-verify-signatures-mode") {
  gitVerifySignaturesMode = fluxsync.VerifySignaturesModeAll
}
```

In addition, can you please move this a bit higher up (`L307`) below the other backwards maintaining `if`-statement?

pkg/daemon/daemon_test.go
```diff
@@ -15,10 +15,10 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	"github.com/fluxcd/flux/pkg/api/v10"
-	"github.com/fluxcd/flux/pkg/api/v11"
-	"github.com/fluxcd/flux/pkg/api/v6"
-	"github.com/fluxcd/flux/pkg/api/v9"
+	v10 "github.com/fluxcd/flux/pkg/api/v10"
```
hiddeco: Probably some linter you have enabled, but all these import aliases are obsolete.

pkg/sync/provider.go
```diff
@@ -12,6 +13,40 @@ const (
 	NativeStateMode = "secret"
 )
 
+// VerifySignaturesMode represents the strategy to use when choosing which commits to GPG-verify between the flux sync tag and the tip of the flux branch
+type VerifySignaturesMode string
+
+const (
+        // VerifySignaturesModeDefault - get the default behavior when casting
```
hiddeco:
```suggestion
	// VerifySignaturesModeDefault - get the default behavior when casting
```

cmd/fluxd/main.go
```diff
@@ -345,8 +347,27 @@ func main() {
 	}
 
 	// Import GPG keys, if we've been told where to look for them
+
+	if err != nil {
```
hiddeco: What does this log?

jstevans: Woops, I was reorganizing my changes and must have included this twice 😄

jstevans:
@hiddeco done!

hiddeco:
One additional request I forgot to mention, can you please remove references to PR from the commit message and the description?

The GitHub feature is amazing for most use-cases but it causes endless pings (including from forks) to the issues whenever the commit is in some way touched and seen as 'new'.

jstevans:
@hiddeco Done! Is there anywhere I _should_ add a reference to the issue?

hiddeco:
@jstevans PR message itself so that the PR is linked to the issue (which you have already done). So all good now.

cmd/fluxd/main.go
```diff
@@ -308,6 +310,18 @@ func main() {
...
 
+	// Maintain backwards compatibility with the --git-verify-signatures
+	// flag, but only if the --git-verify-signature-mode is not set to a
+	// custom (non default) value.
+	var gitVerifySignaturesMode fluxsync.VerifySignaturesMode
+	if *gitVerifySignatures && !fs.Changed("git-verify-signatures-mode") {
+		gitVerifySignaturesMode = fluxsync.VerifySignaturesModeAll
+	} else {
+		if gitVerifySignaturesMode, err = fluxsync.ToVerifySignaturesMode(*gitVerifySignaturesModeStr); err != nil {
+			logger.Log("error", err.Error())
```
hiddeco: Having doubts about the fact that this does not make Flux exit given this all is a safety feature. I think it may be better to:

```suggestion
			logger.Log("error", err.Error())
			os.Exit(1)
```

to prevent things from getting applied without validation, as changing the default value automatically means that you want some type of validation to happen.

jstevans:
GitHub PR flow takes some getting used to, I hope this is correct :)

PR #2813, opened 2020-02-03
Rename --registry-scanning to --registry-disable-scanning & keep it independent from --git-readonly

### LINKED ISSUE #2790: Reformulate `--registry-scanning`
The flag `--registry-scanning` is well-motivated, but can be refined further.

 1. It should be `--disable-registry-scanning` (or `--registry-disable-scanning`)

In general, boolean flags that default to `true` should be used sparingly if at all. Such a flag has to be given a value to have an effect, which breaks the expectation that a boolean flag is provided without a value. (And there are other problems with default-true boolean values; for example, if you have a go struct defining the config, the field value will default to false rather than true, so you have to be extra careful about how you populate it).

You can argue that it's awkward to have a `--disable-X` flag -- yes it is. Designing from scratch, you would have `--enable-registry-scanning` to switch that particular optional thing on. But the starting point is that fluxd already does a bunch of things, so you have to go the other way around, and switch things off -- and the less surprising way to do it is to use flags which default to false.

 2. I don't think `--registry-scanning=false` should imply `--git-readonly`.

It doesn't _logically_ imply it -- it's totally possible to use a writable git repo while switching registry scanning off.

But there's no particular motivation for making git read-only when you're not scanning image registries, either. Automation of image updates isn't possible if you are not looking at image metadata, but that doesn't mean you don't want to write to the git repo -- you may still use a tag as the high-water mark, and you may still want to be able to e.g., change automation policies.

A better factoring would be to have `--registry-scanning=false` imply disabling _automation_, whether or not that's controlled by a switch itself.

2opremio:
Fixes #2790


squaremo:
You didn't fix the boolean-flag-defaulting-to-true problem from #2790.

2opremio:
> You didn't fix the boolean-flag-defaulting-to-true problem from #2790.

I don't really see it as a problem, but I can change it if you see it as a showstopper.

squaremo:
> > You didn't fix the boolean-flag-defaulting-to-true problem from #2790.
> 
> I don't really see it as a problem, but I can change it if you see it as a showstopper.

Well, perhaps either implement it as suggested in the issue, or mount an argument against doing so. My argument is essentially that it's convention to have boolean flags default to `false`, so having one that defaults to `true` will surprise some people; and, secondarily, it's a trap for future players, since boolean values (in libraries, structs, variables) default to false, and eventually someone will forget that this one flag is different, and produce unexpected results. There would probably _not_ be a lot of fallout, but why not avoid it if you can?

2opremio:
My only argument is that double negatives are not intuitive.

squaremo:
> My only argument is that double negatives are not intuitive.

Double negatives don't arise -- you either supply

    --disable-registry-scanning

to turn scanning off, or omit that flag to let it default to `false`, i.e., to keep the status quo.

(If the status quo were that registry scanning was _off_ and you had to explicitly switch it _on_, then the recommendation would be to have `--registry-scanning`.)

2opremio:
Fair enough. I will change it later today

2opremio:
The `e2e` tests caught a pretty basic error which unit tests wouldn't had been able to catch (it happened in the initialization of `main()`). I am so happy about this :)

2opremio:
> Fair enough. I will change it later today

@squaremo I am changing the flag now and found another reason for not having the `--disable` prefix: consistency. All the other registry-related flags start with `--registry`

I will continue changing it anyhow since you fell strongly in favor of it. Just leaving it for the record.

squaremo:
> The `e2e` tests caught a pretty basic error which unit tests wouldn't had been able to catch

Also a wonderful illustration of why not to introduce a boolean field which needs to be `true` for behaviour to stay the same. (We've all done it!)

squaremo:
> another reason for not having the `--disable` prefix: consistency. All the other registry-related flags start with `--registry`

You can change the name, it's the default value I feel strongly about. `--registry-disable-scanning`, e.g.

2opremio:
Ah, that's better.

2opremio:
> Also a wonderful illustration of why not to introduce a boolean field which needs to be `true` for behaviour to stay the same. (We've all done it!)

For the record, I inherited that part of the code :)

cmd/fluxctl/install_cmd.go
```diff
@@ -48,7 +48,7 @@ fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' --gi
 		"whether to enable manifest generation")
 	cmd.Flags().StringVar(&opts.Namespace, "namespace", "",
 		"cluster namespace where to install flux")
-	cmd.Flags().BoolVar(&opts.RegistryScanning, "registry-scanning", true,
+	cmd.Flags().BoolVar(&opts.RegistryDisableScanning, "registry-scanning", true,
```
squaremo:
```suggestion
	cmd.Flags().BoolVar(&opts.RegistryDisableScanning, "registry-disable-scanning", false,
```

cmd/fluxctl/install_cmd.go
```diff
@@ -48,7 +48,7 @@ fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' --gi
...
 		"scan container image registries to fill in the registry cache")
```
squaremo:
```suggestion
		"disable scanning of container image registries")
```

cmd/fluxd/main.go
```diff
@@ -561,7 +560,7 @@ func main() {
 	// Registry components
 	var imageRegistry registry.Registry = registry.ImageScanDisabledRegistry{}
 	var cacheWarmer *cache.Warmer
-	if *registryScanning {
+	if !*registryDisableScanning {
```
squaremo: Ah! Here is a double-negative, and it does make one look twice -- but I am less concerned about this one than the UI itself, having brought it up ..

cmd/fluxd/main.go
```diff
@@ -660,7 +659,7 @@ func main() {
 		"sync-tag", *gitSyncTag,
 		"state", *syncState,
 		"readonly", *gitReadonly,
-		"registry-scanning", *registryScanning,
+		"registry-disable-scanning", *registryDisableScanning,
```
squaremo: I don't think this line is was really necessary in the first place (all the other things are about the git repo and syncing)

docs/faq.md
```diff
@@ -340,27 +340,28 @@ If the replicas field is not present in Git, Flux will not override the replica
...
 
 You can completely disable registry scanning by using the
-`--registry-scanning=false` flag. This allows deploying Flux without Memcached.
+`--registry-disable-scanning` flag. This allows deploying Flux without
+ Memcached.
 
 
 If you only want to disable scanning for certain images, you can keep
-`--registry-scanning` set to `true` (its default value) and tell Flux
+`--registry-disable-scanning` set to `false` (its default value) and tell Flux
```
squaremo: It might be clearer to advise leaving the flag out altogether

pkg/install/install.go
```diff
@@ -38,7 +38,7 @@ func FillInTemplates(params TemplateParameters) (map[string][]byte, error) {
 		if info.IsDir() {
 			return nil
 		}
-		if (params.GitReadOnly || !params.RegistryScanning) && strings.Contains(info.Name(), "memcache") {
+		if params.RegistryDisableScanning && strings.Contains(info.Name(), "memcache") {
```
squaremo: The comment after suggests that memcache will be omitted if --git-readonly, which I think is incorrect. Either the comment needs changing, or the comment and the code ..

pkg/install/install_test.go
```diff
@@ -48,38 +48,38 @@ func unmarshalDeployment(t *testing.T, data []byte) *v1.Deployment {
...
+		GitBranch:               "branch",
+		GitPaths:                []string{"dir1", "dir2"},
+		GitLabel:                "label",
+		GitUser:                 "User",
+		GitEmail:                "this.is@anemail.com",
+		Namespace:               "flux",
+		GitReadOnly:             false,
+		ManifestGeneration:      true,
+		AdditionalFluxArgs:      []string{"arg1=foo", "arg2=bar"},
+		RegistryDisableScanning: true,
```
squaremo: This should not pass -- you just toggled a flag that's supposed to affect the generated templates!

pkg/install/templates/flux-deployment.yaml.tmpl
```diff
@@ -184,8 +184,8 @@ spec:
         # Serve /metrics endpoint at different port;
         # make sure to set prometheus' annotation to scrape the port value.
         - --listen-metrics=:3031
-{{- if not .RegistryScanning }}
-        - --registry-scanning=false
+{{- if .RegistryDisableScanning }}
```
squaremo: :+1:

test/e2e/fixtures/kustom/14_release_image/release_image_patch.yaml
```diff
@@ -1,6 +1,6 @@
 - op: add
   path: /spec/template/spec/containers/0/args/-
-  value: --registry-scanning=true
+  value: --registry-scanning=false
```
squaremo: No longer a flag

cmd/fluxctl/install_cmd.go
2opremio: I updated it in a different way (consistent with the rest)

pkg/install/install.go
2opremio: It's the comment

2opremio: And this is why I missed it

Screen Shot 2020-02-04 at 5 07 13 PM

Goland collapses`if` blocks only including a return statement by by default (which I think is reasonable) but it also hides any comments inside them.

test/e2e/fixtures/kustom/14_release_image/release_image_patch.yaml
2opremio: I missed the prefix, thanks

cmd/fluxd/main.go
2opremio: Uhm, I think it's useful to debug problems with Flux (as important as the readonly mode).

Flux's behavior changes quite a bit.

squaremo: If it's not reporting whether scanning is disabled anywhere else in the log, I guess you could put it here. It's a bit random, though.

squaremo: Or coming from the other direction -- this line has already diverged from its original purpose, which was to report the git-specific parameters, so why not :-)

pkg/install/install_test.go
2opremio: true, it shouldn't. I need to work on re-adding the install package tests later on though

2opremio:
@squaremo I think I've addressed all your comments, PTAL

pkg/install/install_test.go
```diff
@@ -58,35 +58,27 @@ func TestFillInTemplatesAllParameters(t *testing.T) {
...
-		GitURL:           "git@github.com:fluxcd/flux-get-started",
-		GitBranch:        "branch",
-		GitPaths:         []string{},
-		GitLabel:         "label",
-		RegistryScanning: true,
+		GitURL:                  "git@github.com:fluxcd/flux-get-started",
+		GitBranch:               "branch",
+		GitPaths:                []string{},
+		GitLabel:                "label",
+		RegistryDisableScanning: true,
```
squaremo: Presumably, this should now omit memcached deployment, and thereby have fewer manifests?

squaremo: Indeed:

```
(cd pkg/install; go test -v .)
```

fails.

2opremio: Fixed. I really need to fix the install package testing

pkg/install/go.mod
```diff
@@ -8,4 +8,6 @@ require (
 	github.com/shurcooL/vfsgen v0.0.0-20181202132449-6a9ea43bcacd
 	github.com/stretchr/testify v1.4.0
 	golang.org/x/tools v0.0.0-20200121210457-b3205ff6fffe // indirect
+	k8s.io/api v0.17.2
```
squaremo: Any idea how these dependencies end up here? On the surface of it, it doesn't seem like changes in the module code would prompt the change here.

2opremio: Darn, that was probably because of #2652 

The only reason for this was unmarshalling a deployment. I think I could use an anonymous struct with the relevant fields. But I would rather do it in a separate PR.

2opremio: It's getting hard to keep `pkg/install` kubernetes-free

squaremo: OK, in what order do you want to remedy these things

2opremio: Let's merge this first

I fixed it at https://github.com/fluxcd/flux/pull/2824

2opremio: Or rather, I don't care about the ordering. I will deal with conflicts if I have to.

2opremio: I rebased on master. So this is fully resolved now.

Issue #2814, opened 2020-02-03
Implementing Redis as alternative for image registry cache


prometherion:
**Describe the feature**
A top feature of `flux` is the registry scanning, a component that scans container image registries to populate a cache and speed up the retrieval of the most recently released tag.

This feature would like to implement the `redis` strategy beside `memcached` one due to its popularity among the K/V stores landscape.

From an operational perspective, implementing a tool means taking full ownership of the direct dependencies: according to the current Flux state, it means that any Ops team has to know how to take care of the `memcached`, although it's just used as a non-persistent cache.

Last but not least, with this feature, the feature will scaffold the project to allow additional backing services requested by the community.

**What would the new user story look like?**

0. A first draft of the PoC allowing Redis as new repository cache backing service
1. Adding the following new flags:
   * `--cache-backend` flag as _enum_ (`redis`, `memcached`) defaulted to `memcached` for backward compatibility
   * `--cache-request-timeout` as a generic flag to handle request timeout despite the selected strategy
2. Marking the `--memcached-timeout` flag as deprecated in favor of `--cache-request-timeout`
3. Update the `fluxctl` command to allow new Kubernetes manifests creation to spin-up deployment according to the selected cache backing service
4. Update the _Kustomize_ base manifests
5. Update the documentation available in `docs` folder that generates the `ReadTheDocs`

**Expected behavior**

Allow the chance to use `redis` as container images registry backing service instead of `memcached`.

kingdonb:
I appreciate the PR you contributed. Flux v1 is in maintenance mode now, and is not adding any new features unless they are critical.

As Flux contrib efforts have been focused on Flux v2, the Flux project has moved to a new repo, [fluxcd/flux2](/fluxcd/flux2)

In the interest of reducing the number of open issues not directly related to supporting Flux v1 in maintenance mode, and respecting you may have moved on already, I will go ahead and close out this issue for now.

PR #2815, opened 2020-02-03
Redis as alternative storage for container images repository cache

prometherion:
Still a WIP for the feature filled in #2814: allowing `redis` as an alternative for the container images repository cache.

Providing new flags (`--cache-{backend,request-timeout}`) in order to allow backend selection and abstract the backend requests timeout (deprecating `--memcached-timeout` providing backward compatibility along with deprecation notice).

This first contribution aims to scaffold the main file also for future backend cache storages but keeping high backward compatibility to don't break things.

TODO:
- [x] Redis unit tests
- [x] Redis integration tests
- [x] Docs update
- [x] README.md changes
- [x] Scaffolding Helm Chart to allow Redis
- [x] Align `fluxctl install` to allow the Redis experimental feature
- [x] e2e tests


squaremo:
:wave: This is a pretty substantial contribution in the making, thank you for getting involved!

May I ask, what are the advantages of Redis over memcached here? I have never been very convinced of memcached's suitability (it was something of an arbitrary choice), so I am interested to hear your take on it.

prometherion:
Honestly, there are any personal preferences over `memcached` or `redis`: both got their pros and cons, advantages or trade-off. I like `redis` because of low memory footprint but, in the end, it's just something personal, as preferred _flavor_ :)

2opremio:
@prometherion Thanks a lot for the effort! I will try to review this early next week. In the meantime, would you mind adding an e2e test? extending the `fluxctl list-images` and `image_release`  tests to execute both on memcached and redis would be a good start.

prometherion:
I'm working on the refactoring to allow test with multiple cache backend but I need to perform first another refactoring regarding the files generated by `fluxctl install`.

Since I followed the actual implementation, we end up with the following file naming:

- `{redis,memcache}-dep.yaml`
- `{redis,memcache}-svc.yaml`

To allow _bats_ testing we would need various fixtures for each backend because [kustomize doesn't allow to use wildcards on resources names](https://github.com/kubernetes-sigs/kustomize/issues/484)

A possible workaround could be output just `cache-{dep,svc}.yaml` instead of specifying the backend kind.

2opremio:
The Kustomize files in the tests allow templating, in case that helps

.circleci/config.yml
```diff
@@ -101,7 +101,8 @@ jobs:
       - run: make test TEST_FLAGS="-race -tags integration -timeout 5m"
       - run: make check-generated
       - run: make all
-      - run: E2E_KIND_CLUSTER_NUM=4 make e2e
+      - run: E2E_KIND_CLUSTER_NUM=4 CACHE_BACKEND=memcached make e2e
+      - run: E2E_KIND_CLUSTER_NUM=4 CACHE_BACKEND=redis make e2e
```
2opremio: This is overkill because tests take more than 10 minutes and only some of them use the cache (those with --registry-disable-scanning=false)

prometherion: what about parallelize?

prometherion: @2opremio I guess we can achieve parallelism using the _workflow_ section of CircleCI, please take a look [here](https://github.com/fluxcd/flux/pull/2815/files#diff-1d37e48f9ceff6d8030570cd36286a61R207-R209).

Actually, tests are running: let's see how they end up.

squaremo:
@prometherion Do you mind squashing some of the fixup commits? I'm thinking of `Wrong package name` and the like.

prometherion:
sure @squaremo: just done!

Morriz:
Really hoping to see this one land! So we can get redis pubsub on new registry images...yay!

.circleci/config.yml
prometherion: @2opremio in case you missed: I implemented the e2e tests for each image cache backing service, named `e2e-memcached` and `e2e-redis`: since they're parallel, we're not anymore wasting time during the CI.

Although, if you don't like this approach, we could abstract the `bats` test cases using a function to run all the assertions. However, I feel this solution a bit tricky because it looks more complex for those who never contributed.

Waiting for your feedback, I'd really like to get it merged :)

2opremio: I still think it's overkill because most of the e2e tests run without using a cache (because we run them with `--registry-disable-scan`). I would much rather have duplicate runs only for the tests using the cache.

prometherion:
@2opremio so, in the end, agreed on discharge the idea of running parallel CircleCI jobs.

Actually, implemented testing of both backing services (`redis` and `memcached`):
* `14_release_image-{memcached,redis}`
* `15_fluxctl_install-{memcached,redis}`

If it's ok I'm going to squash commits to keep it cleaner.

2opremio:
Sorry to be picky, but I would factor out the test code and call it using a different backend indtead

prometherion:
No problem :) Going to refactor this

prometherion:
@2opremio just pushed: I moved the test logic inside the `lib/{release_image,fluxctl_list}.bash` files and just calling them from the single BATS files.

I'm still using prefix backend cache service name strategy since I have to ensure to perform the `teardown` also in case of errors of a single test.

I would write something like (pseudo-code)

```bats
@test "Fluxctl list-workloads on memcached" {
  setup 'memcached'
  warmup
  # Test fluxctl list-services
  run fluxctl --k8s-fwd-ns "${FLUX_NAMESPACE}" list-workloads --namespace "${DEMO_NAMESPACE}"
  list_workloads "${status}" "${output}"
  teardown 'memcached'
}

@test "Fluxctl list-workloads on redis" {
  setup 'redis'
  warmup
  # Test fluxctl list-services
  run fluxctl --k8s-fwd-ns "${FLUX_NAMESPACE}" list-workloads --namespace "${DEMO_NAMESPACE}"
  list_workloads "${status}" "${output}"
  teardown 'redis'
}
```

But honestly, doesn't look good to me since I wouldn't be able to execute tests per backing service (e.g.: `E2E_TESTS='*-redis*' make e2e`)

Morriz:
Any updates here?

ianmuscat:
Since Redis offers authentication out of the box, perhaps it might be worth considering allowing users to have the ability to configure Flux to use password-based authentication with Redis. Right now it seems that this value is hardcoded to an empty string -- https://github.com/fluxcd/flux/blob/e655c99055adbd32194ae24c05b2db8f76c6b226/pkg/registry/cache/redis.go#L58

prometherion:
Would be great have a feedback from maintainers too about this feature: still waiting a review since a month.

Morriz:
@prometherion any change to get this to pass?

prometherion:
Just rebased over `master` and it has been a pain: `.circleci/config.yml` has been refactored, let's wait CI tests...

prometherion:
@2opremio please, can I request you a further review?
@Morriz can't wait to get this PR merged :sweat_smile:

2opremio:
Unfortunately I don't have time for it, sorry.

prometherion:
NP, who's in charge or available to take a look at this?

cmd/fluxctl/install_cmd.go
```diff
@@ -52,6 +52,7 @@ fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' --gi
 		"do not scan container image registries to fill in the registry cache")
 	cmd.Flags().StringVarP(&opts.outputDir, "output-dir", "o", "", "a directory in which to write individual manifests, rather than printing to stdout")
 	cmd.Flags().BoolVar(&opts.AddSecurityContext, "add-security-context", true, "Ensure security context information is added to the pod specs. Defaults to 'true'")
+	cmd.Flags().StringVar(&opts.CacheBackend, "cache-backend", "memcached", "Select the docker images repository cache backend. Defaults to 'memcached'")
```
squaremo, line 55: Despite the fact that other random flags have found their way in here (I'm looking at you, `--add-security-context`), I don't think it's necessary for the cache backend to be an option here, especially not while it's experimental.

cmd/fluxd/main.go
```diff
@@ -145,11 +144,17 @@ func main() {
 		dryGC        = fs.Bool("sync-garbage-collection-dry", false, "only log what would be garbage collected, rather than deleting. Implies --sync-garbage-collection")
 		syncState    = fs.String("sync-state", fluxsync.GitTagStateMode, fmt.Sprintf("method used by flux for storing state (one of {%s})", strings.Join([]string{fluxsync.GitTagStateMode, fluxsync.NativeStateMode}, ",")))
 
-		// registry
+		// registry scanning cache backend
+		cacheBackend = fs.String("cache-backend", "memcached", "select the registry cache backend (supported: memcached, redis)")
```
squaremo, line 148: Since these `"memcached", "redis"` are used in a switch below, I think it's worth having consts for them.

cmd/fluxd/main.go
```diff
@@ -60,7 +59,7 @@ const (
 
 	// This is used as the "burst" value for rate limiting, and
 	// therefore also as the limit to the number of concurrent fetches
-	// and memcached connections, since these in general can't do any
+	// and cache backend connections, since these in general can't do any
```
squaremo, line 62: I don't think it's used for Redis though?

pkg/install/templates/cache-svc.yaml.tmpl
```diff
@@ -0,0 +1,31 @@
...
+    - name: memcached
+      port: 11211
+  selector:
+    name: memcached
+{{- else if eq .CacheBackend "redis" -}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: redis
```
squaremo, line 21: Wouldn't it be better if this corresponded with the default value for the service name? That way there is less to accidentally get wrong.

squaremo: (notwithstanding I advise above to remove this from `fluxctl install` entirely)

.circleci/config.yml
```diff
@@ -162,4 +162,4 @@ workflows:
             branches:
               ignore: /.*/
             tags:
-              only: /^chart.*/
+              only: /^chart.*/
```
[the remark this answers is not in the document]
prometherion: solved by 374bbedb

chart/flux/README.md
```diff
@@ -262,12 +262,21 @@ The following tables lists the configurable parameters of the Flux chart and the
...
 | `kube.config`                                     | [See values.yaml](/chart/flux/values.yaml#L200-L212) | Override for kubectl default config in the Flux pod(s).
 | `priorityClassName`                               | `""`                                                 | Set priority class for Flux
+| `redis.enabled`                                   | `false`                                              | _Experimental_: create a redis deployment and service.
+| `redis.pullSecret`                                | `None`                                               | _Experimental_: Image pull secret
+| `redis.repository`                                | `redis`                                              | _Experimental_: Image repository
+| `redis.tag`                                       | `5.0.7`                                              | _Experimental_: Image tag
+| `redis.resources`                                 | `None`                                               | _Experimental_: CPU/memory resource requests/limits for reds
```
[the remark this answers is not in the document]
prometherion: solved by 374bbedb

chart/flux/README.md
```diff
@@ -262,12 +262,21 @@ The following tables lists the configurable parameters of the Flux chart and the
...
+| `redis.securityContext`                           | [See values.yaml](/chart/flux/values.yaml#L211-L214) | _Experimental_: Container security context for reds
+| `redis.nodeSelector`                              | `{}`                                                 | _Experimental_: Node Selector properties for the redis deployment
+| `redis.tolerations`                               | `[]`                                                 | _Experimental_: Tolerations properties for the redis deployment
+| `kube.config`                                     | [See values.yaml](/chart/flux/values.yaml#L235-L247) | Override for kubectl default config in the Flux pod(s).
```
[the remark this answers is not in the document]
prometherion: solved by 374bbedb

chart/flux/templates/memcached.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (eq .Values.memcached.enabled true) (eq .Values.registry.disableScanning false)  }}
+{{- if and (and (eq .Values.memcached.enabled true) (eq .Values.redis.enabled false)) (eq .Values.registry.disableScanning false) }}
```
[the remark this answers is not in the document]
prometherion: I just followed the same comparison applied previously, but we change simplify it if you wish: just tried to don't make too many big diffs to make PR easier to review

cmd/fluxd/main.go
prometherion: indeed, but we could use the same option for the `cache.RedisConfig` __MaxConns__ property: WDYT?

cmd/fluxd/main.go
prometherion: fixed by 6767d842

chart/flux/templates/deployment.yaml
```diff
@@ -191,6 +191,7 @@ spec:
           - --ssh-keygen-dir=/var/fluxd/keygen
           - --ssh-keygen-format={{ .Values.ssh.keygen.format }}
           - --k8s-secret-name={{ .Values.git.secretName | default (printf "%s-git-deploy" (include "flux.fullname" .)) }}
+          - --cache-backend={{ include "flux.cache-backend" . }}
           - --memcached-hostname={{ .Values.memcached.hostnameOverride | default (printf "%s-memcached" (include "flux.fullname" .)) }}
```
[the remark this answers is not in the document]
prometherion: if we want, we could wrap the entire line in the _helper_ printing the flag just if it differs from `memcached`

cmd/fluxctl/install_cmd.go
prometherion: from an end-user perspective, if I want to give a try to a different cache service, I would really enjoy this flag: furthermore: at first glance, it looks like a Helm alternative and I get your point, probably `flux install` should just provide a battle-tested installation without so many parameters.

If we want to remove the `cache-backend` flag, I guess the same fate has to be reserved for `add-security-context`.

pkg/install/templates/cache-svc.yaml.tmpl
prometherion: keep in mind that removing from `fluxctl install` will require additional time since I'll have to find a better solution on how to hack the _e2e_ tests to evaluate also the Redis server.

prometherion:
> Rebased this. @prometherion Are you OK if I remove the `fluxctl install` flag, and possibly make other, minor amendments?

I can take care of it because I totally get your point and probably we have to hack a bit tests with `kustomize`, although that could increase the test complexity :(

Morriz:
Bumping this as it would help us a lot

Morriz:
@prometherion can you please respond to the latest questions? Progression here would help a lot

Morriz:
@prometherion can you maybe take a look at this again? Progression here would help a lot

prometherion:
> @prometherion can you maybe take a look at this again? Progression here would help a lot

Should we proceed with this?

I mean, it seems Flux will be deprecated in favor of the GitOps Toolkit, not sure this PR is till relevant.

@squaremo could you give us feedback?

kingdonb:
I see a lot of work has gone into this PR and I think there is some recognition in the discussion here that it is now a longshot for inclusion in a next release of Flux v1.

In the new image reflector controller, if I understand correctly image data are maintained in the ImageRepository custom resource and memcached is no longer used, I am not certain I understand how this scales as usage grows, but it seems it will no longer be a concern of redis vs. memcached. Will that moot the entire issue, or should we continue the discussion here?

kingdonb:
This possibly related discussion has been opened: https://github.com/fluxcd/flux2/discussions/802

> **Deprecate "select latest by build time" for image automation**

I have not yet evaluated the design of image-reflector-controller for scalability, but I think the folks who are more deeply embedded than myself have chosen this design to address scalability concerns exposed by flux v1. Especially if you have an opinion about this, if you have time to look, we would appreciate your comment there!

kingdonb:
In the interest of reducing the number of open issues not directly related to supporting Flux v1 in maintenance mode, and respecting you may have moved on already, I will go ahead and close out this issue for now.

If you have a use case for Flux that isn't covered well in the new Flux v2 (which is a total rewrite), we want to hear about it.

If you've been following our development efforts then of course we hope you are able to upgrade, here's more info on how to find support with that: https://fluxcd.io/support/

It looks like the decision has been made, and heavyweight image metadata caches in Flux are going to be a thing of the past soon, (for many users in production this is a thing of the past already!) Thanks for making this contribution. Closing.

PR #2817, opened 2020-02-03
e2e: extend fluxctl tests

### LINKED ISSUE #2593: Add end-to-end tests for fluxctl and manifest generation

2opremio:
* Add test for `fluxctl sync`
* Add test for `fluxctl` (de)automate/(un)lock/policy
* Add test for `fluxctl list-images` and `fluxctl list-workloads`

With this all the `fluxctl` commands are covered except `fluxctl save` (and `fluxctl version`)

The ordering of tests (i.e. `15_` `16_` `17_`) is not definite. I am waiting for #2544 to reshuffle them again (for instance, it doesn't make sense to test the policy commands after using them in other tests)

Closes #2593

### DIFF
--- a/test/e2e/16_fluxctl_sync.bats
+++ b/test/e2e/16_fluxctl_sync.bats
@@ -1,0 +1,53 @@
+#!/usr/bin/env bats
+
+clone_dir=""
+
+function setup() {
+  load lib/env
+  load lib/install
+  load lib/poll
+  load lib/defer
+
+  kubectl create namespace "$FLUX_NAMESPACE"
+  # Install flux and the git server, allowing external access
+  install_git_srv git_srv_result
+  # shellcheck disable=SC2154
+  export GIT_SSH_COMMAND="${git_srv_result[0]}"
+  # Teardown the created port-forward to gitsrv and restore Git settings.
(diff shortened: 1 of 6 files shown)


test/e2e/16_fluxctl_sync.bats
```diff
@@ -0,0 +1,53 @@
...
+  defer rm -rf "'$clone_dir'"
+  git clone -b master ssh://git@localhost/git-server/repos/cluster.git "$clone_dir"
+  # shellcheck disable=SC2164
+  cd "$clone_dir"
+}
+
+@test "fluxctl sync" {
+
+  # Sync
+  poll_until_true 'fluxctl sync succeeds' "fluxctl --k8s-fwd-ns ${FLUX_NAMESPACE} sync"
```
hiddeco: IIRC this command returns the short commit hash of the applied commit, wouldn't that be better to use in the comparison below?

2opremio: Uhm I just want to make sure that sync tag is aligned with the latest commit. I think the test below is enough (and it doesn't require extra parsing).

2opremio:
@hiddeco I forgot to add one of the test files. PTAL

PR #2822, opened 2020-02-05
Make sure we test all the local modules

2opremio:
This, in particular includes the new pkg/install module

Followup from #2778

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ CURRENT_OS=$(shell go env GOOS)
 CURRENT_OS_ARCH=$(shell echo $(CURRENT_OS)-`go env GOARCH`)
 GOBIN?=$(shell echo `go env GOPATH`/bin)
 
-MAIN_GO_MODULE:=github.com/fluxcd/flux
+MAIN_GO_MODULE:=$(shell go list -m -f '{{ .Path }}')
 LOCAL_GO_MODULES:=$(shell go list -m -f '{{ .Path }}' all | grep $(MAIN_GO_MODULE))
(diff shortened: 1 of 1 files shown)


hiddeco:
@2opremio if the amount of time you have available allows you to, can you please port this to the Helm operator?

2opremio:
Sure! Let's get it merged first though.

Makefile
```diff
@@ -24,6 +24,8 @@ CURRENT_OS=$(shell go env GOOS)
 CURRENT_OS_ARCH=$(shell echo $(CURRENT_OS)-`go env GOARCH`)
 GOBIN?=$(shell echo `go env GOPATH`/bin)
 
+MAIN_GO_MODULE:=github.com/fluxcd/flux
```
hiddeco: Bit disappointed to discover that there does not seem to be a Go utility to extract this from the `module` declaration in a `go.mod` file.

2opremio:
It's worth noting that, due to the (overly) sophisticated way to calculate build dependencies, the build wasn't impacted by this.

That said, I think we should simplify the dependency calculation since it takes a long time, probably longer than invoking `go` systematically (letting it figure out if anything changed). Regardless it's a pain for `$ make ...` autocompletion (it takes seconds right now).

Makefile
2opremio: Uhm, I think it's doable, let me check

2opremio: Done

squaremo:
> That said, I think we should simplify the build dependency calculation since it takes a long time

Yes, it probably takes as long as just invoking `go build` and letting go decide.

2opremio:
> Yes, it probably takes as long as just invoking `go build` and letting go decide.

Then I think I will make the build targets `PHONY` and invoke go systematically. I will do that in a separate PR.

Issue #1991, opened 2019-04-29
Sync fails due to metrics.k8s.io discovery error


hamid2013:
Hello All,

While following the steps, i am getting the below error on the flux and not able to run any workload using the flux:

```ts=2019-04-29T04:06:36.714545618Z caller=images.go:18 component=sync-loop msg="polling images"
ts=2019-04-29T04:06:36.714622818Z caller=images.go:28 component=sync-loop msg="no automated workloads"
ts=2019-04-29T04:06:43.293711087Z caller=loop.go:103 component=sync-loop event=refreshed url=git@github.com:xxxxxxx/xxxxxx branch=master HEAD=c38943df8dde8b4ffda98a050e6f643ec712ccc3
ts=2019-04-29T04:06:43.528694435Z caller=main.go:199 type="internal kubernetes error" err="couldn't get resource list for metrics.k8s.io/v1beta1: the server is currently unable to handle the request"
ts=2019-04-29T04:06:43.528789635Z caller=loop.go:210 component=sync-loop err="collating resources in cluster for sync: not found"
ts=2019-04-29T04:06:43.531861734Z caller=loop.go:90 component=sync-loop err="collating resources in cluster for sync: not found"
ts=2019-04-29T04:11:36.714944593Z caller=images.go:18 component=sync-loop msg="polling images"
ts=2019-04-29T04:11:36.818370548Z caller=main.go:199 type="internal kubernetes error" err="couldn't get resource list for metrics.k8s.io/v1beta1: the server is currently unable to handle the request"
ts=2019-04-29T04:11:44.168625836Z caller=loop.go:210 component=sync-loop err="collating resources in cluster for sync: not found"
ts=2019-04-29T04:11:44.171171938Z caller=loop.go:90 component=sync-loop err="collating resources in cluster for sync: not found"
ts=2019-04-29T04:11:44.172498339Z caller=loop.go:103 component=sync-loop event=refreshed url=git@github.com:xxxxxx/xxxxxxxx branch=master HEAD=c38943df8dde8b4ffda98a050e6f643ec712ccc3
ts=2019-04-29T04:16:36.825719462Z caller=images.go:18 component=sync-loop msg="polling images"
ts=2019-04-29T04:16:36.939251772Z caller=main.go:199 type="internal kubernetes error" err="couldn't get resource list for metrics.k8s.io/v1beta1: the server is currently unable to handle the request"
[...]

PR #2823, opened 2020-02-05
Log start of cluster sync

borancar:
It's helpful to operators to understand at which commit the cluster is
stuck should any errors occur with cluster sync, such as k8s internal
issues:
- https://github.com/fluxcd/flux/issues/1991
- https://github.com/jetstack/cert-manager/issues/2273

### DIFF
--- a/pkg/daemon/sync.go
+++ b/pkg/daemon/sync.go
@@ -70,6 +70,8 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 		return err
 	}
 
+	d.Logger.Log("info", "trying to sync git changes to the cluster", "old", c.oldTagRev, "new", c.newTagRev)
+
 	// Run actual sync of resources on cluster
 	syncSetName := makeGitConfigHash(d.Repo.Origin(), d.GitConfig)
 	resourceStore, err := d.getManifestStore(working)


borancar:
To give some more context, I spent time looking at all the wrong places why the cluster wasn't getting updated until I finally got Delve and narrowed down the issue. The following 2 messages were actually not very helpful in understanding why the images were stuck at their old version:
```
ts=2020-02-04T20:29:52.312079303Z caller=main.go:267 type="internal kubernetes error" kubernetes_caller=k8s.io/client-go@v11.0.0+incompatible/discovery/cached/memory/memcache.go:199 err="couldn't get resource list for admission.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request"
ts=2020-02-04T20:29:52.329078854Z caller=main.go:267 type="internal kubernetes error" kubernetes_caller=k8s.io/client-go@v11.0.0+incompatible/discovery/cached/memory/memcache.go:111 err="couldn't get resource list for admission.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request"
```
Having those messages after a message saying "Trying/Attempting sync..." with specific git versions would have helped me massively narrow down on the issue.

pkg/daemon/sync.go
```diff
@@ -70,6 +70,8 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 		return err
 	}
 
+	d.Logger.Log(fmt.Sprintf("Trying to sync git changes to cluster, from commit %s to commit %s", c.oldTagRev, c.newTagRev))
```
hiddeco:
```suggestion
	d.Logger.Log(fmt.Sprintf("Trying to sync git changes to cluster, from commit '%s' to commit '%s'", c.oldTagRev, c.newTagRev))
```

Would be easier to read in a blob of logs imo.

pkg/daemon/sync.go
```diff
@@ -70,6 +70,8 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 		return err
 	}
 
+	d.Logger.Log(fmt.Sprintf("Trying to sync git changes to cluster, from commit '%s' to commit '%s'", c.oldTagRev, c.newTagRev))
```
hiddeco: I missed the fact that you did not include a key for the message, I think you have two options:

```suggestion
	d.Logger.Log("info", fmt.Sprintf("trying to sync git changes to the cluster, from commit '%s' to commit '%s'", c.oldTagRev, c.newTagRev))
```

or

```suggestion
	d.Logger.Log("info", "trying to sync git changes to the cluster", "old", c.oldTagRev, "new", c.neTagRev)
```

The latter matches https://github.com/fluxcd/flux/blob/4ed1b29667bbb0a82872c7201ce965016e3c68a7/pkg/daemon/loop.go#L220

borancar: Thanks, the latter makes more sense. Just need to fix a typo (c.newTagRev vs c.neTagRev) and pushing.

hiddeco: Can you please smash your commits together and rebase on `master` too?

hiddeco:
Can you please smash your commits together, edit the commit message and remove the issue links from it (to prevent endless pings to the issues from our friendly OSS neighbours), and rebase the result on `master`?

PR #2827, opened 2020-02-06
Run `go mod tidy`

2opremio:
Maybe we should do something to do this automatically (or at least make sure PR creators do it).

Maybe piggyback on `make check-generated`?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,6 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-kit/kit v0.9.0
 	github.com/gogo/googleapis v1.3.1 // indirect
-	github.com/gogo/protobuf v1.3.1
 	github.com/gogo/status v1.1.0 // indirect
 	github.com/golang/gddo v0.0.0-20190312205958-5a2505f3dbf0
 	github.com/google/go-containerregistry v0.0.0-20200121192426-b0ae1fc74a66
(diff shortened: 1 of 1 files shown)


stefanprodan:
Hmm what about go.sum? I see the Kubernetes code-gen in there

hiddeco:
Isn't this the result of Go Mod operations on `/pkg/install` and `/` fighting each other?

squaremo:
All invocations of go build (or test, or probably other things) should use `-mod=readonly`, so that they will fail if go would otherwise modify go.{mod,sum}.

I attempted to make this the case a while ago, but the code generation in pkg/install got in the way. It may be feasible now.

squaremo:
> Isn't this the result of Go Mod operations on `/pkg/install` and `/` fighting each other?

It should not be -- those operate independently.

2opremio:
> Hmm what about go.sum? I see the Kubernetes code-gen in there

God point, but I don't know the answer. That's what `go mod tidy` did.

2opremio:
> All invocations of go build (or test, or probably other things) should use `-mod=readonly`, so that they will fail if go would otherwise modify go.{mod,sum}.

You mean in CI, right? But then ... how do you fill the cache in?

squaremo:
> how do you fill the cache in?

`go` will still fetch things it doesn't have.

2opremio:
> You mean in CI, right? But then ... how do you fill the cache in?

I completely misunderstood how `-mod=readonly` works sorry.

That won't work for all cases (this case in particular), because building doesn't lead to the `go.mod` changes in this PR, only `go mod tidy` does.

squaremo:
> That won't work for all cases (this case in particular), because building doesn't lead to the `go.mod` changes in this PR, only `go mod tidy` does.

No, but it'll force people to be conscious of adding dependencies, so they might get in the habit of running `go mod tidy` when things change :-)

2opremio:
can we merge this though? :)

stefanprodan:
I think we can run `tidy` after the code gen check and fail the build if go.mod or sum changes. We can show a message in CI like "please run go mod tidy".

2opremio:
> I think we can run `tidy` after the code gen check and fail the build if go.mod or sum changes. We can show a message in CI like "please run go mod tidy".

Yeah, that's what I had in mind. I will do it in a separate PR.

PR #2833, opened 2020-02-07
Add support for bash/zsh completion

### LINKED ISSUE #1593: fluxctl bash completion?
Hello,

As fluxctl seems to be leveraging cobra, it should be quite easy to achieve by making use of this cobra feature: https://github.com/eparis/cobra/blob/master/bash_completions.md

I would be happy to implement it but at the same time I'd like to have some feedback from the team before starting.

Would you be interested into getting bash completion setup for fluxctl?

Thanks!

yiannistri:
- Add support for bash and zsh completion
- Currently using the builtin cobra generated scripts but can be extended to support nouns

Fixes #1593

### DIFF
--- a/cmd/fluxctl/completion_cmd.go
+++ b/cmd/fluxctl/completion_cmd.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	completionShells = map[string]func(out io.Writer, cmd *cobra.Command) error{
+		"bash": runCompletionBash,
+		"zsh":  runCompletionZsh,
+	}
+)
+
+func newCompletionCommand() *cobra.Command {
+	shells := []string{}
+	for s := range completionShells {
+		shells = append(shells, s)
+	}
+
+	return &cobra.Command{
+		Use:                   "completion SHELL",
+		DisableFlagsInUseLine: true,
+		Short:                 "Output shell completion code for the specified shell (bash or zsh)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return newUsageError("please specify a shell")
+			}
+
+			if len(args) > 1 {
+				return newUsageError(fmt.Sprintf("please specify one of the following shells: %s", strings.Join(shells, " ")))
+			}
+
+			run, found := completionShells[args[0]]
+			if !found {
+				return newUsageError(fmt.Sprintf("unsupported shell type %q.", args[0]))
+			}
+
+			return run(cmd.OutOrStdout(), cmd.Parent())
+		},
+		ValidArgs: shells,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


cmd/fluxctl/completion_cmd.go
```diff
@@ -0,0 +1,53 @@
...
+		Use:                   "completion shell",
+		DisableFlagsInUseLine: true,
+		Short:                 "Output shell completion code for the specified shell (bash or zsh)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return newUsageError("please specify a shell")
+			}
+
+			if len(args) > 1 {
+				return newUsageError("please specify either bash or zsh")
```
hiddeco: Making use of `shells` here would make this error maintenance free.

yiannistri: Good point! 👍

yiannistri:
@hiddeco seems like the docker container running kubeadm to set up the e2e tests run out of memory 😕Any idea if this is a transient failure and if so could we trigger the build again?

hiddeco:
It is a transient error we are working on with the folks of Kind, I have rebased your PR to get rid of the commits from master that were accidentally pulled in, which has also triggered a new CI build :crossed_fingers:

cmd/fluxctl/completion_cmd.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func newCompletionCommand() *cobra.Command {
+	shells := []string{}
+	for s := range completionShells {
+		shells = append(shells, s)
+	}
+
+	return &cobra.Command{
+		Use:                   "completion shell",
```
squaremo: Do we want to distinguish between `completion` which is literal, and `shell` which is user input? (or indeed, give the possible values)

squaremo:
(needs a rebase before it can be merged; if you don't get to it @yiannistri then someone will do it eventually :-)

cmd/fluxctl/completion_cmd.go
hiddeco: I think it would make sense here to utilize the `shells` (again), and put `<>` around the list to highlight the fact that they are options.

yiannistri: Good idea, only issue is that the `Use` field is also used by the root command as a way to distinguish which of the commands need to skip port forward. So the completion command would have to get instantiated there (or come up with a different way of detecting which commands are run). Alternatively, we could change it to `completion SHELL` to distinguish them, similar to `kubectl`.  What do you think?

hiddeco: Given people are hopefully already familiar with `kubectl`, I think that’s a good alternative.

hiddeco:
@yiannistri can you please rebase on top of `master` (and maybe smash your commits together)?

hiddeco:
Thank you @yasserisa! :sunflower:

PR #2834, opened 2020-02-07
Add JSON output option for fluxctl's list-images and list-workloads

trevrosen:
## What
A JSON output option for `fluxctl list-workloads` and `fluxctl list-images`

## Why
Building tooling (such as chatbot functionality) is much easier with a machine-readable output.

## Note
I added [Testify](https://github.com/stretchr/testify) to the test build b/c I think that the single-line assertions make test code really easy to read. If this is controversial, I can remove. It tends to be the single thing that I use outside of stdlib for tests.

### DIFF
--- a/cmd/fluxctl/format.go
+++ b/cmd/fluxctl/format.go
@@ -2,10 +2,19 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 	"text/tabwriter"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/fluxcd/flux/pkg/registry"
+
+	v6 "github.com/fluxcd/flux/pkg/api/v6"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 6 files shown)


cmd/fluxctl/format.go
```diff
@@ -29,3 +45,142 @@ func makeExample(examples ...string) string {
...
+		if f == format {
+			return true
+		}
+	}
+	return false
+}
+
+// outputImagesJson sends the provided ImageStatus info to the io.Writer in JSON formatting, honoring limits in opts
+func outputImagesJson(images []v6.ImageStatus, out io.Writer, opts *imageListOpts) error {
+	if opts.limit < 1 {
```
hiddeco: Making `opts.limit == 0` equal to an infinite amount of results would be an better option in my opinion. This also makes it easier to programmatically make use of the CLI flag when you sometimes do want to limit the result.

trevrosen: Agreed - also that's what the usage note says anyway, so... my bad 😸 

Fixed.

trevrosen: ahh actually one sec need to do a bit more

hiddeco: In addition I would make the returned error a bit more humane to not step on anyone's toes, while also informing them about the check that failed.

My suggestion would be something like the following: 

> opts.limit is less than 0 but only supports non-negative integers

trevrosen: Done!

Issue #2836, opened 2020-02-09
AWS IAM for service accounts working?


derrickburns:
Has anyone actually tested the support for AWS IAM for service accounts?

I ask that because there are two things needed to support it.  

First, you need to bind with a version of the Go libraries that support it.  I see that you do that in 1.17 and beyond.

Second, you need to provide a `securityContext` that allows one to runs as a group that can read the token, e.g:
```     
securityContext:                                                                                                                                                                           
  fsGroup: 65534
```
I do not see that in the helm chart and I do not see a way of adding a `securityContext`.

If this is needed, could you please add a way of providing a `securityContext` to the pod.

stefanprodan:
Flux runs are root, I think the FS group is needed only for non-root containers. We have confirmation that Flux works with IAM for service accounts see https://github.com/fluxcd/flux/pull/2722#issuecomment-572066310

PR #2837, opened 2020-02-09
Allow user to provide pod security context.

derrickburns:
This is needed to enable Amazon IAM for Service Accounts.


stefanprodan:
This is not needed, @derrickburns can you please close this and #2836

PR #2843, opened 2020-02-11
Add image-registry and image-pull-secrets flags to fluxctl

### LINKED ISSUE #2541: Fluxctl install flux using a private docker registry
**Describe the feature**
A clear and concise description of the feature.
being able to set up a flag or an argument to use flux image living in the private docker registry in the fluxctl install command.

As an engineer, I would like to deploy and use flux in my k8s cluster living in a private network that cannot talk to the public docker registry. 
It would be fluxctl install --image-repository <your docker registry endpoint> --imagePullSecrets <secret name> ...

Feel free to add a diagram if that helps explaining things.

**Expected behavior**
A clear and concise description of what you expect to happen.
No changes in behavior. Just using the private docker registry to pull the fluxcd image.

phillebaba:
This change adds the flags `--image-repository` and `--image-pull-secrets` flags to the `fluxctl install` command. The image repository parameter will change the repository prefix of the flux and memcached in the generated deployment. The image pull secrets parameter will be appended to both of the deployments `imagePullSecrets` field.

Not setting these flags will cause fluxctl to behave like normal.

Fixes #2541


pkg/install/templates/flux-deployment.yaml.tmpl
```diff
@@ -66,7 +66,9 @@ spec:
         # There are no ":latest" images for flux. Find the most recent
         # release or image version at https://hub.docker.com/r/fluxcd/flux/tags
         # and replace the tag here.
-        image: docker.io/fluxcd/flux:1.18.0
+        image: {{ if .ImageRepository }}{{ .ImageRepository }}flux:1.18.0{{ else }}docker.io/fluxcd/flux:1.18.0{{ end }}{{ if .ImagePullSecrets }}
```
hiddeco: I think making the `if`-statement multi-lined would improve readability. If you agree and decide to change this, keep in mind to use the appropriate whitespace controllers (e.g. `{{-`).

pkg/install/templates/memcache-dep.yaml.tmpl
```diff
@@ -20,7 +20,9 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
       - name: memcached
-        image: memcached:1.5.20
+        image: {{ .ImageRepository }}memcached:1.5.20{{ if .ImagePullSecrets }}
```
hiddeco: I think it would be better to use whitespace controllers instead of in-lining the `if`.

phillebaba:
Yeah you are right, I guess I just followed the patterns of what is already there.

hiddeco:
@squaremo can you please give this an additional round of review? As you seem to be more opinionated about the supported flags in the install package than me.

squaremo:
Ah! Sorry, I am late to the party.

Elsewhere I have resisted adding more flags to `fluxctl install`, on the basis that there is already a way to install flux with all the bells and whistles (the Helm chart) and a way to customise the output of `fluxctl install` (kustomize). Its only requirement, really, is to provide a sane base configuration for the latter purpose -- and it should probably be renamed, for that reason.

Regarding the particular bug addressed in this PR, I think the way to address it is with a "how to" document showing how to produce a kustomization with the desired images change, and apply it.

squaremo:
@phillebaba Are you content with the alternative of using kustomize to change the image used? You can even refer to the deploy/ directory in this project as a base, if I understand correctly.

phillebaba:
Yeah I am fine with that. I will close the PR as it is no longer needed.

PR #2846, opened 2020-02-13
cmd/fluxctl: Support passing additional arguments to fluxd via fluxctl install

chancez:
Adds the ability to pass arbitrary additional args to `fluxd` via `fluxctl install`. Useful for setting flags like `--sync-garbage-collection`.

### DIFF
--- a/cmd/fluxctl/install_cmd.go
+++ b/cmd/fluxctl/install_cmd.go
@@ -52,6 +52,7 @@ fluxctl install --git-url 'git@github.com:<your username>/flux-get-started' --gi
 		"do not scan container image registries to fill in the registry cache")
 	cmd.Flags().StringVarP(&opts.outputDir, "output-dir", "o", "", "a directory in which to write individual manifests, rather than printing to stdout")
 	cmd.Flags().BoolVar(&opts.AddSecurityContext, "add-security-context", true, "Ensure security context information is added to the pod specs. Defaults to 'true'")
+	cmd.Flags().StringSliceVar(&opts.AdditionalFluxArgs, "additional-flux-args", nil, "A comma separated list of additional arguments to pass to fluxd")
 
 	// Hide and deprecate "git-paths", which was wrongly introduced since its inconsistent with fluxd's git-path flag
 	cmd.Flags().MarkHidden("git-paths")


hiddeco:
@squaremo if I recall correctly there was a reason we did not add this to the command, is this memory of mine correct?

squaremo:
> if I recall correctly there was a reason we did not add this to the command, is this memory of mine correct?

The argument was keeping it simple -- the aim of `fluxctl install` is to have a simple way to do a bare bones installation.

There's any number of things you might want to do with the resources that get generated, so instead of trying to anticipate all of them, we can just let people write to files then do whatever (e.g., use them as the base of a kustomization).

stefanprodan:
When we add the Flux config to the install command any flag could be set that way.

chancez:
This all seems reasonable, but with that kind of argument you could easily say `fluxctl install` shouldb;t exist, since you can use the raw manifests, helm, or otherwise. I suppose I'll just stop using `fluxctl` if it's not intended to actually be used for production configurations.

stefanprodan:
@chancez for production deployments I would use the helm chart either with `helm install` or `helm template | kubectl apply -f -`. The Flux helm chart allows you to supply [any Flux arg](https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L225), you can setup init containers, your own service account, git config, kube config, env vars, annotations and more.

chancez:
I was using the helm chart previously, but going forward we're trying to completely avoid helm, so I may just use kustomize for bootstrapping flux.

squaremo:
> with that kind of argument you could easily say fluxctl install shouldn't exist, since you can use the raw manifests, helm, or otherwise

@chancez That's a fair observation. `fluxctl install` is handy for generating base configurations with some minor variations -- this is how it's used for end-to-end testing, for example -- but not much further than that. perhaps it ought to be deprecated in favour of a more specialised command. `fluxctl base-config` or something (off the top of my head).

I'm going to close this PR, to take it off the radar, since the conclusion seems to be that there's a way for you to get what you need via Kustomize. Thank you for taking time to prepare the PR and talk it through with us, it's appreciated.

PR #2852, opened 2020-02-17
Chart and documentation updates for `--registry-include-image`.

squaremo:


docs/faq.md
```diff
@@ -354,10 +354,25 @@ You can completely disable registry scanning by using the
...
 
-To exclude images from Docker Hub and Quay.io, use:
+ * `--registry-include-image` takes patterns to be included; no values
+   (the default) means "include everything". If you provide a pattern,
+   _only_ images matching the pattern will be included (less any that
+   are explicitly excluded); and,
+ * `--registry-exclude-image` takes patterns to be excludes; the
+   default is to exclude the Kubernetes base images (`k8s.gcr.io/*`).
+
+To include only images from Docker Hub organisation `exampledotcom`,
```
hiddeco:
```suggestion
To **include** only images from Docker Hub organisation `exampledotcom`,
```

would be my suggestion based on the overall look-and-feel of the document when pretty-printed.

squaremo: I think I should rephrase so it's not symmetrical with the explanation of exclude (since they do not behave symmetrically).

hiddeco: I like that option too.

squaremo:
Thanks for the reviews Hidde!

Issue #2841, opened 2020-02-11
Disable garbage collection for specific resource


phillebaba:
**Describe the feature**
Garbage collection works great but there are sometimes scenarios where it could be very inconvenient if a resource was accidentally deleted. This is especially for resources that map to external dependencies such as LBs or other cloud resources create by for example [aws-servicebroker](https://github.com/awslabs/aws-servicebroker). In a perfect world this would not happen but the reality is that mistakes happen. Especially as the amount of manifests grows and large amounts of resources can be removed from being generated by small line changes. So it would be nice to be able to selectively disable garbage collection for certain resources.

**What would the new user story look like?**
1. Provide a new annotation that could be added to any resource. It could look like `fluxcd.io/garbage-collection=false` or `fluxcd.io/gc=false` indicating that the resource should not be garbage collected.
2. User starts up Flux using the `--sync-garbage-collection` flag
3. New manifest is created with the new annotation and committed to git
4. Flux syncs with git repository deploys the resource
5. Manifest is removed and committed to git
6. Flux skips garbage collection as the resource 

**Expected behavior**
When the resource is removed from git and is marked as an orphaned resource by flux during the sync process it should be skipped by the garbage collector.

squaremo:
I can see the use case for this -- well explained.

There's an annotation `fluxcd.io/ignore` that tells flux to ignore a resource with respect to syncing. It's explained here: https://docs.fluxcd.io/en/1.18.0/faq.html#can-i-temporarily-make-flux-ignore-a-manifest. I think you are describing something different, but let's check: how would `fluxcd.io/gc=false` interact with fluxcd.io/ignore=true`?

phillebaba:
I had a look at the ingore annotation also but it has the downside that any changes would not be synced. Using the ignore annotation would be a two step process. You would first have to commit your manifests without the annotations, and wait for flux to sync them to the cluster. Then add the annotation and commit once again. That would disable any changes in the future, while I just want to stop the GC from removing the resource, but still allowing it to be updated.

Seems like it would be feasible to implement [here](https://github.com/fluxcd/flux/blob/3f0a9621405a5ece7ff5362daf917f618a5cd66c/pkg/cluster/kubernetes/sync.go#L149-L151) where the sync checks if dry run is enabled or not, so it would also skip resources with the annotation.

[...]

Issue #2671, opened 2019-12-09
Have a single document that has a nice list of all annotations.


jdfalk:
**Describe the feature**
Just like you want a clear and concise description of the feature, I want a single page where I can see ALL annotations. I don't want to have to grok them out from the examples. I want a place I can point the devs to and say "here's the list, use them appropriately." 

**What would the new user story look like?**
A single doc in the reference section that lists all the annotations. Look at the Istio one for an example of how to do it: https://istio.io/docs/reference/config/annotations 

**Expected behavior**
I can say to new dev on the team, "Hey go to https://docs.fluxcd.io/en/stable/references/annotations.html and that should answer your questions."

This was brought up before in ticket #1178 but wasn't completely fixed. I appreciate examples, but for those of us who need a quick reference sheet it's so much easier to have one page where everything is listed. I would also require that if new annotations are added, it be a pre-submit or some level of merge that they update the docs.

2opremio:
@jdfalk would you be willing to give it a go?

jdfalk:
Sure. I won't be able to get to it until next week however.

2opremio:
Great! :tulip: There is no rush

kingdonb:
This issue has gone stale, Flux v1 is now in maintenance mode, and will be superseded by Flux v2 soon... but this document might still be useful even now, especially if it was written in the form of a piece of transitionary documentation.

I can't promise this will get prioritized, but it would absolutely make a lot of sense to have a list of annotations, (maybe even including those that were once used but are already deprecated and removed, from before FluxCD org... as in the Istio example page linked in the issue report.) ... any annotation that serves a function in Flux v1, with a note about what replaces this functionality in Flux v2.

It will not be as straightforward as

| feature | Flux 1 annotation | Flux 2 annotation |
| image automation | `fluxcd.io/automated` | `YYY` |
| garbage collection | `fluxcd.io/garbage-collection` | `ZZZ` |

... as the right-hand column won't generally refer to an Annotation in Flux v2, all of the required functionality is implemented on CRDs in the new APIs, I think generally a new feature corresponds with a given field in the spec, and annotations on downstream manifests are not used anymore (except for kyaml setters / comments that drive image automation controller.)

kingdonb:
I don't think we're aiming to have any more direct comparisons with Flux v1, or call additional attention to Flux v1 if we can avoid it, since the focus is now squarely on delivering Flux v2 to GA here at Flux HQ.

[...]

PR #2858, opened 2020-02-19
Add DisableGarbageCollect annotation

phillebaba:
This change implements an annotation to disable garbage collection for specific resources.

https://github.com/fluxcd/flux/issues/2841

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -341,7 +341,7 @@ how/where to undo the change (cf [flux#1211](https://github.com/fluxcd/flux/issu
 annotation exists in either the cluster or in git, it will be respected, so you may need to remove
 it from both places.
 
-Additionally when garbage collection is enabled. Flux will not garbage collect resources in the cluster
+Additionally, when garbage collection is enabled, Flux will not garbage collect resources in the cluster
 with the ignore annotation if the resource is removed from git.
 
 ### Can I disable garbage collection for a specific resource?
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -234,7 +234,7 @@ Now restart `fluxd` to re-read the k8s secret (if it is running):
 
 `kubectl delete $(kubectl get pod -o name -l name=flux)`
 
-If you have installed flux through Helm, make sure to pass 
+If you have installed flux through Helm, make sure to pass
 `--set git.secretName=flux-git-deploy` when installing/upgrading the chart.
 
 ### How do I use a private git host (or one that's not github.com, gitlab.com, bitbucket.org, dev.azure.com, or vs-ssh.visualstudio.com)?
@@ -340,6 +340,18 @@ Mixing both kinds of annotations (in git, and in the cluster), can make it a bit
 how/where to undo the change (cf [flux#1211](https://github.com/fluxcd/flux/issues/1211)). If the
 annotation exists in either the cluster or in git, it will be respected, so you may need to remove
 it from both places.
+
+Additionally when garbage collection is enabled. Flux will not garbage collect resources in the cluster
+with the ignore annotation if the resource is removed from git.
+
+### Can I disable garbage collection for a specific resource?
+
+Yes. By adding the annotation below to a resource Flux will sync updates from git, but it will not
+garbage collect when the resource is removed from git.
+
+```yaml
+    fluxcd.io/ignore: sync_only
+```
 
 ### How can I prevent Flux overriding the replicas when using HPA?
 
--- a/pkg/cluster/kubernetes/sync.go
+++ b/pkg/cluster/kubernetes/sync.go
@@ -145,8 +145,14 @@ func (c *Cluster) collectGarbage(
 
 		switch {
 		case !ok: // was not recorded as having been staged for application
-			if res.Policies().Has(policy.DisableGarbageCollect) {
-				c.logger.Log("info", "not deleting resource; garbage colelction annotation in file", "resource", res.ResourceID())
+			if res.Policies().Has(policy.Ignore) {
+				c.logger.Log("info", "skipping GC of cluster resource; resource has ignore policy true", "dry-run", dryRun, "resource", resourceID)
+				continue
+			}
+
+			v, ok := res.Policies().Get(policy.Ignore)
+			if ok && v == "sync_only" {
+				c.logger.Log("info", "skipping GC of cluster resource; resource has ignore policy sync_only ", "dry-run", dryRun, "resource", resourceID)
 				continue
 			}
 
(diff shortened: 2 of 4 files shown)


pkg/cluster/kubernetes/sync_test.go
```diff
@@ -464,6 +485,18 @@ metadata:
 		test(t, kube, "", ns1+defs1+defs2+ns3+defs3, false)
 	})
 
+	t.Run("sync adds and GCs skips resources", func(t *testing.T) {
```
squaremo: This test does verify what it claims to; but I think it should also verify that resources are updated (i.e., not ignored) when they have a gc_disabled annotation.

pkg/cluster/kubernetes/sync.go
```diff
@@ -145,6 +145,11 @@ func (c *Cluster) collectGarbage(
 
 		switch {
 		case !ok: // was not recorded as having been staged for application
+			if res.Policies().Has(policy.DisableGarbageCollect) {
+				c.logger.Log("info", "not deleting resource; garbage colelction annotation in file", "resource", res.ResourceID())
```
squaremo:
```suggestion
				c.logger.Log("info", "not deleting resource; garbage collection disabled by annotation", "resource", res.ResourceID())
```

(typo, and the annotation is in the cluster resource, not the file)

squaremo:
Thinking aloud: I wonder if this could be value of "ignore", instead of its own annotation.

Here are the combinations of annotations and effects, when there are both:

| ignore \ disable_gc | false | true |
|---------------------------|--------|--------|
| **false**                  | sync and gc | sync, no gc |
| **true**                    | no sync, no gc | no sync, no gc |

There's only three different behaviours there, so they don't operate independently -- one of them is redundant. So you could represent those three possibilities as three values:

    fluxcd.io/ignore: "true" | "false" | "gc"

One counter-argument is that it's weird to have a three-valued boolean like that (albeit they are only stringly pictures of booleans).

wdyt @phillebaba ?

phillebaba:
There would be some benefit with reducing the number of annotations, the risk is that having three values would confuse people. So documentation of the behavior is important.

The `ingore: true` + `disable_gc: false` combination would only disable gc if we assume ignore works as it was intended from the beginning.

Does these descriptions match the intended behavior of the different values? Assuming that the end user has enabled GC.
| value | description | 
| --- | --- | 
| false | Resources are synced when added to git and garbage collected when removed from git |
| no_gc | Resources are synced when added to git but will not be garbage collected when removed from git. |
| true | Resources are not synced when added to git. Additionally if the value is set manually outside of flux the resource will no longer be synced by flux, and it will not be garbage collected if removed from git. |

Am I missing something in the description?

squaremo:
> There would be some benefit with reducing the number of annotations, the risk is that having three values would confuse people. So documentation of the behavior is important.

Yes, I agree on both points. There's an opportunity to change the formulation slightly, so long as it's backward-compatible -- say, `ignore: {false,none}|sync_only|{true,sync_and_gc}`. If that would mitigate the confusion ..

> The `ingore: true` + `disable_gc: false` combination would only disable gc if we assume ignore works as it was intended from the beginning.

Right, that's presupposed. I find it hard to imagine that someone is relying on it _not_ working like that (even I had difficulty getting my head around the fact that it didn't).

> Resources are synced when added to git but will not be garbage collected when removed from git.

I would say "synced _from_ git", since updates in git will still be applied to the cluster. Similarly, in the description of "true".

squaremo:
OK so I think the goal for this PR should be to make it work as we've settled on above -- are you happy to adjust it, @phillebaba? Alternatively, I could make ignore mean ignore-for-gc-too in another PR, then you could work on top of that.

phillebaba:
I can implement both parts, don't think that is an issue. You are right about going with the multi value annotation solution as it would simplify logic.

I have a lot to do at work this week so I can't get this done until the weekend.

squaremo:
Cool, no worries @phillebaba, any time you lend to this project is appreciated :-)

phillebaba:
I change the behavior so that you disable GC by setting `sync_only` as the value for the ignore annotation.

I also added some more info in the FAQ, but i think that getting a single page that documents all annotations as described in https://github.com/fluxcd/flux/issues/2671 would be great!

A table like this would describe the ignore annotations behavior pretty well.
| Value | Sync | GC |
| --- | --- | --- |
| false | x | x |
| sync_only | x |  |
| true |  |  |

docs/faq.md
```diff
@@ -341,6 +341,18 @@ how/where to undo the change (cf [flux#1211](https://github.com/fluxcd/flux/issu
 annotation exists in either the cluster or in git, it will be respected, so you may need to remove
 it from both places.
 
+Additionally when garbage collection is enabled. Flux will not garbage collect resources in the cluster
```
squaremo:
```suggestion
Additionally, when garbage collection is enabled, Flux will not garbage collect resources in the cluster
```

docs/faq.md
```diff
@@ -341,6 +341,18 @@ how/where to undo the change (cf [flux#1211](https://github.com/fluxcd/flux/issu
...
+with the ignore annotation if the resource is removed from git.
+
+### Can I disable garbage collection for a specific resource?
+
+Yes. By adding the annotation below to a resource Flux will sync updates from git, but it will not
+garbage collect when the resource is removed from git.
```
squaremo: :+1:

pkg/cluster/kubernetes/sync.go
```diff
@@ -145,6 +145,17 @@ func (c *Cluster) collectGarbage(
 
 		switch {
 		case !ok: // was not recorded as having been staged for application
+			if res.Policies().Has(policy.Ignore) {
+				c.logger.Log("info", "skipping GC of cluster resource; resource has ignore policy true", "dry-run", dryRun, "resource", resourceID)
+				continue
+			}
+
+			v, ok := res.Policies().Get(policy.Ignore)
+			if ok && v == "sync_only" {
```
squaremo: Could do with a constant for this

pkg/cluster/kubernetes/sync_test.go
```diff
@@ -703,7 +704,74 @@ spec:
...
+  name: dep1
+  namespace: foobar
+  annotations: {flux.weave.works/ignore: "sync_only"}
+`
+
+		// sync namespace and deployment
+		test(t, kube, ns1+dep1, ns1+dep1, false)
+
+		// sync namespace only but expect deployment to not be GC
+		test(t, kube, ns1, ns1+dep1, false)
```
squaremo: Nice and simple test case :+1:

Issue #959, opened 2018-02-22
fluxctl release --exclude should support wildcards


marccarre:
Instead of having to do:
```
fluxctl --token $TOKEN release --update-all-images --all \
    --exclude=foo:deployment/bar \
    --exclude=foo:deployment/baz \
    --exclude=foo:deployment/cat \
    --exclude=foo:deployment/cow \
    --exclude=foo:deployment/dog \
    --exclude=foo:deployment/fly
```
I wish I could do:
```
fluxctl --token $TOKEN release --update-all-images --all \
    --exclude=foo:deployment/*
```

rndstr:
Somewhat related to https://github.com/weaveworks/flux/issues/902 which I've mildly been looking into. I thought about `--controller=foo:deployment/*` instead of `--all --namespace=foo`but the problem with using 
The problem I see with `*` is it will be expanded by the shell, so you would need to quote it which is quite error-prone.

PR #2860, opened 2020-02-19
Added exclude wildcard support for fluxctl release

sysdevguru:
Addresses #959

### DIFF
--- a/cmd/fluxctl/release_cmd.go
+++ b/cmd/fluxctl/release_cmd.go
@@ -137,6 +137,19 @@ func (opts *workloadReleaseOpts) RunE(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+		// check wildcard exclude
+		ns, _, srv := s.Components()
+		if srv == "*" {
+			ctx := context.Background()
+			cs, err := opts.API.ListServices(ctx, ns)
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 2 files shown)


squaremo:
Thanks for putting in a PR :+1:

Can you describe here the intended behaviour?

sysdevguru:
`fluxctl release --update-all-images --all --exclude=<namespace>:<workload>/*` will exclude `workload` in `namespace` from releasing.

Users can face case that they need this wildcard exclude support.
Related to #959

kingdonb:

I like the intent of this change, and am inclined to consider it for merge in spite of the feature-ness, due to the non-breaking nature, but due to structural differences in the CI/merge process today, I cannot even try to merge it without first adding your DCO to satisfy the CI checks.

If you are still interested in merging this, please rebase and submit again, following the DCO bot's instructions to use `--sign-off`. I cannot promise that it will get merged as it technically represents a new feature, but I could take another look. I want to let you know also that Flux v2 is now beyond the Feature Parity milestone, in case you haven't heard the news.

Users of Flux v1 are strongly encouraged to begin upgrading soon, if you have not already. Image Automation has been updated and now has a new design which includes some breaking changes. The design is not final until some time later, it would be considered finalized around GA or 1.0 release, and so there is still plenty of time to suggest improvements.

Thanks again for your contributions. I'm closing this for now.

PR #2862, opened 2020-02-20
Reduce SSH config complexity in Helm chart

jeff-minard-ck:
The previous logic of "has newline, split and loop each line at right indentation" can be trimmed to a single `nindent` function call. Additionally, the second `if` check is redundant.


kingdonb:
There are merge conflicts, I think this PR still has promise but the section that it updates has grown

@jeff-minard-ck are you still interested in merging this, if so would you be willing to rebase and try resolving the conflict?

It looks like the same pattern this change clears up has been used again, so now there might be room for two `nindent` calls.

kingdonb:

PR #2865, opened 2020-02-21
Dedicated error for unresolvable Git hostname

### LINKED ISSUE #1116: "git repo not ready" error message not exactly useful
Upon setting up Flux I mistyped the git repo URL (`github` instead of `github.com`) and read "git repo not ready" over and over again in the log. The error message in question is [declared here](https://github.com/weaveworks/flux/blob/8c0f7946d538a6cbdbcabc32773c478dd6e5d223/git/repo.go#L28).

My first hunch was that the deploy key was wrong or not set up - that's how I understood "ready". Maybe we could change the error message to `can't connect to git repo %s.` or some such? It'd be good to spell out exactly which git repo we're trying to connect with, so you get another chance to see your mistake in the logs.

sysdevguru:
Fixes #1116

Just made another PR because the origin branch was deleted.
Please review the change, @hiddeco 
Thanks in advance

### DIFF
--- a/cmd/fluxctl/sync_cmd.go
+++ b/cmd/fluxctl/sync_cmd.go
@@ -45,6 +45,8 @@ func (opts *syncOpts) RunE(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("no git repository is configured")
 	case git.RepoReady:
 		break
+	case git.RepoUnreachable:
+		return fmt.Errorf("can not connect to git repository with URL %s\n\nFull error message: %v", gitConfig.Remote.URL, gitConfig.Error)
 	default:
 		if gitConfig.Error != "" {
 			return fmt.Errorf("git repository %s is not ready to sync\n\nFull error message: %v", gitConfig.Remote.URL, gitConfig.Error)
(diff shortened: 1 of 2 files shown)


cmd/fluxctl/sync_cmd.go
```diff
@@ -45,6 +45,8 @@ func (opts *syncOpts) RunE(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("no git repository is configured")
 	case git.RepoReady:
 		break
+	case git.RepoUnreachable:
+		return fmt.Errorf("can't connect to git repo with url %s, please check your DNS or repository url\n\nFull error message: %v", gitConfig.Remote.URL, gitConfig.Error)
```
hiddeco:
```suggestion
		return fmt.Errorf("can not connect to git repository with URL %s\n\nFull error message: %v", gitConfig.Remote.URL, gitConfig.Error)
```

To make it align with other errors.

hiddeco: Once taken care of, can you please also smash your commits together into one? This keeps our `master` history clean.

sysdevguru: Yes, I have changed the error message as you suggested.

sysdevguru:
> Thank you @sysdevguru ⭐️

You are welcome

PR #2872, opened 2020-02-22
chart: support priority class

ArchiFleKs:
Add `priorityClassName` to set priority class for Flux.


chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.18.0"
-version: 1.2.0
+version: 1.3.0
```
stefanprodan: Please undo this, the chart version bump is done separate PR.

PR #2876, opened 2020-02-25
fix: update pluralization of yaml to non-disputable version

mauricemeyer:
# Changelog version

This PR updates the pluralization of YAML in an error message from „yamels“ to „yaml files“.

# Long version

We were debugging an issue with failing git repository clones today. While doing so, the pluralization of „yaml“ to „yamels“ was bugging me extremely, so I thought about a version of it that was possibly agreeable to more people. 

Therefore, I changed it to „yaml files“.

### DIFF
--- a/pkg/cluster/kubernetes/resource/load.go
+++ b/pkg/cluster/kubernetes/resource/load.go
@@ -30,7 +30,7 @@ func Load(base string, paths []string, sopsEnabled bool) (map[string]KubeManifes
 	for _, root := range paths {
 		err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
 			if err != nil {
-				return errors.Wrapf(err, "walking %q for yamels", path)
+				return errors.Wrapf(err, "walking %q for yaml files", path)
 			}
 
 			if charts.isDirChart(path) {


hiddeco:
This PR erases signature history :cry: 

https://github.com/fluxcd/flux/search?q=yamel&type=Commits
https://github.com/fluxcd/flux/search?q=yamel+is%3Aissue&type=Issues
https://github.com/fluxcd/flux/search?q=yamel+is%3Aissue+is%3Apr&type=Issues

shibumi:
> This PR erases signature history 

Well, everything beautiful needs a point where it starts. I would appreciate the term **yaml files** instead of **yamels**.

Why?

* When **yaml files** is used, it's clear that both parties (if human or not) talk about **yaml**
* **yamels** is confusing, when I had a look on this term the first time, I thought it's a new fancy file format or whatever (not kidding).

hiddeco:
No worries, I 100% understand the PR from a UX perspective.

Having known @squaremo for a while, I know both `yamel` and `yamels` are persistent in his vocabulary. I actually had a good laugh while Googling for `"yamels"` as `squaremo/kubeyaml` (the tool we use for replacing i.e. image tags while making image updates) is the only YAML related result on the first page.

squaremo:
I bless this retro-corrective to my vocabulary

PR #2878, opened 2020-02-26
Deprecate fluxctl install in favour of fluxctl base-config

squaremo:
Lately we have had requests to add flags to `fluxctl install` for
various options; this indicates people would like a non-Helm way to
install flux.

However `fluxctl install` isn't really it -- flags are not a good way
to control the complexity of even a modest Kubernetes configuration
like Flux's.

A better thing to do is to give people a way to get a workable base
config which they can then customise -- indeed, kustomize -- to their
hearts' content.

This commit in effect renames `fluxctl install` to `fluxctl
base-config`, to make it clearer that the expected use is to generate
a base config for further modification. `fluxctl install` remains, for
backward-compatibility, but outputs a deprecation notice to stderr.

TODO:
 - [ ] update docs that refer to `fluxctl install`
 - [ ] consider adding howtos for using `fluxctl base-config` with kustomize (e.g., to cover https://github.com/fluxcd/flux/issues/2541)

### DIFF
--- a/cmd/fluxctl/baseconfig_cmd.go
+++ b/cmd/fluxctl/baseconfig_cmd.go
@@ -1,0 +1,110 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/fluxcd/flux/pkg/install"
+)
+
+type baseConfigOpts struct {
+	install.TemplateParameters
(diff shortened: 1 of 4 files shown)


kingdonb:
I think this has gone stale and merging it would probably represent a breaking change at this point.

`fluxctl install` is superseded by `flux install`, which is actually the supported way to install Flux v2 now 👍 

Closing this as won't merge

Issue #2857, opened 2020-02-19
Flux refreshing image


MilanDasek:
We're experiencing lot of calls from Flux to AWS API.

Especially these calls:
GetDownloadUrlForLayer
BatchGetImage
ListImages

I have checked those calls are for ECR images that belongs to EKS deployments where we have these annotations:
```
kind: Deployment
metadata:
  annotations:
    fluxcd.io/automated: "false"
    fluxcd.io/ignore: "true"
```

We have these versions:
```
image: docker.io/fluxcd/flux:1.18.0
image: docker.io/fluxcd/helm-operator:0.10.1
image: memcached:1.5.15
```

We have these flux settings:
```
    spec:
      containers:
      - args:
        - --log-format=fmt
        - --ssh-keygen-dir=/var/fluxd/keygen
        - --k8s-secret-name=flux-git-deploy
        - --memcached-hostname=flux-memcached
        - --sync-state=git
        - --memcached-service=
        - --git-url=git@bitbucket.org:%redacted%
        - --git-branch=%redacted%
        - --git-path=%redacted%
        - --git-readonly=false
        - --git-user=DevOps Team
        - --git-email=%redacted%
        - --git-set-author=false
        - --git-poll-interval=3m
        - --git-timeout=5m
        - --sync-interval=3m
        - --git-ci-skip=false
        - --git-label=%redacted%
        - --automation-interval=10m
        - --registry-poll-interval=10m
        - --registry-rps=100
        - --registry-burst=125
        - --registry-cache-expiry=2h
        - --registry-trace=false
        - --registry-ecr-region=us-east-1
        - --registry-ecr-include-id=%redacted%
        - --registry-ecr-exclude-id=%redacted%
        - --connect=ws://fluxcloud
```

Logs for example
```
ts=2020-02-19T14:12:53.792550178Z caller=warming.go:198 component=warmer info="refreshing image" image=%redacted%.dkr.ecr.us-east-1.amazonaws.com/%redacted% tag_count=124 to_update=5 of_which_refresh=5 of_which_missing=0
ts=2020-02-19T14:12:56.44288336Z caller=warming.go:198 component=warmer info="refreshing image" image=%redacted%.dkr.ecr.us-east-1.amazonaws.com/%redacted% tag_count=465 to_update=324 of_which_refresh=324 of_which_missing=0
```

I tried `--registry-disable-scanning` which stopped those checks, but we would like to keep some of them.

Please advise.

squaremo:
[...]

PR #2888, opened 2020-03-02
Updating Docs to ignore workloads

prometherion:
`fluxcd.io/ignore` has been deprecated but doc still points to it: the right group should be `flux.weave.works/ignore`

Related to #2857


hiddeco:
It is actually the other way around, `flux.weave.works` has been deprecated and `fluxcd.io` is the new annotation domain.

prometherion:
Ouch, awkward :D

Going to close the PR due to my bad!

Issue #2905, opened 2020-03-11
Scaling workloads


Willem1987:
**Describe the feature**
We want to keep both the image used and the number of replica's in sync for a workload. When we scale to 0 we don't want flux to undo this.

**Expected behavior**
fluxctl release --workload=namespace:deployment/dep_name --user=flux --message="Scale to 0" --update-scale=0

Either through update-scale or update-replicas i would like to be able to update the replicas. 

A workaround could be excluding the replicas tag in the patch during the sync?

Willem1987:
Related to https://github.com/fluxcd/flux/pull/2908

kingdonb:
Responded on the PR #2908 that you submitted. Closing, as Flux v1 is in maintenance mode.

Thanks for using Flux!

PR #2908, opened 2020-03-12
#2905 Allow changing workload scale with release command:

Willem1987:
Added number of replicas to list-workloads to view current scale
Added --update-scale flag to be able to track replicas in scm

### DIFF
--- a/cmd/fluxctl/error.go
+++ b/cmd/fluxctl/error.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 )
 
 type usageError struct {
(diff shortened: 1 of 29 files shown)


kingdonb:
Hi! Thank you for your contribution.

I am sorry this has gone so long without a review. There appears to be quite a lot of work here. If you've been able to use this in your own environment and benefit from the change you've made, that's great!

Flux v1 is in maintenance mode and is not adding new features at this point. We hope you're able to upgrade with us to Flux v2. I'm not aware of this feature in the new version, I think it's against the grain to add features that already exist elsewhere, like in kubectl – there is a `kubectl scale` verb which can be used with Flux deployments that omit the `replicas` count.

Many of the Flux and Flagger examples do this, eg. to allow the scale to be set by `HorizontalPodAutoscaler` which can raise and lower the scale in response to events. For this reason I don't think the feature would be accepted by Flux maintainers.

Thanks for using Flux, and thanks for sharing your contribution! We appreciate our users and hope you will upgrade with us to the next major release of Flux. The feature parity milestone has been announced as of Release 0.8, and users are being encouraged to upgrade to help validate that all use cases are better served by the new rewritten version of Flux.

Closing for now, but always happy to answer further questions if you have them.

PR #2937, opened 2020-03-25
e2e: Update gitsrv to v1.0.0

stefanprodan:
- update [gitsrv](https://github.com/fluxcd/gitsrv) deployment to v1.0.0
- download known_hosts from GitHub release
- use the downloaded known_hosts in tests


test/e2e/run.bash
```diff
@@ -20,6 +21,11 @@ BATS_EXTRA_ARGS=""
 source "${E2E_DIR}/lib/defer.bash"
 trap run_deferred EXIT
 
+function download_known_hosts() {
+  mkdir -p "${FLUX_ROOT_DIR}/bin"
```
hiddeco: I think the `bin` directory must either be renamed to `cache` or the `known_hosts` should be stored elsewhere.

PR #2939, opened 2020-03-26
Support multiple git-path arguments in helm chart

MihaiAnei:
This adds support for https://github.com/fluxcd/flux/pull/1297 which is something we use in order to be able to use the same repo for multiple K8S clusters (environments actually) that have slight differences between them. Those differences are in separate folders and flux watches a different set of folders depending on cluster. It easier than having multiple flux instances per cluster. The change was tested.

The following short checklist can be used to make sure your PR is of good
quality, and can be merged easily:
- [x] optional but much appreciated;


hiddeco:
Instead of passing multiple flags to `fluxd`, you can pass a comma separated string to a single flag and it will have the same effect. This is already possible _as is_ by performing e.g.:

```ssh
helm upgrade -i flux flucd/flux --set git.path="path1/,path2/"
```

MihaiAnei:
> Instead of passing multiple flags to `fluxd`, you can pass a comma separated string to a single flag and it will have the same effect. This is already possible _as is_ by performing e.g.:
> 
> ```
> helm upgrade -i flux flucd/flux --set git.path="path1/,path2/"
> ```

I did not know that. Thanks!

PR #2940, opened 2020-03-26
ci: Migrate tests to GitHub actions

stefanprodan:
GitHub Actions migration (ref: #2944):
- [x] run unit tests
- [x] verify codegen
- [x] build binaries and containers
- [x] run e2e tests (four Kubernetes Kind clusters matrix)

GH Actions specific changes:
- cache go modules
- run one Kubernetes cluster per VM
- upload fluxd container image, fluxctl, kubectl, kustomize and sops to GH artifacts 
- download the artifacts in e2e jobs to avoid rebuilding binaries and containers
- cleanup artifacts daily 

Conclusions:
- the random kind/docker/kubeadm failures that we experienced with CircleCI are gone
- no more `sudo killall apt-get`, apt-get doesn't hang  like in CircleCI
- the e2e tests take the same amount of time on GH Actions free as in CircleCI paid account
- the major downside to GH Actions is the lack of docker image caching between jobs
- you have to use docker save/load and upload/download artifacts between jobs
- you can't delete artifacts during a GH workflow run 
- GH Actions UI is unreliable and buggy, most of the time the logs are not displayed until the job finishes 

Overall I'm happy with GitHub Actions even though it was a painful migration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,7 @@ jobs:
       - run: make test TEST_FLAGS="-race -tags integration -timeout 5m"
       - run: make check-generated
       - run: make all
+      #- run: E2E_KIND_CLUSTER_NUM=4 make e2e
       - run: E2E_KIND_CLUSTER_NUM=4 make e2e
       - save_cache:
           key: cache-{{ checksum "Makefile" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,8 +101,6 @@ jobs:
       - run: make test TEST_FLAGS="-race -tags integration -timeout 5m"
       - run: make check-generated
       - run: make all
-      #- run: E2E_KIND_CLUSTER_NUM=4 make e2e
-      - run: E2E_KIND_CLUSTER_NUM=4 make e2e
       - save_cache:
           key: cache-{{ checksum "Makefile" }}
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,9 +97,6 @@ jobs:
           keys:
             - go-mod-{{ checksum "go.mod" }}
             - go-mod-
-      # Note that the test timeout is per Go package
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -101,7 +101,6 @@ jobs:
       - run: make test TEST_FLAGS="-race -tags integration -timeout 5m"
```
dholbach: With https://github.com/fluxcd/flux/pull/2940/files#diff-60fe268f71ae7f8409aff40d0a3e51d8R27 is above line still necessary?

dholbach:
Well done @stefanprodan - you got this done in next to no time!

.circleci/config.yml
stefanprodan: Good point, I've removed it.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/upload-artifact@v1
+        with:
+          name: cache
+          path: cache
+  e2e-testing:
+    runs-on: ubuntu-latest
+    needs: unit-testing
+    strategy:
+      matrix:
+        test: ['10_* 11_* 12_*', '13_* 14_*', '15_* 16_* 17_*', '20_* 21_* 22_*']
```
hiddeco: Would something like

```
        test: ['1/4', '2/4', '3/4', '4/4']
```

be an option (with additional logic added to `run-gh.bash`)? This would make (newly) added tests run without (forgetting) to add them here.

stefanprodan: I've balanced the tests to take the same amount of time, for example running 15 in shard 2 will take 18m

hiddeco: Ack.

stefanprodan: I think we should refactor the tests and have the balancing logic in bash, but this can come later.

PR #2948, opened 2020-03-30
Fix more links

dholbach:
As a follow-up from https://github.com/fluxcd/helm-operator/pull/347 here is the link checking being updated and a bunch of old links fixed too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -941,8 +941,8 @@ Many thanks for contributions from @2opremio, @AndriiOmelianenko,
 [weaveworks/flux#2138]: https://github.com/weaveworks/flux/pull/2138
 [weaveworks/flux#2140]: https://github.com/weaveworks/flux/pull/2140
 [weaveworks/flux#2142]: https://github.com/weaveworks/flux/pull/2142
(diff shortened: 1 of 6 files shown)


dholbach:
Hohum... not sure what to do about this:

 README.md
	ERROR	https://chattigo.com
		X509: certificate signed by unknown authority
	ERROR	https://img.shields.io/badge/latest-documentation-informational
		 (HTTP error 522)
	ERROR	https://www.blablacar.com/
		Forbidden (HTTP error 403)
	ERROR	https://www.canva.com/
		Forbidden (HTTP error 403)
	ERROR	https://www.underarmour.com
		Forbidden (HTTP error 403)

dholbach:
I never knew that checking links would turn into a minefield like this. What I attempt to do in this PR is
- use `liche` (through the GH Action) for the docs we generate using `mkdocs` (`docs` → `site`) - this works great.
- use `markdown-link-checker` (through GH action) for checking `internal/docs` and the top-level markdown files (the latter is where we have LOTS OF links that cause us to get all kinds of spurious results)

Might need some help here. :heart_eyes:

dholbach:
I'll make this PR just about fixing links, so we can merge it ASAP. #2956 is about checking links.

Issue #2945, opened 2020-03-27
Drop support for HelmRelease v1beta1 and v1alpha2


stefanprodan:
I think it's time to remove the [integrations](https://github.com/fluxcd/flux/tree/master/integrations) packages and drop support for `flux.weave.works/v1beta1` and `helm.integrations.flux.weave.works/v1alpha2`.

It's been 9 months since we released `flux.weave.works/v1`, those that haven't upgraded will probably never do.

Having these clientsets in Flux makes it hard to upgrade the Kubernetes client-go since we can't re-generate them as the codegen was removed from Flux repo when we moved Helm Operator to its own repository. Currently this is blocking the update to Kubenetes client-go 1.18.0.

hiddeco:
As already discussed in private, +1 for this.

PR #2949, opened 2020-03-30
Remove support for HelmRelease alpha and beta

stefanprodan:
This PR removes the HelmRelease `v1beta1` and `v1alpha2` clients from Flux codebase. This means Flux will no longer update container images for:
* `helmreleases.helm.fluxcd.io/v1beta1`
* `fluxhelmreleases.helm.integrations.flux.weave.works/v1alpha2` 

The Helm Operator users who upgraded to `helmreleases.helm.fluxcd.io/v1` will not be impacted by this change. 

Fix: #2945


stefanprodan:
@hiddeco I removed the last traces of `FluxHelmRelease`.

Issue #2951, opened 2020-03-31
Bring portforward into Flux codebase 


stefanprodan:
Flux uses https://github.com/justinbarrick/go-k8s-portforward this project hasn't been maintained and has no sem ver release compatible with go mod. As it's just a couple of lines of code I propose we bing this into Flux code base.

PR #2952, opened 2020-03-31
Add portforward package

stefanprodan:
Based on https://github.com/justinbarrick/go-k8s-portforward

Fix: #2951

### DIFF
--- a/pkg/portforward/portforward.go
+++ b/pkg/portforward/portforward.go
@@ -1,0 +1,207 @@
+package portforward
+
+// based on https://github.com/justinbarrick/go-k8s-portforward
+
+import (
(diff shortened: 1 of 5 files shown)


pkg/portforward/portforward.go
```diff
@@ -0,0 +1,207 @@
+package portforward
+
+// based on https://github.com/justinbarrick/go-k8s-portforward
```
squaremo, line 3: "licensed under the Apache License 2.0, and adapted in the following ways:
..."

(the latter bit only if you changed anything significant)

stefanprodan: I changed only the tests

PR #2982, opened 2020-04-07
Add AWS China region domain to recognized ECR hosts

### LINKED ISSUE #2964: Error of fetching image metadata from AWS ECR(China region)
**Describe the bug**

My EKS is running on AWS China region(cn-north-1) with version 1.15, and I have successfully set up Flux using Helm installation. The Flux works well and it could access Git with deploy key and create/update applications thru configurations stored in Git repo. The issue was that Flux cannot fetch app's image metadata from AWS ECR, it turned out that when the newer image was created in ECR repo, the Flux couldn't update configurations accordingly in Git repo then update the applications.
I have checked my EKS node's IAM role, ECR ReadOnly policy was attached properly. And I have even attached ECR FullyAccess policy, but it didn't work. I have also tried IRSA, bonded an IAM role for Flux service account with ECR FullyAccess policy, but still failed.
The weird thing is that I installed EKS+Flux in AWS region us-east-1 with the same configurations in China region(without modify anything like node IAM role or IRSA), the Flux works very well with EKS&ECR, no issue came out...
And I also just checked the source code of "aws.go" https://github.com/fluxcd/flux/blob/master/pkg/registry/aws.go in Flux repo, I noticed that one of const ecrHostSuffix is set to ".amazonaws.com", but for AWS China region the ECR registry URLs should looks like "account-id.dkr.ecr.region.amazonaws.com.cn" according to AWS documents https://docs.amazonaws.cn/en_us/aws/latest/userguide/endpoints-Beijing.html, https://docs.amazonaws.cn/en_us/aws/latest/userguide/ecr.html and https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/aws/aws_credentials.go. So I guess maybe this the main cause for this issue. But I am not 100% sure whether I am right.

**To Reproduce**
Strictly follow the guide [https://docs.fluxcd.io/en/latest/tutorials/get-started-helm/](url) to install Flux thru Helm v3, and EKS is running on AWS China Region: cn-north-1

**Expected behavior**

Expected someone to help point me how to fix this issue.

**Logs**

ts=2020-03-31T18:23:19.267382623Z caller=images.go:159 component=sync-loop err="fetching image metadata for 634072196806.dkr.ecr.cn-north-1.amazonaws.com.cn/gitops-app: item not in cache, last error: Get https://634072196806.dkr.ecr.cn-north-1.amazonaws.com.cn/v2/gitops-app/tags/list: no basic auth credentials"
ts=2020-03-31T18:23:23.195082293Z caller=warming.go:180 component=warmer canonical_name=634072196806.dkr.ecr.cn-north-1.amazonaws.com.cn/gitops-app auth={map[]} err="requesting tags: Get https://634072196806.dkr.ecr.cn-north-1.amazonaws.com.cn/v2/gitops-app/tags/list: no basic auth credentials"

**Additional context**

- Flux version: 1.18.0
- Kubernetes version: AWS EKS 1.15
- Git provider: Github
- Container registry provider: AWS ECR

hiddeco:
Fixes #2964.

### DIFF
--- a/pkg/registry/aws.go
+++ b/pkg/registry/aws.go
@@ -34,6 +34,7 @@ const (
 	embargoDuration = 10 * time.Minute
 
 	EKS_SYSTEM_ACCOUNT = "602401143452"
+	EKS_SYSTEM_ACCOUNT_CN = "918309763551"
 )
 
 // AWSRegistryConfig supplies constraints for scanning AWS (ECR) image
--- a/pkg/registry/aws.go
+++ b/pkg/registry/aws.go
@@ -23,7 +23,9 @@ import (
 
 const (
 	// For recognising ECR hosts
-	ecrHostSuffix = ".amazonaws.com"
+	awsPartitionSuffix = ".amazonaws.com"
+	awsCnPartitionSuffix = ".amazonaws.com.cn"
+
 	// How long AWS tokens remain valid, according to AWS docs; this
 	// is used as an upper bound, overridden by any sooner expiry
 	// returned in the API response.
@@ -47,6 +49,15 @@ func contains(strs []string, str string) bool {
 		if s == str {
 			return true
 		}
+	}
+	return false
+}
+
+func validECRHost(domain string) bool {
+	switch {
+	case strings.HasSuffix(domain, awsPartitionSuffix):
+	case strings.HasSuffix(domain, awsCnPartitionSuffix):
+		return true
 	}
 	return false
 }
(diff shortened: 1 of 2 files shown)


pkg/registry/aws.go
```diff
@@ -51,6 +54,15 @@ func contains(strs []string, str string) bool {
 	return false
 }
 
+func validECRHost(domain string) bool {
+	switch {
+	case strings.HasSuffix(domain, awsPartitionSuffix):
```
squaremo, line 59: No implicit fallthrough in Go!

bboreham: #3002

Issue #2984, opened 2020-04-07
Flux should continue despite failure to connect to fluxcloud


derrickburns:
If Flux cannot connect to fluxcloud, it will not make progress.  The problem is that one cannot fix this problem using GitOps.

Instead, Flux should continue performing updates.

hiddeco:
Related to #2963.

hiddeco:
Can you describe in what way Flux was unable to connect to fluxcloud? I scaled down fluxcloud and although Flux was chatty about being unable to connect it did still synchronize a new made commit.

```
ts=2020-04-07T20:17:34.193599919Z caller=upstream.go:118 component=upstream err="executing websocket ws://fluxcloud/v11/daemon: dial tcp 10.96.225.106:80: connect: connection refused"
ts=2020-04-07T20:17:37.367327299Z caller=loop.go:133 component=sync-loop event=refreshed url=ssh://git@github.com/hiddeco/test-repo branch=master HEAD=b701a53c921d5db7190138410e8c1b81723f6233
ts=2020-04-07T20:17:37.403196892Z caller=sync.go:73 component=daemon info="trying to sync git changes to the cluster" old=c03c3ef5f5880a146df60ffb81d676d866135077 new=b701a53c921d5db7190138410e8c1b81723f6233
ts=2020-04-07T20:17:37.601686244Z caller=sync.go:539 method=Sync cmd=apply args= count=2
ts=2020-04-07T20:17:37.703718153Z caller=sync.go:605 method=Sync cmd="kubectl apply -f -" took=101.96328ms err=null output="namespace/bar created\ncronjob.batch/delete-me configured"
ts=2020-04-07T20:17:37.707404026Z caller=daemon.go:704 component=daemon event="Sync: b701a53, <cluster>:namespace/bar" logupstream=true
ts=2020-04-07T20:17:38.737594231Z caller=sync.go:395 component=daemon err="executing HTTP request: executing HTTP request: Post http://fluxcloud/v6/events: dial tcp 10.96.225.106:80: connect: connection refused"
```

awh:
@derrickburns could you provide us with some logs demonstrating this behaviour?

derrickburns:
I didn't capture any logs when the problem occurred. 

If it recurs, I will.

kingdonb:
Flux v1 is formally superseded since the GitOps Toolkit APIs have been declared stable:

https://fluxcd.io/docs/migration/timetable/

[...]

PR #2985, opened 2020-04-07
Update ratchet revision right after successful sync

### LINKED ISSUE #2963: Flux appears unable to proceed when connected for event emission and that fails
**Describe the bug**

We have our flux daemon connected to a local fluxcloud instance which is setup to post to both slack and a webhook.  When fluxcloud is unable to complete one of those things it gives the flux daemon a 500 error and this causes flux to stop proceeding the constantly retry the same commit over and over and over.  In out specific case the slack payload was working but the webhook was throwing a 500.  This caused flux from moving forward in our git history and retrying (succeessfully) to apply the same commit over and over and over.

**To Reproduce**

Deploy flux with fluxcloud and setup a webhook in flux cloud to something that will return a 500.  This will prevent flux from moving forward.

**Expected behavior**

It seems like the event emission aspect of flux should be "fire and forget" and if the downstream event handler doesn't do something properly it shouldn't block flux from moving forward.

**Logs**

```
ts=2020-04-02T15:41:28.486961895Z caller=sync.go:393 component=daemon err="executing HTTP request: 500 Internal Server Error Could not post to Webhook, status: 500\n"
fluxcloud Could not post to Webhook, status: 500 Internal Server Error
```

**Additional context**

- Flux version: 1.18
- Kubernetes version: 1.11
- Git provider: Github

hiddeco:
Fixes #2963, #2984

### DIFF
--- a/pkg/daemon/sync.go
+++ b/pkg/daemon/sync.go
@@ -83,6 +83,13 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 		return err
 	}
 
+	// Move the revision the sync state points to
+	if ok, err := ratchet.Update(ctx, c.oldTagRev, c.newTagRev); err != nil {
+		return err
+	} else if !ok {
+		return nil
+	}
+
 	// Determine what resources changed during the sync
 	changedResources, err := d.getChangedResources(ctx, c, d.GitTimeout, working, resourceStore, resources)
 	serviceIDs := resource.IDSet{}
@@ -112,13 +119,6 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 			// Abort early to ensure at least once delivery of events
 			return err
 		}
-	}
-
-	// Move the revision the sync state points to
-	if ok, err := ratchet.Update(ctx, c.oldTagRev, c.newTagRev); err != nil {
-		return err
-	} else if !ok {
-		return nil
 	}
 
 	err = refresh(ctx, d.GitTimeout, d.Repo)


squaremo:
Not moving the high water mark when the event fails to be posted isn't a bug, it's the design. That is exactly what the high water mark marks -- the last event that was sent successfully.

hiddeco:
Can you elaborate a bit on why this is the design? I would think the fact that it has been successfully applied on the cluster is what counts (as do the users who filed the issues), and not the fact that we have informed a 3rd party this has happened.

squaremo:
> Can you elaborate a bit on why this is the design?

The idea is to keep track of the point to which you have successfully notified about events. If it were moved before even trying to post an event, then you would lose events. This is an "at least once" design -- each event will be posted once, and maybe more if there are failures.

NB the high water mark is all about the events. Syncing always syncs to the HEAD of whichever branch.

You can argue (as Kyle does I think) that events should be "fire and forget" (or "at _most_ once"). But that's a different _design_, not a bug. A better way to avoid repeated notifications would be to deduplicate at the event receiver (i.e., fluxcloud). And if people still wanted at most once notifications, it'd be reasonable to add that behind a flag, I think.

rocktavious:
I would argue that the communication of what the "git tag" on the repository's meaning is not clear then.  Also as an enduser keeping track of the point to which a successful event is notified feels like a terrible design as the import job that flux does is keep the cluster in-sync with the state of the git repo.  The fact that a notification event didn't submit correctly feels like its not really part of the core functionality that flux serves.  I'd much rather have a metric tracking when failed event notifications happen so i can choose to alert on that or not.  Checking out the where the git tag is is much much more helpful indicator of where all my k8s clusters flux are at.  Specifically we run over 12 flux's and being able to do a git fetch on the tags and see where they are in the git history is a quick troubleshooting indicator that something is amiss.  When this is held up by event notifications its very misleading.

So if this truely is the route to be stayed with for the git tags then i think it needs to be HIGHLY communicated that the git tag's purpose is this (and i would bet most flux users do not know this or think otherwise)

Issue #2891, opened 2020-03-03
Rate limit reduced then raised within seconds


bboreham:
**Describe the bug**

In fluxd logs, rate limit to one site is reported being reduced then raised within seconds.

**To Reproduce**

Unclear, but several Flux maintainers have access to the cluster - it's Weave Cloud staging.

**Expected behavior**

After reducing the rate limit, leave it down for at least a few minutes, maybe an hour or more.

**Logs**

I wasn't sure if the whole logs might contain private information so filtered to just lines mentioning rate limiting.

```
ts=2020-03-02T12:34:07.462645791Z caller=rate_limiter.go:71 component=ratelimiter info="reducing rate limit" host=quay.io limit=5.00
ts=2020-03-02T12:34:07.463004422Z caller=warming.go:180 component=warmer canonical_name=quay.io/uswitch/kiam auth={map[]} err="requesting tags: error parsing HTTP 429 response body: invalid character '<' looking for beginning of value: \"<html>\\r\\n<head><title>429 Too Many Requests</title></head>\\r\\n<body bgcolor=\\\"white\\\">\\r\\n<center><h1>429 Too Many Requests</h1></center>\\r\\n<hr><center>nginx/1.12.1</center>\\r\\n</body>\\r\\n</html>\\r\\n\""
ts=2020-03-02T12:34:10.203194851Z caller=rate_limiter.go:88 component=ratelimiter info="increasing rate limit" host=quay.io limit=7.50
ts=2020-03-02T12:34:10.642936171Z caller=rate_limiter.go:88 component=ratelimiter info="increasing rate limit" host=quay.io limit=10.00
...
ts=2020-03-02T12:57:11.48896275Z caller=rate_limiter.go:71 component=ratelimiter info="reducing rate limit" host=quay.io limit=5.00
ts=2020-03-02T12:57:11.489245675Z caller=warming.go:180 component=warmer canonical_name=quay.io/damien.lespiau/quay-exporter auth={map[]} err="requesting tags: error parsing HTTP 429 response body: invalid character '<' looking for beginning of value: \"<html>\\r\\n<head><title>429 Too Many Requests</title></head>\\r\\n<body bgcolor=\\\"white\\\">\\r\\n<center><h1>429 Too Many Requests</h1></center>\\r\\n<hr><center>nginx/1.12.1</center>\\r\\n</body>\\r\\n</html>\\r\\n\""
[...]

PR #2986, opened 2020-04-07
Introduce cooldown period in rate limiter

hiddeco:
Addresses #2891.


ercpereda:
Is it a fixed value or there is a way to set it up?

PR #2989, opened 2020-04-07
Set `/dockercfg/` path for custom dockerconfig files

sebastianortizs4n:
The chart mounts the dockercfg.secretName in the `/dockercfg/`, but the path is not setted for the `configFileName`, so it does not takes the right configuration when using it. Otherwise the `configFileName` needs to be setted with the full path `/dockercfg/dockerconfigfile.json`

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -268,7 +268,7 @@ spec:
           - --registry-ecr-exclude-id={{ .Values.registry.ecr.excludeId }}
           {{- end }}
           {{- if .Values.registry.dockercfg.enabled }}
-          - --docker-config={{ .Values.registry.dockercfg.configFileName }}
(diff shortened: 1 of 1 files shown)


marchmallow:
See PR when this was changed last year. It is documented in the list of values that configFileName is a path and file name actually.. see the default value for it already includes `dockercfg` path and can be changed as need when deploying

stefanprodan:
@sebastianortizs4n we can't accept such a change to the chart as it will break all current Flux installs.

marchmallow:
When installing flux, you need to set eg: `--set registry.dockercfg.configFileName=/dockercfg/.dockerconfigjson` to set the full path and file name if the default not valid in your env. Though I agree that the parameter could have been better named as it may not be clear that the file name includes the path.

sebastianortizs4n:
Understood. It was really confusing using it and made me have to look the chart source to know what whas happing.

PR #2992, opened 2020-04-08
Make clear dockercfg.secretName and configFileName

sebastianortizs4n:
This patches tries to make clear the documentation of how the custom dockercfg via secret works.


sebastianortizs4n:
I haven't changed any links... looks like the check is not working right.

stefanprodan:
@dholbach can you please look at why the link checker is failing on all PR?

dholbach:
→ https://github.com/kubernetes/minikube/blob/master/docs/addons.md 404
It's in the logs!

Will send a PR.

stefanprodan:
@dholbach I see no logs here:
Screenshot 2020-04-14 at 13 13 10

dholbach:
Try the step before.

dholbach:
@stefanprodan https://github.com/fluxcd/flux/pull/3000

dholbach:
It should be fixed in master. If you rebase now, it should be fine. Looks like some links of the minikube docs changed. :man_shrugging:

PR #3001, opened 2020-04-14
Fix typo into fluxyaml-config-files.md doc

tux-00:


hiddeco:
Thank you @tux-00 :tulip:

PR #3035, opened 2020-04-27
Update chart git.path docs

gtseres:
As seen in the `git-path` [flag source code](https://github.com/fluxcd/flux/blob/8b4c115fd211089e2f8718b43c7a11e8efc7ceb0/cmd/fluxctl/install_cmd.go#L37), it can take more than one string values (in the form `git-path=path1,path2,...,pathN`)
This PR enhances the `git.path` documentation in the Helm chart README file, so that it is explicitly mentioned that more than one git paths can be defined.


hiddeco:
Thank you @gtseres :star:

PR #3039, opened 2020-04-30
direct comparison of last-synced and newly-synced resources

### LINKED ISSUE #2494: Spurious notifications for resources in multi-resource manifest files
**Describe the bug**
I have a setup where I have two identical clusters (with same configurations) and both of them are syncing using flux. I have also set up justinbarrick's fluxcloud for notifications in slack / teams.

I have a periodic upgrade to these clusters where I update a deployment's template with a newer image every time.

In the flux logs i always see that message regarding the commit with:
`deployment.apps/myapp configured`

But flux cloud notification always contain arbitrary resource that wasn't changed for instance:
```
Resources updated:
myapp:service/myapp
```

Note that the notification doesn't contain the updated deployment.

Note that all of the resources are defined in a single yaml in the git repo and are only updated with the image version.

I'm not using helm-operator

**Expected behavior**
To get a proper notification that deployment was updated without unrelated resources that were not updated.

**Additional context**
- Flux version: 1.14.2
- Fluxcloud: 0.3.8
- Kubernetes version: 1.13.10
- Git provider: VSTS
- Container registry provider: Azure Container Registry

### LINKED ISSUE #2981: Flux incorrectly shows what resources have been updated when using Kustomize
**Describe the bug**

When using Kustomize (or any other manifest generator it seems) Flux incorrectly shows which resources have been updated both in the logs and in the associated sync event metadata. This means that weave cloud/fluxcloud also posts incorrect information regarding the event.

**Expected behavior**

I have a repository for two environments, `development` and `staging`. Each resides in its own namespace. Each file contains one resource associated with the given application/service, usually a `Deployment` manifest. The repository structure looks like follows:

```
├── common
│   ├── .flux.yaml
│   ├── ambassador.yaml      
│   ├── cert-manager.yaml     
│   ├── fluxcloud.yaml     
│   ├── flux.yaml     
│   ├── helm-exporter.yaml     
│   ├── helm-operator.yaml     
│   └── prometheus-operator.yaml
├── development
│   ├── .flux.yaml
│   ├── app1.yaml      
│   ├── app2.yaml        
│   ├── app3.yaml      
│   ├── proxy.yaml        
│   ├── keycloak-post.yaml  
│   ├── keycloak-pre.yaml      
│   ├── keycloak.yaml         
│   └── sftp.yaml         
├── development-secrets
│   ├── .flux.yaml
│   ├── .sops.yaml
│   └── secrets.enc.yaml    
├── staging                  
│   ├── .flux.yaml
│   ├── app1.yaml   
│   ├── app2.yaml        
│   ├── app3.yaml      
│   ├── proxy.yaml        
│   ├── keycloak-post.yaml  
│   ├── keycloak-pre.yaml      
│   ├── keycloak.yaml         
│   └── sftp.yaml         
└── staging-secrets   
    ├── .flux.yaml
    ├── .sops.yaml
    └── secrets.enc.yaml
```

Manifest generation and SOPS support is enabled in Flux, and everything works great. When `app1` in `development` is updated, Flux generates events properly. This can be seen in the logs, ie. after changing `app1` in `development`:

```
ts=2020-04-06T16:26:29.222506302Z caller=daemon.go:704 component=daemon event="Sync: 17f533a, development:helmrelease/development-app1" logupstream=true
```

**Broken behaviour**

Problems started to appear after I decided to use Kustomize to reduce code duplication. Following https://github.com/fluxcd/flux-kustomize-example I changed `.flux.yaml` in `staging` and `development` directories to:

```
version: 1
patchUpdated:
  generators:
    - command: kustomize build .
  patchFile: flux-patch.yaml
```

I also added a new overlay directory, each environment got its `kustomization.yaml` file, and the newly generated manifests using Kustomize have the same content as the old static files.

The problem I've started encountering is the fact that flux shows completely random resources in its Sync events. For example, doing a simple change in the `app1` manifest in `development` now shows:

```
ts=2020-04-01T20:40:01.10921144Z caller=daemon.go:704 component=daemon event="Sync: b208daf, development:helmrelease/development-app1, staging:helmrelease/staging-app2" logupstream=true
```

The set of shown changed releases looks pretty random - sometimes it shows different resources from another environment, sometimes 2-3 resources from the same one, I haven't found a rule to that. Sometimes even the resource which triggered the update isn't shown in the list (`development/app1` in this case). It never showed resources from the `common` or `secrets` directory, which weren't handled by Kustomize.

To make sure that the problem isn't in my manifests, I browsed the source code a bit and found the following gem:

https://github.com/fluxcd/flux/blob/3e01223d90a5aa9b71cc9d7342ba18e1e2090f01/pkg/daemon/sync.go#L216-L225

Does this mean that all files that hold more than one resource are affected? Not only ones generated by Kustomize? Any workarounds that can be applied for the time being?

**Additional context**

- Flux version: 1.18 and 1.19
- Kubernetes version: 1.13.6, 1.15.9
- Git provider: Github
- Container registry provider: GCR

ordovicia:
`getChangedResources()` in `pkg/daemon/sync.go` could not determine what
resources are modified when a manifest file has multiple resource
definitions. This commit replaces `getChangedResources()` with new
`compareResources()` function, which determines resource updates and
deletions by direct comparison of last-synced and newly-synced clones.

Fixes #2494
Fixes #2981

### DIFF
--- a/pkg/daemon/sync.go
+++ b/pkg/daemon/sync.go
@@ -130,10 +130,7 @@ func (d *Daemon) getLastResources(ctx context.Context, rat ratchet) (map[string]
 		return make(map[string]resource.Resource), nil
 	}
 
-	// First sync
-	d.Logger.Log("info", "resource map is not initialized, despite repo has been cloned")
-
-	// Load resources from clone of currentRevision
+	// Fist sync -- load resources from clone of currentRevision
 	lastResourcestore, cleanup, err := d.getManifestStoreByRevision(ctx, currentRevision)
 	if err != nil {
 		return nil, errors.Wrap(err, "reading the repository checkout")
--- a/pkg/daemon/sync.go
+++ b/pkg/daemon/sync.go
@@ -1,15 +1,16 @@
 package daemon
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
+	"time"
+
 	"github.com/fluxcd/flux/pkg/metrics"
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
-	"path/filepath"
-	"time"
 
 	"github.com/fluxcd/flux/pkg/cluster"
 	"github.com/fluxcd/flux/pkg/event"
@@ -20,13 +21,15 @@ import (
 	"github.com/fluxcd/flux/pkg/update"
 )
 
-// revisionRatchet is for keeping track of transitions between
+// ratchet is for keeping track of transitions between
 // revisions. This is slightly more complicated than just setting the
 // state, since we want to notice unexpected transitions (e.g., when
 // the apparent current state is not what we'd recorded).
-type revisionRatchet interface {
-	Current(ctx context.Context) (string, error)
-	Update(ctx context.Context, oldRev, newRev string) (bool, error)
+type ratchet interface {
+	CurrentRevision(ctx context.Context) (string, error)
+	CurrentResources() map[string]resource.Resource
+
+	Update(ctx context.Context, oldRev, newRev string, resources map[string]resource.Resource) (bool, error)
 }
 
 type eventLogger interface {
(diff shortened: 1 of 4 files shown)


squaremo:
I really like the idea of this. Generating manifests tends to be quite expensive; can the set of applied manifests be kept around from the last sync, for comparison?

ordovicia:
> Generating manifests tends to be quite expensive

Good point.
I hesitated to maintain the last applied manifests to keep the ratchet lightweight, but I am convinced that keeping the last manifests is better.
I will try on changing this PR to that direction within several days.

ordovicia:
Updated. Now `lastKnownSyncState` maintains synced resource map and `Daemon.Sync()` just refers the resource map instead of cloning and generating manifests on every sync.

pkg/daemon/sync.go
```diff
@@ -41,67 +44,103 @@ type changeSet struct {
...
 
-	// Determine what resources changed during the sync
-	changedResources, err := d.getChangedResources(ctx, c, d.GitTimeout, working, resourceStore, resources)
-	serviceIDs := resource.IDSet{}
-	for _, r := range changedResources {
-		serviceIDs.Add([]resource.ID{r.ResourceID()})
-	}
+	// Determine what resources changed and deleted during the sync
+	updatedIDs, deletedIDs := compareResources(lastResources, resources)
+	// TODO(ordovicia): include deleted resources in sync events
```
ordovicia, line 129: I will address this TODO in another PR.

ordovicia:
Hi @squaremo , could you please take a look this at your convenience? Thank you.

pkg/daemon/sync.go
```diff
@@ -20,13 +21,15 @@ import (
...
+// ratchet is for keeping track of transitions between
 // revisions. This is slightly more complicated than just setting the
 // state, since we want to notice unexpected transitions (e.g., when
 // the apparent current state is not what we'd recorded).
-type revisionRatchet interface {
-	Current(ctx context.Context) (string, error)
-	Update(ctx context.Context, oldRev, newRev string) (bool, error)
+type ratchet interface {
+	CurrentRevision(ctx context.Context) (string, error)
+	CurrentResources(ctx context.Context) (map[string]resource.Resource, error)
```
squaremo, line 30: Can you think of any circumstances in which this will need a context?
(The current revision needs one because it does git operations)

pkg/daemon/sync.go
```diff
@@ -41,67 +44,103 @@ type changeSet struct {
...
 // Sync starts the synchronization of the cluster with git.
-func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string, ratchet revisionRatchet) error {
-	// Make a read-only clone used for this sync
+func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string, ratchet ratchet) error {
+	currentRevision, err := ratchet.CurrentRevision(ctx)
+	if err != nil {
+		return err
+	}
+
 	ctxt, cancel := context.WithTimeout(ctx, d.GitTimeout)
```
squaremo, line 53: Creating this context here and using it in (potentially) two places below is odd. The purpose of making a new context (rather than using the one passed as an argument) is to give each git operation a timeout; so, _each_ git operation should get a context with a deadline, just before it's attempted.

pkg/daemon/sync.go
```diff
@@ -41,67 +44,103 @@ type changeSet struct {
...
-	defer func() {
-		if err := working.Clean(); err != nil {
-			d.Logger.Log("error", fmt.Sprintf("cannot clean sync clone: %s", err))
+	// Resource map is not initialized
+	if lastResources == nil {
+		if currentRevision == "" {
+			// Initial clone
+			lastResources = make(map[string]resource.Resource)
+		} else {
+			// Clone last-synced revision
```
squaremo, line 73: Are there circumstances in which a revision was set, but not the resources associated with that revision? If there are, I think it's worth enumerating them here to justify this special case. I would expect the revision and resources to be set atomically -- either both, or neither.

pkg/daemon/sync.go
```diff
@@ -41,67 +44,103 @@ type changeSet struct {
...
+			lastSynced, err := d.Repo.Export(ctxt, currentRevision)
+			if err != nil {
+				return err
+			}
+			defer cleanupClone(lastSynced)
+
+			lastResourceStore, err := d.getManifestStore(lastSynced)
```
squaremo, line 80: When processing the current revision, there is a step which unlocks secrets. Is that also necessary here?

pkg/daemon/sync.go
ordovicia: Actually, not.
I wrote this function signature just for alignment with the signature of `CurrentRevision()`.

I have just fixed the signature not to require context.
Thank you for your comment!

pkg/daemon/sync.go
ordovicia: I understood and fixed. Thank you.

pkg/daemon/sync.go
ordovicia: Thank you. You are right.
The revision and resources should be set atomically.

But if I understand correctly, there is a special case -- in a test that uses `daemon.WithWorkingClone()`, only revision is set and resource map is left uninitialized.
I think a proper solution for addressing this special case is tweaking test code to make them not a special case, but I do not have enough experience with Flux code to do that.
Instead, I added a comment in `Daemon.Sync()` and made it emit a warning when entering the special case.

pkg/daemon/sync.go
ordovicia: Fixed. Thank you.

pkg/daemon/sync.go
squaremo: Now I look into those tests, I realise that the revision comes from the sync state -- either a tag, or an annotation on the secret -- so most of the time when the loop _starts_, it will have a current revision without having a corresponding set of resources. Sorry, I got that wrong in my review.

So, a changed assessment: I think if you factor a method out for getting the resources for a revision, this could be a bit tidier. (Neither of the exported git repos are used after calculating the manifests, so the method could create and tidy those up, I think.)

ordovicia: Ah, I re-read `daemon` package and I realized that, as you said, the sync loop starts with the "current revision" being set most of the time, not only in tests.

I have factored out some methods for loading `manifests.Store` and last-synced resources.

ordovicia:
Rebased and squashed some commits.

pkg/daemon/sync.go
```diff
@@ -125,13 +113,86 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
...
+		return nil, err
+	}
+
+	// Repo has never been cloned yet
+	if currentRevision == "" {
+		return make(map[string]resource.Resource), nil
+	}
+
+	// First sync
+	d.Logger.Log("info", "resource map is not initialized, despite repo has been cloned")
```
squaremo, line 134: Since we're no longer thinking of this as an unexpected situation, it doesn't need a log message. But if you still want one, I would have a message like "initialising last sync resources from last revision". The "despite" makes it sound exceptional, but it will happen every time fluxd starts.

ordovicia: I agree. Removed the log. Thank you!

ordovicia:
Thank you for reviewing many times.

Issue #1327, opened 2018-08-31
Proposal: Run flux as unprivileged user


dkerwin:
The official docker images and Kubernetes manifests run flux as `root` user. Running as `root` is not a requirement for flux to operate and I would prefer to run with minimal privileges. Switching to a unprivileged user-account would require these steps:

- Add user account in Dockerfile
- Set `USER` in Dockerfile
- Update flux to handle ssh files differntly

The main issues that would result from the privilege drop is that flux will no longer be able to update the permissions of ssh-related files. Here's what I propose as a solution:
 
1. Mount custom ssh-config to a location outside of `/home/flux/.ssh`
2. During the startup of flux copy the identity and other configs to `/home/flux/.ssh`
3. Update permissions of `/home/flux/.ssh` and it's contents

What are your thoughts on this?

squaremo:
Copying files that are mounted on startup is not adequate; the daemon needs to see changes to the mounted volumes, especially the SSH key. It might be workable just to change the location of these things so that they fall inside /home/flux, and are owned by the flux user. If it's not possible to mount volumes such that they are readable by the flux user (and have the permissions as desired by SSH), this is a non-starter.

The other thing I can think of that may be affected is the access to /var/run/kubernetes, which lets flux connect to the API server. From a quick `kubectl exec -- /bin/sh` it looks that's world-readable; but worth checking.

stefansedich:
I have picked this up and am having a go at getting to a working solution that will work for everyone and so far I have a working proof of concept for the flux operator which is working for git sync however I am still yet to test gpg signing/verification at this stage.

https://github.com/weaveworks/flux/compare/master...stefansedich:non-root

One downside with this approach and as mentioned by @squaremo we lose the ability to simply update the ssh/gpg secrets and have things update and we would require the user to restart the pod for things to take affect. However is this a trade-off we are willing to accept so that we can run as a non-root user?

Hoping to start a discussion now that I have something up and running and before I go sink any more time into it and if we decide pod reboots are 100% unacceptable I can then continue to dig further.

jacobsin:
Verified just now this approach works - thanks @stefansedich

Although I really have hoped the `git-key` secrets mount to have the correct `0400` permission at the first place, then there is no need to chmod at the first place , but think I am hitting another open issue with openshift https://github.com/openshift/origin/issues/16424.

Btw, I got this to work without initContainer but just lifecycle.postStart hook

stefansedich:
[...]

PR #3051, opened 2020-05-11
add securitycontext for flux pod

alemairebe:
When working with restrictive PSP, it enables to enforce lower permission with the securitycontext. A small example is provided in values.yaml

Signed-off-by: Adrien le Maire <adrien_lemaire@trimble.com>

### DIFF
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -298,3 +298,10 @@ hostAliases: {}
 #    hostnames:
 #    - "foo.remote"
 #    - "bar.remote"
+
+podSecurityContext: {}
+#  allowPrivilegeEscalation: false
+#  fsGroup: 1001
+#  runAsUser: 1001
+#  runAsGroup: 1001
+#  runAsNonRoot: true
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -277,6 +277,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `syncGarbageCollection.dry`                       | `false`                                              | If enabled, fluxd won't delete any resources, but log the garbage collection output (see [garbage collection](../../docs/references/garbagecollection.md))
 | `manifestGeneration`                              | `false`                                              | If enabled, fluxd will look for `.flux.yaml` and run Kustomize or other manifest generators
 | `hostAliases`                                     | `{}`                                                 | Additional hostAliases to add to the Flux pod(s). See <https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/>
+| `podSecurityContext`                              | `{}`                                                 | SecurityContext to add to the Flux pod(s). See <https://kubernetes.io/docs/tasks/configure-pod-container/security-context/>
 
 [memcached-ref]: https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L201-L204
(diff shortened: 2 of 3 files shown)


alemairebe:
@stefanprodan please review this PR

chart/flux/values.yaml
```diff
@@ -298,3 +298,10 @@ hostAliases: {}
 #    hostnames:
 #    - "foo.remote"
 #    - "bar.remote"
+
+podSecurityContext: {}
+#  allowPrivilegeEscalation: false
+#  fsGroup: 1001
+#  runAsUser: 1001
+#  runAsGroup: 1001
+#  runAsNonRoot: true
```
stefanprodan, line 307: None of these work with Flux as it runs as root.

alemairebe: It forces the container to run as non-root. I couldn't find in the documentation any reason for flux to run as root, so I try to run it as a normal user.
```
docker run --rm --user 1001 docker.io/fluxcd/flux:1.19.0 
ts=2020-05-11T09:38:53.944207902Z caller=main.go:259 version=1.19.0
```
It could work, maybe it needs additional emptyDir to work with git pulled directories, to avoid permission issues

stefanprodan: Making Flux run as non-root is a non-trivial task. The SSH keygen, secrets mounts and gpg setup need to be rethink and addressed both in code and configs.

See:
- https://github.com/fluxcd/flux/issues/1327
- https://github.com/fluxcd/helm-operator/issues/233

alemairebe: Thanks for this explanation and links.
I run Flux only with git over HTTPS so I do not face any of theses issues. I understand it needs to stay backwards compatible and therefor can't be merged as-is.
I will check if I can help on pushing some other code to get it working.

marmelin: Hi @alemairebe, can pls explain how you do that "Flux only with git over HTTPS so I do not face any of theses issues". Thx

alemairebe: Hi @marmelin ,
I do not use SSH keys nor GPG, I pass git https credentials from a secret with the helm chart value : "env.secretName" then on "git.url" I have "https://$(username):$(password)@my_git_repo.local/repo_project/my_repo.git" where $(username) and $(password) are fetched form the secret.
This is well described in the documentation : https://github.com/fluxcd/flux/tree/master/chart/flux#flux-with-git-over-https

marmelin: Do you run flux with root or non-root? Because I thought root is only required for ssh git connection.

alemairebe: I run flux with `runAsUser: 1001` could be any number but 0

kingdonb:
Thank you for contributing, this PR cannot be merged into Flux v1 as-is, as it has been explained there are too many breaking issues, and Flux v2 has addressed the concern of running as non-root user. We hope that Flux v2 can satisfy all your use cases, and you will consider upgrading, or reporting the deficiencies as issues if there are any blocking that!

I will recommend to close this one out, but you are welcome to write back with concerns, or bring any outstanding issues to the Flux 2 Discussions board.

kingdonb:
As some time has passed since this was submitted and there has been no follow-up for a while, I think it's safe to close. Thanks for using Flux!

PR #3053, opened 2020-05-11
Support annotations in chart SSH secret

maruina:
We're using Spinnaker to deploy `flux`.

By default, Spinnaker version `secrets` and `configmap` unless they are annotated. See `strategy.spinnaker.io/versioned` at https://www.spinnaker.io/reference/providers/kubernetes-v2/#strategy

This PR allows to add annotations to the secret so you can disable Spinnaker versioning.

~Created as draft while testing.~ Tested and deployed successfully.

### DIFF
--- a/chart/flux/templates/secret.yaml
+++ b/chart/flux/templates/secret.yaml
@@ -3,5 +3,8 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ template "flux.fullname" . }}-git-deploy
+  {{- if .Values.ssh.secret.annotations }}
+  annotations: {{ toYaml .Values.ssh.secret.annotations | nindent 4 }}
+  {{- end }}
 type: Opaque
 {{- end -}}
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -234,6 +234,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `gpgKeys.configMapName`                           | `None`                                               | Kubernetes config map with public GPG keys the Flux daemon should import
 | `sops.enabled`                                    | `false`                                              | If `true` SOPS support will be enabled
 | `ssh.known_hosts`                                 | `None`                                               | The contents of an SSH `known_hosts` file, if you need to supply host key(s)
+| `ssh.secret.annotations`                          | `{}`                                                 | Additional secret annotations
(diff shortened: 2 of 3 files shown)


chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.19.0"
-version: 1.3.0
+version: 1.3.1
```
hiddeco, line 3: We bump the chart version during the release.

chart/flux/templates/secret.yaml
```diff
@@ -3,5 +3,8 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ template "flux.fullname" . }}-git-deploy
+  {{- if .Values.serviceAccount.annotations }}
```
hiddeco, line 6: This does not seem to be the right `.Values` given the addition to the `values.yaml`.

chart/flux/Chart.yaml
maruina: Reverted

chart/flux/templates/secret.yaml
maruina: You are correct, I did some copy and paste and I missed this.

hiddeco:
@maruina can you please smash your commits together? This makes the  change in `master` clean, given it's a single addition to the chart.

maruina:
@hiddeco done :)

chart/flux/templates/secret.yaml
```diff
@@ -3,5 +3,8 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ template "flux.fullname" . }}-git-deploy
+  {{- if .Values.secret.annotations }}
```
hiddeco, line 6: I am on the fence if the `.secret` provides sufficient context about what secret the annotation is added to. Looking at the way things are composed based on the configured and/or default values, I am wondering if i.e. `.ssh.secret.annotations` would make more sense (as `.git.secret` would conflict with the git-secrets feature).

maruina: I don't have any strong opinion, and I'm happy to go with `ssh.secret.annotations`.

hiddeco: :+1:

maruina: Done. Please tell me if you want me to improve the comment or description.

Issue #3044, opened 2020-05-05
flux stuck and does not recover


runningman84:
**Describe the bug**

After a few days flux got stuck and did not sync anymore. I killeded the pod and everything was up and running again.

**To Reproduce**

I have no idea... maybe the network connection was brokene for a few minutes or the file system got corrupted...

**Expected behavior**

flux should terminate if it cannot reach the git repo for a long time like 1 hour. This would allow k8s to restart the container which gives visibility and might also solve the issue.

**Logs**

```
│ ts=2020-05-05T09:17:13.783823769Z caller=checkpoint.go:24 component=checkpoint msg="up to date" latest=1.19.0                                                                                                                                 │
│ ts=2020-05-05T09:25:36.751714104Z caller=loop.go:107 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository., full output:\n Cloning into bare repository '/tmp/flux-gitclone406024532 │
│ ts=2020-05-05T10:25:36.752085036Z caller=loop.go:107 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository., full output:\n Cloning into bare repository '/tmp/flux-gitclone406024532 │
│ ts=2020-05-05T11:25:36.752559107Z caller=loop.go:107 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository., full output:\n Cloning into bare repository '/tmp/flux-gitclone406024532 │
│ ts=2020-05-05T12:25:36.753326379Z caller=loop.go:107 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository., full output:\n Cloning into bare repository '/tmp/flux-gitclone406024532 │
│ ts=2020-05-05T13:25:36.754100308Z caller=loop.go:107 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository., full output:\n Cloning into bare repository '/tmp/flux-gitclone406024532 │
│ ts=2020-05-05T14:25:36.754661578Z caller=loop.go:107 component=sync-loop err="git repo not ready: git clone --mirror: fatal: Could not read from remote repository., full output:\n Cloning into bare repository '/tmp/flux-gitclone406024532 
```

**Additional context**

- Flux version: 1.19
- Kubernetes version: v1.17.4+k3s1
- Git provider: github
- Container registry provider: na

brenix:
[...]

PR #3054, opened 2020-05-11
Fail liveness checks if daemon loop is stuck

sebastianblunt:
Adds a dedicated /liveness endpoint that fails the liveness check if the
daemon's main loop has not run for \<interval\> + \<timeout\>.

This provides a safety net for anything that may cause the main loop to
get stuck, as such (yet unexplained) cases have been observed in the
wild. For example https://github.com/fluxcd/flux/issues/3044

The issue of the main loop getting stuck is something we've been
observing happening infrequently but consistently over a prolonged
period, each time requiring a manual restart of the pod. This would
avoid requiring manual intervention to fix this state, whatever the
cause might be.


pkg/daemon/loop.go
```diff
@@ -70,7 +76,21 @@ func (d *Daemon) Loop(stop chan struct{}, wg *sync.WaitGroup, logger log.Logger)
 	d.AskForSync()
 	d.AskForAutomatedWorkloadImageUpdates()
 
+	// How long until we consider a loop to be "stuck", in seconds, for
+	// failing liveness checks, which is given to be the smallest interval
+	// + corresponding timeout, and no less than 1 minute.
+	loopTimeout := (d.SyncInterval.Milliseconds() + d.SyncTimeout.Milliseconds()) / 1000
+	// FIXME: What timeout should we use for AutomationInterval?
+	if x := (d.AutomationInterval.Milliseconds() * 2) / 1000; x < loopTimeout {
```
sebastianblunt, line 84: I wasn't sure what timeout to use for this, so I just set it to double the interval.

sebastianblunt:
Thanks for reviewing.

Through further experimentation locally it's still managing to get stuck with this change, but pulling in latest master seems to have fixed the issue causing it to get stuck anyway.

I'll be sure to implement your suggestions if I reopen this.

PR #3076, opened 2020-05-24
Preserve namesace when running helm template

glothriel:
Helm chart now preserves target namespace when running 'helm template'
For motivation, please read this thread: https://github.com/bitnami/charts/issues/2006, it's basically the same.

The commit allows deploying flux itself in a gitops-friendly way using `helm template`, it has no effect when using `helm install` /  `helm update` etc commands.
I didn't bump chart version nor update the changelog, because I was not sure how to do it. If an issue is required, please let me know, I will create one.

I will be happy to discuss if there are questions.

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ template "flux.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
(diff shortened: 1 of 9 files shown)


squaremo:
This seems reasonable to me; @stefanprodan or @hiddeco, are there any hidden traps with putting an explicit namespace in the templates?

glothriel:
What are next steps guys?

hiddeco:
@glothriel your PR required a rebase before it could be merged. The change will be included in the next chart release. Thank you for your contribution! :tulip:

PR #3079, opened 2020-05-25
Parameter name was incorrect

jaydeland:
The fluxd process parameters for skipping CI need to be updated


squaremo:
First-time contributor high five ✋

Issue #2944, opened 2020-03-27
Migrate CI to GitHub Actions


stefanprodan:
Tasks:
- [x] unit tests #2940
- [x] codegen #2940
- [x] binaries and containers builds #2940
- [x] end-to-end testing #2940
- [x] release helm charts #3085
- [x] pre-release container images
- [x] release binaries upload
- [x] release container images
- [ ] run CVE scan on release

kingdonb:
These issues all seem to be resolved, with the exception of CVE scanning which broke and had to be reverted. The scanning for CVEs is performed manually (with Snyk) between releases and for new releases.

Pre-release container images are published when the PR merges to master, and since the release cadence is likely going to be deliberately slowed down during maintenance, to minimize the effort required for downstreams, and before releases can be published, we will usually recommend folks who are still depending on Flux v1 to find the prerelease images here:

https://hub.docker.com/r/fluxcd/flux-prerelease/tags?page=1&ordering=last_updated

Closing out this issue.

kingdonb:
Gack, I missed that some of these CI tasks that were left unchecked are still driven from CircleCI.

*  pre-release container images
* release binaries upload
* release container images

Let me know if I should reopen this issue. I don't think it is a high priority to resolve anymore, (Flux v1 will not be maintained forever, and was formally superseded by Flux v2 when the GitOps Toolkit API stability was announced about a month ago.)

PR #3086, opened 2020-05-28
ci: Run CVE scanning for latest release and master build

stefanprodan:
This PR adds CVE scanning with trivy for Flux container images (latest release/pre-release).
The scan runs every day in GitHub Actions and should email the Flux CD team if CVEs are found.


stefanprodan:
@squaremo the release workflow is still on CircleCI. I've add the CVE scan to the migration list here https://github.com/fluxcd/flux/issues/2944

PR #3101, opened 2020-06-02
[chart] Add support for headless service for Flux, similar to Memcached

stylius:
New configuration setting for the Helm Chart to allow deployments with Headless service.

Memcached already has this feature but is not documented. This change adds the same functionality to Flux service. This PR would document the Flux part though.

Problem description: Every deployment uses by default at least 2 service IPs - one for Flux and another for Memcached. In our setup Flux is deployed for each namespace to monitor different repos for different teams. This quickly exhausts the Service IP network just for Flux, where it is not really necessary to use such Service IP. This change allows deployment with a Headless Kubernetes Service or a service without dedicated IP address.

Users affected: Not many to be honest, it seems to be a corner case. Still this PR does not change the default behaviour of the chart and only allows the new feature to be enabled only if necessary.

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -255,6 +255,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `serviceAccount.annotations`                      | ``                                                   | Additional Service Account annotations
 | `clusterRole.create`                              | `true`                                               | If `false`, Flux will be restricted to the namespaces given in `allowedNamespaces` and the namespace where it is deployed
 | `service.type`                                    | `ClusterIP`                                          | Service type to be used (exposing the Flux API outside of the cluster is not advised)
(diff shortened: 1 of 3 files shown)


kingdonb:
This doesn't look like a breaking change, thank you for contributing it.

I can review it further if you are still interested in merging it, if you can please rebase and sign off your commit. Flux v1 is in maintenance, and this looks like technically a new feature. I'm not sure it can merge.

Are you still interested? If so please rebase and sign off your commit.

stylius:
Yes, I am still interested in merging it if it would be allowed. Using off the shelf Flux is better than maintaining local copy of the chart just for the ability to spare service IPs where they are not necessary.

Code was rebased and signed off.

kingdonb:
Sorry for the delay in merging.

I think this change is harmless and it can go into the 1.23.0 release. It should be versioned as Chart 1.9.0, since it is a new feature.

kingdonb (approved):
LGTM

Issue #3108, opened 2020-06-04
Add value 'extraEnvsFrom' to Helm Chart


KadenLNelson:
It would be nice to be able to template the `envFrom` in the Flux container spec.

PR #3109, opened 2020-06-04
Add value 'extraEnvsFrom' to Helm Chart

KadenLNelson:
It would be nice to be able to template the envFrom in the Flux container spec.

fixes https://github.com/fluxcd/flux/issues/3108


KadenLNelson:
@hiddeco may this be merged?

Issue #2390, opened 2019-08-21
Suggestion: Continuous Fuzzing


yevgenypats:
Hi, I'm Yevgeny Pats Founder of [Fuzzit](https://fuzzit.dev) - Continuous fuzzing as a service platform.

We have a free plan for OSS and I would be happy to contribute a PR if that's interesting.
The PR will include the following
- [go-fuzz](https://github.com/dvyukov/go-fuzz) fuzzers (This is generic step not-connected to fuzzit)
- Continuous Fuzzing of master branch which will generate new corpus and look for new crashes
- Regression on every PR that will run the fuzzers through all the generated corpus and fixed crashes from previous step. This will prevent new or old bugs from crippling into master.

You can see our basic example [fuzzitdev/example-go](https://github.com/fuzzitdev/example-go) and you can see an example of "in the wild" integration [prometheus/prometheus](https://github.com/prometheus/prometheus), [coredns/coredns](https://github.com/coredns/coredns), [google/syzkaller](https://github.com/google/syzkaller).

Let me know if this is something worth working on.

Cheers,
Yevgeny

kingdonb:
Flux v1 is in maintenance mode now, and is not adding any new features unless they are critical.

As Flux contrib efforts have been focused on Flux v2, the Flux project has moved to a new repo, [fluxcd/flux2](/fluxcd/flux2)

In the interest of reducing the number of open issues not directly related to supporting Flux v1 in maintenance mode, and respecting you may have moved on already, I will go ahead and close out this issue for now. Thanks for using Flux!

PR #3112, opened 2020-06-04
Added 3 fuzzers

AdamKorcz:
If merged, this PR adds 3 fuzzers to Flux. 

The fuzzers can be run locally and integrated into regression testing. In addition, I can confirm that they run on oss-fuzz's infrastructure, and I will be glad to integrate Flux into their platform. If accepted, Google will run these fuzzers continuously and notify maintainers via email in case of any bugs. The service is free and the project agrees to fix the bugs that oss-fuzz finds.

This PR adds new functionality but not to the core functionality. Nor does it modify the core functionality.
An issue has been raised earlier regarding fuzzing Flux (https://github.com/fluxcd/flux/issues/2390), but it has not been discussed by the community.
I do not know if these fuzzers are related to Helm Chart.

Signed-off-by: AdamKorcz <adam@adalogics.com>

### DIFF
--- a/pkg/cluster/kubernetes/resource/fuzz.go
+++ b/pkg/cluster/kubernetes/resource/fuzz.go
@@ -1,0 +1,9 @@
+package resource
+
+func Fuzz(data []byte) int {
+	_, err := ParseMultidoc(data, string(data))
+	if err != nil {
+		return 0
+	}
+	return 1
+}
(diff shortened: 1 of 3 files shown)


kingdonb:
If there is someone keen to maintain something like this, I can reopen, but if Flux v1 is in maintenance mode, I don't think it can merge. Thanks for your contribution!

Issue #3015, opened 2020-04-21
Unable to fetch tags from ECR 


pbn4:
**Describe the bug**

I have a fluxcd running inside EKS with Fargate. There is a workload for which I have automated update set up with repository being the AWS ECR. Roles for fargate pods give them permissions to access ECR. 

Also none of the issues and solution I found in this repository are solving my problem, thus new thread.

**To Reproduce**

Set up an EKS cluster with fargate and flux and observe flux logs.

**Expected behavior**

Flux should update the workload with new image. 

**Logs**

`ts=2020-04-21T11:11:15.679624161Z caller=warming.go:180 component=warmer canonical_name=xxxx.dkr.ecr.eu-west-1.amazonaws.com/repo auth={map[]} err="requesting tags: Get https://xxxx.dkr.ecr.eu-west-1.amazonaws.com/v2/repo/tags/list: no basic auth credentials"
`

I did some experimentation with aws-cli docker image inside flux namespace. I was trying to get list of accounts e.g. `configure list` is yielding an empty configuration for this newly created pod and I'm not sure if this is correct. How can I check if arbitrary pod has the access to this URL flux is trying to fetch (I assume flux is specifying authentication somehow)? 

**Additional context**

- Flux version: 1.19
- Kubernetes version: 1.15
- Git provider: Github
- Container registry provider: ECR

pbn4:
Some additional logs:

`
ts=2020-04-21T13:30:52.413063388Z caller=aws.go:124 component=aws error="fetching region for AWS" err="EC2MetadataRequestError: failed to get EC2 instance identity document\ncaused by: RequestError: send request failed\ncaused by: Get http://169.254.169.254/latest/dynamic/instance-identity/document: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"
`

admssa:
```
ts=2020-04-27T18:16:34.152506667Z caller=warming.go:180 component=warmer canonical_name=... auth={map[]} err="requesting tags: Get https://.../tags/list: no basic auth credentials"
```
I'm having the same issue. (Access to metadata from pods was blocked to prevent using node roles by pods) 
But it doesn't seem flux able to use aws web identity role and token. (EKS IRSA)

@pbn4  What about remounting new secret each renewing interval in this solution  - https://github.com/fluxcd/flux/issues/2708#issuecomment-619915270 ?

pbn4:
[...]

PR #3124, opened 2020-06-09
skip aws api check when regions are passed from config

evq:
fixes https://github.com/fluxcd/flux/issues/3015

### DIFF
--- a/pkg/registry/aws.go
+++ b/pkg/registry/aws.go
@@ -23,7 +23,7 @@ import (
 
 const (
 	// For recognising ECR hosts
-	awsPartitionSuffix = ".amazonaws.com"
+	awsPartitionSuffix   = ".amazonaws.com"
 	awsCnPartitionSuffix = ".amazonaws.com.cn"
 
 	// How long AWS tokens remain valid, according to AWS docs; this
@@ -33,7 +33,7 @@ const (
 	// how long to skip refreshing a region after we've failed
 	embargoDuration = 10 * time.Minute
 
-	EKS_SYSTEM_ACCOUNT = "602401143452"
+	EKS_SYSTEM_ACCOUNT    = "602401143452"
 	EKS_SYSTEM_ACCOUNT_CN = "918309763551"
 )
 
@@ -120,27 +120,30 @@ func ImageCredsWithAWSAuth(lookup func() ImageCreds, logger log.Logger, config A
 					"exclude-ids", fmt.Sprintf("%v", config.BlockIDs))
 			}()
 
-			// This forces the AWS SDK to load config, so we can get
-			// the default region if it's there.
-			sess := session.Must(session.NewSessionWithOptions(session.Options{
-				SharedConfigState: session.SharedConfigEnable,
-			}))
-			// Always try to connect to the metadata service, so we
-			// can fail fast if it's not available.
-			ec2 := ec2metadata.New(sess)
-			metadataRegion, err := ec2.Region()
-			if err != nil {
-				preflightErr = err
-				if config.Regions == nil {
-					config.Regions = []string{}
+			if config.Regions != nil {
+				okToUseAWS = true
+				logger.Log("info", "using regions from local config")
+			} else {
+				// This forces the AWS SDK to load config, so we can get
+				// the default region if it's there.
+				sess := session.Must(session.NewSessionWithOptions(session.Options{
+					SharedConfigState: session.SharedConfigEnable,
+				}))
+				// Always try to connect to the metadata service, so we
+				// can fail fast if it's not available.
+				ec2 := ec2metadata.New(sess)
+				metadataRegion, err := ec2.Region()
+				if err != nil {
+					preflightErr = err
+					if config.Regions == nil {
+						config.Regions = []string{}
+					}
+					logger.Log("error", "fetching region for AWS", "err", err)
+					return
 				}
-				logger.Log("error", "fetching region for AWS", "err", err)
-				return
-			}
 
-			okToUseAWS = true
+				okToUseAWS = true
 
-			if config.Regions == nil {
 				clusterRegion := *sess.Config.Region
 				regionSource := "local config"
 				if clusterRegion == "" {


squaremo:
The preflight check is used to initialise the AWS state when that's needed, but it's also used if `--registry-require=ecr` is given (it's like an assertion, meaning "crash if you can't get creds via AWS"). This change would break the latter use, since it will succeed without checking, if you've supplied regions.

I'm not sure why using ECR from Fargate (as in #3015) is different to using ECR from anywhere else. If you can give an explanation why this is the right change to fix 3015, I'd appreciate it.

The way forward then would be to separate the preflight check from the region calculation, so the check can still be called as an assertion.

evq:
I don't think this is a complete change to fix ECR on Fargate, I just figured I would push it just in case anyone else ran into this issue and needed a quick workaround ( assuming they cared more about having it work than having it not work since this breaks the assertion ).

I'm not actually running this on Fargate at all, but my setup reproduces the same error mentioned in #3015. I'm using [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) to provide IAM permissions to this pod. The pod is running on an EC2 nodegroup where [I have blocked access to the EC2 metadata service in able to prevent the pods from accessing the instance credentials, per best practice](https://docs.aws.amazon.com/eks/latest/userguide/restrict-ec2-credential-access.html).

So the issue arises when flux is unable to access the EC2 metadata service. This can be verified by [looking at the error pbn4 reported](https://github.com/fluxcd/flux/issues/3015#issuecomment-617186590). Currently the preflight check doesn't actually have any bearing on whether you'll be able to actually connect to ECR, only whether you can connect to the EC2 metadata service - a totally separate and unrelated service.

I believe there are two changes that should be made to properly fix this:
1. The region lookup via metadata service should be separated from the preflight check and made optional.
2. The preflight check ideally would attempt to retrieve ECR credentials, such a check would be more accurate and could even be enforced on Fargate / locked down EC2

Unfortunately I don't think 2 always works if I understand some of the use cases right. Originally I was wondering if we could simply call `fetchAWSCreds(region, []string{accountID})` for each region / account id permutation in the preflight (using config.AccountIDs). However, AccountIDs is optional - as it seems they can be automatically picked up from the ECR repository URL of an image at runtime? Also there's no good way to know which of the combinations of {region, accountID} are supposed to work, so we would simply have to check whether ANY of them work. There doesn't really seem to be enough information at runtime to do a proper preflight check, so let know how you would like to proceed.

squaremo:
That's a good diagnosis, I understand it much better now.

So paraphrasing a little, there's two problems to be solved, potentially:

 - make ECR auto-authorisation work in the case that EC2 metadata cannot be accessed (this would solve #3015)
 - make the preflight check more accurate by testing access to ECR (not necessary for #3015, since it's possible to run without `--registry-require=ecr`)

.. and the latter is tricky because it may not be possible to check ECR access before knowing an account or region to check with.

I would be happy for the first problem to be solved, without the second; people that are using `--registry-require=ecr` will have the same behaviour as they get now, and people that need #3015 fixed will get a working system.

yannick:
this is also a problem when running flux locally, e.g. in minikube and the containers are in ECR.

kingdonb:
Greetings,

If you can pull `--rebase` the master branch and squash your commits to amend with `--sign-off` per the [DCO bot requirement](https://github.com/fluxcd/flux/pull/3124/checks?check_run_id=1975574996), I will review them and include them for merge whenever possible! Please let me know if you are still interested in merging this.

I cannot merge or adopt your commits without the DCO, please rebase and squash, and sign off

Thank you for your contribution!

kingdonb (changes requested):
Did you mean to make any changes in `.dockerignore`, `.github/workflows`, or `Dockerfile`?

When you rebase, could you remove these changes first?

evq:
Happy to do so tomorrow, that most recent commit was just for an internal build - will prune it so we can get this merged

kingdonb:
I'm not especially comfortable reviewing this change by myself since I don't have an environment that I can test it in, to prove to my satisfaction that it has the desired effects.

But it seems like the discussion has already been talked out and with a maintainer satisfied that the change is OK, I would be happy to merge and include it in the release. The DCO is check is unfortunately still blocked.

But... I can apparently also bypass it now. Either someone with higher permissions than +v is listening and has found a way to grant me the ability to mark the DCO check as passing, or... 👻

If you're having trouble rebasing, please let me know, I can help; there is still time to merge in for the next release!

Thanks again.

kingdonb:
It's no problem, we can have a 1.22.1 release out soonish if this is still needed.

Whenever you get around to it is fine.

evq:
sorry for the delay - i've removed the last commit, rebased on origin/master and amended with `--signoff`. it appears the DCO is passing now - although there is a FOSSA failure related to some rate limit

kingdonb:
I was going to rebase this and retry it against the latest commit in `master` but I noticed the PR does not have "maintainers can edit this" set. Is it possible to edit the PR and check that box, so I don't have to come back and ask you to rebase again?

Alternatively, if you can `git pull --rebase origin master` and `git push --force`, that should induce the PR checks to run again. I think the failure was most likely a transient one that will not come back the next time.

I can't merge or take any further action until one of these actions is taken. Thanks again for your contribution.

kingdonb:
We're working our way down to zero backlogged PRs. This is one that I think we wanted to merge.

If you can set the "maintainers with write access can edit this branch" flag on the PR, I can try to move it forward.

Thanks for using Flux!

kingdonb:
I think the DCO sign-off is what is supposed to empower me to reopen this as a new PR in my name without asking for permission, so I can run the tests again and make any changes needed, since this submitter left off the "maintainers can edit this PR" flag.

At this time I don't have any changes only a clean rebase on v1.22.2 version of Flux release and am pushing a new PR with head revision eebe032b so that it can be re-tested for FOSSA which failed with rate limiting errors.

I'm not stating any opinion or approval by reopening this PR, just trying to move things along (and also get it out of the queue if possible, since it seems to be stuck here.)

kingdonb:
Closing in favor of #3485, which is expected to be passing in CI

PR #3130, opened 2020-06-11
Add SupplyStack as Flux user

BitProcessor:
Adds SupplyStack to the list of companies using Flux


squaremo:
This drops an entry from the table (UK Hydrographic Office) because it pushes it into an extraneous column. The unfortunate side-effect of using a table is that everything has to be reflowed into three columns 😞 . Do you mind either doing that reflow, or opening a PR from a branch I can push to, so I can edit the commit? Thanks!

BitProcessor:
Oopsie - committed a fix, please verify @squaremo

PR #3131, opened 2020-06-11
[chart] Allow specifying k8s-default-namespace in the helm chart

ninadpage:
In the helm chart, allow specifying `--k8s-default-namespace` flag for fluxd - the namespace flux should use for resources where a namespace is not specified.

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -250,6 +250,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `rbac.create`                                     | `true`                                               | If `true`, create and use RBAC resources
 | `rbac.pspEnabled`                                 | `false`                                              | If `true`, create and use a restricted pod security policy for Flux pod(s)
 | `allowedNamespaces`                               | `[]`                                                 | Allow flux to manage resources in the specified namespaces. The namespace flux is deployed in will always be included
+| `defaultNamespace`                                | `""`                                                 | The namespace flux should use for resources where a namespace is not specified. If none is provided here, the default namespace in kubeconfig is used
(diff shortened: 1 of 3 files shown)


kingdonb:
Hi @ninadpage, I see that the build failed 24 days ago, but I suspect this is from #3392 which was closed 15 days ago, and that it would pass if retested.

I think this change is probably fine, I don't know if there would be philosophical objections but since the option is present in fluxd, I don't know why it isn't exposed. Generally I have seen issues with resources that are missing metadata namespace causing image automation to fail. I was not aware of k8s-default-namespace.

Can you say if this option has been tested or maybe has issues in cases when image automation is also used? I have not tested it myself just yet.

kingdonb:
Test this improvement for possible including in v1.23.0

(I think it won't make the cut, as there are too many possible interactions to test everything in time, but it would help if I have an active collaborator who can say if this change works reliably.)

@ninadpage Have you been using a build with this change in your environment, or waiting for it to merge here?

ninadpage:
@kingdonb My apologies for responding so late - the notification of your first reply fell through the cracks.

> Have you been using a build with this change in your environment, or waiting for it to merge here?

Yes, I tested the helm chart from my feature branch, with flux (appVersion) `1.22.1`. It adds the non-namespaced resources to the specified namespace as expected.

> And if so, do you also use image automation?

No we don't, so I have not been able to test this in conjunction with image automation. However, since this change is for the helm chart which just passes an already supported flag to flux, I believe it should not have any impact on image automation.

kingdonb:
OK, that's a great point. I'll plan on including this in the next chart release. 👍

kingdonb (approved):
LGTM

PR #3133, opened 2020-06-12
Notify upstream of manifest generation error

ordovicia:
With this PR, `Daemon.Sync()` now notifies upstreams when it failed to generate manifests.
This PR adds `GenerationError` field to `SyncEventMetadata` and populates the field when `doSync()` returns an error that indicates manifest generation failure.

This PR introduces a breaking change because it adds a field to `SyncEventMetadata`.
Upstreams that assume the current `SyncEventMetadata` structure may fail to parse new events.

PR #3136, opened 2020-06-15
Emit logs when listing resources in a namespace is forbidden

ordovicia:
With this PR, Flux daemon emits logs when listing k8s resources in a namespace (or "all namespaces") is forbidden.
This would help users know what's wrong and fix their issues.

### DIFF
--- a/pkg/cluster/kubernetes/sync.go
+++ b/pkg/cluster/kubernetes/sync.go
@@ -8,12 +8,13 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
-	"github.com/ryanuber/go-glob"
 	"io"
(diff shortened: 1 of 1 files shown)


kingdonb:
Hello. Same as #3133 - just a friendly note about this PR which had gone stale. We require a DCO for contributions now.

If you are still interested in merging this, I'll need you to follow the DCO check details and sign off your commit.

Generally logging changes are being considered as breaking changes, but if this is an important event that is missed in the logs and it caused you trouble, it could also be considered a bug and candidate for fixing in maintenance mode.

Please let me know, thanks

ordovicia:
Thank you for your comment. I no longer need this PR to be merged. Closing.

PR #3137, opened 2020-06-15
private-git-host-deploying-via-helm

andrew-lee-metaswitch:
Added better instructions for how to deploy Flux with Helm using a private git server. The key thing here is *not* to create a ConfigMap `flux-ssh-config` manually, but rather to specify the contexts of that ConfigMap in `ssh.known_hosts` in the Helm Chart.

There's a possible WIBNI to allow the HelmChart to reference a pre-existing  `flux-ssh-config` ConfigMap, but I can't get too excited about that.

The contribution criteria says: 
`      do you think many users would profit from a dedicated setting  for this functionality in the Helm chart?`

I think this is something for discussion.

### DIFF
--- a/docs/guides/use-private-git-host.md
+++ b/docs/guides/use-private-git-host.md
@@ -25,7 +25,7 @@ Host key verification failed.
 fatal: Could not read from remote repository
 
 container$ # ^ that was expected. Now we'll try with a modified known_hosts
-container$ ssh-keyscan $GITHOST >> ~/.ssh/known_hosts
+container$ ssh-keyscan $GITHOST >> /root/.ssh/known_hosts
 container$ git clone $GITREPO
 Cloning into '...'
 ...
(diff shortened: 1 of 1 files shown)


docs/guides/use-private-git-host.md
```diff
@@ -85,3 +89,11 @@ metadata:
 You will need to explicitly tell `fluxd` to use that service account by
 uncommenting and possible adapting the line `# serviceAccountName:
 flux` in the file `flux-deployment.yaml` before applying it.
+
+## If using Helm
+
+If you are deploying flux via a Helm chart, then you will need to add 
+the contents of you known_hosts file as a Value `ssh.known_hosts`.
```
kingdonb, line 96:
```suggestion
the contents of your known_hosts file as a Value `ssh.known_hosts`.
```

kingdonb:
Thank you for making a contribution here!

There is a DCO check that failed to execute, I'm not sure if you're able to `git pull --rebase origin master` and `git push --force`, the DCO bot should run again and let you know if you've followed the DCO requirements.

(That is, to commit with `--sign-off`)

kingdonb:
I'm afraid I've messed up the DCO for this, by trying to rebase it, and somehow the original sign-off isn't present anywhere in my git reflog history.

I'm just going to leave it like this for now, and close out, as I'm not sure what else is being proposed in this PR. We can reopen it if you're still around and have further suggestions. Thanks for contributing to Flux!

Issue #3143, opened 2020-06-17
Specifying automated containers in a HelmRelease


austinbv:
**Describe the bug**

We are running into a weird issue with the `HelmOperator` - We are trying to configure the automated deployments of custom images in the Airflow Helm Chart in the `gitsync` containers and `initContainers`. We have added annotations to our helm release and it looks like the below. We are not seeing any logs about the `initContainer` or `gitSync` and new images are not detected.

```yaml
kind: HelmRelease
metadata:
  name: airflow
  namespace: airflow
  annotations:
    fluxcd.io/automated: true

    filter.fluxcd.io/airflow: glob:airflow-dev-*

    filter.fluxcd.io/initContainer: glob:git-*
    repository.fluxcd.io/initContainer: dags.initContainer.image.repository
    tag.fluxcd.io/initContainer: dags.initContainer.image.tag

    filter.fluxcd.io/gitSync: glob:git-*
    repository.fluxcd.io/gitSync: dags.git.gitSync.image.repository
    tag.fluxcd.io/gitSync: dags.git.gitSync.image.tag
spec:
  releaseName: airflow
  chart:
    repository: https://kubernetes-charts.storage.googleapis.com
    version: 7.1.x
    name: airflow
  values:
    airflow:
      image:
        repository: ****.dkr.ecr.****.amazonaws.com/****
        tag: airflow-dev-63b3d5ed649cb3e41611c8fc5eb8dfc583433604
    dags:
      initContainer:
        enabled: true
        image:
          repository: ****.dkr.ecr.****.amazonaws.com/****
          tag: git-41d01deba5c2a08edd5e006fddd8fe8497539a73
      git:
        gitSync:
          enabled: true
          image:
            repository: ****.dkr.ecr.****.amazonaws.com/****
            tag: git-41d01deba5c2a08edd5e006fddd8fe8497539a73
```

**Additional context**

- Helm Operator version: `1.1.0`
- Kubernetes version: `v1.16.8-eks-e16311`
- Git provider: `github`
- Helm repository provider: `stable/airflow`

----

⚠️ ⚠️ ⚠️ ⚠️  

[...]

PR #3144, opened 2020-06-18
Specific documentation for HelmRelease Automation

austinbv:
I wanted to make explicit that you need to use a string value for `fluxcd.io/automated: "true"`. 

This resolves https://github.com/fluxcd/helm-operator/issues/464 and resolves https://github.com/fluxcd/flux/issues/3143


austinbv:
Done

Issue #3151, opened 2020-06-24
Enhancing flux to clone repo using git@ (git over ssh) behind a http proxy


Abhishek-Srivastava:
**Describe the feature**
Flux, with [this](https://github.com/fluxcd/flux/pull/2279)  enhancement, flux supported git:// [git protocol](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols). However, the downside of the Git protocol is the lack of authentication. 
The git over ssh protocol works fine with till we aren't behind a corporate proxy. To make Flux we require to have those extra stanzas in the config file in the root's ssh directory in the flux container. Something similar to:
```
Host github.com
    ProxyCommand socat STDIO PROXY:<proxyIP>:%h:%p,proxyport=<proxyport>,proxyauth=<proxyAuth>
    User git
    Hostname ssh.github.com
    Port 443
    IdentityFile /etc/fluxd/ssh/identity
```

The changes involved would only be required in the [ssh.yaml](https://github.com/fluxcd/flux/blob/master/chart/flux/templates/ssh.yaml) to accept the config which would then automatically mount this new file in the expected ssh directory. The contents of the config would be then supplied via the [ssh-section](https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L218) in values.yaml

**Expected behavior**
This enhancement would then allow flux to clone repositories using the git@ (git with ssh) protocol behind a corporate proxy.

Abhishek-Srivastava:
Fixed with #3152. Thanks to @stefanprodan for the quick review and guidance.

PR #3152, opened 2020-06-24
Supporting git clone over ssh behind http proxy

Abhishek-Srivastava:
This fix refers to this enhancement
#3151 

Flux, with [this](https://github.com/fluxcd/flux/pull/2279)  enhancement, flux supported git:// [git protocol](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols). However, the downside of the Git protocol is the lack of authentication. 
The git over ssh protocol works fine with till we aren't behind a corporate proxy. To make Flux we require to have those extra stanzas in the config file in the root's ssh directory in the flux container. Something similar to:
```
Host github.com
    ProxyCommand socat STDIO PROXY:<proxyIP>:%h:%p,proxyport=<proxyport>,proxyauth=<proxyAuth>
    User git
    Hostname ssh.github.com
    Port 443
    IdentityFile /etc/fluxd/ssh/identity
```

### DIFF
--- a/chart/flux/templates/ssh.yaml
+++ b/chart/flux/templates/ssh.yaml
@@ -15,4 +15,14 @@ data:
         {{ .Values.ssh.known_hosts }}
       {{- end }}
     {{- end }}
+  config: |
+    {{- if .Values.ssh.config }}
+      {{- if contains "\n" .Values.ssh.config }}
+        {{- range $value := .Values.ssh.config | splitList "\n" }}
+          {{ print $value }}
+      {{- end }}
+      {{- else }}
+          {{ .Values.ssh.config }}
+      {{- end }}
+    {{- end }}
 {{- end -}}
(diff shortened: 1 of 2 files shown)


chart/flux/templates/ssh.yaml
```diff
@@ -15,4 +15,14 @@ data:
         {{ .Values.ssh.known_hosts }}
       {{- end }}
     {{- end }}
+  config: |
+    {{- if .Values.ssh.config }}
```
stefanprodan, line 19: When a config is not specified, this creates an empty file. Can you please move the `if` to include the the config entry.

Abhishek-Srivastava: That makes total sense. Pushed a new commit. Done a quick validation too. Thanks a ton reviewing.

Abhishek-Srivastava:
> LGTM
> 
> Thanks @Abhishek-Srivastava please rebase with master so we can merge it.

Done!!! Thanks a ton again.

Abhishek-Srivastava:
@stefanprodan can we merge it?

Issue #3058, opened 2020-05-13
Flux not watching private docker registry


nadendlaprasad:
```
# create git secret
kubectl create secret generic flux-git-auth --namespace fluxcd --from-literal=GIT_AUTHUSER=myaccount --from-literal=GIT_AUTHKEY=readacted

# create docker registry secret
kubectl create secret --namespace fluxcd docker-registry docker-config \
  --docker-server="myprivatedockerregistry.com" \
  --docker-username="admin" \
  --docker-password="redacted"

# applying dock-config secret  for default service account image pullsecrets
kubectl patch serviceaccount flux -p '{"imagePullSecrets": [{"name": "docker-config"}]}' -n fluxcd

# installing flux

helm upgrade -i flux fluxcd/flux \
--set git.url='https://$(GIT_AUTHUSER):$(GIT_AUTHKEY)@my-gitlab.com/myaccount/flux-helm-operator.git' \
--set env.secretName=flux-git-auth \
--set registry.dockercfg.configFileName="/dockercfg/config.json" \
--set registry.dockercfg.enabled=true \
--set registry.dockercfg.secretName=docker-config \
--set-file ssh.known_hosts=/tmp/flux_known_hosts \
--set allowedNamespaces="app-dev" \
--set serviceAccount.create=false \
--set serviceAccount.name=default \
--set registry.trace=true \
--namespace fluxcd

# installing helm operator
helm upgrade -i helm-operator fluxcd/helm-operator --wait \
--namespace fluxcd \
--set image.pullSecret=docker-config \
--set git.ssh.secretName=flux-git-deploy \
--set helm.versions=v3 \
--set allowNamespace="app-dev"


```

I can see it is able to pull git repo from private git host and deploy resources via helm operator. But It is not able to watch private registry.

## interesting message from the flux pod logs
warning="--docker-config not used; pre-flight check failed" err="open /dockercfg/config.json: no such file or directory"

Issue:
Flux is not watching the private docker registry. But it is able to connect to my private gitlab host and able to deploy resources via helm operator. 
FYI, I am using the latest chart.

Is there anything wrong with my config ?

pmatheson-greenphire:
It looks like the helm chart mounts docker-credentials to /dockercfg/ instead of /dockercfg/config.json
https://github.com/fluxcd/flux/blob/c02b716fc36e8a3b2cb663d67eb34aea00deb926/chart/flux/templates/deployment.yaml#L84-L88

[...]

PR #3161, opened 2020-06-26
Fix dockercfg mountPath

pmatheson-greenphire:
https://github.com/fluxcd/flux/issues/3058
make default dockercfg mountPath the same as default value registry.dockercfg.configFileName

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -158,7 +158,7 @@ spec:
           {{- end }}
           {{- if .Values.registry.dockercfg.enabled }}
           - name: docker-credentials
-            mountPath: /dockercfg/
+            mountPath: /dockercfg/config.json 
             readOnly: true
           {{- end }}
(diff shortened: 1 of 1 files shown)


kingdonb:
If I am understanding correctly, the `registry.dockercfg` section of Flux chart values is totally unusable without this fix, is that correct?

kingdonb:
I haven't lost track of this. I am going to try to include this change in the next release, since we are trying to catch up any dependencies and this targets the latest release of Helm Operator, it might fit.

I may rebase this and include it through another PR, to simplify the review workflow. I have just used the github "Update branch" feature to drag it along for now, so we can get a fresh CI pass.

kingdonb:
If you can, please rebase and satisfy the DCO check before the next release window 👍

kingdonb:
Seems the submitter has lost interest, I am closing this.

Will be happy to reopen or accept a different pull request if this is still affecting someone. I just can't accept it without a DCO.

PR #3164, opened 2020-06-27
Add the --git-ci-skip-message parameter.

mmorejon:
This changes will include the `--git-ci-skip-message` parameter into the helm chart.

Signed-off-by: Manuel Morejon <manuel@mmorejon.io>

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -221,7 +221,8 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `git.signingKey`                                  | `None`                                               | If set, commits will be signed with this GPG key
 | `git.verifySignatures`                            | `false`                                              | If set, the signatures of the sync tag and commits will be verified
 | `git.label`                                       | `flux-sync`                                          | Label to keep track of sync progress, used to tag the Git branch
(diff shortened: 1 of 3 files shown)


kingdonb:
Thanks for contributing this, it looks helpful

I think it might not be possible to merge this unfortunately without breaking peoples' existing configurations or requiring those upgrading for stable fixes to read and follow change notes; at the minimum it will of course require a new Helm chart release. I've noted this and I'm not sure what action to take for now, please bear with me as I police the remaining open PRs.

kingdonb:
Thank you for your contribution. Flux v1 is in maintenance mode and is no longer adding new features unless they are critical. If you are still looking for a way to include this option in the Helm chart, you can use `additionalArgs`, it is already supported in a way. Since this change would require Flux v1 users who have set `ciSkip` to change their `values.yaml`, it is a breaking change and cannot be merged as-is.

If this change is still important to you, then we can review it once again, however the expected reasons why changes can be merged to Flux v1 now that it is in maintenance mode are limited to security and maintenance concerns as well as bug fixes. Since there is a way to pass this option through `values.yaml` (however poorly documented) it would not seem to be a critical change, I think we will leave it behind.

I will close this for now, but happy to discuss it further, and hope you have been able to migrate to Flux v2, or are considering it soon! Thanks again.

Issue #2792, opened 2020-01-27
Provide Grafana Dashboard with Flux Prometheus metrics


2opremio:
Users have been requesting a Grafana dashboard for quite a while (see https://github.com/fluxcd/flux/issues/385 ).

There is an unofficial (outdated) dashboard at https://grafana.com/grafana/dashboards/10475 but maybe we should supply an official one.

Further information about metrics https://docs.fluxcd.io/en/1.17.1/references/monitoring.html

2opremio:
To keep the dashboards up to date along with new metrics we should probably use https://github.com/weaveworks/grafanalib or https://github.com/jkcfg/grafana

lucasvmiguel:
I see the value of using something like [grafanalib](https://github.com/weaveworks/grafanalib) to create the dashboard, but I believe is going to create more complexity (setting up python env, generating the dashboard json, etc) than helping us to make it easier to keep it updated.

btw, I'm happy to work on that if there is no one yet.

lucasvmiguel:
Is that a good start point?

Let me know if I'm missing any fundamental metrics.

![Screenshot 2020-06-26 at 17 01 04](https://user-images.githubusercontent.com/7236505/85877180-abb2f700-b7ce-11ea-98ec-5ad3ab54113e.png)

tux-00:
Is it possible to have the number of failed sync ? Maybe the number of manifests synced is 0 when errors ?

lucasvmiguel:
The first graph shows `Synced Manifests` and `Synced Manifest Failures` (didn't have any failures at that moment though). On Monday, I can open a PR with those metrics (and any other that you folks suggest).

I just want to confirm if it's ok to create just a config map with that (something like [here](https://github.com/istio/istio/blob/a671e62bd1ca3e6fd09bd0fbe03545c136b0f3f5/manifests/charts/istio-telemetry/grafana/templates/configmap-dashboards.yaml))

MihaiAnei:
Looking forward for the dashboard :)

I just want to add that it would be nice for the dashboard to be "compiled" (if using grafanalib) and available on a public URl just like 
```
https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/grafana/dashboards/nginx.json
```
because the sidecar container of the grafana helm chart used in prometheus-operator chart, for some time now, support loading dashboards from URLs. A simple Grafana restart would load the latest version.

stefanprodan:
[...]

PR #3169, opened 2020-06-30
Add new grafana dashboard for flux

lucasvmiguel:
Add new grafana dashboard for flux, this dashboard was discussed on the following issue: https://github.com/fluxcd/flux/issues/2792

85877180-abb2f700-b7ce-11ea-98ec-5ad3ab54113e

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -278,6 +278,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `syncGarbageCollection.dry`                       | `false`                                              | If enabled, fluxd won't delete any resources, but log the garbage collection output (see [garbage collection](../../docs/references/garbagecollection.md))
 | `manifestGeneration`                              | `false`                                              | If enabled, fluxd will look for `.flux.yaml` and run Kustomize or other manifest generators
 | `hostAliases`                                     | `{}`                                                 | Additional hostAliases to add to the Flux pod(s). See <https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/>
+| `dashboard.enabled`                               | `false`                                              | If enabled, flux will create a configmap with a dashboard in json that's going to be picked up by grafana (see [sidecar.dashboards.enabled](https://github.com/helm/charts/tree/master/stable/grafana#configuration))
 
 [memcached-ref]: https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L201-L204
 [kubeconfig-ref]: https://github.com/fluxcd/flux/blob/master/chart/flux/values.yaml#L232-L244
--- a/chart/flux/dashboards/flux-dashboard.json
+++ b/chart/flux/dashboards/flux-dashboard.json
@@ -1,0 +1,524 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    },
+    {
+      "name": "DS_LOKI",
+      "label": "Loki",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "loki",
+      "pluginName": "Loki"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "7.0.3"
+    },
+    {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph",
+      "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "logs",
+      "name": "Logs",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "loki",
+      "name": "Loki",
+      "version": "1.0.0"
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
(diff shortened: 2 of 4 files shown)


lucasvmiguel:
As you can see on the pic, it worked just fine in my cluster, but it would be good to see if it works in someone else's cluster as well.

chart/flux/values.yaml
```diff
@@ -312,3 +312,7 @@ hostAliases: {}
 #    hostnames:
 #    - "foo.remote"
 #    - "bar.remote"
+
+dashboards:
+  # If enabled, flux will create a configmap with a dashboard in json that's going to be picked up by grafana
+  enabled: true
```
stefanprodan, line 318: Add a new line here. Also please set it to false as this depends on prom annotations that are disabled by default.

chart/flux/templates/configmap-dashboards.yaml
```diff
@@ -0,0 +1,20 @@
...
+    grafana_dashboard: "1"
+    app: {{ template "flux.name" $ }}
+    chart: {{ template "flux.chart" $ }}
+    release: {{ $.Release.Name }}
+    heritage: {{ $.Release.Service }}
+data:
+  {{ base $path }}: '{{ $files.Get $path }}'
+---
+{{- end }}
+{{- end -}}
```
stefanprodan, line 20: New line here

chart/flux/README.md
```diff
@@ -278,6 +278,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `syncGarbageCollection.dry`                       | `false`                                              | If enabled, fluxd won't delete any resources, but log the garbage collection output (see [garbage collection](../../docs/references/garbagecollection.md))
 | `manifestGeneration`                              | `false`                                              | If enabled, fluxd will look for `.flux.yaml` and run Kustomize or other manifest generators
 | `hostAliases`                                     | `{}`                                                 | Additional hostAliases to add to the Flux pod(s). See <https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/>
+| `dashboard.enabled`                               | `true`                                               | If enabled, flux will create a configmap with a dashboard in json that's going to be picked up by grafana
```
stefanprodan, line 281: Default should be false

chart/flux/templates/configmap-dashboards.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.rbac.create -}}
```
stefanprodan, line 1: Why is this conditioned by RBAC?

chart/flux/dashboards/flux-dashboard.json
```diff
@@ -0,0 +1,468 @@
...
+      "xaxis": {
+        "buckets": null,
+        "mode": "time",
+        "name": null,
+        "show": true,
+        "values": []
+      },
+      "yaxes": [
+        {
+          "format": "short",
```
billimek, line 222: Consider making this `"format": "s",` so that the axis values will be labeled as seconds instead of just a raw short number.

chart/flux/dashboards/flux-dashboard.json
```diff
@@ -0,0 +1,468 @@
...
```
billimek, line 318: Consider making this `"format": "s",` so that the axis values will be labeled as seconds instead of just a raw short number.

billimek:
@lucasvmiguel thanks so much for submitting this.  I did notice when I added the dashboard json directly to my grafana, it had the logs panels set to `Prometheus` (which is my default datasource) and not `Loki`.

I wonder, is the dashboard json the 'raw' dashboard json data from your grafana, or was it 'exported for sharing'?

lucasvmiguel:
@billimek I didn't use this to export the JSON, I'm gonna use that and see if it fixes the issue

chart/flux/templates/configmap-dashboards.yaml
lucasvmiguel: I think it was an accident, I'll remove it

billimek:
Importing the dashboard json from this PR to my grafana server results in a perfectly-working dashboard, FWIW:
image

MihaiAnei:
Is the chart released with these changes? I did not see a version increase in Chart.yaml

tux-00:
@MihaiAnei I was wondering the same thing.

By the way maybe we can add a namespace value ? I don't remember if grafana is looking at other namespaces than its own namespace.

MihaiAnei:
```
grafana:
      sidecar:
        dashboards:
          searchNamespace: ALL
        datasources:
          searchNamespace: ALL
```
as long as the dashboard has 
```
grafana_dashboard: "1"
```

With the above settings in prom-operator or your separate grafana deployment, it should load any dashboard in any configmap in all namespaces :)

I like to keep the resources in the same ns they are deployed so I don't think this is required.

lucasvmiguel:
@stefanprodan is there any place that we find how it works the release process? (related to @MihaiAnei question)

tux-00:
@MihaiAnei I use more or less default prometheus operator to manager Grafana and dashboards. I tried to deploy the Flux helm chart 1.4.0 (the one with dashboard modifications), it appears that grafana is not reading dashboards outsite of its own namespace.
The dashboard was created in my flux namespace first and there was no Flux dashboard in Grafana, then I exported the dashboard into my Grafana namespace and it worked.
I can open a PR to add the possibility to change the namespace of the dashboard.

MihaiAnei:
@tux-00 can I see your grafana values you used to deploy it?
Are you sure you added this?
```
grafana:
  sidecar:
    dashboards:
      searchNamespace: ALL
    datasources:
      searchNamespace: ALL
```

tux-00:
Grafana values for prometheus operator are like this, I don't have changed the searchNamespace value:
```
grafana:
  additionalDataSources: []
  adminPassword: ...
  defaultDashboardsEnabled: true
  enabled: true
  extraConfigmapMounts: []
  ingress:
    ...
  namespaceOverride: ""
  plugins:
  - grafana-piechart-panel
  service:
    portName: service
  serviceMonitor:
    interval: ""
    metricRelabelings: []
    relabelings: []
    selfMonitor: true
  sidecar:
    dashboards:
      enabled: true
      label: grafana_dashboard
    datasources:
      annotations: {}
      createPrometheusReplicasDatasources: false
      defaultDatasourceEnabled: true
      enabled: true
      label: grafana_datasource
  tolerations:
  - effect: NoSchedule
    key: key
    operator: Equal
    value: node-role.monitoring
```
But I would like to push the dashboard in the same place as the others. I'm currently making a PR to do this.

PR #3140, opened 2020-06-16
Limit git commit message to first 10 images

alex-shpak:
In order to address error mentioned in #1857
```
ts=2020-06-16T14:06:12.217892013Z caller=loop.go:151 component=sync-loop jobID=2e650640-7405-54e9-1e77-43fa9663e01c state=done success=false err="fork/exec /usr/bin/git: argument list too long"
```

This PR will change message in case if there are over 10 docker images updated to:
```
Auto-release multiple (11) images

 - docker.io/image:v1
 - docker.io/image:v10
 - docker.io/image:v11
 - docker.io/image:v2
 - docker.io/image:v3
 - docker.io/image:v4
 - docker.io/image:v5
 - docker.io/image:v6
 - docker.io/image:v7
 - docker.io/image:v8
 ...
```

In our case issue was that after few failed attempts to commit to git repo there was large amount of automated releases to apply, which resulted in mentioned error.

I hope to merge this fix, it has high impact on our environment.
Thank you!

PR #3179, opened 2020-07-08
Put git messages into tmp files

alex-shpak:
Unfortunately #3140 was not enough to fix problem.

This PR will partially revert #3140 and also put git message for `git commit` and `git note` into TMP files. Preventing `args too long` error.


alex-shpak:
As for commit message, biggest I got so far was 21 images, it is bearable to read. For notes, for big cluster it gets multiple kb of json.

P.S. Updated branch.

PR #3208, opened 2020-07-21
chart: add kube.externalConfig option

MaxRink:
This Flags adds the Option to skip the generation of an default
kubeconfig and KUBECONFIG ENV Var to provide both on your own.
This enables for example bringing in a kubeconfig from an external
secret via ExtraEnv and ExtraVolumes/Mounts

- [x] if it introduces a new functionality or configuration flag;
      did you document this in the references or guides?

- [ ] optional but much appreciated;
      do you think many users would profit from a dedicated setting
      for this functionality in the Helm chart?
--> This is a niche use-case.
We basically use Flux in conjunction with Cluster API and use flux to do inital provisioning and updatesof stuff like the CNI and other Flux instances. We basically have everything encapsulated in Helm charts and are consuming Flux as a subchart in our Cluster chart. At that point Cluster-API only outputs the kubeconfig as secret which we currently cant use with the Flux Helm chart, only with an self-templated deployment which is not Ideal.
This PR basically adds ne necassesary featureflag for us to use the Flux helm chart and not having to maintain a fork of just this one feature-flag.

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -267,6 +267,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `memcached.nodeSelector`                          | `{}`                                                 | Node Selector properties for the memcached deployment
 | `memcached.tolerations`                           | `[]`                                                 | Tolerations properties for the memcached deployment
 | `memcached.priorityClassName`                     | `""`                                                 | The name of the priority class to assign to the memcached pod.
+| `kube.externalConfig`                             | `false`                                              | If enabled, no kubeconfig and env var pointing to the kubeconfig will be created. You need to provide both on your own.
 | `kube.config`                                     | [See values.yaml][kubeconfig-ref]                    | Override for kubectl default config in the Flux pod(s).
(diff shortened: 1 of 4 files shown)


chart/flux/README.md
```diff
@@ -267,6 +267,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `memcached.nodeSelector`                          | `{}`                                                 | Node Selector properties for the memcached deployment
 | `memcached.tolerations`                           | `[]`                                                 | Tolerations properties for the memcached deployment
 | `memcached.priorityClassName`                     | `""`                                                 | The name of the priority class to assign to the memcached pod.
+| `kube.externalConfig`                             | `false`                                              | If enabled, no kubeconfig and ENV Var pointing to the kubeconfig will be created. You need to provide both on your own.
```
hiddeco, line 270:
```suggestion
| `kube.externalConfig`                             | `false`                                              | If enabled, no kubeconfig and env var pointing to the kubeconfig will be created. You need to provide both on your own.
```

MaxRink:
Done,

PR #3209, opened 2020-07-22
Introduce advanced templating for use as subchart

MaxRink:
This introduces advanced templating that allowsyou to easier use the Flux chart as a subchart in another helm v2 Chart.
This shouldn't break anything as tpl already is a supported feature with Helm 2.
It adds the benefit that the usage of flux as a subcharts becomes easier as you can now reference global values within your values.yaml, thus you dont need to define the same value multiple times and can just crossreference it.
Example:

```yaml
global:
  tenant:
    name: tenant_1
    environment: packet
    git: "git.example.com"
    token: "somethingCryptic"
    clustername: "workload-cluster-1"

flux:
  git:
    url: https://x-access-token:{{ .Values.global.tenant.token }}@{{ .Values.global.tenant.git }}/schiff/platform/infrastructure/alpha3-environments.git
    branch: "master"
    path: "platform/tenantClusterComponents/{{ .Values.global.tenant.environment }}/{{ .Values.global.tenant.name }}/{{ .Values.global.tenant.clustername }}/core-components"
    user: "{{ .Values.global.tenant.clustername }}-flux"
    email: "{{ .Values.global.tenant.clustername }}-flux@clusters.das-schiff.telekom.de"
```
Addendum:
imo it doesnt make sense to introduce this to every option, as you propably wont overwrite many of the values from your umbrella chart. This PR basically just introduces what /we/ needed, but im sure that there are other values where it also would make sense to switch to this style.

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -301,3 +301,30 @@ Update Flux version with:
 helm upgrade --reuse-values flux fluxcd/flux \
 --set image.tag=1.17.1
 ```
+
+### Use as a subchart
+
+This chart allows certain variables to be centrally declared in the umbrella chart and then resused in the [subchart](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/).
+
+example:
+```yaml
+global:
+   git:
+      url: "your_git_host.example.com"
+
+umbrellachart:
+   git-url: "{{ .Values.global.git.url }}"
+
+flux:
+   git:
+      url: "{{ .Values.global.git.url }}"
+```
+
+This is currently possible for this variables:
+* extraVolumes
+* git.url
+* git.branch
+* git.path
+* git.readonly
+* git.user
+* git.email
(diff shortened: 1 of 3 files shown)


MaxRink:
In what section would you see the documentation?
In the README and values.yaml or just in one of them?

hiddeco:
@MaxRink `README.md` of the chart would be the best place I think.

MaxRink:
I've added my initial draft of docs for this feature and actually found a few more Variables that are propably woth handeling in this mannor, like HostAliases.
I'll addd them to this PR once i find time to write the necassesary helper code for nested lists.

kingdonb:

I have seen Flux used in Umbrella charts, like in this example: https://github.com/okteto/charts/tree/master/flux

I think that means this change is not needed, but it's possible there is something I have misunderstood about this.

As some time has passed since this was submitted and there has been no follow-up for a while, I think it's safer to close.

PR #3227, opened 2020-07-30
Add a secretDataKey parameter to the Flux Helm chart

### LINKED ISSUE #3226: Add a Helm chart parameter for the SSH secret's data key
**Describe the feature**
Add a `git.secretDataKey` parameter to the Helm chart that enables the user to override the default deploy key secret's data key `identity`.   They can then create the secret from a file with a different name (like `id_rsa`) or to use a pre-existing secret that stores the deploy key's secret key value under a different data key.

1. User either creates a secret to store their deploy key's private key using an existing id_rsa key file or decides to use a deploy key stored in an existing secret.
2. User adds a secretDataKey field to the Helm chart's values.yaml to override the default key name `identity`.
3. User deploys Flux using the Helm chart and their values.yaml overrides.
4. Flux connects to Git using the private key stored in the Secret described by `git.secretName` and the data key described by `git.secretDataKey` values in the values.yaml.

**Expected behavior**
The Flux Helm chart should enable the user to set the data key of the deploy key secret in a `git.secretDataKey` field in their values.yaml.

sbernheim:
Adding a secretDataKey parameter to the Flux Helm chart enables the user to override the default data key `identity` and thereby potentially create the secret from a file with a different name (such as id_rsa) or use a pre-existing secret that stores the key value using a different data key.

Expanding the Helm chart's documentation on the subject of ssh keys to add some clarity and cover this parameter.

resolves #3226

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -199,7 +199,9 @@ spec:
           - --ssh-keygen-dir=/var/fluxd/keygen
           - --ssh-keygen-format={{ .Values.ssh.keygen.format }}
           - --k8s-secret-name={{ .Values.git.secretName | default (printf "%s-git-deploy" (include "flux.fullname" .)) }}
+          {{- if .Values.git.secretDataKey }}
           - --k8s-secret-data-key={{ .Values.git.secretDataKey | default (printf "identity" ) }}
+          {{ - end }}
           - --memcached-hostname={{ .Values.memcached.hostnameOverride | default (printf "%s-memcached" (include "flux.fullname" .)) }}
           - --sync-state={{ .Values.sync.state }}
           {{- if .Values.sync.timeout }}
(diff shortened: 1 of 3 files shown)


chart/flux/templates/deployment.yaml
```diff
@@ -199,6 +199,7 @@ spec:
           - --ssh-keygen-dir=/var/fluxd/keygen
           - --ssh-keygen-format={{ .Values.ssh.keygen.format }}
           - --k8s-secret-name={{ .Values.git.secretName | default (printf "%s-git-deploy" (include "flux.fullname" .)) }}
+          - --k8s-secret-data-key={{ .Values.git.secretDataKey | default (printf "identity" ) }}
```
hiddeco, line 202: I think putting an if-statement around this instead of defaulting would be a better approach, as it prevents a double default declaration.

sbernheim: Added an if statement to check if the secretDataKey value is set.  Also changed the default value in values.yaml from "identity" to a blank string "" (although the documentation remains the same).

sbernheim: @hiddeco - Please review and resolve this conversation if you're satisfied with the change.

hiddeco: Due to the added `if`-statement, the `default` is now no longer necessary.

```suggestion
          - --k8s-secret-data-key={{ .Values.git.secretDataKey }}
```

hiddeco (approved):
Thank you @sbernheim :1st_place_medal: :tulip:

Issue #3049, opened 2020-05-08
--sync-timeout is capped at 60s


awwithro:
**Describe the bug**

We're running a sync that is taking longer than a minute and hits a timeout. We tried to raise the timeout to account for this but we still get a `context deadline exceeded` error. If we lower the timeout to less than 60s we see timeouts sooner. I suspect there is another timeout that is being hit that is overriding the effective sync-timeout

**To Reproduce**

1. Create a job that will take longer than 60s (sleep 300)
2. Run flux with a --sync-duration greater than the job's run time

**Expected behavior**

The sync timeout should be respected 

**Logs**

our args
```
      - --log-format=fmt
      - --ssh-keygen-dir=/var/fluxd/keygen
      - --ssh-keygen-format=RFC4716
      - --k8s-secret-name=flux-git-deploy
      - --memcached-hostname=fluxcd-memcached
      - --sync-state=git
      - --sync-timeout=2m
      - --memcached-service=
      - --git-url=ssh://{{our repo}}
      - --git-branch=master
      - --git-path={{ our path }}
      - --git-readonly=true
      - --git-user={{our user }}
      - --git-email={{ our email }
      - --git-verify-signatures=false
      - --git-set-author=false
      - --git-poll-interval=5m
      - --git-timeout=20s
      - --sync-interval=5m
      - --git-ci-skip=false
      - --manifest-generation=true
      - --automation-interval=5m
      - --registry-rps=200
      - --registry-burst=125
      - --registry-trace=false
      - --registry-disable-scanning
```

```
ts=2020-05-08T15:37:36.922163159Z caller=sync.go:73 component=daemon info="trying to sync git changes to the cluster" old=99ccd2c99ce8a771caf3139db6807d8af2b5b78e new=d4fc25c5a4a2d545007cff0e3433b59a452308d2
ts=2020-05-08T15:38:40.902810713Z caller=loop.go:107 component=sync-loop err="loading resources from repo: error executing generator command \"../../bin/flux-generator.sh\" from file \"../../.flux.yaml\": context deadline exceeded\nerror output:\n\ngenerated output:\n"
```

**Additional context**

- Flux version: 1.19.0
- Kubernetes version: 1.18.2
- Git provider: AWS code commit
- Container registry provider: bintray

saravanakumar-periyasamy:
the command execution timesout at 60s and is set here - https://github.com/fluxcd/flux/blob/v1.19.0/pkg/manifests/configfile.go#L24

[...]

PR #3228, opened 2020-07-31
feat: optional command timeout field in config file that defaults  to syncTimeout

marshallford:
Adds the ability to set an optional timeout duration for all commands that defaults to the sync timeout instead of `time.Minute`. Attempts to resolve: #3049. Keep in mind that in order to fix the issue described, the sync timeout **must be** greater than or equal to the sum of the generator command timeouts. I am open to ideas or to scraping this completely. Thanks Flux team!

Example:

```yaml
version: 1
patchUpdated:
  generators:
  - command: kustomize build .
    timeout: 3m # if field is not provided, sync timeout is used
  patchFile: flux-patch.yaml
```

Couple of other notes:

1. ~~I looked into parsing `timeout` from the config yaml as the type `time.Duration` instead of a string but the use of the `ghodss/yaml` package made that a bit confusing (for a new golang user anyway). It would have been nice to return an error in `ParseConfigFile` if an invalid duration was given instead of using the syncTimeout as a default without notice.~~

Solved by C&P this snippet: https://stackoverflow.com/a/48051946

2. I'm not super happy with adding the field `defaultTimeout` to the `configAware` struct. Open to suggestions on how to improve passing through the sync timeout.

TODO:

- [ ] Investigate better way of passing around/through sync timeout flag
- [ ] Update docs once this method of setting a command timeout per generator is approved

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -73,7 +73,7 @@ type repo interface {
 func (d *Daemon) getManifestStore(r repo) (manifests.Store, error) {
 	absPaths := git.MakeAbsolutePaths(r, d.GitConfig.Paths)
 	if d.ManifestGenerationEnabled {
-		return manifests.NewConfigAware(r.Dir(), absPaths, d.Manifests)
+		return manifests.NewConfigAware(r.Dir(), absPaths, d.Manifests, d.SyncTimeout)
 	}
 	return manifests.NewRawFiles(r.Dir(), absPaths, d.Manifests), nil
 }
--- a/pkg/manifests/configfile.go
+++ b/pkg/manifests/configfile.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	jsonschema "github.com/xeipuuv/gojsonschema"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/fluxcd/flux/pkg/image"
 	"github.com/fluxcd/flux/pkg/resource"
@@ -21,7 +22,6 @@ import (
 
 const (
 	ConfigFilename = ".flux.yaml"
-	CommandTimeout = time.Minute
 )
 
 // This is easier to read as YAML, trust me.
@@ -29,8 +29,12 @@ const configSchemaYAML = `
 "$schema": http://json-schema.org/draft-07/schema#
 definitions:
   command:
-    type: object
     required: ['command']
+    type: object
+    properties:
+      command: { type: string }
+      timeout: { type: string }
+    additionalProperties: false
   version: { const: 1 }
 type: object
 oneOf:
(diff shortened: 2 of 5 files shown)


marshallford:
@stefanprodan could I get a review on this? Would you be open to a schema change like this?

marshallford:
@stefanprodan ping

marshallford:
ping @hiddeco

rdubya16:
@stefanprodan Any movement on this? This is blocking us as well.

mike-stewart:
@squaremo We're running into this bug as well. We are running a build of this PR in one of our clusters now, and it seems to be resolving the issue. Any chance this PR could be merged?

pkg/daemon/daemon.go
```diff
@@ -12,10 +12,10 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/fluxcd/flux/pkg/api"
-	"github.com/fluxcd/flux/pkg/api/v10"
-	"github.com/fluxcd/flux/pkg/api/v11"
-	"github.com/fluxcd/flux/pkg/api/v6"
-	"github.com/fluxcd/flux/pkg/api/v9"
+	v10 "github.com/fluxcd/flux/pkg/api/v10"
```
squaremo, line 15: Was this (and the ff) changed by an IDE? I don't think it's necessary, or relevant to this PR.

pkg/manifests/configfile.go
```diff
@@ -84,6 +88,35 @@ func mustCompileConfigSchema() *jsonschema.Schema {
 	return schema
 }
 
+type Duration struct {
```
squaremo, line 91: I think `metav1.Duration` (k8s.io/apimachinery/pkg/apis/meta/v1) could be used here.

pkg/manifests/configfile.go
```diff
@@ -110,39 +143,29 @@ type ConfigFile struct {
...
-	Generators []Generator `json:"generators"`
-	Updaters   []Updater   `json:"updaters,omitempty"`
+	Generators []Command `json:"generators"`
+	Updaters   []Updater `json:"updaters,omitempty"`
 }
 
 // Generator is an individual command for generating manifests.
-type Generator struct {
-	Command string `json:"command,omitempty"`
+type Command struct {
```
squaremo, line 151: The preceding comment needs updating.

pkg/manifests/configfile.go
```diff
@@ -427,52 +454,56 @@ func (cf *ConfigFile) execGenerators(ctx context.Context, generators []Generator
...
 
-func (cf *ConfigFile) execCommandsWithCombinedOutput(ctx context.Context, env []string, commands []string) []ConfigFileCombinedExecResult {
+func (cf *ConfigFile) execCommandsWithCombinedOutput(ctx context.Context, env []string, commands []Command, defaultTimeout time.Duration) []ConfigFileCombinedExecResult {
 	env = append(env, "PATH="+os.Getenv("PATH"))
 	result := []ConfigFileCombinedExecResult{}
 	for _, c := range commands {
 		stdOutAndErr := bytes.NewBuffer(nil)
-		err := cf.execCommand(ctx, env, stdOutAndErr, stdOutAndErr, c)
+		timeout := c.Timeout
+		if &c.Timeout == nil || timeout.Duration < time.Second {
```
squaremo, line 503: `if &c.Timeout == nil` does not make sense as a comparison. It says "is the _address_ of a field in a value `nil`", which will never be true. If you want the timeout value to be optional, make the type of the field a pointer.

squaremo (changes requested):
Thanks for putting time into this! I appreciate the need, and I'm happy with the idea. Passing the default timeout is a bit awkward, as you mention; I'm fine with it though, at least it's obvious what's happening.
There's some bits in the implementation that need another look, which I've flagged up.

pkg/manifests/configfile.go
marshallford: I chose to create my own type so that I could marshal and unmarshal the duration.

pkg/daemon/daemon.go
marshallford: Removed. VS Code + the Go extension was making this change on save.

pkg/manifests/configfile.go
marshallford: fixed.

pkg/manifests/configfile.go
marshallford: I think I fixed this?

pkg/manifests/configfile.go
squaremo: `metav1.Duration` has its own marshalling code, so it parses a value of e.g., `5m`. If it's a requirement to allow floats here, I guess you are stuck with implementing your own. Is that a hard requirement?

squaremo:
I've rebased this, and added a commit for using `metav1.Duration`. Once it passes CI I'll squash those early few commits and merge.

pkg/manifests/configfile.go
marshallford: Oh ok, perfect. I'm fine with no floats.

marshallford:
> I've rebased this, and added a commit for using `metav1.Duration`. Once it passes CI I'll squash those early few commits and merge.

Awesome Thanks!

squaremo (approved):

squaremo:
@marshallford Can you do me a quick favour? We now [require a DCO](https://github.com/fluxcd/flux/pull/3228/checks?check_run_id=1309845179) (as of today!). It's pretty easy to apply it -- instructions are in that link, and you will need to git reset your local branch head first, to pick up my rebase.

squaremo:
Brill, thank you @marshallford!

Issue #2904, opened 2020-03-11
Consider Supporting git-crypt


ekristen:
**Describe the feature**

git-crypt is simple and straightforward pgp encryption of files in a git repo.

SOPS while powerful is overly complicated for a small team and git-secret doesn't support diff-ing.

git-crypt is a better version of git-secret with diff support.

xUnholy:
This indeed would be an ideal feature to support - I've been looking into how viable this might be for my own setup.

https://github.com/raspbernetes/k8s-gitops/issues/57

ekristen:
After taking more time to review SOPS and to use it a bunch it is very powerful and useful for teams of any size. It's definitely designed for configuration data and secrets only, while it can handle random data it's not the best. I do find diff-ing is also not ideal, but the fact that you can encrypt with different keys, including KMS makes it worth it in my opinion.

kingdonb:
Flux v1 is formally superseded since the GitOps Toolkit APIs have been declared stable:

https://fluxcd.io/docs/migration/timetable/

The repo will remain in maintenance for some time, but no new features can be accepted. Bugs can be addressed if they are critical and there is a PR to resolve it, but soon only CVEs can be addressed in Flux v1, and new users are all recommended to use Flux v2 for some time now.

Thanks for using Flux!

PR #3239, opened 2020-08-09
Add git-crypt support

robpou:
This commit tries to add support for git-crypt. While I'm aware that both git-secret and sops are already supported, our team (and I'm sure many others) rely on git-crypt and since the additional support is quite trivial, I'm asking for including it. 

We are in progress of moving our processes to flux, and as we saw that you support git-secret our first thought was simply "ok let's change our repo to it then". But the more we had to work with it the less we liked it. The main reasons are:
- every commit (`git secret reveal` && editing something && `git secret hide`) results by default in a re-encryption of *all* of the secret files (in our case more than 100 files) which pollutes history
- git-secret relies on .gitignore entries, which actually hide the uncommited changes e.g. from `git status`
- git diff is useless with git-secret (the workaround painful e.g. https://github.com/msilvestre/git-secret-diff)

First https://github.com/fluxcd/flux/issues/1676 asked for crypto support, but that ended up implementing git-secret in https://github.com/fluxcd/flux/pull/2159. This commit would also solve conversations like https://github.com/fluxcd/flux/issues/2904

Please tell me if I can improve something and I will.


robpou:
I saw the CI error and will fix the issue with the missing function.
Should I close this and create a new MR or update this one?

chart/flux/templates/deployment.yaml
```diff
@@ -226,7 +226,10 @@ spec:
           - --git-verify-signatures={{ .Values.git.verifySignatures }}
           - --git-set-author={{ .Values.git.setAuthor }}
           {{- if .Values.git.secret.enabled }}
```
cedi, line 228: this line should be deleted, as it causes a syntax error:
two opening if's but only one end.
This line is made obsolete by the next line :)

PR #3248, opened 2020-08-15
Filter files before SOPS decryption

stefanprodan:
Changes:

* Decrypt YAMLs with SOPS only if the file has a 'sops' field. This should speed up the apply time for large repositories. 
* Update SOPS package and binary to v3.6.0


pkg/cluster/kubernetes/resource/load.go
```diff
@@ -207,7 +208,7 @@ func loadFile(path string, sopsEnabled bool) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	if sopsEnabled {
+	if sopsEnabled && strings.Contains(string(bytes), "sops:") {
```
squaremo, line 211: `bytes.Contains(bytes, []byte("sops:"))` (or better still, make that latter a var)

squaremo: (you'll have to rename the variable of course)

pkg/cluster/kubernetes/resource/load.go
```diff
@@ -203,14 +203,14 @@ func ParseMultidoc(multidoc []byte, source string) (map[string]KubeManifest, err
...
 // it will try to decrypt it before returning the data
 func loadFile(path string, sopsEnabled bool) ([]byte, error) {
-	bytes, err := ioutil.ReadFile(path)
+	fileBytes, err := ioutil.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
-	if sopsEnabled {
-		return softDecrypt(bytes)
+	if sopsEnabled && bytes.Contains(fileBytes, []byte("sops:")) {
```
redbaron, line 210: unlikely to be triggered IRL, but this disables sops for JSON-like content of yaml files

PR #3256, opened 2020-08-18
fix: automated image update for public registries

moshloop:


squaremo:
Can you say a bit about what you're trying to improve with this PR?

moshloop:
We have a workload using a Harbor Registry (1.10 and 2.0) with a public repository (i.e no imagePullSecrets) with an "automated"  update policy that only gets updated when this patch applied.

* Image scanning assumes there are credentials -  I tried using an explicit `--docker-config` but that didn't seem to work either. 
* I just realized that that I can replace the `d.ImageRefresh` hack with a `c.warm` call here https://github.com/fluxcd/flux/blob/master/pkg/registry/cache/warming.go#L130

kingdonb:
Thank you for your contribution.

If this change is still important to you, then we can review it once again, however the expected reasons why changes can be merged to Flux v1 now that it is in maintenance mode are limited to security and maintenance concerns as well as bug fixes. If you have already moved on to Flux v2 and are no longer interested in merging this change, then we can leave it closed.

The change is out-of-date with the base branch, and CI has been changed within the past months such that changes must have a DCO sign-off in order to be considered for merging. If you are still hoping to get this merged, we can consider it, but first I will need to have it rebased with a DCO sign-off added that satisfies the DCO bot. Please let me know if that is the case.

PR #3273, opened 2020-09-01
faq: --git-readonly implies --registry-disable-scanning

Frederik-Baetens:
I lost a bit of time figuring this out, and this note would have helped me.


kingdonb:
Thanks for your contribution here,

I think there are formatting changes that I would add, but if you give your Sign-Off then I can try to include this

Can you amend your PR to please include a sign-off for the DCO bot?

kingdonb:
Thank you for your contribution. I cannot consider any changes for merge unless they include a DCO sign-off.

As some time has passed since this was submitted and there has been no follow-up for a while, respecting that you may have moved on and no longer are concerned about this, I think it's safe to close.

PR #3274, opened 2020-09-02
add quivr to the flux user list in readme

Frederik-Baetens:
We've been very happy with it so far! Thanks for making this great tool.

We're a group of studens for which a tool like this provides a lot of time savings, reliability, and most importantly, makes our infra accessible to our new members without having to give everyone cluster access.


dholbach (approved):
Thanks a lot for the PR, @Frederik-Baetens - great to have you in our community!

PR #3276, opened 2020-09-03
Improve error when path from `--git-path` doesn't exist

michaelbeaumont:
The error implies not only that a config file needs to be found but precisely that error is caught in the previous lines.


kingdonb:
Thank you for your contribution, and sorry for the delay getting it attended to! Flux v2 has now crossed Feature Parity and is where the majority of development efforts are focused now.

If you are still interested in merging this, please rebase and add your DCO sign-off to satisfy the DCO bot.

kingdonb:
Thanks for your contribution. I can reopen, or you can resubmit, if you are still interested in this change.

Hope you've been able to upgrade to Flux v2, (in which case this can stay closed.) Thanks for using Flux!

PR #3281, opened 2020-09-07
Clarify misleading FAQ entry.

bricef:
Based on user feedback, the _Does Flux automatically sync changes back to git?_ FAQ entry was misleading and led to incorrect assumptions. This new entry should better reflect the current state of Flux.


stefanprodan (approved):
LGTM

Please squash your commits.

bricef:
Commits squashed!

PR #3290, opened 2020-09-14
Add Belvo as Flux user

mcanaves:
We have been using Flux in production for a year. Thanks to everyone for all your excellent work!


hiddeco (approved):
Thank you @mcanaves! :sun_with_face:

PR #3293, opened 2020-09-15
Add value for memcached pod labels

KadenLNelson:
Users will now be able to add labes to the memcached pods using the
value `.Values.memcached.podLabels`. This value is currently available
for the Flux pods via `.Values.podLabels`.


KadenLNelson:
@hiddeco @dholbach may this be reviewed?

kingdonb:
This looks like a non-breaking change to the helm chart and it passes CI. The `blocked-dco` label should be added (or the remaining CI check can be automatically disposed, @KadenLNelson, if you add your signoff with `git commit --amend -s` and `git push --force` to your branch memcached-podlabels)

I am not certain how rapidly this would be added to a release, thank you for your patience, there may be more releases but as you may have heard Flux v1 is in maintenance mode, we would love it if you try Flux v2 and see if it can meet your use case.

I am currently reviewing all of the open PRs and have noted that merging this one will require a new chart to be published as well. If there are other PRs to merge which also affect the chart, perhaps we can get them all in at once so that only one new Chart release will be needed. Thanks again for your contribution.

kingdonb:
This user vanished before DCO, the PR can be closed now if nobody else wanted to see it merged.

Closing.

PR #3296, opened 2020-09-16
Align antecedent annotation in flux and helm-operator

alex-shpak:
Hi!

I noticed that `Antecedent` field in fluxcd API is not showing resource ID that created by HelmOperator. Apparently annotation name is different in flux and helm-operator. This pull request will align them.

https://github.com/fluxcd/flux/blob/edadbf4d49ed23173ff60b932f6b99da0163bc3a/pkg/cluster/kubernetes/resourcekinds.go#L26
https://github.com/fluxcd/helm-operator/blob/master/pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go#L18


squaremo:
Good spot! Would be it sensible to make this backward compatible, do you think? In other words, to look for either the new or the old annotation. Also: can the const in the helm API package be used for the new annotation, or does that cause dependence problems?

alex-shpak:
Hi!
I can make it check both, if needed. Was not sure is that annotation is used at all now. Your call.

Regarding helm-operator dependency. Flux imports `RC` version of helm-operator
https://github.com/fluxcd/flux/blob/master/go.mod#L12-L17

and `AntecedentAnnotation` was introduced in `1.2.0`. When I update dependency to `1.2.0` it compiles, but I'm not sure is it safe to do.

I will give it a try and see if tests are still passing.

alex-shpak:
Had to force-push to branch for DCO check.

alex-shpak:
Well, tests has passed.

kingdonb:
I have rebased this and ran `go mod tidy` on my own fork, should this be merged (or does it somehow represent a breaking change?) My fork at https://github.com/kingdonb/flux/tree/patch-alex-shpak-1 -- this looks like a mostly housekeeping change to me.

kingdonb:
I didn't understand fully whether this change represents breaking something, or is still needed.

@alex-shpak do you still want to have this included in a forthcoming release of Flux v1?

alex-shpak:
Hi! Yes, would be great to include it in order to be able to connect created workloads to helm release.
There is chance that someone somewhere uses old annotation, but probably pretty low.

I will rebase.

kingdonb:
OK, I have rebased this for inclusion in the next release. It is included in #3442, the Omnibus Branch for 1.22.0.

kingdonb (approved):
LGTM 👍

alex-shpak:
Great!

PR #3300, opened 2020-09-19
Add zwift.com as Flux user in Production

lucioveloso:
Add Zwift.com in the Flux Production companies list


hiddeco (approved):
Thank you @lucioveloso 🌻

PR #3303, opened 2020-09-24
Correct FAQ typo: fluxcd.io/ignore value must be string

dewe:
Docs, FAQ

Fix typo: `fluxcd.io/ignore` value need quotes to force string value.


stefanprodan (approved):
LGTM

Thanks @dewe

PR #3311, opened 2020-09-29
Fixup configmap in Git GPG documentation

kingdonb:
When I tried `kubectl create configmap generic [...]` I got an error. I believe the "generic" template is for secrets, not configmaps.

This is a docs-only update – I believe this was just a typo.


hiddeco (approved):
Good catch, thanks @kingdonb :sunflower:

PR #3315, opened 2020-10-01
Adds Wildlife Studios to the production users list

douglasquintanilha:
Changes:
* Adds Wildlife Studios in the production users list.

We have been using Flux on production since the beginning of this year, thanks for this great project!


hiddeco (approved):
Thank you @douglasquintanilha :heart:

PR #3252, opened 2020-08-16
Notify upstream of sync failures

ordovicia:
This PR is a substitution of #3145 since the fork of #3145 was deleted.
From #3145 description:

> Sync can fail for many other reasons besides the manifests not being
> ok. Some of these errors could be due to some pre processing scripts
> in the .flux.yaml file or some git checkout errors. This commits logs
> sync erros and allows upstream log event providers to properly display
> the errors
> 
> Signed-off-by: Marcos Lilljedahl marcosnils@gmail.com 

I cherry-picked the commit of #3145 and added commits 7c84304e974f2da371c43fc1a1e33a2082284f37 to include commit info in sync failure event.

PR #3316, opened 2020-10-02
Notify upstream when sync fails

marcosnils:
This PR is a replacement of  of https://github.com/fluxcd/flux/pull/3252 since it has been closed inappropriately 

cc @fdpeiter @ordovicia

### DIFF
--- a/pkg/daemon/sync.go
+++ b/pkg/daemon/sync.go
@@ -55,6 +55,7 @@ func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string
 				EndedAt:   time.Now().UTC(),
 				LogLevel:  event.LogLevelError,
 				Message:   rerr.Error(),
+				Metadata:  &event.SyncEventMetadata{},
 			}
 
 			if changeSet != nil {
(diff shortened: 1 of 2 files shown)


kingdonb:
I am a bit new at this, do we consider Sync as a public method is part of our API, and would changing the signature of a public method constitute a breaking change?

In that vein, is this diff considered to be changing the signature, or is it handled as the same signature given both declarations have only one single return (of the same type)? Please forgive a somewhat novice Go question, I am still learning 📚

```diff
-func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string, rat ratchet) error {
+func (d *Daemon) Sync(ctx context.Context, started time.Time, newRevision string, rat ratchet) (rerr error) {
```

If this is to be merged, @marcosnils, it will need to have a `--signoff` added and probably should be rebased as well.

kingdonb:
I cannot merge anything without a DCO signoff. Closing for now.

Please resubmit if this is still interesting for you, and add the sign-off requested by DCO check in the Details link there.

PR #3331, opened 2020-10-15
Use fossa-contrib/fossa-action instead

smorimoto:
As the title says...


stefanprodan (approved):
LGTM

Thanks @smorimoto

smorimoto:
@stefanprodan Anytime :)

PR #2851, opened 2020-02-13
Improve logging

cmanzi:
Fixes https://github.com/fluxcd/flux/issues/2806, and partially addresses https://github.com/fluxcd/flux/issues/2812. 

We are heavy users of FluxCD in a multi-tenant environment, and the default logging was not cutting it for us, as mentioned in other issues.

It adds the following flags:
`--log-level` - set the log level for Zap logging output
`--log-devel` - set logging to Zap logging 'development' mode (trace, effectively)

Our solution was to replace the current logger with Zap structured logging (we use JSON, but it still supports `logfmt` by way of a compatibility library). We use JSON logging:
```
{"level":"error","ts":1581636184.4995604,"caller":"cache/warming.go:181","msg":"error requesting tags","component":"warmer","canonical_name":"quay.io/jetstack/cert-manager-controller","error":"error parsing HTTP 429 response body: invalid character '<' looking for beginning of value: \"<html>\\r\\n<head><title>429 Too Many Requests</title></head>\\r\\n<body bgcolor=\\\"white\\\">\\r\\n<center><h1>429 Too Many Requests</h1></center>\\r\\n<hr><center>nginx/1.12.1</center>\\r\\n</body>\\r\\n</html>\\r\\n\"","stacktrace":"github.com/fluxcd/flux/pkg/registry/cache.(*Warmer).warm\n\tgithub.com/fluxcd/flux/pkg/registry/cache/warming.go:181\ngithub.com/fluxcd/flux/pkg/registry/cache.(*Warmer).Loop\n\tgithub.com/fluxcd/flux/pkg/registry/cache/warming.go:124"}
[...]

Issue #2806, opened 2020-01-31
Leveled logging


prometherion:
**Describe the feature**

Actually, filtering logs per level is quite tricky: users that want to get only `error` or `warning` messages have to parse output via _grep_ or any other viable solution, even worse if they want to get messages above a selected threshold (e.g.: just _warn_ and _err_).

Since `flux` is already using the dependency [go-kit](https://github.com/fluxcd/flux/blob/0810999e759bccb31f132bb1c2d3987e18581c39/go.mod#L44) also for [logging](https://github.com/fluxcd/flux/blob/0810999e759bccb31f132bb1c2d3987e18581c39/cmd/fluxd/main.go#L243), would be nice to implement this feature decorating the logger thanks to the `level` [package](https://godoc.org/github.com/go-kit/kit/log/level) of the same dependency.

`level` allows several logging levels according to the [code base](https://github.com/go-kit/kit/blob/master/log/level/level.go#L69-L97) and this feature wouldn't be breaking: further, no needs of test refactoring or a new test suite.

**What would the new user story look like?**

1. providing the new `--verbosity` flag with a default value for getting actual logging expectations in order to don't break backward compatibility or users expecting of getting a full and detailed log, as it is right now.
2. allowed values for `--verbosity` flag would be shadowed to `level` ones since they're reasonable and easy to understand at first glance, so:
   - `allow-debug` to get all the logs (actual situation)
   - `allow-info` to get _info_, _warn_ and _error_ messages
   - `allow-warn` to get _warn_ and _error_ messages
   - `allow-error` to get only _error_ messages
3. refactoring the codebase

**Expected behavior**

Get the ability to get log messages just for a given level without the need for various _grep_ or _jq_ piping in case of JSON format output.

alexanderbuhler:
+1 on this one. We've just integrated with datadog and flux messages are being treated as errors by default. Some flux logs contain `info=` or `err=` so we can map them to a log level, others are missing any clue about level/verbosity, like this one:

```
ts=2020-02-11T12:00:49.890827678Z caller=warming.go:206 component=warmer updated=datadog/agent successful=0 attempted=2
```

kingdonb:
[...]

PR #3337, opened 2020-10-20
feat: log update manifest errors

salzig:
Find it hard to describe if this is something that anyone else would love to see, but i just saw that `calculateChanges` will happily log problems related to images, but the returned error by `d.UpdateManifest` is dropped. 

Right now i'm in a situation where i can see that flux is endlessly finding new images and wants to update the manifest, but the manifest is never updated. I expect this change to give me clues of whats happening.

### DIFF
--- a/pkg/daemon/images.go
+++ b/pkg/daemon/images.go
@@ -43,7 +43,11 @@ func (d *Daemon) pollForNewAutomatedWorkloadImages(logger log.Logger) {
 	changes := calculateChanges(logger, candidateWorkloads, workloads, imageRepos)
 
 	if len(changes.Changes) > 0 {
-		d.UpdateManifests(ctx, update.Spec{Type: update.Auto, Spec: changes})
+		_, err := d.UpdateManifests(ctx, update.Spec{Type: update.Auto, Spec: changes})
+		if err != nil {
+			logger.Log("error", errors.Wrap(err, "updating manifests"))
+			return
+		}
 	}
 }
 


yebyen:
This sounds like a good intentioned change to me, we don't want to swallow errors. But I've had endlessly updating loop before where the same image is tagged twice with the same timestamp (since it is the same image), and flux handles it well. There is no git commit loop and there is no flood in the logs. Is there any risk that the increased log volume will create a problem, where there is no problem now? I say this because of Flux and ArgoCD, I think Flux handles this edge case better.

It probably consumes some resources to be trapped in this update loop, even if Flux is not updating, but it seems wrong to just log the error (we should find out what's the problem and fix it, instead!) Maybe there can be a canonical ordering that can be applied to images with the same timestamp that prevents the update cycle... for that matter, maybe there already is.

I'm not sure if you think it's a bug that needs to be exposed, but I'm happy to work through with you and find out why this is causing you trouble! When there are multiple images having the same tag, I think that right now Flux v1 does the right thing. I did not code it though and am still new here, so it is possible there is something to address.

As for ArgoCD, I've found when my CI can tag the same image multiple times, Argo gets caught in an update loop (ArgoCD also does not persist the update to git as Flux does, so it is through observability of Argo that I became aware of this.) I find that when Argo hits this update loop, the K8s controller and the extra pod startup times together consume more than enough CPU cycles over time to put my T3.large instances over the CPU credit limit and start causing problems, if this is not noticed and addressed by an operator right away, then we have some downtime.

Neither here nor there, as I have not experienced this issue myself with Flux v1 itself.

yebyen:
I guess you probably compiled it on your own, @salzig can you report back if you think this change is still needed (or something else?)

kingdonb:
@salzig When you found this issue, was there already any indication of trouble in the error log?

I'm thinking this would make sense with a log level of debug? But in #2806 and #2851, I got the idea there is no such log level in flux at this point. I plan to ask the group whether it makes sense to merge this, I don't think it's a bad idea myself. But I don't know if it should be error or info level, or debug.

salzig:
@kingdonb as far as i understood, based on discussion in slack, is the source code considered fragile, beside that did this code change not result in the expected result.

PR #3341, opened 2020-10-23
Update Getting Help section

mewzherder:
Point to featured talks in the GOTK page as a way to learn more about Flux v2.


mewzherder:
Actually, realized that there needs to be more general pointing to the GOTK. Revoke and redo.

PR #3346, opened 2020-10-29
sets memcached namespace as required by flux deployment

joaovitor:
# sets memcached namespace as required by flux deployment

The deployment of helm chart requires the memcache service in the same namespace of flux deployment.
This small change sets the namespace in the service deployment.

----
https://github.com/fluxcd/flux/blob/master/chart/flux/templates/deployment.yaml#L208

```
- --memcached-hostname={{ .Values.memcached.hostnameOverride | default (printf "%s-memcached" (include "flux.fullname" .)) }}
```

### DIFF
--- a/chart/flux/templates/memcached.yaml
+++ b/chart/flux/templates/memcached.yaml
@@ -75,6 +75,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ template "flux.fullname" . }}-memcached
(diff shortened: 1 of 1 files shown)


squaremo (approved):
Makes sense to be explicit about the namespace (and maybe people are using `helm template` ...). Thank you ✋ <-- first time contributor high-five

PR #3355, opened 2020-11-05
fix completion

lalyos:
It seems that the completion command is not working.

Actually what happens that the generated shel completion script is only generated for the "completion" command not for the root "fluxctl" command.

## Reproduce the issue

See how the generated functions tries to create a new completion for the command  "completion"
```
$ fluxctl completion bash | tail -7

if [[ $(type -t compopt) = "builtin" ]]; then
    complete -o default -F __start_completion completion
else
    complete -o default -o nospace -F __start_completion completion
fi
```

## Fix

After applying this patch , see the newly generated completion function:
```
$ ./fluxctl completion bash | tail -7

if [[ $(type -t compopt) = "builtin" ]]; then
    complete -o default -F __start_fluxctl fluxctl
else
    complete -o default -o nospace -F __start_fluxctl fluxctl
fi
```

### DIFF
--- a/cmd/fluxctl/completion_cmd.go
+++ b/cmd/fluxctl/completion_cmd.go
@@ -49,13 +49,13 @@ func newCompletionCommand() *cobra.Command {
 }
 
 func runCompletionBash(out io.Writer, fluxctl *cobra.Command) error {
-	return fluxctl.GenBashCompletion(out)
+	return fluxctl.Root().GenBashCompletion(out)
 }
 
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Good catch 🐟

One small thing, we require commits to be signed-off before we can merge. The failed action will tell you how to do this: https://github.com/fluxcd/flux/runs/1359439198

lalyos:
@hiddeco DCO is fixed now ...

Issue #3366, opened 2020-11-19
Chart 1.5.0: Cannot use hostAliases and extraContainers concurrently.


alanjcastonguay:
**Describe the bug**

Installing Flux with helm chart 1.5.0 and leveraging both `hostAliases` and `extraContainers` produces an invalid pod spec. The yaml ordering & indentation around these two fields is wrong.

https://github.com/fluxcd/flux/blob/12c473fee8f131b1442bb159872bf242d5ec6d38/chart/flux/templates/deployment.yaml#L299-L308 

**To Reproduce**

Using this `bug-values-extra.yaml`:

```yaml
resources:
  limits:
    cpu: 1234m

hostAliases:
  - ip: "127.0.0.1"
    hostnames:
    - "foo.bar.local"

# Run https://github.com/fluxcd/flux-recv as a sidecar in the flux deployment
extraContainers:
  - name: extra-container-name
    image: fluxcd/flux-recv

dnsPolicy: ClusterFirst
```

Templates to this output.

```
helm template fluxcd/flux --version 1.5.0 --values bug-values-extra.yaml | grep -C 6 hostAliases
```

```yaml
          resources:
            limits:
              cpu: 1234m
            requests:
              cpu: 50m
              memory: 64Mi
      hostAliases:
        - hostnames:
          - foo.bar.local
          ip: 127.0.0.1
        - image: fluxcd/flux-recv
          name: extra-container-name
      dnsPolicy: ClusterFirst
```

**Expected behavior**

Everything from inside `extraContainers` lands inside the `spec.template.spec.containers[]` array, not inside the `spec.template.spec.hostAliases[]` array.

**Logs**

n/a

**Additional context**

- Flux version: chart 1.5.0
- Helm version: version.BuildInfo{Version:"v3.3.0-rc.1", GitCommit:"5c2dfaad847df2ac8f289d278186d048f446c70c", GitTreeState:"dirty", GoVersion:"go1.14.4"}
- Kubernetes version: n/a
- Git provider: n/a
- Container registry provider: n/a

alanjcastonguay:
This was introduced in https://github.com/fluxcd/flux/pull/2705

alanjcastonguay:
This is how https://github.com/instrumenta/kubeval/ complains without and with the above change & `bug-values-extra.yaml`

```
helm template fluxcd/flux --version 1.5.0 --values bug-values-extra.yaml | docker run -i garethr/kubeval -f - --strict
[...]

PR #3367, opened 2020-11-19
Enable helm chart usage of hostAliases and extraContainers concurrently

alanjcastonguay:
Bug #3366

Content of extraContainers is listed below/inside pod.spec.hostAlias[] array instead of in the pod.spec.containers[] array.

Bug created by #2705 but only manifests if both options are used at the same time.

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -298,12 +298,12 @@ spec:
           {{- end }}
           resources:
 {{ toYaml .Values.resources | indent 12 }}
+{{- if .Values.extraContainers }}
+{{ toYaml .Values.extraContainers | indent 8}}
+{{- end }}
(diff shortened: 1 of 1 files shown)


alanjcastonguay:
@squaremo we talked about this today

squaremo (approved):
Yep, I see what the problem is -- this fixes it neatly. Thanks @alanjcastonguay 🍍

squaremo:
@alanjcastonguay Can you add a DCO sign-off to the commit pls --

```
git commit --amend --signoff --no-edit
git push --force-with-lease
```

should do it, I believe. (We updated the DCO requirements when installing https://github.com/fluxcd/community/blob/main/GOVERNANCE.md, to be in line with CNCF expectations)

alanjcastonguay:
Amended commit message with DCO signoff & force-pushed.

PR #3371, opened 2020-11-23
snap: move to core20

dholbach:
Also add `curl` as build-dep.


squaremo (approved):
Seems reasonable :+1: Thanks Daniel :apple:

PR #3373, opened 2020-11-24
Upgrade json-patch to v4.9.0

stefansedich:
This PR upgrades json-patch to v4.9.0 to fix the issues we are experiencing after upgrading to cert-manager 1.x, we began to see lockups in the sync-loop attempting to perform image upgrades.

This thread sums up the issues experienced https://cloud-native.slack.com/archives/CLAJ40HV3/p1601469463086700 by both myself and another user when flux experiences image upgrades with the cert-manager CRDs being present.

After some digging I saw that Flux is using the version of json-patch (v4.5.0) before the fixes (v4.6.0) addressing the exponential complexity for nested objects https://github.com/kubernetes/kubernetes/issues/91615 which was causing issues with the cert-manager CRDs.

After upgrading json-patch to 4.9.0 and running a custom build of flux in our cluster everything is happy again and our upgrades are no longer blocking the sync loop. It would be nice if we could get the PR merged and a release pushed so that others do not experience this issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/bradfitz/gomemcache v0.0.0-20190329173943-551aad21a668
 	github.com/cheggaaa/pb/v3 v3.0.2
 	github.com/docker/distribution v2.7.1+incompatible
-	github.com/evanphx/json-patch v4.5.0+incompatible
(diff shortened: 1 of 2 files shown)


squaremo (approved):
Looks straight-forward and reasonable to me. (And passes tests)
Thank you @stefansedich 🍰

PR #3387, opened 2020-12-22
fix monitoring example in documentation

anasinnyk:
Hi,
we think you are mean `increase` here except `delta`

### DIFF
--- a/docs/references/monitoring.md
+++ b/docs/references/monitoring.md
@@ -20,7 +20,7 @@ The following metrics are exposed:
 
 Flux sync state can be obtained by using the following PromQL expressions:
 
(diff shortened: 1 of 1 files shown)


kingdonb:
Thank you for your contribution. Is this change correct? I think it makes sense to me, but it has sat waiting for review for some time and I'm not quite expert in the monitoring of Flux.

I'm updating the branch to satisfy the review checks. We can merge it if there's consensus that it's an error.

kingdonb:
This change regrettably still isn't clear to me. I'd love to see it demonstrated for example, I just haven't spent a lot of time monitoring Flux with Prometheus yet. Hopefully I will be able to get around to this soon.

kingdonb:
I'm going to close this because it has gone stale, but with the understanding that it can be reopened if the issues are addressed.

PR #3394, opened 2021-01-06
Change location of Helm repo "stable"

### LINKED ISSUE #3392: CI fails with charts not found
```
# Adding stable repo with URL: https://kubernetes-charts.storage.googleapis.com 
# Error: error initializing: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Failed to fetch https://kubernetes-charts.storage.googleapis.com/index.yaml : 403 Forbidden
# Error: could not find tiller
```

squaremo:
The stable chart repo is being deprecated, but it's also moved to
https://charts.helm.sh/stable/, and the old location (built into the
Helm client) has been removed.

Fixes #3392.


squaremo:
If this passes the e2e tests, it's a win.

stefanprodan (approved):

PR #3397, opened 2021-01-11
Tighten pod security policy and allow custom role rules

langecode:
Thank you very much for a great product. I know the current Flux version is in maintenance mode and it may not make sense to do this change. However, as we have done it I thought we should give you the opportunity to include it upstream should you so decide.

The PR tightens the pod security policy for Flux a bit and also adds security context to the pod specifications to make sure the most correct pod security policy is chosen (see [Policy Order](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order)). Furthermore the PR will make the set of rules configured for the namespace or cluster wide role configurable making it possible to tighten the permissions applied to Flux. The default is complete access as it is in the current chart version.

Note the cluster wide objects has been prefixed with the namespace as Helm v3 works with release names as "namespaced" hence if Flux were to be deployed into multiple namespaces using the same release name there could be a clash in the naming of non-namespaced objects.


kingdonb:
I'm not sure it can be merged as-is, there are also some indentation regressions I'd like to fix in here. Thank you for contributing this, I will bring it to the Flux developers meeting and see how the maintainers feel about adding a configuration for pod security policy to Flux v1

In general the disposition towards changes in Flux v1 at this point is, there must be no further breaking changes as Flux v1 is in maintenance mode. After looking this over, I can't say whether tightening the policy might represent a breaking change, but it is a change in behavior and maybe should be undertaken separately from the change that simply parameterizes the PSP.

What do you think about splitting this PR in two parts @langecode ? One part which parameterizes the PSP, and another that

Related to #3409 
This PR makes a more aggressive change, does not assume only hostNetwork is what you want to enable. This PR also has more knobs, I think I prefer the approach where in values.yaml, only one utterance of `hostNetwork: true` is needed to reconfigure both the PSP and the pod spec together, and there is no need at configuration time to ensure the configs are consistent.

I will check Flux v2 project to solve this problem for the controllers, where they may need configuration such as hostNetwork rules. It may be good to adopt that strategy if it can be re-implanted here in a non-breaking way.

kingdonb:
Thank you for your contribution. Flux v1 is in maintenance mode and is no longer adding new features unless they are critical. Also my understanding is that `podSecurityPolicy` is deprecated in K8s 1.21 due out soon, and will be unavailable some time later. It seems based on that information that the time is better spent elsewhere.

langecode:
No problem closing this. We have our own fork implementing the change and
have been deploying based on the fork instead.  But we are moving to
flux2/gotk so it is not a big issue for us anymore.

Issue #3382, opened 2020-12-07
Version mismatch between flux version and update-checker


BitProcessor:
**Describe the bug**

Version mismatch between Flux version and update-checker version

```
ts=2020-12-07T14:41:01.117494825Z caller=main.go:259 version=1.21.0
...
ts=2020-12-07T14:41:02.408813856Z caller=checkpoint.go:24 component=checkpoint msg="up to date" latest=1.20.1
```

**To Reproduce**
Install flux 1.21.0

kingdonb:
I have an older version of Flux running somewhere and I can confirm, the update checker is outputting the wrong version to upgrade to in Flux v1.x

```
ts=2021-01-18T21:32:29.06877884Z caller=main.go:259 version=1.19.0
...
ts=2021-01-20T22:17:40.356923655Z caller=checkpoint.go:21 component=checkpoint msg="update available" latest=1.20.1 URL=https://github.com/fluxcd/flux/releases/tag/1.20.1
```

kingdonb:
I suspect this is related to #3400 and it will resolve when that one is merged

kingdonb:
I found the answer, and am working on the resolution for this today.

kingdonb:
This is resolved now:

```
ts=2021-01-22T23:50:14.724327112Z caller=checkpoint.go:21 component=checkpoint msg="update available" latest=1.21.1 URL=https://github.com/fluxcd/flux/releases/tag/1.21.1
```

kingdonb:
@BitProcessor can you see if the update checkpoint code is returning a sensible release now; if so please go ahead and close this. Thanks for bringing this to our attention! I've also checked Helm Operator to ensure the latest release is being reported.

BitProcessor:
```
ts=2021-01-26T19:08:18.924082067Z caller=main.go:259 version=1.21.1
...
 ts=2021-01-26T19:08:19.309232596Z caller=checkpoint.go:24 component=checkpoint msg="up to date" latest=1.21.1
```
Looking good 👍

PR #3400, opened 2021-01-18
Housekeeping: merge release branch 1.21.x into master

squaremo:


kingdonb:
I think this should be merged, as v1.21.1 is already published and circulating... there are some dependency CVEs fixed by the upgrade from 18 to 20 base image, by #3371. The update checker API is still reporting v1.21.0 as the latest version (#3382).

(Is there anything else needed, before this merge?)

hiddeco (approved):

Issue #3384, opened 2020-12-09
Flux is not able to get the EC2Metadata to detect the AWS region when switching the EKS EC2-instances Metadata-Service-Version from IMDSv1 to IMDSv2.


raviranjithkumar:
**Describe the bug**

Flux is not able to get the EC2Metadata to detect the AWS region when switching the EKS EC2-instances Metadata-Service-Version from IMDSv1 to IMDSv2. Due to this the deployments are getting impacted.

Switching from IMDSv1 to IMDSv2 is necessary because of the Security issues that we have with [IMDSv1](https://medium.com/@shurmajee/aws-enhances-metadata-service-security-with-imdsv2-b5d4b238454b)

In our Org, we use flux in production and there is a security concern to update to IMDSv2 but the current Flux is not supporting it so this fix is critical for us. We need the community help to address this issue.

I see the flux code is using ec2metadata: https://github.com/fluxcd/flux/blob/2f8e2075784fdc217c4359a5f25286d46230eda8/pkg/registry/aws.go#L130

*About IMDSv2: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

**To Reproduce**

Steps to reproduce the behaviour:

1. Disable IMDSv1 and start using IMDSv2 in EC2 instances of the EKS cluster

**Expected behavior**

Flux should be able to get the AWS region information and do the deployments as expected

**Logs**

`{"caller":"aws.go:137","component":"aws","err":"EC2MetadataRequestError: failed to get EC2 instance identity document\ncaused by: EC2MetadataError: failed to make EC2Metadata request\n\tstatus code: 401, request id: \ncaused by: ","error":"fetching region for AWS","ts":"2020-12-09T12:37:34.324397636Z"}
`
**Additional context**

- Flux version: 1.20.2
- Kubernetes version: v1.15.11-eks-065dce
- Git provider: Github
- Container registry provider: AWS-ECR

relu:
I suspect this might have something to do with the outdated version of the go aws-sdk [flux depends on (v1.27.1)](https://github.com/fluxcd/flux/blob/v1.20.2/go.mod#L34). I wasn't able to pinpoint in its changelog exactly when or if this particular issue was resolved, but I did find some references to potentially related fixes introduced in more recent versions:
- https://github.com/aws/aws-sdk-go/blob/master/CHANGELOG.md#sdk-bugs-18
[...]

PR #3401, opened 2021-01-18
Update AWS SDK and SOPS

squaremo:
The AWS SDK in use is pretty old; a newer version may work better with IMDSv2 and thereby help with #3384. However, there is reason to stick to the version SOPS uses; so, update SOPS to its newest patch release, then update the AWS SDK to the (much newer) version that SOPS depends on.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ replace github.com/fluxcd/flux/pkg/install => ./pkg/install
 require (
 	github.com/Jeffail/gabs v1.4.0
 	github.com/Masterminds/semver/v3 v3.0.3
-	github.com/aws/aws-sdk-go v1.36.28
+	github.com/aws/aws-sdk-go v1.33.18
 	github.com/bradfitz/gomemcache v0.0.0-20190329173943-551aad21a668
 	github.com/cheggaaa/pb/v3 v3.0.2
 	github.com/docker/distribution v2.7.1+incompatible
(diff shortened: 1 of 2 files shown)


stefanprodan:
Hopefully this will not break SOPS when used with AWS KMS, they are using `github.com/aws/aws-sdk-go v1.33.18`.

squaremo:
> Hopefully this will not break SOPS when used with AWS KMS, they are using `github.com/aws/aws-sdk-go v1.33.18`.

We can reel it back to that version if necessary -- it's still somewhat newer than 1.27.

stefanprodan:
@squaremo I would update SOPS to v3.6.1 and keep the AWS SDK on the same version as SOPS.

stefanprodan (approved):
LGTM

I would change the PR name to reflect sops version bump.

yiyan-wish:
@stefanprodan The newest release does not have this commit, help to release a new version

kingdonb:
@yiyan-wish I will help with this, thanks for bringing attention to it

kingdonb:
#3421 is open now which will release this update in a new binary when approved, merged, and tagged, it should be released as 1.21.2 👍

bquantump:
I don't believe this was ever fixed. Seems the sops version is still 3.6.0, no?
https://github.com/fluxcd/flux/blob/master/docker/sops.version

yebyen:
This PR was merged, but it did not update that file @bquantump - can you submit another PR with the change still needed?

Will be happy to get it merged and get a patch version released soon, as long as we can be reasonably sure it doesn't break anything.

kingdonb:
It looks like this was the expected upgrade, I'll get it released.

PR #3406, opened 2021-01-25
Rebased #3296

kingdonb:
This PR is just #3296, rebased for re-testing

(Closes: #3296)


kingdonb:
Because of my write access to the repo, we can just use #3296 now. Closing this one instead.

PR #3409, opened 2021-01-27
support hostnetwork for flux pod

yiyan-wish:
Support hostnetwork for flux pod

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -306,7 +306,7 @@ spec:
 {{ toYaml .Values.hostAliases |indent 8 }}
 {{- end }}
 {{- if .Values.hostNetwork }}
-      dnsPolicy: {{ .Values.hostNetwork }}
+      hostNetwork: {{ .Values.hostNetwork }}
 {{- end }}
 {{- if .Values.dnsPolicy }}
       dnsPolicy: {{ .Values.dnsPolicy }}
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -305,6 +305,9 @@ spec:
       hostAliases:
 {{ toYaml .Values.hostAliases |indent 8 }}
 {{- end }}
+{{- if .Values.hostNetwork }}
+      dnsPolicy: {{ .Values.hostNetwork }}
+{{- end }}
 {{- if .Values.dnsPolicy }}
       dnsPolicy: {{ .Values.dnsPolicy }}
 {{- end }}
--- a/chart/flux/values.yaml
+++ b/chart/flux/values.yaml
@@ -77,6 +77,8 @@ extraVolumes: []
 #     - name: ndots
 #       value: "1"
 
+hostNetwork: false
+
 gpgKeys:
   # These keys will be imported into GPG in the Flux container.
   secretName: ""


kingdonb:
Thank you for submitting, can you set `.Values.dnsPolicy` to hostNetwork instead? I think this may be a typo.

I'm not sure if this patch will work as written. Flux v2 is also very soon intended to replace this project, please find https://fluxcd.io/community/ for more information

If something you want that is not handled currently, it will be better if possible for your use case to be handled there.

Please give Flux v2 a try, or propose it in the Discussion page for Flux v2 before submitting as a PR. New feature requests are unlikely to be included in Flux v1 unless they are compelling or meet an important need without introducing breaking changes.

yiyan-wish:
@kingdonb 
Pods with cluster network on eks are restricted to access the aws service.
And pod with hostnetwork can be authenticated by the aws.

kingdonb:
Ah, thank you that helps clarifies the use of this patch. It does not look like a breaking change, since hostNetwork is false by default and you've preserved that behavior.

I will make a note about it and include it in my review list, if it is not breaking then perhaps we can include it in a new release. There is a short list of not breaking changes that people have asked for, I think there is a chance can still be merged and make into a new Flux v1 release.

Since this is a chart values change, but appears to be non-breaking I don't see any reason why it wouldn't be possible to include it. There may be a separate release for non-breaking changes that require a new chart release. Please squash the commits into one if you can, I will see if I can get it included.

kingdonb:
This will conflict with the podsecuritypolicy implemented by #2225, https://github.com/fluxcd/flux/pull/2225/files#diff-4e7a89b4334804eb6d1632eb29220f116895532df3ab3743209ad32875325053R16

That `rbac.pspEnabled` is an optional configuration which is also disabled by default. I'm not sure that would mean we can't merge it, but it seems good to note this. You might also patch the psp with `hostNetwork: true` when enabling hostNetwork, so it will not break when Flux also has psp enabled.

kingdonb:
@yiyan-wish Do you want to take a shot at correcting the issue [mentioned earlier](https://github.com/fluxcd/flux/pull/3409#issuecomment-768679143)

I will prefer not to merge a change that introduces a new pair of options that can be unintentionally set in conflict with each other, but it should be easy to solve.

kingdonb:
I'm afraid this change will not be included in the 1.21.2 release without some updates, I was hoping I could recommend the Helm post-rendering feature added in v3.1.0 to you, which can be used to make small changes in helm charts with a post-renderer like Kustomize, but I don't know if you're using Helm Operator to manage Flux or not, and this feature is not in v1.

The fluxcd/helm-operator#365 issue that asked to add support for this was belayed for Flux v2. There, support has been added now and is forthcoming in the next release of Helm Controller - see fluxcd/helm-controller#202 and fluxcd/helm-controller#104.

If this feature is important to you, I can help make the necessary changes for a next release, say Flux 1.21.3 that might be another week or two down the road. Some testing will be needed to verify the pod security policy settings won't be in conflict.

I don't think it's major changes needed to merge this pull request, I am trying to make sure that everyone who has minor non-breaking changes that are important to them can get them included while Flux v1 remains in maintenance mode; it will help if you can tell me if this is important. A new release is coming imminently, so that all docs landing pages are updated with the appropriate verbiage about Flux v1 in maintenance mode and admonishment recommending to upgrade to the next version.

If you have by any chance already moved onto Flux v2, then there may be no need for this anymore and we can close the issue. Let me know please!

yiyan-wish:
@kingdonb Thanks, and support hostnetwork is important for me, in our aws eks environment, only support imdsv2, and the response hop is limited to 1. So only using the hostnetwork, the flux pod can access the aws service.

kingdonb:
@yiyan-wish do you think that #3384 characterizes the same issue that is approached and resolved by this PR?

We can prioritize it for inclusion in the next Flux v1 release, but that might still take some time. I want to be able to recommend a course of action to the reporter in #3384, if this change is all it needs they might be able to test and provide a second or third party verification for us. That will help me show stronger case this is a good candidate for the next release.

kingdonb:
Other users have reported the issue in #3384 should be resolved by upgrading KIAM to 4.0 and adding this setting to your KIAM config. I don't know if you are using KIAM or not, but it seems that enabling `hostNetwork` is not a way that we should promote, since it goes around rather than thru authorization protocols that have been set up to enforce IAM in pods.

```
# Ref: https://github.com/uswitch/kiam/issues/359
      allowRouteRegexp: "/latest/*"
```

What do you think @yiyan-wish, are you using KIAM or similar and does this help at all?

kingdonb:
Greetings,

We decided since this is not a recommended configuration we want to promote to users, to avoid giving folks the wrong idea that they should use host networking and it would be a supported configuration, that this change cannot be merged as-is.

https://helm.sh/docs/topics/advanced/#post-rendering - if you want to support this in your cluster but are having trouble because the Helm chart doesn't have this option, you can add it in a way that is safe and will be repeatable on future versions of Flux without forking, by using a post-renderer; the link provided shows how to use eg. kustomize for post-rendering.

Thanks again for your contribution, and if you are still struggling with this issue please feel free to visit #flux on the CNCF slack to ask for help. We hope that you can upgrade to Flux v2 and you find that your use cases and needs are all met well through the next version of Flux.

Issue #2143, opened 2019-06-10
Fix critical/high CVEs


willholley:
The `flux:1.12.3` image as published on Docker Hub is currently reporting vulnerable to the following CVEs:

Vulnerability ID|   Severity|   Affected Packages|   How to Resolve
--|-|-|-
CVE-2019-5435|      Low|          libcurl|             Upgrade libcurl to >= 7.64.0-r2
CVE-2019-5436|      Low|          libcurl|             Upgrade libcurl to >= 7.64.0-r2
CVE-2019-5018|      High|          sqlite-libs|         Upgrade sqlite-libs to >= 3.28.0-r0
CVE-2019-8457|      Critical|          sqlite-libs|         Upgrade sqlite-libs to >= 3.28.0-r0

All issues have been addressed by the upstream packages that introduce them to the container and building the Dockerfile from source resolves the problem.

This is mostly a note to inform the changelog for the next release (assuming a patch release would be created rather than republishing under the existing tag).

squaremo:
Thanks for bringing this up @willholley. Maybe we should have a `vulnerabilities addressed` section in the changelog -- something to consider.

squaremo:
@willholley What did you use to do the scan?

squaremo:
(cleared the milestone so that I can close it; but we can add this info to the changelog in retrospect)

willholley:
@squaremo this was using the IBM Cloud Container Registry, which automatically applies a vulnerability scanner - https://cloud.ibm.com/docs/services/Registry?topic=va-va_index). The latest image is currently reporting no issues:

```
Checking security issues for 'flux:1.13.0'...

Image 'flux:1.13.0' was last scanned on Thu Jun 13 12:02:39 UTC 2019
The scan results show that NO ISSUES were found for the image.

OK
```

hiddeco:
@willholley thanks for the info!

We had a scan running when we were still using Quay.io (with a notification to the maintainers), but as we had to move away from Quay.io, this is no longer in place :disappointed:

I will put some thought into an alternative solution.

nabadger:
@hiddeco if it's of any use, the clair-scanner project is pretty useful for this 

This project builds a daily DB of vulns, so you can do it all locally via CI.

- https://github.com/arminc/clair-local-scan

ChipWolf:
@awh could the priority be bumped on this please?

squaremo:
[...]

PR #3416, opened 2021-02-04
Release chart 1.6.1

kingdonb:
Includes a bump from Flux Daemon v1.21.0 to v1.21.1 with release notes here:

https://github.com/fluxcd/flux/blob/master/CHANGELOG.md#1211-2021-01-06

Notable changes include update from snap core18 to core20, which includes
updates that silence several CVE warnings (#2143), and a chart fix setting the
memcached service's namespace from Release.Namespace that was omitted before.

(The snap core update as I understand it has no effect on what is deployed by Helm, the snap base is for use with fluxctl.)

Signed-off-by: Kingdon Barrett <kingdon@weave.works>

### DIFF
--- a/chart/flux/CHANGELOG.md
+++ b/chart/flux/CHANGELOG.md
@@ -1,3 +1,15 @@
+## 1.6.1 (2021-02-04)
+
+### Improvements
+
+ - Updated Flux to `1.21.1`
+   [fluxcd/flux#3393](https://github.com/fluxcd/flux/pull/3393)
+
+### Fixes
+
+ - Sets namespace on memcached service
+   [fluxcd/flux#3346](https://github.com/fluxcd/flux/pull/3346)
+
 ## 1.6.0 (2020-11-20)
 
 ### Improvements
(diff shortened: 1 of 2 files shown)


chart/flux/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 1.6.1 (2021-02-04)
+
+### Improvements
+
+ - Updated Flux to `1.21.1`
+   [fluxcd/flux#3393](https://github.com/fluxcd/flux/pull/3393)
+ - snap: move to core20 [fluxcd/flux#3371](https://github.com/fluxcd/flux/pull/3371)
```
stefanprodan, line 7: We track here the Helm chart changes not snap. Please remove all changes that have no affect on the Helm chart.

stefanprodan (changes requested):

chart/flux/CHANGELOG.md
kingdonb: OK. fixed

stefanprodan (approved):
LGTM

Thanks @kingdonb

Issue #3422, opened 2021-02-10
create a giant warning banner on flux v1 docs


alisondy:
Suggestion

place a giant warning banner on  the top of old flux docs, similar to what kubernetes does see this image

![Screen Shot 2021-02-10 at 14 01 19](https://user-images.githubusercontent.com/19580905/107519783-909fc680-6ba8-11eb-8805-98aa26858166.png)

New users may accidentally follow tutorials on the wrong docs, getting halfway through a tutorial and then realizing it's for a different version of flux.

---

I was browsing this page

https://docs.fluxcd.io/en/1.21.1/tutorials/get-started/#set-up-flux

After doing a google, for a flux guide, and i thought initially that it was the flux v2 docs,

But it wasn't until I looked at the flux-get-started repo that I was looking at docs for an old version of flux 🤦‍♀️

dholbach:
https://squidfunk.github.io/mkdocs-material/setup/setting-up-the-header/#announcement-bar maybe?

dholbach:
If we enable this globally, we can maybe revert https://github.com/fluxcd/flux/pull/3419?

PR #3423, opened 2021-02-10
use announce section in material theme

dholbach:
Fixes: #3422

cc @alisondy @kingdonb @hiddeco 

Screenshot from 2021-02-10 15-33-35


docs_theme_override/main.html
```diff
@@ -0,0 +1,22 @@
...
+    approach to GitOps, implements some long-requested features of v1 and is generally
+    more flexible. It will be GA soon.
+</p>
+
+<p>
+    We strongly recommend you familiarise yourself with Flux v2 and start looking at
+    <a href="https://github.com/fluxcd/flux2/discussions/413">your migration path</a>.
+</p>
+
+{% endblock %}
```
hiddeco, line 22:
```suggestion
{% endblock %}

```

dholbach:
The thing is, @alisondy pointed out that this is just going to be reflected in our last `/stable/` release. I checked if there's a global readthedocs way to resolve this, but I couldn't find it. This might be the only way to do it. The most global it will easily get. :shrug:

hiddeco:
@dholbach I think we may be able to enable the feature that warns people if they are not looking at the latest version of the documentation (with a likewise banner), that would guide them towards the right message.

alisondy:
https://docs.readthedocs.io/en/stable/versions.html#version-warning ? 👀

dholbach:
> https://docs.readthedocs.io/en/stable/versions.html#version-warning ? eyes

Looks like it's only available for Sphinx projects (we're using `mkdocs`).

hiddeco:
Then there is little we can do I am afraid :-(

alisondy:
https://github.com/shopsys/readthedocs-version-warning-mkdocs-plugin
maybe ?
👀

dholbach:
The way I understand https://github.com/shopsys/readthedocs-version-warning-mkdocs-plugin/blob/master/mkdocs_version_warning/__init__.py#L41-L50 is that the banner text is fixed?

But thanks for looking. Maybe we'll find another plugin?

dholbach:
https://docs.readthedocs.io/en/stable/faq.html#how-can-i-avoid-search-results-having-a-deprecated-version-of-my-docs might also be helpful.

dholbach:
Maybe we land this change first, as it will at least resolve things for ALL pages of `/stable/` and `/latest/` and next we worry about old releases?

Is the look of it too disruptive?

hiddeco (approved):

PR #3436, opened 2021-02-26
go mod update

kingdonb:
Generally updating dependencies to the latest possible patch versions. Testing for probable inclusion in 1.22.0 release.


squaremo (approved):
Seems fine :+1: There's a line for k8s.io/apimachinery 0.19.8 and a replacement for k8s.io/apimachinery 0.17.17, which is confusing -- but only because they are both changed here.

Issue #2070, opened 2019-05-22
Revert ECR-outside-AWS default behaviour


squaremo:
Before #1863, if you were using ECR from _outside_ AWS, you would see a warning in the log (from the region detection), then it would proceed as for regular images -- that is, attempt to get credentials from imagePullSecrets etc.

After #1863, if fluxd cannot detect an AWS region _and is not told one explicitly_, it will ignore all images from ECR.

This change in behaviour comes from the separation of region detection and image inclusion/exclusion (filtering). Before, if a region wasn't detected, no image filtering took place. After, image filtering is done whether or not a region is detected or supplied. This gives rise to the question of what to do if a region is neither detected nor supplied -- and the somewhat arbitrary decision was to treat that as excluding all regions, rather than _including_ all regions. 

Thus: the surprising situation in which a working cluster stopped scanning ECR images, when upgraded from flux 1.10 to 1.12, without any config changes.

gtseres:
Hello @squaremo! We have been using Flux for quite some time and I have recently started learning Go, so I thought that I could give this a shot if no concern! 

Reading through the issue details, I understand that the desired behaviour would be to populate the regions configuration [here](https://github.com/fluxcd/flux/blob/d4f2f58aeba5c0c2c94d524ad1e0eb65db8d09f0/pkg/registry/aws.go#L122) with a value. Would that value be all the available ECR regions? If that is the case, would using the [SSM service](https://docs.aws.amazon.com/sdk-for-go/api/service/ssm/) to get all the available ECR regions make sense? For example, when using the `aws` cli, the output is the following:
```
$ aws ssm get-parameters-by-path   --path /aws/service/global-infrastructure/services/ecr/regions --output json |   jq .Parameters[].Value
"ap-east-1"
"ap-northeast-1"
"ap-south-1"
"ap-southeast-1"
"ap-southeast-2"
"cn-north-1"
"eu-west-2"
"eu-west-3"
"me-south-1"
"sa-east-1"
"ca-central-1"
"cn-northwest-1"
"eu-central-1"
"eu-west-1"
"us-east-1"
"us-east-2"
"us-gov-east-1"
"us-gov-west-1"
"us-west-1"
"us-west-2"
"ap-northeast-2"
"eu-north-1"
```

Is this idea along the lines with what you have in mind for this or is it in the wrong direction?

If this is in the right direction, I could implement a solution that involves the `github.com/aws/aws-sdk-go/service/ssm` Go package to get the available regions.

[...]

PR #3438, opened 2021-02-28
Improve ecr outside aws

lexysoda:
This solves https://github.com/fluxcd/flux/issues/2070

We currently ignore all ECR images if regions aren't set, and we are outside of an AWS cluster. (And do this silently and rather undocumented!)

The reasoning for this was cost saving: https://github.com/fluxcd/flux/pull/1619/commits/8ba016f849d49bbce2829e89edb5b641dfa117a2 which can be dismissed if we are outside of AWS anyway.

### DIFF
--- a/pkg/registry/aws.go
+++ b/pkg/registry/aws.go
@@ -134,9 +134,6 @@ func ImageCredsWithAWSAuth(lookup func() ImageCreds, logger log.Logger, config A
 			metadataRegion, err := ec2.Region()
 			if err != nil {
 				preflightErr = err
-				if config.Regions == nil {
-					config.Regions = []string{}
-				}
 				logger.Log("error", "fetching region for AWS", "err", err)
 				return
 			}
--- a/pkg/registry/aws.go
+++ b/pkg/registry/aws.go
@@ -45,7 +45,10 @@ type AWSRegistryConfig struct {
 	BlockIDs   []string
 }
 
-func contains(strs []string, str string) bool {
+func containsOrNil(strs []string, str string) bool {
+	if strs == nil {
+		return true
+	}
 	for _, s := range strs {
 		if s == str {
 			return true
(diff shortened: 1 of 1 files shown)


kingdonb:
Thank you for the contribution! Can you tell me if this changes a default behavior?

I see this is to address an existing issue report which you have referenced, it will also help for the merge-ability of this PR if you can help me understand this change in terms of what mis-behavior it fixes. The reason I ask this is because the original report was marked as an enhancement; if this is a bug fix, then it can be considered for merge without too many questions.

Currently Flux is in maintenance mode, as I suspect you already know, so no breaking changes can be merged. If this changes a default behavior without asking the user, that will be something I probably cannot merge, since nobody is expecting behavior changes inside of the maintenance period. However if the change is opt-in (such that the only way to activate the behavior is a user-configured parameter) then it could more easily be considered.

I have given this a cursory review only at this point and it looks like you've done about what @squaremo suggested. I will come back and consider this for inclusion in the next release! Thanks again.

kingdonb:
If you can explain this change in a one-liner, something in a format that I can include as a description of what was merged in the `CHANGELOG`, then that will also help. I can appreciate that you put more information into the commit messages, and split the change into digestible parts (I am new here and I apologize if I'm a bit slow on the uptake, but I am still learning my way through the codebase.)

kingdonb:
I am generally reluctant to merge this because if we're wrong about this, it sounds like it has the potential to cost people money.

One thing I'd like to understand: if this behavior is designed to operate the way it does on purpose (documented or undocumented), eg. it is meant to prevent inter-region traffic because it costs money, wouldn't a machine with no region, being outside of the AWS network altogether, mean that now the permitted traffic is egress (which also costs money?)

lexysoda:
Hey there, thanks for your comments. Let me summarize:
Yes, this changes default behavior. But the change would only change running systems for a very niche subset of users, for which the following is true:
1. Run a cluster outside AWS
2. Have workloads with ECR images attached
3. Have not set the `--registry-ecr-region` flag
and therefore have all ECR image-scanning disabled by default.  For these users the ECR images would suddenly all get scanned. (But I do have the feeling I'm the only user stupid enough to use ECR outside of AWS :facepalm: )

Anyway, I do understand if you don't want any changes to default behavior. The thing is, from just reading the docs, I expected this to be the default behavior. So in that case I would propose the following:
1. Keep the current behavior but add a note to the `--registry-ecr-region` flag explanation what happens when running outside AWS.
2. Add verbose logging that prints out all images that are not being scanned because no region is set. (I had no clue why my images weren't updated by flux as there is 0 log output and I had to read the source code to find out.)

What do you say?

kingdonb:
I'm going to close this because it has gone stale, but with

PR #3439, opened 2021-03-02
Update the discovery cache to apiextensions v1

stefanprodan:
This PR changes the client version used to query the Kubernetes discovery API, instead of `apiextensions.k8s.io/v1beta1` Flux will be using the `apiextensions.k8s.io/v1` client, making it compatible with future versions of Kubernetes. 

Changes:
- Update Kubernetes packages to `v0.17.17`
- Update the CRD API version from `v1beta1` to `v1` (`v1beta1 `will be removed in Kubernetes v1.22)

**Warning**: After this change, Flux will no longer work on Kubernetes **<1.16.0**


kingdonb:
Thank you! Since we had planned to upgrade client-go in Flux v1.22.0 in facilitation of this requirement for K8s v1.22.0, should I plan on including this change in the upcoming release?

stefanprodan:
client-go can't be upgraded as will conflict with this: https://github.com/fluxcd/flux/blob/master/go.mod#L40 You can update kubectl omly.

squaremo (approved):
Changes look nice and straight-forward, and it doesn't even need to change the minor version of client-go. Well pin-pointed Stefan 🥭 
If this passes e2e testing I think it's good to go. I'll leave it to @kingdonb to decide where to plug it into the release schedule.

kingdonb:
Because this represents a breaking change for old Kubernetes versions (in the name of keeping forwards compatibility, which cannot be achieved any other way), we will keep this PR open and block merging it until Kubernetes 1.22-beta.1 is released.

kingdonb:
Kubernetes 1.22-beta.1 is expected to be released on Thu July 8 according to https://www.kubernetes.dev/resources/release/

hiddeco:
The release of the changes in this PR should be coordinated with the changes in https://github.com/fluxcd/helm-operator/pull/599.

kingdonb:
I've tested fluxcd/helm-operator#599 and added to it with fluxcd/helm-operator#618. Helm Operator with K8s v1.22.x looks good. I'd like to cut a pre-release very soon, and we can proceed rapidly to final release if everything is in order.

I am testing this together with the modernized Helm Operator the rest of today. It should be ready to cut at least pre-release of both projects on Wednesday, (maybe even sooner if there are no additional blockers that are discovered.)

kingdonb:
[Flux v1.23.0-rc1](https://github.com/fluxcd/flux/releases/tag/1.23.0-rc1) is building/publishing now that includes these changes. I will rebase so this PR merges out when the release is finalized. 👍

kingdonb:
Helm operator next release (1.3.0 and 1.4.0) are ready in fluxcd/helm-operator#622 and fluxcd/helm-operator#623

I expect they will probably both drop on Monday, unless we find some issue. Flux v1.23.0 is out.

PR #3452, opened 2021-03-26
Remove deprecated GH workflows

### LINKED ISSUE #3451: Remove deprecated GH workflows
We need to delete the ci-cleanup and cve-scan as they are failing. We could replace trivy with snyk in the future.

kingdonb:
Resolves #3451

Signed-off-by: Kingdon Barrett <kingdon@weave.works>


kingdonb:
I will investigate adding snyk, this is filed in response to #3451 

There is no urgency to merge this IMHO, if we can add a Snyk scan to this PR and replace what is being removed at the same time. I will find time for this on Monday.

If it comes to a point where this is blocking PRs from merging soon, we can go ahead and merge without the replacement, but I don't think this will take long to resolve.

stefanprodan:
We can't add Snyk since the CNCF key is for flux2 repo only,  we've opened a ticket. I would merge this as it.

stefanprodan (approved):
LGTM

Thanks @kingdonb

Issue #3275, opened 2020-09-02
Upgrade memcached:1.5.20 image into memcached:1.6.6 or latest 


PKalhara:
**Describe the feature**

Currently flux repository using a little bit older  **_memcached_** image  which is version **_1.5.20_** in following places
[memcache-dep.yaml](https://github.com/fluxcd/flux/blob/8609b01765984664cad3615801efa550417442ff/deploy/memcache-dep.yaml#L23) and [values.yaml](https://github.com/fluxcd/flux/blob/8609b01765984664cad3615801efa550417442ff/chart/flux/values.yaml#L206). We need to upgrade this memcached image in to the latest available tag in the docker hub.  

**What would the new user story look like?**

This memcached image upgrade will fix the vulnerabilities in the existing image and essentially this will help us to pass the twistlock vulnerability scan as we are blocked in this place due to the high number of security vulnerabilities in the exsting docker image tag **_memcached_:_1.5.20_**

**Expected behavior**

Please update [memcache-dep.yaml](https://github.com/fluxcd/flux/blob/8609b01765984664cad3615801efa550417442ff/deploy/memcache-dep.yaml#L23) and [values.yaml](https://github.com/fluxcd/flux/blob/8609b01765984664cad3615801efa550417442ff/chart/flux/values.yaml#L206) with the latest available docker image tag in [memcached:1.6.6](https://hub.docker.com/layers/memcached/library/memcached/1.6.6/images/sha256-8f521bdbfb7f948840c4aaeebd1d0b51bce37f9aae033bbf0417c2ccb40df5ad?context=explore)

kingdonb:
@PKalhara Have you been able to test this configuration and seems to work fine?

This can be a priority for the next Flux v1 milestone, if you want to submit it as a pull request that will help expedite it and you will get all the credit for having resolved this issue in the glorious release notes.

Apologies about the length of time that has elapsed since your inquiry.

kingdonb:
I will try to make sure this gets addressed in 1.22.2, sorry for the delay.

kingdonb:
Actually we can still address it with #3456 today, taking a look at this now 👍

PR #3455, opened 2021-03-31
Release 1.22.1

### LINKED ISSUE #3453: Bump to gnutls 3.6.15-r1 or later in docker build and base image
**Describe the bug**
There is a security vulnerability in the `gnutls` library that has been resolved in the 3.6.15-r1 release.  `gnutls` is used in the alpine base image as well as the [flux dockerfile](https://github.com/fluxcd/flux/blob/21debdeec3012371937348b12434543b196925aa/docker/Dockerfile.flux#L5).  The issue was reported to alpine [here](https://gitlab.alpinelinux.org/alpine/aports/-/issues/12543) and was resolved [before the 3.12.5 release](https://gitlab.alpinelinux.org/alpine/aports/-/commits/3.12-stable) was created.

**Expected behavior**

We should be using reducing security vulnerabilities as much as possible, including bumping to newer versions of packages and newer versions of base images when possible.  We should bump the apline image constraint in the dockerfile to 3.12.5 as well as the version constraint for `gnutls` to `>3.6.15-r1`.

**Additional context**

- Flux version: 1.22.0
- Kubernetes version:
- Git provider:
- Container registry provider:

kingdonb:
Simply fixes CI with #3452, and then building again which rolls the alpine stack.

As of this morning this is enough to mitigate a CVE in #3453 which came from an end-user report. (Fixes #3453)


stefanprodan:
@kingdonb the CI changes are not relevant to users, I would mention the CVE fix only.

chart/flux/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "1.21.2"
-version: 1.6.2
+appVersion: "1.22.0"
```
hiddeco, line 2: Shouldn't this be `1.22.1`, or is this the result of some other dance between release branches?

yebyen:
You are both correct, thanks for catching these issues. I will amend and re-request the review.

chart/flux/Chart.yaml
kingdonb: That's right, the update in this branch is housekeeping from the last chart-bump. A new chart-bump branch is ready now and will merge this update once the release is published, in a separate PR.

That will be in #3456, which I just opened as a draft.

hiddeco (approved):

kingdonb:
Thanks for your responsiveness. I'm going to cut this, and unfortunately push out #3275 to the next release as I haven't had a PR submitted for that yet and so I have had no chance to test the memcached update.

Running up against the end of the day now in the GMT area, and I don't want to delay this release (which would make the dates in the CHANGELOG incorrect, don't want that!)

PR #3456, opened 2021-03-31
Release chart 1.8.0

### LINKED ISSUE #3453: Bump to gnutls 3.6.15-r1 or later in docker build and base image

kingdonb:
Related to #3455

Fixes #3453


kingdonb:
I will rebase this to run the tests again after 1.22.1 image is published. The end-to-end chart tests will probably read as failing before that actually happens, at least that's what I'd think will happen.

chart/flux/values.yaml
```diff
@@ -203,7 +203,7 @@ memcached:
   enabled: true
   hostnameOverride:
   repository: memcached
-  tag: 1.5.20
+  tag: 1.6.9-alpine
```
hiddeco, line 206: Having doubts about combining a MINOR version update of a dependency with a PATCH update of the chart itself.

kingdonb:
I have squashed everything, and tested the alpine version of memcached image works OK, with fewer CVEs in scan results.

chart/flux/values.yaml
kingdonb: If you think it makes sense, I will amend the chart bump and call it 1.8.0

kingdonb: (Done)

hiddeco (approved):
Thank you @kingdonb :rainbow:

PR #3461, opened 2021-04-12
Update Verifa to the Production flux users

maniankara:
Add Verifa Oy to the list of companies running FluxCD in production


kingdonb:
Thanks for using Flux!

kingdonb (approved):
👍

PR #3471, opened 2021-04-22
Update kustomize to v3.8.10

kingdonb:
There is #3458 but we will require a MINOR version bump if the dependencies have also had one, and I am not ready for Flux 1.23.0 just yet.

I think there is a chance that 1.22.2 comes out as much as two weeks before 1.23.0, and since someone contributed this change (originally formed as a PATCH-only update to Kustomize) and is currently waiting for it to go into a release...

If 1.23.0 won't be ready for another week or two, I think we can still at least upgrade Kustomize patch version in the next Flux v1 patch release.

~Kustomize v3.8 series actually went all the way up to v3.8.10, so I will probably fix this PR to use that version before merging.~ done


kingdonb:
Some users were asking for Kustomize updates, I planned to update to the latest (to match Flux v2) 3.10.0, but want to get a release out today for docs and CVE warnings, @stefanprodan how does this approach sound? OK to merge this and cut a new (patch-level) release?

stefanprodan:
> OK to merge this and cut a new (patch-level) release?

Yes sounds good to me.

stefanprodan (approved):

PR #3481, opened 2021-05-05
move list of production users

dholbach:
We moved our list of production users to <https://fluxcd.io/adopters#flux-v1>. Explain how to add yourself.


kingdonb (approved):
Thanks @dholbach 👍

PR #3482, opened 2021-05-06
Always honor the allowedNamespaces setting (or absence), for use with dynamically created namespaces

schizoid90:
Found an issue where, when providing your own clusterRole, flux is not able to provision in namespaces other than it's own. The work around is to add all namespaces to the allowedNamespaces value but this change will not add that option if a clusterRole name has been provided

### DIFF
--- a/chart/flux/templates/deployment.yaml
+++ b/chart/flux/templates/deployment.yaml
@@ -195,7 +195,7 @@ spec:
               name: {{ .Values.env.secretName }}
           {{- end }}
           args:
-          {{- if not .Values.clusterRole.create }}
+          {{- if and (not .Values.clusterRole.create) ( .Values.allowedNamespaces) }}
           - --k8s-allow-namespace={{ join "," (append .Values.allowedNamespaces .Release.Namespace) }}
           {{- end}}
           {{- if .Values.defaultNamespace }}


kingdonb:
Thank you for your contribution!

I'm trying to understand any potential unforeseen implications of making a change like this. What will be the impact of this change, on users who have provided a clusterrole and who have not set allowedNamespaces? It already looks like a bit of a mess in here. I'm not sure this change is needed, after spending some time looking into it. Explained below, hope this helps.

Current state (as I understand it):

If the `clusterRole.create` is false then no cluster role is created at install time, and the setting for k8s-allow-namespace is included: `.Values.allowedNamespaces` and `.Release.Namespace` are appended together and those are the allowed namespaces.

If the `clusterRole.create` is true then a cluster role is created, with cluster-wide access, and the `allowedNamespaces` setting is ignored to match the clusterRole, (which is not restricted unless ... it's not obvious what conditions `clusterRole.create=true` are actually restricted, I think that Flux will always create a clusterrole with broad permissions unless this is set to false.)

Neither of these conditions should limit the access of Flux to only the namespace it is deployed in, unless allowedNamespaces actually hasn't been set at all.

Can you provide the values you installed Flux with? I think they must be wrong.

I just tried this for myself, and got this result (snipped from flux deployment):

```yaml
    spec:
      containers:
      - args:
        - --k8s-allow-namespace=bar,baz,bep,fluxcd
...
```

Here is my flux `values.yaml` from running `helm upgrade --install -n fluxcd --create-namespace flux fluxcd/flux --values flux-values.yaml`:

```yaml
clusterRole:
  create: false
  name: foo
allowedNamespaces:
  - bar
  - baz
  - bep
```

I haven't incorporated your change, and got these results with no problem... I think most likely, you have invoked allowedNamespaces incorrectly which would explain why only the flux install namespace is in this list. If you are using `helm upgrade` or `helm install` with `--set`, it is actually quite complicated to set array values correctly. (I would suggest using a values override file instead, to avoid the potential for misapplied/incorrectly formatted values getting discarded accidentally.)

Is it possible that you passed in a comma-separated string, or somehow other incorrectly populated allowedNamespaces with some data that Helm has discarded, instead of the expected array of namespace strings?

schizoid90:
I am also seeing the allowed namespaces before incorporating my change. I am solving the situation where we are using lots of namespaces that can be created by many users as required, we would have to be updating the list of allowed namespaces constantly.

Perhaps I am misunderstanding then. In the event that the named clusterRole exists does Flux ignore the k8s-allow-namespaces setting?

kingdonb:
Hello friend, I am rebasing a few PRs that remain outstanding so they are caught up to master. I found yours had a merge conflict so I squashed the commits and resolved it.

Please take a look if you have time, to help me be certain that I did it correctly.

I am forced pushing over your branch from a2e7a40bea493582a4077d1cadad41db8b45e14b to ad2eff55f4dd06a56f7c7ac95728b9df64c4b33b – these commits appear to be identical 👍

kingdonb:
I think we talked this issue out some time ago, it was quite complex.

The change is really simple. It just does what is in the title of this issue.

* **Always honor the allowedNamespaces setting (or absence), for use with dynamically created namespaces**

If you pass in an empty value for `allowedNamespaces`, or `false`, then the `--allowed-namespaces` option will be omitted.

I'll be merging this for the next chart release that will come out with Flux 1.24.0

chart/flux/templates/deployment.yaml
```diff
@@ -195,7 +195,7 @@ spec:
               name: {{ .Values.env.secretName }}
           {{- end }}
           args:
-          {{- if not .Values.clusterRole.create }}
+          {{- if and (not .Values.clusterRole.create) ( .Values.allowedNamespaces) }}
```
kingdonb, line 198: Testing this by hand with `helm template`

With a values of:
```
clusterRole:
  create: false
allowedNamespaces: []
```

there is no `--allow-namespaces` string present at all in the template output. It is not in the Flux daemon's deployment `args`, (this is the state that could not be configured before this change is merged.)

With a value of:
```
clusterRole:
  create: false
allowedNamespaces:
  - abc
  - def
```

the `--allow-namespace` option is as follows:

```
- --k8s-allow-namespace=abc,def,default
```

(since the Flux namespace is `default` in this case as I am installing with the helm chart in the default namespace, I believe this is correct)

In the template output by 1.10.2 Flux Helm Chart version, this outcome is impossible. The allow-namespace would still get passed with the Flux install namespace, something like this:

```
- --k8s-allow-namespace=default
```

The allow-namespace directive is still used when the `clusterRole.create` is false, because it is assumed you will list the namespaces that Flux can access. I think this actually suggests there is another patch needed for your use case, where you have already provided a clusterRole and it need not be created for you by Flux's helm chart.

Not to delay the merging of this patch any longer, since it has been blocked from merging for long enough.

kingdonb: I think I have that a bit backwards, with `clusterRole.create` set to false, allow-namespace can be used, but with `clusterRole.create` set to true, allow-namespace is never used. This behavior is the same as before, as Flux can apparently assume that when a ClusterRole is created, it will have unfettered access to all namespaces.

I'm not sure this is correct. Mulling it over out loud while I decide whether this can merge as-is or if it needs something additional. We can always patch it again, but I'd rather be sure I'm not breaking anything at all inside of the Helm chart, as I'm sure you'll understand

kingdonb: I guess it makes sense. You can create a clusterRole that is allowed access to only certain namespaces, but Flux won't create one. Flux only creates a clusterRole that has unfettered access. So it would be wrong to change that, as it will be a breaking change for people who have that configuration. I'm going to leave it alone.

My judgment is this PR can merge as-is 👍 LGTM

kingdonb (approved):

kingdonb:
Hello, there is some discussion in #3552 from a user who had `clusterRole.create` set to false, and didn't pass `allowedNamespaces` – they alerted me that the upgrade past 1.24.0 completely disabled image update automation for them.

I just wanted to check with you,

1. Are you still using Flux v1 (or have you moved on)
2. Did you notice if the Image Update Automation feature works or not with this configuration (`allowedNamespaces: []`)
3. Do you have an alternate approach that could preserve backwards compatibility for the issue described in #3552?

I am afraid we may have to revert this PR, if it has spoiled backwards compatibility. There are lots of Flux users who disable `clusterRole.create` so that they can use Flux v1 within a single namespace, and I'm afraid we didn't consider their use case.

If we can preserve the intent of your original PR without breaking image updates for those users, I will happily keep the behavior. If you moved on, though, since Flux v1 maintenance is likely nearing its end, and this change would represent more new development on a feature that wasn't present before maintenance, I would prefer to simply revert the change, preserving backwards compatibility.

It seems this bug was deeper than it looks.

https://github.com/fluxcd/flux/blob/4358a3b96e54a466f169741330b5bc99596cb605/pkg/cluster/kubernetes/kubernetes.go#L353-L356

I can see here the intent is, when allowedNamespaces is zero length, all namespaces should be allowed. So, the jury is out on where the fault lies in the code now. I am leery to make any further changes in this thread unless we are certain they will solve all associated issues with this change.

schizoid90:
@kingdonb I'm in favour of reverting the PR if it's causing issues. I've moved away from v1 myself so not in a position to try anything different to fix the issue.

PR #3485, opened 2021-05-23
Skip AWS Metadata API pre-flight check when regions are passed via config

### LINKED ISSUE #3015: Unable to fetch tags from ECR

kingdonb:
Replaces #3124

Reportedly fixes #3015

Thanks to @evq and sorry for the delay. ~I'll see if we can get this in 1.23.0.~ This is planned to be included now in 1.23.2.

Since this issue sat for a while and had lots of discussion, I had to research it again. The best summary of who is affected is here:

https://github.com/fluxcd/flux/pull/3124#issuecomment-644832582

The short version is, if you are following best-practices with EKS+IRSA, and still using Flux v1 with its image update automation feature on ECR, then it probably affects you. I don't have any reason to think there's any potential this fix could represent a breaking change.

### DIFF
--- a/pkg/registry/aws.go
+++ b/pkg/registry/aws.go
@@ -23,7 +23,7 @@ import (
 
 const (
 	// For recognising ECR hosts
-	awsPartitionSuffix = ".amazonaws.com"
+	awsPartitionSuffix   = ".amazonaws.com"
 	awsCnPartitionSuffix = ".amazonaws.com.cn"
 
 	// How long AWS tokens remain valid, according to AWS docs; this
@@ -33,7 +33,7 @@ const (
 	// how long to skip refreshing a region after we've failed
 	embargoDuration = 10 * time.Minute
 
-	EKS_SYSTEM_ACCOUNT = "602401143452"
+	EKS_SYSTEM_ACCOUNT    = "602401143452"
 	EKS_SYSTEM_ACCOUNT_CN = "918309763551"
 )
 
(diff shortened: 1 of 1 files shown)


gonzalobarbitta:
I was running into the same issue as [#3124](https://github.com/fluxcd/flux/pull/3124), but using ec2 instead of Fargate. This change solved my problem too.

kingdonb:
I don't know the root cause, but I see this strictly as a bugfix with no breaking potential, and I will plan to include it in the next patch release.

hiddeco (approved):
LGTM, thanks for the combined effort @kingdonb @gonzalobarbitta :100:

PR #3490, opened 2021-06-11
Fix broken link - Push updates to different branch

kingdonb:
This link was good in the original version of this doc page, when it was to be published in the new docs.

https://docs.fluxcd.io/en/1.22.2/faq/#flux-v2-covers-new-use-cases
You can see the bad link here, under this heading ☝️ 

Thanks @juozasg for pointing this out


stefanprodan (approved):

kingdonb:
This PR is no longer needed here, the doc moved and this is fixed in fluxcd/website#352

PR #3512, opened 2021-07-25
Go mod updates

kingdonb:
This branch is equivalent to 06f868c09e6a93cdf61b1cf5792e9ad8974f2ea2, minus f16dfe79

I don't (yet) want to enable the renovatebot on the fluxcd/flux repo proper, as it can generate a lot of noise, but it would be good to get some of the updates it recommended into a release.

The PRs were merged individually at [kingdonb/flux](/kingdonb/flux) however many tests failed because of Docker Hub rate limiting and likely also because my account isn't provisioned with a paid CircleCI membership. Some of the test failures I experienced were likely due to breaking changes, that I reverted and squashed out of the history, you won't see them here.

I will decline to include any major version bumps unless there's a compelling reason for it, as they would have to be carefully evaluated to be sure we haven't leaked any of their breaking changes through to our users.

I worked the PR over for a while and I believe I have incorporated all of the non-breaking upgrades from dependencies. 👍


kingdonb:
It looks like the breaking change was in `weaveworks/common`, which is an internal package that is not versioned according to semver. Reverting that update causes this PR to pass. I'm reinstating the Alpine update reverted in:

```
try reverting alpine update … 9c67658
```

since it was apparently not the issue here!

kingdonb:
Closing in favor of #3537

PR #3515, opened 2021-07-27
Update Kubernetes Go client to v0.21.3 (latest)

kingdonb:
There remained some pins in the go.mod, which I required a release of helm-operator to be published (v1.4.0) in order to progress.

I've removed those pins and resolved all of the compile errors. We should be able to trivially bump the k8s-cli go version to 1.22.0 after merging this.


kingdonb:
This will conflict with #3436 , I will make them both as drafts PRs now and probably have to resolve them in a separate PR.

kingdonb:
Closed in favor of #3537

PR #3519, opened 2021-07-29
Use longer slice of Revision hash for ShortRevision

### LINKED ISSUE #3517: 🎂 `fluxctl sync` fails when more than 1 commit shares the first 7 letters of commit hash
We at @taskworld got this error while running `fluxctl sync`.

```
+ fluxctl sync --k8s-fwd-ns=[redacted]
Synchronizing with ssh://git@github.com/[redacted]/[redacted].git
Revision of master to apply is f1965f6

Waiting for ff965f6 to be applied ...
== Error ==

Error: short SHA1 ff965f6 is ambiguous, full output:
 error: short SHA1 ff965f6 is ambiguous
hint: The candidates are:
hint:   ff965f65e commit 2021-07-29 - [redacted]
hint:   ff965f6d5 commit 2020-11-18 - Auto-release [redacted]/[redacted]:[redacted]
fatal: bad revision 'ff965f65e8b8f689527b5557cd4c14[redacted]..ff965f6'


We don't have a specific help message for the error above.

It would help us remedy this if you log an issue at

    https://github.com/fluxcd/flux/issues

saying what you were doing when you saw this, and quoting the message
at the top.
```

It seems like `fluxctl` has been hardcoded to use only the first 7 letters of the commit name.

https://github.com/fluxcd/flux/blob/2fd24cc5d909930c7745ec884d9060f1239b7291/cmd/fluxctl/sync_cmd.go#L75-L82

Over 1.5 year our FluxCD-managed git repository has 10000+ commits. Using the [**🎂 birthday paradox formula,**](https://en.wikipedia.org/wiki/Birthday_problem#Calculating_the_probability) there is ~17% chance that at least one 7-character-long prefix matches more than one commit. I would suggest making this a bit longer, or just use the full commit hash.

kingdonb:
Fix #3517 

There have been several reports from groups with 10,000 commits or more in their repo that `fluxctl` has trouble when the Git SHA has a collision, because only the first 7 characters are used.

For future-proofing, I chose 12 characters. I bet 10 would be enough. We could do the birthday problem calculation to decide on an appropriate number, I just want to see if tests still all pass when I've fixed all the instances of that `[:7]` that I could find throughout the codebase, or if there might be one hidden in a dependency somewhere outside the project.

### DIFF
--- a/cmd/fluxctl/await.go
+++ b/cmd/fluxctl/await.go
@@ -33,7 +33,7 @@ is safe to retry operations.`)
 		update.PrintResults(stdout, result.Result, verbosity)
 	}
 	if result.Revision != "" {
-		fmt.Fprintf(stderr, "Commit pushed:\t%s\n", result.Revision[:7])
+		fmt.Fprintf(stderr, "Commit pushed:\t%s\n", result.Revision[:12])
 	}
 	if result.Result == nil {
 		fmt.Fprintf(stderr, "Nothing to do\n")
@@ -54,7 +54,7 @@ to run a sync interactively.`)
 			}
 			return err
 		}
-		fmt.Fprintf(stderr, "Commit applied:\t%s\n", result.Revision[:7])
+		fmt.Fprintf(stderr, "Commit applied:\t%s\n", result.Revision[:12])
 	}
 
 	return nil
--- a/cmd/fluxctl/sync_cmd.go
+++ b/cmd/fluxctl/sync_cmd.go
@@ -72,7 +72,7 @@ func (opts *syncOpts) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	rev := result.Revision[:7]
+	rev := result.Revision[:12]
 	fmt.Fprintf(cmd.OutOrStderr(), "Revision of %s to apply is %s\n", gitConfig.Remote.Branch, rev)
 	fmt.Fprintf(cmd.OutOrStderr(), "Waiting for %s to be applied ...\n", rev)
 	err = awaitSync(ctx, opts.API, rev, opts.Timeout)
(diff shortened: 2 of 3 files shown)


kingdonb:
Depending on core maintainer availability for reviews, and our disposition toward this change as PATCH or MINOR, this could be released as early as Monday, in 1.23.2, or some time after that in 1.24.0.

To say it another way, I have it marked as 1.24.0 milestone for now, but this could change.

kingdonb:
We can consider this for including in Flux v1.24.0

I do not know if it would be considered a breaking change, it is clearly a breaking change between the fluxctl CLI and the flux daemon, but I think users are expected to upgrade their `fluxctl` binary when they upgrade `flux`, so I'm not sure if that means it should be excluded.

It's not a particularly urgent bug for us to fix, the workaround is quite simple but also a bit ugly. (Just push another commit that does not have a "birthday neighbor" as explained in #3517, so nobody is waiting desperately for this to be released as a fix.)

I'm OK letting it sit for a while, I have not marked it for the 1.24.0 milestone, but if we are going to merge this we probably will want it to be on a MINOR release even if it is a bug fix, so that people will at least likely remember to upgrade `fluxctl`.

kingdonb:
Per our discussion on #3517 this PR is deliberately not included in 1.24.0

This will be the only PR left 🕯️ on the Flux v1 repo, assuming everything else that remains in the [1.24.0 milestone](https://github.com/fluxcd/flux/milestone/32) can get approved for including and merged for the next release as planned 🎉

pjbgf:
This project is in Migration and security support only, so unfortunately this PR won't be merged. We recommend users to migrate to Flux 2 at their earliest convenience.

More information about the Flux 2 transition timetable can be found at: https://fluxcd.io/docs/migration/timetable/.

PR #3521, opened 2021-07-29
Release 1.23.2

kingdonb:
This release only changes one thing, a bug fix:
* #3015, closed via #3485

Reviewer, please see also the Chart update PR in:
* https://github.com/fluxcd/flux/pull/3522


kingdonb:
We can consider adding

* #3446 and
* #3156 

in this release as well, but if there are issues with approving either of those straight away...

I would rather release this version from this PR as-is, because it has already been confirmed as a fix for current users who are experiencing the issue described from #3015 in the wild.

kingdonb:
Based on the roundabout steps I had to take to attempt to triage #3523 this morning (which is probably the same issue fixed by this PR, #3485), I think this release should probably not be published without including some specific new documentation.

Perhaps adding an FAQ entry or two would make it better, about ECR and the common modes of failure we have seen, and the workarounds that are possible. The workaround is really small, but the fact that it requires a docs update and users to change their behavior in order to take advantage, suggests to me this change may be a MINOR rather than PATCH update.

Users have to know, for example, that in many common cases the Flux daemon's ECR pre-flight checks will not succeed, and they will need to provide region information manually to enable scanning of ECRs, even if they are properly permitted by IAM.

kingdonb:
Approved, via #3522

stefanprodan (approved):

PR #3524, opened 2021-08-01
Run e2e against v1.22.0-rc.0

kingdonb:
The kubectl version and helmrelease CRD from our test scaffold needed to be upgraded. I have not been able to run a full e2e test against the release candidate until now, only by hand smoke testing and with beta and alpha releases. These changes allow e2e itself to pass, given there is a satisfactory version of Kubernetes cluster prepared.

After the release of v1.22.0 final, Kind will be updated, and the review machinery for PRs can be updated to run against the new version as well.

In the mean time, I have prepared a v1.22.0-rc.0 cluster with Kubeadm release candidate and Weave Net, which shows the e2e passing locally, having added these changes to our most recent release. 👍


kingdonb:
The v1.22.0 release of Kubernetes is due out today, hopefully Kind will release a new version soon after that 👍

kingdonb:
Replaced by #3530

PR #3527, opened 2021-08-05
Add GitHub issue template

kingdonb:
Based on the issue template from the Flux v2

@hiddeco suggested I make this improvement to reduce the number of questions I need to ask when an issue report is opened, and to stem the tide of folks who might be in the wrong place (meant to report an issue about Flux v2) / still reporting issues about Flux v1 but who are unaware of the maintenance mode status and implications of that with respect to breaking changes or new feature requests and how they can be handled.


hiddeco (commented):
The preview displays a possible error for me that I think needs to be sorted

```
There is a problem with this template

YAML error: (): found character that cannot start any token while scanning for the next token at line 50 column 18.
```

kingdonb:
Thank you, I composed this by hand because I couldn't figure out the UI.

```
description: "`flux` daemon version, Flux's helm chart version, and/or `fluxctl version`"
```
This line needed to be quoted, it makes sense that the backtick is not a valid start character for YAML string. I have fixed it and forced pushed over.

kingdonb:
The CI failure was from downloading Helm CLI, which Helm apparently just shut down their Google Storage buckets within the past 48 hours. I cherry-picked a commit from #3530 to resolve it here.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,79 @@
+---
+name: Bug report
+description: Create a report to help us improve Flux
```
hiddeco, line 3:
```suggestion
description: Create a report to help us improve Flux Legacy
```

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,79 @@
...
+- type: input
+  attributes:
+    label: Container Registry provider
+    description: If applicable, add the Container Registry provider you are having problems with, e.g. DockerHub, GitHub Packages, Quay.io, etc.
+- type: textarea
+  attributes:
+    label: Additional context
+    description: Add any other context about the problem here. This can be logs, environment specific caveats, etc.
+- type: checkboxes
+  id: terms
```
hiddeco, line 65:
```suggestion
  id: maintenance
```

hiddeco:
I think the preview page has become more intelligent, as from what I recall, I used an empty string for the `assignees` as well which did not result in a warning notice.

kingdonb:
How are you validating/previewing? I think there must be some admin controls on the repo that I don't have.

On the Settings page, my options are very limited. (That's fine though, thanks for your help!)

hiddeco (approved):
Template looks good to me, thanks @kingdonb. :100:

hiddeco:
@kingdonb by viewing the file in rendered mode, e.g. https://github.com/fluxcd/flux/blob/flux-v1-issue-template/.github/ISSUE_TEMPLATE/bug_report.yaml. All issues are resolved now however, and nothing shows anymore.

kingdonb:
Thanks!

PR #3537, opened 2021-08-17
Update kubernetes go client (@ 0.22.0), other go mods

kingdonb:
This PR combines #3515 and #3512, merging them correctly (which they would have generated a conflict otherwise) and also should complete the Kubernetes CLI upgrade, from 1.17.17 to 1.21.3, and now to 1.22.0.

* #3515 
* #3512 

which are both about go module updates.

The upgrade to Kubernetes Client 1.21.3 required some changes to the code, I have to admit I don't know much about Kubernetes context, and I've basically added Context.TODO() calls in every location that required a context, ...

Since taking advantage of any context for a feature reason would have represented doing some feature work for Flux v1, I did not do that. We can consider taking advantage of context to fix bugs in the future, but I don't know when that will happen or what form that effort would have to take. It's not work I have planned myself.

I think it's fine to leave all of these Context.TODO() stubs around forever, I'd be happy to be told by someone that knows more if that's OK (or even if that's not strictly true).

I will keep this as a draft for now until it shows tests passing and until I have verified there are no more stragglers to upgrade before we call it for the 1.24.0 Flux milestone release.

PR #3538, opened 2021-08-18
Update kubernetes go client (@ 0.21.3), other go mods

kingdonb:
After today's dev meeting, I understand better what impact there could be for users from #3537 and why we cannot upgrade Flux's kubernetes go client quite that far.

It makes sense to me that we should upgrade to the last release before the breaking change, since it will still be compatible with two versions forward and two versions backwards. I think is the change we should merge instead of #3537.


kingdonb:
The notes in this comment are still the most relevant starting point for deciding what PRs to merge for this release:

https://github.com/fluxcd/flux/pull/3537#issuecomment-900510673

stefanprodan:
In go.mod the version is set to `go 1.15` should be 1.16 like in CI.

stefanprodan (approved):
LGTM

Thanks @kingdonb

kingdonb:
Thanks @stefanprodan I addressed that in 734022c pushed just now :thumbsup:

Issue #3552, opened 2021-09-09
AllowedNamespaces parameter removed and breaks flux


andrewhertog:
### Describe the bug

We have flux configured in a tenanted format where we have a flux installation per namespaces/tenant.

We just upgraded flux from 1.18.0 to 1.24.1 and the chart from 1.2.0 to 1.11.1 

We noticed quite quickly that flux stopped updating images in that namespace it was watching. 

I went back to chart version 1.2.0 and used flux docker image for 1.24.1 and saw that flux immediately started working. At that point I upgraded the chart one minor version at a time to 1.11.0, and determined that it was the change from 1.10.2 to 1.11.0 that broke the chart.

I'm assuming that this change -https://github.com/fluxcd/flux/pull/3482 is not backwards compatible if `allowedNamespaces` is not set in the values file. 

We've set this value and 1.24.1 is working again with the latest helm chart.

### Steps to reproduce

Install chart version 1.10 with app version 1.24.1, and the following values in NOT the default namespace

```yaml
USER-SUPPLIED VALUES:
additionalArgs:
- --git-label=flux
- --git-sync-tag=flux
annotations:
  ad.datadoghq.com/tags: '{"tenanted-flux":"true"}'
  sidecar.istio.io/inject: "false"
clusterRole:
  create: false
createCRD: false
git:
  branch: main
  path: src/
  pollInterval: 1m
  readonly: false
  secretName: flux-git
  timeout: 40s
  url: <<gitops repo url>>
  user: flux
image:
  repository: fluxcd/flux
  tag: 1.24.1
manifestGeneration: true
memcached:
  annotations:
    ad.datadoghq.com/tags: '{"tenanted-flux":"true"}'
    sidecar.istio.io/inject: "false"
  nodeSelector:
    nodegroup: system
  repository: memcached
  tag: 1.6.10-alpine
nodeSelector:
  nodegroup: system
registry:
  trace: true
sync:
  timeout: 2m
```

Then upgrade to chart version 1.11. Push a new container image to test registry scanning, and see that it stops working with this upgrade

### Expected behavior

Flux upgrade should be backwards compatible. Fixing this required us to set the `allowedNamespaces` parameter.

### Kubernetes version / Distro / Cloud provider

1.18 - EKS

### Flux version

Flux 1.24.1 Chart 1.11.0

### Git provider

Github

### Container Registry provider

Jfrog Artifactory

### Maintenance Acknowledgement

- [X] I am aware of Flux v1's maintenance status

[...]

PR #3553, opened 2021-09-10
Revert "Check for allowed namespaces when providing a clusterrole"

kingdonb:
This reverts commit 2ddd7e4bcec72bb9b499c2d13fcbc74b899e0042, per discussion from #3552 and #3482.

It looks like there's a class of users who only use one namespace and do not set allowedNamespaces that are turning up now, reporting this change wasn't backwards compatible for them and stopped image automation from working anymore.


hiddeco (approved):
Seems like a valid call, thank you @kingdonb :sunflower:

Issue #3554, opened 2021-09-21
Flux produces warning logs about deprecated API versions for kubernetes objects that do not exist


TheeGreenBee:
### Describe the bug

Having upgraded to using flux 1.24.1 from 1.23.1 we have started seeing warning logs from flux telling us that we are using deprecated API versions, for a number of kubernetes objects, each time flux syncs with the Git repository e.g.

```
W0921 14:33:34.146163       7 warnings.go:70] v1 ComponentStatus is deprecated in v1.19+
W0921 14:33:34.241446       7 warnings.go:70] apiregistration.k8s.io/v1beta1 APIService is deprecated in v1.19+, unavailable in v1.22+; use apiregistration.k8s.io/v1 APIService
W0921 14:33:34.327825       7 warnings.go:70] certificates.k8s.io/v1beta1 CertificateSigningRequest is deprecated in v1.19+, unavailable in v1.22+; use certificates.k8s.io/v1 CertificateSigningRequest
W0921 14:33:34.352626       7 warnings.go:70] networking.k8s.io/v1beta1 IngressClass is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 IngressClassList
W0921 14:33:34.358693       7 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0921 14:33:34.363110       7 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0921 14:33:34.403549       7 warnings.go:70] rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRoleBinding
W0921 14:33:34.411424       7 warnings.go:70] rbac.authorization.k8s.io/v1beta1 ClusterRole is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRole
W0921 14:33:34.416192       7 warnings.go:70] rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
[...]

PR #3555, opened 2021-09-21
WIP: handle noisy deprecation warnings

kingdonb:
There is almost certainly more to do, this follows the discussion in #3554 

We would like for Flux users to be able to get their deprecation warnings through Flux logs if it's the only way they have, however we would prefer that the GC sweep does not emit deprecation warnings for resource kinds that are deprecated but aren't even used on the cluster.

I believe this PR currently only suppresses all warnings, it is WIP.

### DIFF
--- a/cmd/fluxd/main.go
+++ b/cmd/fluxd/main.go
@@ -421,6 +421,7 @@ func main() {
 			logger.Log("err", err)
 			os.Exit(1)
 		}
+		restClientConfig.WarningHandler = rest.NoWarnings{}
 		restClientConfig.QPS = 50.0
 		restClientConfig.Burst = 100
 	}


kingdonb:
I can confirm from testing ([here](https://github.com/kingdon-ci/jenkins-infra/pull/47/files)) that this change does suppress all the warnings related to deprecations, though it seems a bit ham-handed, if we can assume the warnings are able to be received some other way, and they are just noise in the Flux logs, then this could be merged.

Given those constraints are not necessarily true, this is not merge-ready IMHO, however for users who are affected by this, if you want to squelch the messages and/or are unable to build this change for yourself, you can test it out now:

```
          image: docker.io/kingdonb/flux:suppress-noisy-deprecations-3d8b1524
```

I will bring this issue up for discussion at the Flux Dev Meeting later this week, meanwhile it will have to remain pending.

kingdonb:
It sounds like we may not have actually been emitting these warnings during the main loop `kubectl apply` but only at the GC sweep. I will have to look at this again later on this week.

dimbleby:
I've submitted #3558, I think that's what you're intending.

dimbleby:
> The warning messages that we care most about, the ones that belong to users' manifests that are emitted at apply time, are not suppressed by this change.

While this is true, those warnings are _already_ suppressed - by virtue of the fact that flux simply doesn't log them.  As I mentioned at https://github.com/fluxcd/flux/issues/3554#issuecomment-924366279, `kubectl apply` sends those warnings to stderr, and flux does not log that output.

I don't think that makes any difference at all to whether this MR should be accepted.  But if we really do care about warnings associated with user manifests, then there's a separate piece of work to be done for those.

kingdonb:
Closing in favor of #3558

PR #3572, opened 2021-11-17
Release 1.24.3

kingdonb:


kingdonb:
This PR should be good for review, I'm ready to cut the next release. If it gets pushed until tomorrow, I will update the CHANGELOG so that it has the proper release date in it. :+1: Still no urgency I think.

kingdonb:
I've just recalled that cutting a release does not require a review, I can merge this and publish 1.24.3 today.

There is no need, or end-user reports that would justify another release now. I still plan to make the release by tomorrow at latest. (About once every 3 weeks is also a good cadence for security fixes in the base image, FWIW.)

stefanprodan (approved):

PR #3591, opened 2022-02-21
Required upgrade for circleci

kingdonb:
"We are deprecating Ubuntu 16.04-based machine images on CircleCI in preparation for an EOL on Tuesday, May 31..."

PR #3596, opened 2022-03-28
Required upgrade for circleci

### LINKED ISSUE #3592: CircleCI base image 1604 is outdated and will retire soon
### Describe the bug

The CircleCI base image has moved, Flux v1 has to upgrade in order to remain current.

The build failed when I tried upgrading with 4355bb0 – I'm not sure why

The error:

```
Makefile:40: *** target pattern contains no '%'.  Stop.
```

### Steps to reproduce

Just look at the build results

### Expected behavior

It should build on the latest base image

### Kubernetes version / Distro / Cloud provider

N/A

### Flux version

N/A

### Git provider

_No response_

### Container Registry provider

_No response_

### Additional context

I will submit a PR when I figure this out. (For inclusion in v1.24.4 Flux release.)

### Maintenance Acknowledgement

- [X] I am aware of Flux v1's maintenance status

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kingdonb:
Fix #3592

We are deprecating Ubuntu 16.04-based machine images on CircleCI in preparation for an EOL on Tuesday, May 31

(Replaces #3591)

The reason for a change in `Makefile` is that newer CircleCI images effectively prohibit us from overriding GOPATH in the `environment` block of `.circleci/config.yml` – the `GOPATH` is set to have two directories by default, we can't override it, and only the first is used for inferring a default `GOBIN`.

I am not certain if allowing `GOPATH` to be set as defaulted can have any other negative side-effects, but this seems to be the least impact change possible that will allow us to go on building after May 31.

Tested a few different approaches to solving this on my fork at `kingdonb/flux` and pretty much settled that other approaches have drawbacks that make this the best option. I experimented with using an orb called `persist-env` and I think that would be rejected for security concerns (though it also worked to solve the issue, when I bypassed a security warning.)

Those tests are on my personal fork, I have not bypassed any security warnings for the Flux PR to merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ ifeq ($(ARCH),)
 endif
 CURRENT_OS=$(shell go env GOOS)
 CURRENT_OS_ARCH=$(shell echo $(CURRENT_OS)-`go env GOARCH`)
-GOBIN?=$(shell echo `go env GOPATH`/bin)
+GOBIN?=$(shell echo `go env GOPATH|cut -d: -f1`/bin)
 
 MAIN_GO_MODULE:=$(shell go list -m -f '{{ .Path }}')
 LOCAL_GO_MODULES:=$(shell go list -m -f '{{ .Path }}' all | grep $(MAIN_GO_MODULE))
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /home/circleci/go/src/github.com/fluxcd/flux
     machine:
-      image: ubuntu-1604:201903-01
+      image: ubuntu-2004:202010-01
     resource_class: large
     environment:
       GO_VERSION: 1.16.7


kingdonb:
The build is passing again, with this PR. I will make some PRs to resolve any outstanding security issues that I can pick up and resolve easily in the base image, then prepare the next release.

I'm not sure if we can move to alpine 3.15, but the alpine 3.14 support with security updates is still good for 1 year + so, it should not be a concern in the short-term.

Makefile
```diff
@@ -22,7 +22,7 @@ ifeq ($(ARCH),)
 endif
 CURRENT_OS=$(shell go env GOOS)
 CURRENT_OS_ARCH=$(shell echo $(CURRENT_OS)-`go env GOARCH`)
-GOBIN?=$(shell echo `go env GOPATH`/bin)
+GOBIN?=$(shell echo `go env GOPATH|cut -d: -f1`/bin)
```
kingdonb, line 25: "It's probably a simple one-line change"

^ Me, three weeks ago. (Yes it was! But how does one know which line needs to change? The error was: https://app.circleci.com/pipelines/github/fluxcd/flux/976/workflows/dd384c61-c331-4a9b-9154-d3ea5dd872af/jobs/11947?invite=true#step-107-132
```
Makefile:40: *** target pattern contains no '%'.  Stop.
```

FFS... I think we might want to consider again modernizing the CircleCI / moving off and adopting GitHub Actions for Flux v1, as originally proposed in #2944, now that I am more familiar with our build tooling and with GitHub Actions, I could consider reopening that issue here.)

Not today, this one-line kludge means we can keep using it for now.

kingdonb:
@stefanprodan Please review and 👍 if you have a moment! 🙇

stefanprodan (approved):
LGTM

Thanks @kingdonb

PR #3603, opened 2022-04-07
Add access to `--git-verify-signatures-mode` from Helm

julienduchesne:
This is retro-compatible with previous versions (won't cause a new deployment without specifically setting the config)
It allows a user to set the `--git-verify-signatures-mode` option instead of the deprecated `--git-verify-signatures` which is currently always set

### DIFF
--- a/chart/flux/README.md
+++ b/chart/flux/README.md
@@ -270,6 +270,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `git.setAuthor`                                   | `false`                                              | If set, the author of git commits will reflect the user who initiated the commit and will differ from the git committer.
 | `git.signingKey`                                  | `None`                                               | If set, commits will be signed with this GPG key
 | `git.verifySignatures`                            | `false`                                              | If set, the signatures of the sync tag and commits will be verified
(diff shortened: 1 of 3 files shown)


kingdonb:
Thanks for submitting a PR! I think this looks OK to merge and include in the next chart release.

There's nothing objectionable or not backwards compatible, and it simply continues advancing forward with the deprecation that was already made in Flux daemon. 👍 

I'm marking this for the next milestone. Unsure when that will be, but likely within the next few weeks.

kingdonb:
This may take some time to get merged and released, as I'll want to do this carefully and make sure I have tested it.

At a glance it does look like it should be backwards-compatible. We can do a chart release without doing a full release and build, but as the releases have got less frequent, I suspect there is likely a fresh backlog of CVE scan warnings I'll need to catch up with, and it might be worthwhile publishing a 1.25.1 release with no changes, only a fresh build that happens against the next newest alpine which is 3.15.4.

kingdonb:
I will rebase and merge this to release with the next chart, coming out together with the 1.25.1 flux daemon image 👍

kingdonb (approved):
LGTM 👍 thanks for your contribution to Flux!

PR #3617, opened 2022-05-26
Housekeeping: merge release branch 1.25.x into master

kingdonb:


kingdonb:
@stefanprodan Should @pjbgf get write access to this repository based on his membership in `core-maintainers`?

kingdonb:
The build reflects as failed because I published the release as `v1.25.2` accidentally (should be `1.25.2`) – this is an easy mistake to make in the GitHub release UI and I've made it before, (it's fixed.)

The artifacts all got pushed out successfully, this build is passed. I'll be testing it out momentarily from the published source just for smoke, then comment on the Helm chart PR when I'm comfortable that it's ready to tag the chart.

stefanprodan:
@kingdonb all good now, I added flux v1 and helm-op to the team.

kingdonb:
Turns out the build has to pass... I have fixed it, the build is passing now.

Thanks @stefanprodan for fixing the repo permissions!

Issue #3320, opened 2020-10-05
Flux v1 is in maintenance mode


dholbach:
Flux v1 (and the Helm Operator v1) are in maintenance mode and effectively have been for a while.

This means we

* are focusing all of our attention on [v2](https://toolkit.fluxcd.io) and
* will only be working on Flux (and the Helm Operator v1) for security fixes

If you are considering helping out with maintenance, please talk to us. We want to make this transition work for everyone.

### Why is this important?

We as a community decided we will invest into the future of GitOps by re-architecting our tooling from the ground up, resulting in

* Easier code
* Better observability
* Ease of integrating
* Composability and extensibility

The effort is well-underway and v2 is already on feature-parity in terms of Flux read-only mode and Helm Operator functionality. (And there’s already loads more in terms of functionality that Flux v1 and friends couldn't do, such as multiple repository support, health assessment of the deployed workloads, structured logging, events and alerts.)

Check out the guide on https://toolkit.fluxcd.io/get-started/ to find out how to get started with the GitOps Toolkit.

### Timeline

Please take a look at [the roadmap for Flux v2](https://toolkit.fluxcd.io/roadmap/). Once we have reached feature-parity across all current feature sets (➊ read-only mode, ➋ helm-operator, ➌ write operations), we will continue to support Flux v1 and Helm Operator v1 for 6 more months.

### I need help with an issue now

Flux is still being maintained and supported, it will just take a bit longer until we get around to issues and PRs. Critical bug fixes have our priority, we are going to be hesitant and look very careful at anything that will extend functionality.

First and foremost, you can help by checking out https://docs.fluxcd.io/en/stable/troubleshooting/ and https://docs.fluxcd.io/en/stable/faq/ very closely and asking for help on `#flux` on [CNCF Slack](https://slack.cncf.io) if necessary.

This is a very important transition period for us as a community, and we firmly believe that we'll come out stronger together on the other side. :rocket:

benyanke:
Can this be renamed to "Flux v1 is in maintenance mode"? The title is a bit deceptive as is.

staceypotter:
Can we add Leigh's upcoming talk for the Flux / GOTK Guide walk-through? 

**Upcoming Talk:** 
[GitOps Toolkit Guide Walk-through](https://www.meetup.com/GitOps-Community/events/273640196/)
Join us October 19 (10am PT / 18:00 BST) for to this special walk-through of the GitOps Toolkit! 
[...]

PR #3634, opened 2022-08-17
nodeselector beta.kubernetes.io/os updated

krouser:
updated nodeselector to kubernetes.io/os

Signed-off-by: Stanlee <east4era@yahoo.com>

### DIFF
--- a/deploy/flux-deployment.yaml
+++ b/deploy/flux-deployment.yaml
@@ -19,7 +19,7 @@ spec:
         name: flux
     spec:
       nodeSelector:
-        beta.kubernetes.io/os: linux
+        kubernetes.io/os: linux
       serviceAccountName: flux
       volumes:
       - name: git-key


kingdonb:
Thanks for your contribution here. I am having trouble finding documentation to support when the `beta.kubernetes.io` was removed, I see it was deprecated a long time ago in 1.14.

Is this broken today if you try to use it in K8s 1.24? (An earlier version?)

I should note that Flux v1 is in maintenance mode (#3320) and we generally try to admonish everyone coming through here to upgrade; if you have some blocking issues that prevent you from upgrading to Flux v2, we'd like to hear about them!

kingdonb:
Hello, we are preparing to archive the Flux repo, per #3645 

I'm going to close this, as it's unlikely we'll be doing any more feature releases, but thank you for being a part of Flux history!

PR #3638, opened 2022-08-29
Add deprecation warnings to Helm Chart and fluxctl

### LINKED ISSUE #3628: Add deprecation notice to fluxctl and Flux v1 Helm chart
To prevent new users to start with v1 instead of v2, we could add a warning to both `fluxctl install` and to `helm install flux`.
For the CLI, the warning should be written to stderr to not affect the install command, for Helm the warning could be place in NOTES.txt.

The deprecation notice must include a link to Flux v2 get started guide.

kingdonb:
Fix #3628

This is for the upcoming release (hopefully we can cut this today 🎉)

I'll be submitting one more PR with the gamut of upgrades, (and we should likely have corresponding changes for the Helm Operator's chart coming out shortly as well.)

### DIFF
--- a/chart/flux/templates/NOTES.txt
+++ b/chart/flux/templates/NOTES.txt
@@ -8,3 +8,12 @@ and running:
 
   fluxctl identity --k8s-fwd-ns {{ .Release.Namespace }}
 
+---
+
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @kingdonb

Issue #3564, opened 2021-11-01
batch/v1beta1 CronJob is deprecated ... use batch/v1 CronJob


kingdonb:
### Describe the bug

```
warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

Related to this comment in #3558:

* https://github.com/fluxcd/flux/pull/3558#issuecomment-956543000

We actually call this API by name; when/if Flux v1 should upgrade the K8s client-go again, it will be a good time to address this by moving on. The CronJob batch/v1 API has only been available since 1.21 though, so it may need to wait a while longer before it can be addressed here.

This issue may also never need to be addressed, depending on future developments in the [Migration Timetable](https://fluxcd.io/docs/migration/timetable/).

(Please upgrade to Flux v2 ASAP!)

### Steps to reproduce

Simply run Flux at v1.24.1, this message comes up repeatedly.

It's not nearly as distressing as the list of warnings addressed in #3558 but it does repeat (every automation interval?)

### Expected behavior

No deprecation warnings in the normal course of Flux's operation.

### Kubernetes version / Distro / Cloud provider

kubeadm v1.21.6 on bare metal

### Flux version

Flux v1.24.1+

### Git provider

N/A

### Container Registry provider

N/A

### Maintenance Acknowledgement

- [X] I am aware of Flux v1's maintenance status

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kingdonb:
Kubernetes 1.24 is due out tomorrow (!)

I think we said we'd target this upgrade to come out with the same timing as about Kubernetes 1.25, so later this year.

Will continue to keep an eye on this and test against the latest releases as they come out.

pjbgf:
This project is in Migration and security support only, so unfortunately this issue won't be fixed. We recommend users to migrate to Flux 2 at their earliest convenience.

More information about the Flux 2 transition timetable can be found at: https://fluxcd.io/docs/migration/timetable/.

kingdonb:
This is the issue to fix, in order to provide support for K8s 1.25 and up. We are probably throwing errors on this version now.

kingdonb:
I will close this issue, as we will be archiving the fluxcd/flux repository soon:

* #3645

PR #3639, opened 2022-08-29
Update dependencies

kingdonb:
Working on upgrades for the next patch release


kingdonb:
I lined up all the release PRs, so when this one gets approved I should be able to knock them all down and release in about an hour 👍

kingdonb:
WDYT @pjbgf ? `kustomize` is pinned/frozen in place at 3.8.4 and `sops` is already at 3.7.3, the latest. I think that's everything, except for Kubernetes. Let me know if you find anything I have missed, I'll add it.

Do you think there are any other version pins we should advance?

K8s 1.25.x is out now, which means K8s 1.21.x series is finally out of support (EOL). We have been pinning this series (1.21.x) and planned on upgrading past the breaking issues from those dependencies again around this milestone, about when there should be no more people still running on K8s 1.16.x, I have lost track of the details about supporting at the edges. I thought we had an issue pinned where we discussed how we plan to handle this, but I'm not seeing it. Will search for it again later.

I think that we might still want to consider publishing that type of upgrade but it should come later, if at all, in a next MINOR v1.26.0. I am not aware of any issues with K8s 1.24 or 1.25 preventing Flux v1 users from advancing on that upgrade path. We can release these upgrades as a patch now, and plan those upgrades for next release. That's what I'd propose.

That being said, I have not yet tested against K8s 1.25.0, I do not have a test matrix of different Kubernetes release versions at my disposal for testing against, as my maintainer duties are not resourced in that way here. I have plenty of ARM64 resources but Flux v1 only runs on AMD64 officially. I will try to make a 1.25.0 cluster and root out any issues with the e2e.

---

We are not expecting to make major changes in the Flux v1 codebase again. I think there are some deprecation warnings that may impact users in the strange place of being on the very latest Kubernetes release but also pinned to Flux legacy v1, there are probably no users in this position. (It seems much more likely that Flux v1 runs on systems which are never upgraded.)

I have not checked the deprecation schedule for this release out yet, but I think we are due for another breaking change requiring upgrade soon from upstream Kubernetes. Maybe someone else is more on top of this information than I am today.

go.mod
```diff
@@ -22,6 +22,9 @@ replace (
 	k8s.io/code-generator => k8s.io/code-generator v0.21.14
 )
 
+// Versions v3.0.0>x>v2.4.0 are susceptible to CVE-2022-28948
+replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
+
```
pjbgf, line 27:
```suggestion

// Fix CVE-2022-1996
replace github.com/emicklei/go-restful => github.com/emicklei/go-restful v2.16.0+incompatible

```

pjbgf (changes requested):
From my checks, we can fix an additional CVE as per below.

pjbgf:
@kingdonb I agree with the approach, let's take the Kubernetes update as a separate thread/release. In the coming days I plan to take a look around Flux 1 and the impact of 1.21.x being EOL and [1.22.x becoming EOL soon](https://kubernetes.io/releases/#release-v1-22).

kingdonb:
I think #3564 is the issue that I was looking for. If there are any more deprecation warnings to address, let's open an issue.

Reopened:
* #3564

pjbgf (approved):
LGTM

Thanks @kingdonb for working on this. :bow:

PR #3640, opened 2022-08-29
Go Yaml v3

kingdonb:
Unsure why this apparently causes tests to fail


stefanprodan:
yaml v3 is not backwards compatible with v2, why are er doing such major changes in Flux v1?

kingdonb:
We aren't, I figured it out in #3639 – it should pass now.

Flux will stay on yaml.v2

Some dependencies have moved on to yaml.v3 but they pin an incorrect version with a CVE.

We can replace it and no more CVE warnings 👍

PR #3647, opened 2022-11-01
Readd docs

dholbach:
This PR re-adds docs from https://fluxcd.io/legacy/flux/

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 # Flux
 
-![Flux v1 Diagram](/img/flux-v1-diagram.png)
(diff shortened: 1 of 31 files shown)


docs/README.md
```diff
@@ -0,0 +1,43 @@
+# Flux
+
+![Flux v1 Diagram](https://fluxcd.io/img/flux-v1-diagram.png)
```
stefanprodan, line 3: Please move the diagram here as well and update the link.

stefanprodan (approved):
LGTM

Thanks @dholbach 🏅

